PR #17, opened 2021-11-17
Several updates to testing

kueda:
* Restructured tests so they are in their own dir at the root
* Added an integration test that mocks the response from inaturalistjs
* Added a unit/component test for ObsList that mocks ObservationProvider
* Instead of mocking Realm, I mocked the Realm config so it writes to an
  in-memory database for tests
* Added test data factories with factoria and faker
* Updated Github Actions config to avoid duplicate builds and notify Slack


tests/factories/RemoteComment.js
```diff
@@ -0,0 +1,10 @@
+import { define } from "factoria";
+
+export default define( "RemoteComment", faker => ( {
+  id: faker.datatype.number( ),
+  uuid: faker.datatype.uuid( ),
+  created_at: "2021-11-09T08:28:05-08:00",
+  updated_at: "2021-11-09T17:22:17-08:00",
+  user: factory( "RemoteUser" ),
```
albullington, line 8: Noticed this one is missing a `factory` import. Should we set up some [pre-commit hooks](https://markoskon.com/create-a-pre-commit-hook-with-husky-and-lint-staged/) to run `eslint` (and maybe `jest`) before every commit?

kueda: Will fix. Def. in favor of a pre-commit eslint. Pre-commit jest is overkill, IMO, because running all the tests takes a while, and we're going to be doing that anyway with Github Actions.

src/components/Observations/ObsCard.js
```diff
@@ -31,7 +31,7 @@ const ObsCard = ( { item, handlePress }: Props ): Node => (
     </View>
     <View>
       <Text style={textStyles.text}>{item.identificationCount}</Text>
-      <Text style={textStyles.text}>{item.commentCount}</Text>
+      <Text style={textStyles.text} testID="ObsList.obsCard.commentCount">{item.commentCount}</Text>
```
albullington, line 34: Question here: how much test coverage should we be aiming for? Is it enough to test that one piece of data like `commentCount` is rendering correctly, or is it better to test a few different things?

tests/factories/LocalObservation.js
```diff
@@ -0,0 +1,14 @@
+import { define } from "factoria";
+
+// TODO use faker for more of these dynamic values. Also, would we really
+// store timeObservedAt like this in Realm?!
```
albullington, line 4: Nope, we would store time in the same format as received from the API. Not a very realistic example.

tests/integration/MyObservations.test.js
```diff
@@ -0,0 +1,55 @@
...
+  ( ) => render(
+    <NavigationContainer>
+      <ObservationProvider>
+        <ObsList />
+      </ObservationProvider>
+    </NavigationContainer>
+  )
+);
+
+test( "renders the number of comments from remote response", async ( ) => {
```
albullington, line 26: I'm curious why you chose to render from the remote response instead of the local, in-memory database. Since we're only rendering the list of observations once they're stored in Realm, wouldn't rendering local responses be a more realistic integration test?

src/components/Observations/hooks/fetchObservations.js
```diff
@@ -220,6 +220,7 @@ const writeToDatabase = useCallback( ( results ) => {
     results.forEach( obs => {
       const newObs = createObservationForRealm( obs );
       realm?.write( ( ) => {
+        // Shouldn't the primary key in realm handle this?
```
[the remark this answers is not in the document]
kueda: Hm, could we perhaps use Realm's [upsert functionality](https://docs.mongodb.com/realm/mongodb/update-documents/#upsert-documents) for this?

src/providers/ObservationProvider.js
```diff
@@ -34,19 +31,19 @@ const ObservationProvider = ( { children }: Props ): Node => {
         setObservationList( localObservations );
       }
 
-      // Live queries and objects emit notifications when something has changed that we can listen for.
-      subscriptionRef.current = localObservations;
+      localObservations.addListener( ( ) => {
+        // If you just pass localObservations you end up assigning a Results
+        // object to state instead of an array of observations. There's
+        // probably a better way...
+        setObservationList( localObservations.map( o => o ) );
```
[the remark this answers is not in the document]
kueda: Hah, I was hoping you could tell me. In practical terms, if I did *not* do this, `observationsList` would get assigned a [Results object](https://docs.mongodb.com/realm-legacy/docs/javascript/latest/api/Realm.Results.html), not an array. In the test environment, that yielded a warning about not wrapping updates to state in `act()`, and indeed, it didn't actually update the state in a way that the test would wait for. In the app itself, everything was fine. You can try it by changing this line and running `MyObservations.test.js`. I'd prefer not to mock parts of Realm since this is an integration test, but it seemed like Realm was performing some asynchronous task that the test wasn't able to wait for, so this was my compromise.

src/components/Observations/ObsCard.js
kueda: It's subjective. If you test every single thing, you have a lot of confidence it's working the way you intended, but you also spent a lot of time writing tests instead of writing the code under test, and you probably made a really brittle test that will require lots of maintenance every time you change the code under test. Snapshots seem like a way to reduce the labor of writing tests, but they don't address the brittleness issue. I was sort of thinking they might be more appropriate later in the development process when components achieve some degree of stability.

Personally, I wanted to test things like "if I provide this data, are the cards I expect getting rendered, AND do the cards have some of the content I'm expecting?" and not "do the cards have ALL the content I'm expecting?" because I imagine the content of the cards is going to change a lot in these early stages of development.

tests/integration/MyObservations.test.js
kueda: Hm, maybe the test description is misleading. What I meant was "render the number of comments from a remote response that gets persisted locally and made available to the ObsList component." That's what's happening, but that seemed excessively wordy.

Again, it's somewhat subjective, but I interpret "integration" to mean integration between the different parts of the code in your app. If we mock the Realm response here, we ignore the code we have for mapping API responses to Realm objects, so we're not integrating as much of our own code as we could and thus not testing something closer to the actual user experience. With this approach, we're only mocking an input from outside our app (the inatjs response), which then gets run through our mapping code to create Realm records, which are then made available to the component through our own context and providers, so this is testing the integration of all of those things that we wrote. It's also integrating some things that we didn't write, like the internals of Realm, but as long as they work quickly and predictively, I think that's actually a good thing.

src/components/Observations/ObsCard.js
albullington: Rad, makes sense to me. The RN testing course I took suggested the same thing about snapshots -- we might want to start using them once we're closer to an MVP and know the code is more stable.

src/providers/ObservationProvider.js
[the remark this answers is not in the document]
albullington: Hmm. I think this is fine for now, but it's something we might keep an eye on to make sure there are no performance implications.

It's also a little strange to me that we're setting state with the Results object on line 31, but resetting state with an array on line 38. Probably better to be consistent?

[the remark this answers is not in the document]
kueda: > Probably better to be consistent?

Not sure. Personally, I think the discrepancy and comment point out that something is wrong here, so it's better to isolate that wrongness in a single place where we can annotate it and hopefully fix it some day.

PR #19, opened 2021-11-18
Pre-commit hooks; make Realm models resemble API more closely; add realm listener for ObsList

albullington:
- set up husky
- added User and Taxon realms
- added realm listener to re-render ObsList when new observations are added to realm
- swapped navigation so Drawer navigator is the root, and the observation Stack navigator is the first item in Drawer
- use `.objectForPrimaryKey` method to look up existing realm objects

probably shouldn't merge this until after `5-testing` branch is merged in, since the realm model overhaul will cause some changes to tests & test data


src/providers/ObservationProvider.js
```diff
@@ -36,6 +36,13 @@ const ObservationProvider = ( { children }: Props ): Node => {
 
       // Live queries and objects emit notifications when something has changed that we can listen for.
       subscriptionRef.current = localObservations;
+      localObservations.addListener( ( arr, changes ) => {
+        // changes object has properties including insertions, modifications, and deletions
+        // so we can decide when we need obslist to rerender here. otherwise, it will listen for all changes
+        if ( changes.insertions.length > 0 || changes.modifications.length > 0 ) {
```
kueda, line 42: Don't we *want* to listen to all changes to this collection, including deletions? Like if you delete an observation locally, shouldn't that be reflected in what you see on My Observations?

albullington: Yep, fixed

src/models/Observation.js
```diff
@@ -1,27 +1,55 @@
...
-      commonName: "string?",
-      createdAt: "string",
+      createdAt: "string?",
       description: "string?",
       identifications: "Identification[]",
-      identificationCount: "int",
       latitude: "double?",
-      location: "string",
       longitude: "double?",
       photos: "Photo[]",
```
kueda, line 48: We should model this as observation.observationPhotos.photo. The ObservationPhoto is where we store things like the position of the photo in the obs (e.g. when you want to re-order them). So you'd need an additional ObservationPhoto model in the app, and the mapping would need to look in the `observation_photos` attribute to get them from the API response.

src/models/Observation.js
```diff
@@ -1,27 +1,55 @@
...
-      placeGuess: "string",
-      qualityGrade: "string",
-      taxonRank: "string",
-      timeObservedAt: "string",
-      userProfilePhoto: "string",
-      userLogin: "string",
-      userPhoto: "string"
+      placeGuess: "string?",
+      qualityGrade: "string?",
+      taxon: "Taxon",
```
kueda, line 51: `taxon` is not required. Does it need a `?` here?

src/models/Taxon.js
```diff
@@ -0,0 +1,24 @@
...
+      name: taxon.name,
+      preferredCommonName: taxon.preferred_common_name,
+      rank: taxon.rank
+    };
+  }
+
+  static schema = {
+    name: "Taxon",
+    properties: {
+      id: "int",
```
kueda, line 15: Shouldn't this be a primary key?

src/models/User.js
```diff
@@ -0,0 +1,22 @@
...
+      iconUrl: user.icon_url,
+      login: user.login,
+      name: user.name
+    };
+  }
+
+  static schema = {
+    name: "User",
+    properties: {
+      id: "int",
```
kueda, line 14: Should be a primary key.

src/navigation/rootNavigation.js
```diff
@@ -0,0 +1,46 @@
...
+const hideHeader = {
+  headerShown: false,
+  label: "my observations"
+};
+
+const Drawer = createDrawerNavigator( );
+
+const App = ( ): React.Node => (
+  <NavigationContainer>
+    <ObservationProvider>
```
kueda, line 22: If we provide the user's observations to the entire app, does that mean they're hanging around in state even if they're spending all their time in Explore or Identify? E.g. you go to My Observations and your obs get loaded from Realm and added to the global state where they're available to My Observations as well as obs detail. But then you leave and go to Identify: are your local observations still hanging around in memory even though you're not using them any more?

src/navigation/rootNavigation.js
```diff
@@ -0,0 +1,46 @@
...
+      <Drawer.Navigator screenOptions={screenOptions} name="Drawer">
+        <Drawer.Screen
+          name="my observations"
+          component={StackNavigator}
```
kueda, line 26: If this is a StackNavigator that *only* manages the stack within My Observations, shouldn't we give it a unique name, like `MyObservationsStackNavigator`? Or is there a way we could reuse it for any drawer?

src/providers/ObservationProvider.js
```diff
@@ -34,19 +31,18 @@ const ObservationProvider = ( { children }: Props ): Node => {
         setObservationList( localObservations );
       }
 
-      // Live queries and objects emit notifications when something has changed that we can listen for.
-      subscriptionRef.current = localObservations;
+      localObservations.addListener( ( ) => {
+        // changes object has properties including insertions, modifications, and deletions
+        // so we can decide when we need obslist to rerender here. otherwise, it will listen for all changes
+        setObservationList( localObservations.map( o => o ) );
```
kueda, line 37: Do you understand why we need `localObservations.map( o => o )` here instead of just `localObservations`, or using the argument from the `addListener` callback? I did that in the testing branch b/c it worked, but I didn't entirely understand why.

Also, do we need to remove this listener?

kueda (commented):
Well, I guess I did have some comments and questions.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -39,24 +38,26 @@ const ObsDetails = ( ): Node => {
       <ScrollView>
       <View style={viewStyles.userProfileRow}>
         <Pressable style={viewStyles.userProfileRow} onPress={navToUserProfile}>
-          <UserIcon uri={observation.userProfilePhoto} />
-          <Text>{`@${observation.userLogin}`}</Text>
+          {/* TODO: fill user icon in with saved current user icon or icon from another user API call */}
+          <UserIcon uri={null} />
+          {/* TODO: fill in text with saved current user login or login from another user API call */}
+          <Text>@username</Text>
```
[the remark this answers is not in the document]
albullington: That's a good idea -- I can make a line item for that in the mobile dev timeline. I've been curious about libraries like [react-native-skeleton-content](https://github.com/alexZajac/react-native-skeleton-content-nonexpo) which generate a skeleton UI, since that seems like the direction most mobile apps have gone instead of loading wheels. But that seemed to me like a design decision and also wouldn't solve for the empty state you're taking about.

package.json
```diff
@@ -7,10 +7,12 @@
     "ios": "react-native run-ios --simulator='iPhone 13'",
     "start": "react-native start",
     "test": "jest",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "prepare": "husky install"
```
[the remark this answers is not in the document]
albullington: Yep, added to README (with a link to the Husky setup instructions, just in case that doesn't work).

src/components/Observations/hooks/fetchObservations.js
```diff
@@ -45,193 +97,27 @@ const useFetchObservations = ( ): Array<Object> => {
...
+        if ( existingObs !== undefined ) {
+          // TODO: modify existing objects when syncing from inatjs
           return;
         }
         realm?.create( "Observation", newObs );
       } );
     } );
-}, [createObservationForRealm] );
+    setLoading( false );
+}, [] );
```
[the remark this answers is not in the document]
albullington: Yep, the empty array is used to make sure the `writeToDatabase` function only runs once here, since it's basically telling React that the effect doesn't need to re-run based on any props/state changes. `useCallback` won't run at all without a second argument.

src/navigation/rootNavigation.js
albullington: Yep, I'm pretty sure they would still be in memory. Maybe a better (more performant) way to do this is to create one context and provider for My Observations and a separate one for Explore.

kueda: Yeah, I think that sounds right, especially considering they'll be serving different kinds of data (presumably): dynamic data from the API in Explore, local data from Realm in My Observations (I think).

PR #40, opened 2022-01-20
i18n with Fluent and i18next

kueda:
Sets up i18n with i18next using Fluent as the translation file format. Fluent provides a lot of flexibility to translators and works well with Crowdin's UI, and the cost of adding complexity for translators. These changes presume we're using the Crowdin CLI to upload source changes and download translations, but this could probably be set up with their Github integration too.

I haven't actually globalized much of the text in the app so far, nor have I set up any automated upload / download functionality via github actions. This is mainly here to see if this seems like a reasonable approach. Maybe we don't even want to start globalizing text until we have something more than the proof-of-concept UI we have right now.

FWIW, another approach we could take is to perform conversion from Fluent to Fluent JSON at runtime. I chose to avoid this to reduce app complexity and avoid performance issues, but that would allow us to theoretically load translations over the wire. Maybe something to keep in mind for the future.


README.md
```diff
@@ -29,5 +29,20 @@ We currently have two kinds of tests
...
+# Upload new strings
+crowdin upload --token YOUR_ACCESS_TOKEN --project-id YOUR_PROJECT_ID
+
+# Download new translations and build for use in the app
+crowdin download --token YOUR_ACCESS_TOKEN --project-id YOUR_PROJECT_ID
+node src/i18n/i18ncli.js build
+git add src/i18n/l10n/*
+git commit -a -m "Updated translations"
+```
+
```
albullington, line 45: I think it would be helpful to spell out that `/src/i18n/strings.ftl` is the source of truth where developers should add new translations, since the only way to know that right now is by digging into the `crowdin.yml` file.

Maybe overkill, but it could also be helpful to spell out some brief guidelines for how to name translation keys in a consistent way. And when to use comments.

Issue #34, opened 2022-01-12
Messages screen


albullington:
Messages should function similarly to the existing Android app. A user should be able to:

- [ ] Search their messages
- [ ] See a list of their most recent messages
- [ ] Tap into messages to see the full details (sender, timestamp, full message thread)
- [ ] Search for iNaturalist users
- [ ] Compose a message (with subject and body) to a specific user

nickm01:
will attempt this one

PR #46, opened 2022-02-11
#34 Show Messages - first attempt

nickm01:
Basic/simple implementation of messages, navigation from footer.  Includes tests.

NOT included:
1. Showing full message data on the screen (right now just shows subject)
2. Go to details
3. Pull-to-reload etc
4. Styling 
5. Storing data in realm (max 100)
6. Paging and/or just show max 100


src/components/Messages/MessageList.js
```diff
@@ -0,0 +1,47 @@
...
+        <Text style={textStyles.projectName}>{item.subject}</Text>
+      // TODO </Pressable>
+    );
+  };
+
+  return (
+    <FlatList
+      data={messageList}
+      renderItem={renderMessages}
+      testID="Message.list" //Q: Is this for automated testing?  How should I use?
```
albullington, line 41: Yes, `testID` is for automated testing. We've been using jest and setting up unit tests to check that some of the basic app functionality is working -- making sure data is rendered correctly on the page, that button taps do what we expect them to, and that a user can navigate away from their current screen.

Since you're already looking at ProjectDetails, here's a link to the ProjectDetails tests where you can see how we're setting up the unit tests.

https://github.com/inaturalist/iNaturalistReactNative/blob/03392f7da3a35952149488a12d4b1d323123d260/tests/unit/components/Projects/ProjectDetails.test.js

src/models/Message.js
```diff
@@ -0,0 +1,50 @@
...
+      ...msg,
+        subject: msg.subject,
+        body: msg.body,
+        fromUser: User.mapApiToRealm( msg.from_user, realm ),
+        toUser: User.mapApiToRealm( msg.to_user, realm ),
+        threadId: msg.thread_id
+    };
+
+    // need to append user after the observation realm object has been created
+    //delete newObs.user; //Q: Why do we need to do this?
```
albullington, line 30: We don't, actually... this was code I wrote before I realized we can upsert data into Realm at the time of creation. This is fixed in the branch I'm working on (30-sound-recorder) and will likely get merged in next week.

src/models/Message.js
```diff
@@ -0,0 +1,50 @@
...
+    return newMsg;
+  }
+
+
+  //Q: https://api.inaturalist.org/v1/docs/#!/Messages/get_messages shows everything is optional - how do we handle?
+  static schema = {
+    name: "Message",
+    primaryKey: "uuid", //Q: UUID is used instead of ID - is that right?
```
albullington, line 38: I'd go with whatever is in the API documentation here. If we're using `id` in the messages API, best to keep it consistent.

Also, it's fine to have most or all fields be optional in Realm.

src/models/Message.js
```diff
@@ -0,0 +1,50 @@
...
+      threadId: msg.thread_id
+    };
+  }
+
+  static createMessageForRealm( msg, realm ) {
+    const newMsg = {
+      ...msg,
+        subject: msg.subject,
+        body: msg.body,
+        fromUser: User.mapApiToRealm( msg.from_user, realm ),
```
albullington, line 24: These should be snake case, right?

nickm01: @albullington was mimic'ing `class Observation`.  A little unsure when we use snake vs camel

src/components/Messages/Messages.js
```diff
@@ -0,0 +1,28 @@
...
+
+import React from "react";
+import type { Node } from "react";
+
+import ViewWithFooter from "../SharedComponents/ViewWithFooter";
+import MessageList from "./MessageList";
+import useMessages from "./hooks/useMessages";
+
+const Messages = ( ): Node => {
+  const userLogin = { user_login: "nickm01" };
```
albullington, line 11: are lines 11-13 necessary here? I think we can just call `useMessages( )`, which will load all messages 1x when the screen first focuses.

src/components/Messages/hooks/useMessages.js
```diff
@@ -0,0 +1,55 @@
...
+
+import { useEffect, useState } from "react";
+import inatjs from "inaturalistjs";
+
+import { getJWTToken } from "../../LoginSignUp/AuthenticationService";
+import { MESSAGE_FIELDS } from "../../../providers/helpers";
+
+// TODO: need to trigger a rerender on change of user. Right now API parameters not being used.
+const useMessages = ( apiParams: Object ): {
+  messageList: Array<Object>,
```
albullington, line 11: This should be `messages`, not `messageList`, since `messages` is what is being returned from the hook. Not sure if you have Flow set up, but if so you should be able to see this Flow error on line 15 of `Messages.js`.

src/components/Messages/hooks/useMessages.js
```diff
@@ -0,0 +1,55 @@
+// @flow
+
+import { useEffect, useState } from "react";
+import inatjs from "inaturalistjs";
+
+import { getJWTToken } from "../../LoginSignUp/AuthenticationService";
+import { MESSAGE_FIELDS } from "../../../providers/helpers";
```
albullington, line 7: This is smart, I would have forgotten we already had this defined elsewhere.

src/navigation/messagesStackNavigation.js
```diff
@@ -0,0 +1,28 @@
...
+
+import Messages from "../components/Messages/Messages";
+
+const Stack = createNativeStackNavigator();
+
+const screenOptions = {
+  headerShown: true
+};
+
+const MessagesStackNavigation = (): React.Node => (
```
albullington, line 14: Can we move this out of its own navigator and into `myObservationsStackNavigation`?

The only way to access messages in iNat mobile is via the MyObservations screen. The user should be able to navigate to Messages from `components/Observations/MessagesIcon`. (Sorry, I probably should have specified this in the Github issue).

src/navigation/rootNavigation.js
```diff
@@ -36,6 +37,11 @@ const App = ( ): React.Node => (
         options={hideHeader}
       />
       <Drawer.Screen name="missions/seen nearby" component={PlaceholderComponent} />
+      <Drawer.Screen
```
albullington, line 40: no need for Messages in the Drawer navigator

tests/unit/components/Messages/Messages.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+it( "displays message subject", ( ) => {
+  const { getByText } = renderMessages( );
+
+  waitFor( ( ) => {
+    expect( getByText( mockMessage.subject ) ).toBeTruthy( );
+  } );
+} );
+
+it( "displays activity indicator", ( ) => {
```
albullington, line 45: Can we change the name of this test? This seems to be testing that the activity indicator is no longer on screen once messages are fetched, so the test name seems misleading.

tests/unit/components/Messages/Messages.test.js
```diff
@@ -0,0 +1,52 @@
...
+import Messages from "../../../../src/components/Messages/Messages";
+
+const mockedNavigate = jest.fn( );
+const mockMessage = factory( "RemoteMessage" );
+
+jest.mock( "../../../../src/components/Messages/hooks/useMessages" , ( ) => ( {
+  __esModule: true,
+  default: ( ) => {
+    return {
+      messageList: [mockMessage],
```
albullington, line 15: This should also be `messages`, not `messageList`, since that's what is being returned from the hook.

albullington (changes requested):

tests/unit/components/Messages/Messages.test.js
```diff
@@ -0,0 +1,49 @@
...
+  <NavigationContainer>
+    <Messages />
+  </NavigationContainer>
+  )
+);
+
+test.todo( "add displays message subject test" );
+//unclear why this fails
+//
+// it( "displays message subject", async ( ) => {
```
nickm01, line 45: @albullington need some help here on why this doesn't work if you have a moment.

src/components/Messages/hooks/useMessages.js
nickm01: @albullington oh how do I get flow setup?

tests/unit/components/Messages/Messages.test.js
nickm01: @albullington havent been able to test that the activity indicator works yet

tests/unit/components/Messages/Messages.test.js
albullington: Two things:

1) like I said, you need to change the returned value of `useMessages` to use the key messages, not messageList. If you have the wrong name of the return value, the test will fail.
2) you don't need any of the async/awaits here, since you're just rendering a component. so renderMessages should look like:

```
const renderMessages = ( ) => render(
  <NavigationContainer>
    <Messages />
  </NavigationContainer>
);
```

and similarly in the test, there should be no async/await:

```
it( "displays message subject", ( ) => {
  const { getByText } = renderMessages( );
```

if you do these two things, the test should pass.

albullington (approved):
This all looks good to me. I'll go ahead and merge in.

PR #54, opened 2022-03-08
Fixes Projects.test.js false positive problem

nickm01:
The `expect` lines never actually completed before the test was terminated.  Switched to approach used in MyObservations.test.js

New to this stuff - so happy for other thoughts!

### DIFF
--- a/tests/unit/components/Projects/Projects.test.js
+++ b/tests/unit/components/Projects/Projects.test.js
@@ -37,25 +37,25 @@ jest.mock( "@react-navigation/native", ( ) => {
   };
 } );
 
-const renderProjects = ( ) => render(
-  <NavigationContainer>
-    <Projects />
-  </NavigationContainer>
+const renderProjects = async () => waitFor(
+  () => render(
+    <NavigationContainer>
+      <Projects />
+    </NavigationContainer>
+  )
 );
 
-test( "displays project search results", ( ) => {
-  const { getByTestId, getByText } = renderProjects( );
+test( "displays project search results", async ( ) => {
+  const { getByTestId, getByText } = await renderProjects( );
 
   const input = getByTestId( "ProjectSearch.input" );
   fireEvent.changeText( input, "butterflies" );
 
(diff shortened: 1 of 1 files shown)


tests/unit/components/Projects/Projects.test.js
```diff
@@ -37,25 +37,25 @@ jest.mock( "@react-navigation/native", ( ) => {
   };
 } );
 
-const renderProjects = ( ) => render(
-  <NavigationContainer>
-    <Projects />
-  </NavigationContainer>
+const renderProjects = async () => waitFor(
```
kueda, line 40: For a unit test, you probably don't need `waitFor` if you've mocked all your effects. If you're waiting for anything it should probably be an integration test because you're testing both the code under test _and_ whatever you're waiting for.

tests/unit/components/Projects/Projects.test.js
```diff
@@ -37,25 +37,25 @@ jest.mock( "@react-navigation/native", ( ) => {
...
+    <NavigationContainer>
+      <Projects />
+    </NavigationContainer>
+  )
 );
 
-test( "displays project search results", ( ) => {
-  const { getByTestId, getByText } = renderProjects( );
+test( "displays project search results", async ( ) => {
+  const { getByTestId, getByText } = await renderProjects( );
```
kueda, line 49: Similarly, if you're not using `waitFor`, you don't need to use any async/await stuff.

kueda (changes requested):
I think the original false positive you reported was because we were incorrectly using `waitFor` without awaiting completion, but I don't think `waitFor` is necessary to begin with here.

For me this passes if I just remove the `waitFor` in the test itself and run everything synchronously. If I insert `expect( getByText( "hello" ) ).toBeTruthy( );` in the test I get a failure as expected. Does that work for you? IMO that tests what we want to test without dealing with async/await complexity.

kueda (approved):

PR #75, opened 2022-04-01
Bug fixes from beta 0.0.1

albullington:


src/i18n/strings.ftl
```diff
@@ -55,5 +60,16 @@ SIMILAR-SPECIES-header
...
 TAXONOMY-header = TAXONOMY
+
+Upload-a-photo-from-your-gallery = Upload a photo from your gallery
+
+Upload-X-photos = Upload {$count ->
+    [one] 1 photo
+    *[other] {$count} photos
+}
+
+You-can = You can:
```
kueda: This could use a context comment. Someone is eventually going to ask "You can what?" b/c either the pronoun or the verb will be inflected differently depending on what follows.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "inaturalistreactnative",
-  "version": "0.0.1",
+  "version": "1.0",
```
kueda, line 3: I might go with 0.1.0 instead. Applying semantic versioning rules to applications (as opposed to component software) isn't perfect, but if it takes the form of major.minor.patch, I use patch to indicate changes that fix problems with existing functionality, minor for changes that add new functionality, and major for re-writes or changes so sweeping that the user feels like they're using a different app.

In this PR, you've fixed a lot of bugs, but you've also added a lot of new functionality, hence the minor version bump. I would save a major version bump for launch, though in this case, we may end up skipping from version 0 to version 3 or something, to make sure that the major version is higher than whatever the existing iPhone and Android apps are right now.

albullington: Ah, I was thinking that I probably wouldn't be able to upload an Android build with a lower version than the previous version. Maybe that only applies to build number though.

PR #77, opened 2022-04-08
Settings Screen

budowski:


src/components/Settings/UserSearchInput.js
```diff
@@ -0,0 +1,72 @@
...
+import Pressable from "react-native/Libraries/Components/Pressable/Pressable";
+
+const UserSearchInput = ( { onUserChanged} ): React.Node => {
+  const [hideResults, setHideResults] = React.useState( true );
+  const [userResults, setUserResults] = React.useState( [] );
+  const [userSearch, setUserSearch] = React.useState( "" );
+  // So we'll start searching only once the user finished typing
+  const [finalUserSearch] = useDebounce( userSearch, 500 );
+
+  useEffect( () => {
```
albullington, line 15: Let's reuse `sharedHooks/useRemoteSearchResults.js` here

src/components/Settings/SettingsTabs.js
```diff
@@ -0,0 +1,39 @@
+// @flow
```
albullington, line 1: This code isn't getting used, is it? I see another component called `SettingsTabs` in `components/Settings/Settings.js` that seems to be the one we're actually rendering.

src/components/Settings/SettingsRelationships.js
```diff
@@ -0,0 +1,341 @@
...
+  }, [accessToken, finalUserSearch, following, page, sortBy, trusted] );
+
+  useEffect( () => {
+    refreshRelationships();
+  }, [accessToken, finalUserSearch, following, trusted, sortBy, page, refreshRelationships] );
+
+  useEffect( () => {
+    const getBlockedUsers = async () => {
+      try {
+        const responses = await Promise.all( settings.blocked_user_ids.map( ( userId ) => inatjs.users.fetch( userId ) ) );
```
albullington, line 68: I'm getting an `undefined is not an object (evaluating 'settings.blocked_user_ids.length')` error here, so I can't load the relationships tab. I'm guessing the same error will pop up for `muted_user_ids`.

src/components/Settings/SettingsProfile.js
```diff
@@ -0,0 +1,68 @@
...
+
+const SettingsProfile = ( { settings, onSettingsModified } ): React.Node => {
+  const profileSource =
+    settings.removeProfilePhoto || !settings.icon ?
+      require( "./../../images/profile.png" ) :
+      { uri: settings.newProfilePhoto ? settings.newProfilePhoto.uri : settings.icon};
+
+  const onImageSelected = ( response ) => {
+    if ( response.didCancel ) {return;}
+    onSettingsModified( { ...settings, newProfilePhoto: response.assets[0], removeProfilePhoto: false } );
```
albullington, line 14: Can we display the new profile photo after a user selects it from `launchImageLibrary`? Otherwise it's hard to know whether the selection actually worked.

src/components/Settings/SettingsProfile.js
```diff
@@ -0,0 +1,68 @@
...
+          style={viewStyles.profileImage}
+          source={profileSource}
+        />
+        <View style={viewStyles.column}>
+          <Button title="Upload New Photo" onPress={() => launchImageLibrary( {}, onImageSelected )} />
+          <Button title="Remove Photo" onPress={() => onSettingsModified( { ...settings, removeProfilePhoto: true } )} />
+        </View>
+      </View>
+      <View style={viewStyles.column}>
+        <Text>Username</Text>
```
albullington, line 31: One thing I'm noticing throughout Settings is that we're not pulling in the user's existing information. We should display the user's current username, email, bio, etc. so they can see their user profile before making changes.

src/components/Settings/Settings.js
```diff
@@ -0,0 +1,198 @@
...
+      setSettings( response.results[0] );
+      setIsLoading( false );
+    } catch ( exc ) {
+      console.error( "exc", exc );
+      console.error( JSON.stringify( exc ) );
+    }
+  }, [accessToken] );
+
+
+  useEffect( () => {
```
albullington, line 111: I'm getting a 404 error with this API call when testing with staging API v2. I also think this should be moved into `SettingsApplications.js`, because we don't need to make this API call or load authorized applications at all unless a user taps that tab.

The `authorizedApps` and `providerAuthorizations` state and `revokeApp` functions can also be moved into the `SettingsApplications` component.

src/components/Settings/Settings.js
```diff
@@ -0,0 +1,198 @@
...
+
+
+  const fetchProfile = useCallback( async () => {
+    try {
+      const response = await inatjs.users.me( {api_token: accessToken} );
+      console.log( "User object", response.results[0] );
+      setSettings( response.results[0] );
+      setIsLoading( false );
+    } catch ( exc ) {
+      console.error( "exc", exc );
```
albullington, line 105: I'm getting this first `console.error` every time I open Settings.

src/components/Settings/Settings.js
```diff
@@ -0,0 +1,198 @@
...
+  const revokeApp = async ( appId ) => {
+    const response = await inatjs.authorized_applications.delete( { id: appId }, {api_token: accessToken} );
+    console.log( "Revoked app", response );
+    const apps = await inatjs.authorized_applications.search( {}, {api_token: accessToken} );
+    console.log( "Authorized Applications", apps.results );
+    setAuthorizedApps( apps.results );
+  };
+
+  return (
+    <SafeAreaView style={viewStyles.container}>
```
albullington, line 173: Can we wrap this component in `SharedComponents/ViewWithFooter.js` so a user can get back to the drawer menu?

albullington (changes requested):
This is looking pretty good overall. I'd like to make sure this works on staging with API v2, that it loads existing user data from `users.me`, and that the 3 crashes I'm seeing get fixed. I'd also recommend using `npm install` instead of yarn, so we can remove the yarn lockfile. I think we're going to run into build issues with yarn since we're using npm for our CI.

index.js
```diff
@@ -8,14 +8,14 @@ import App from "./src/navigation/rootNavigation";
...
 
-inatjs.setConfig( {
-  apiURL: "https://stagingapi.inaturalist.org/v2",
-  writeApiURL: "https://stagingapi.inaturalist.org/v2"
-} );
-
 // inatjs.setConfig( {
-//   apiURL: "https://api.inaturalist.org/v2",
-//   writeApiURL: "https://api.inaturalist.org/v2"
+//   apiURL: "https://stagingapi.inaturalist.org/v2",
```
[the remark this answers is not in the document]
budowski: No, I tested it with v1 - now tested with v2 and did some fixes - however, there are open issues with v2 and the `inaturalistjs` library - will write down in the comment for PR below

package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
```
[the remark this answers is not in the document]
budowski: Merged and tested with latest

src/components/Settings/PlaceSearchInput.js
```diff
@@ -0,0 +1,77 @@
...
+import {inatPlaceTypes} from "../../dictionaries/places";
+
+const PlaceSearchInput = ( { placeId, onPlaceChanged} ): React.Node => {
+  const [hideResults, setHideResults] = React.useState( true );
+  const [placeResults, setPlaceResults] = React.useState( [] );
+  const [placeSearch, setPlaceSearch] = React.useState( "" );
+  // So we'll start searching only once the user finished typing
+  const [finalPlaceSearch] = useDebounce( placeSearch, 500 );
+
+  useEffect( () => {
```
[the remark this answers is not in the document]
budowski: Fixed in latest commit

src/components/Settings/PlaceSearchInput.js
```diff
@@ -0,0 +1,77 @@
...
+          onPlaceChanged( 0 );
+        }}>
+          <Image
+            style={viewStyles.clearSearch}
+            resizeMode="contain"
+            source={require( "../../images/clear.png" )}
+          />
+        </Pressable>
+      </View>
+      {!hideResults && finalPlaceSearch.length > 0 && placeResults.map( ( place ) => (
```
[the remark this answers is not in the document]
budowski: Related to v2 API changes - we'll need to update `inaturalistjs` library to support this - the problem: For v1, there is one endpoint for places search, and for v2 it's another. So we'll need to add some logic that checks if the APIs being used is v1 or v2.

src/components/Settings/SettingsAccount.js
```diff
@@ -0,0 +1,100 @@
+import {Text, View} from "react-native";
+import {viewStyles, textStyles} from "../../styles/settings/settings";
+import React from "react";
+import {Picker} from "@react-native-picker/picker";
+import {colors} from "../../styles/global";
+import Pressable from "react-native/Libraries/Components/Pressable/Pressable";
+import CheckBox from "@react-native-community/checkbox";
+import {inatLanguages} from "../../dictionaries/languages";
```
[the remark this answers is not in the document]
budowski: Not that I'm aware of - @pleary?

src/components/Settings/SettingsNotifications.js
```diff
@@ -0,0 +1,88 @@
...
+  "When a project adds your observations": "prefers_project_added_your_observation_email_notification",
+  "Project curator changes": "prefers_project_curator_change_email_notification",
+  "Taxonomy changes": "prefers_taxon_change_email_notification",
+  "Observations by people I follow": "prefers_user_observation_email_notification",
+  "Observations of taxa or from places that I subscribe to": "prefers_taxon_or_place_observation_email_notification"
+};
+
+
+const EmailNotification = ( { title, value, onValueChange } ): React.Node => {
+  return <Pressable style={[viewStyles.row, viewStyles.notificationCheckbox]}  onPress={() => onValueChange( !value )}>
```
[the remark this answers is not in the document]
budowski: Fixed in latest commit

src/i18n/index.js
```diff
@@ -28,10 +28,14 @@ i18next
   } ) )
   .init( {
     // TODO should default to the system locale, or a user preference
+    compatibilityJSON: "v3",
```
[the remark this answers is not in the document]
budowski: Added in latest commit - basically related to code not working otherwise in Android

src/components/Settings/UserSearchInput.js
budowski: Fixed in latest commit

src/components/Settings/SettingsTabs.js
budowski: Removed in latest commit

src/components/Settings/SettingsRelationships.js
budowski: v2 API issue - `users/me` is missing muted_user_ids and blocked_user_ids

src/components/Settings/SettingsProfile.js
budowski: Weird, it's working for me on Android - still not showing anything after you select an image on iOS?

src/components/Settings/SettingsProfile.js
budowski: Was related to v2 vs v1 API usage - now working

src/components/Settings/Settings.js
budowski: Fixed in latest commit. However, in API v2 there are not endpoints to pull authorized apps and provider authorizations.

src/components/Settings/Settings.js
budowski: Fixed in latest commit

src/components/Settings/Settings.js
budowski: Fixed in latest commit

budowski:
Open issues with v2 APIs (@kueda / @pleary):

* `users/me` - missing in response: email, blocked_user_ids, muted_user_ids, search_place_id, time_zone, locale (if supposed to be under site.locale -> it's null), place_id (if supposed to be under site.place_id -> it's null), all prefers_X settings, preferred_project_addition_by, preferred_X_license

* `places` - no `places/autocomplete` - need to update `inaturalistjs` library with an additional endpoint to do queries (it's `places?q=abcd` instead). However, this means we'll need some logic that checks if it's `v1` or `v2` API being used for this.
* No `authorized_applications` APIs.
* No `provider_authorizations` APIs.

src/components/Settings/PlaceSearchInput.js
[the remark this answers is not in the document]
kueda: @budowski, how about using https://api.inaturalist.org/v2/docs/#/Search/get_search instead? That should work in v1 and v2 with no need to modify inatjs. https://github.com/inaturalist/iNaturalistAPI/issues/321 is an issue with the response sizes, but I don't think it should prevent you from building with that endpoint.

In general I think we should move toward `/search` and away from `/:model/autocomplete` endpoints so people get consistent results.

src/components/Settings/SettingsAccount.js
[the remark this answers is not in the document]
kueda: There are a few reasons we hard-code these in the Android app, but the most important is that we don't want the user choosing a language the app does not support and looking for a file that doesn't exist. I'm pretty sure we can guard against that in this app b/c it is only loading files that do exist, and if another locale is requested we'll just load English, but I agree it would be far preferable to show exactly the same choices we show on the web. So here's a plan

* @budowski: remove time zones from this PR as we are no longer supporting user-editable time zones
* I'll will make issues to have the API serve up locales and language names the app is allowed to use, but for now we can just use the approach in this PR
* For Sites / the Network, we should not hardcode them and use https://api.inaturalist.org/v2/sites instead, BUT since that was never specified here, let's just ignore that here and work on it in the future; I'll make an issue.

src/components/Settings/SettingsProfile.js
kueda: This doesn't work for me in iOS or in Android, as in I don't see the image updated in the app, and the request to stagingapi fails... I think b/c PUT /v2/users/:id is [documented but not implemented](https://github.com/inaturalist/iNaturalistAPI/issues/323). Yaron, while we're working on that, can we get the image to appear even before it uploads so the user can see that something changed?

src/components/Settings/SettingsProfile.js
kueda: Definitely loading everything that's available right now, with the caveat that there's an [API problem](https://github.com/inaturalist/iNaturalistAPI/issues/324) with returning email.

kueda:
Thanks, @budowski.

> users/me - missing in response

https://github.com/inaturalist/iNaturalistAPI/issues/324

> places - no places/autocomplete

Please try `/v2/search?sources=places&q=term` instead. It may have its problems, but I think it's preferable to separate autocomplete endpoints for all records.

> No authorized_applications APIs.

https://github.com/inaturalist/iNaturalistAPI/issues/325

> No provider_authorizations APIs.

https://github.com/inaturalist/iNaturalistAPI/issues/326

Some things remaining from Amanda's review you can work on now:

- [ ] remove `time_zone` from the settings; we don't support that any more
- [ ] show the profile pic the user selected before upload begins
- [ ] use /v2/search for the places autocomplete

src/components/Settings/SettingsProfile.js
budowski: @kueda - Very odd, I just tested the preview-before-upload on Android and it did work for me (meaning, choosing "Upload new photo" then the selected photo gets displayed in the app). It doesn't show the already-uploaded photo, since the v2 staging API shows a URL in the `staticdev.inaturalist.org` domain, which returns S3 error of access denied.
Maybe it's a permissions issue?

kueda: Not sure why it's not working for you, but for me, `const profileSource` is not getting set when `settings.icon` is blank and you choose a new and you choose a new profile photo without removing the old one first. This change works for me:

```javascript
let profileSource;
if ( settings.newProfilePhoto ) { profileSource = { uri: settings.newProfilePhoto.uri }; }
else if ( settings.icon ) { profileSource = { uri: settings.icon }; }
else { profileSource = require( "./../../images/profile.png" ); }
```

budowski: Used your latest code with some modifications, since it didn't work if you chose a photo then removed it

budowski:
@kueda - all three issues addressed in latest commit

kueda (changes requested):
Thanks for addressing those three requests

- [x] Merge from main and resolve conflicts
- [ ] Make sure tests are passing
- [ ] If I change something in Settings, don't save, then go to explore, those unsaved changes should not be there when I return. IMO, once you've made changes in Settings, you should not be allowed to leave until you save or confirm your choice to leave in a modal
- [x] Get everything working with v2 (found & fixed some bugs while reviewing; lmk if you find more)
- [ ] Locale chooser is not showing the current user's locale as selected (seems to just be on the first option, Arabic, even though my locale is set to `en`)

The following could be worked on if you want, but maybe we should focus on merging and split those off into their own issues?
- [ ] Relationships section
- [ ] Applications section

budowski:
@kueda - all issues fixed in latest commit. Just one note regarding showing an alert of "Discard changes" - 
The problem is that we're using `react-navigation` with a side menu / drawer - so the [official](https://reactnavigation.org/docs/preventing-going-back/) method of preventing the user from leaving the screen doesn't work here. 
We can [detect](https://reactnavigation.org/docs/function-after-focusing-screen) when we lose focus this way, but cannot prevent it from happening.
The only option remaining, would be to override each and every footer/side-menu button click, that will call a generic should-allow-leaving callback function of the current screen (if it one exists).

src/components/Settings/SettingsRelationships.js
kueda: Still not working, but for a different reason: you're querying the relationships with values the API doesn't understand.

src/i18n/index.js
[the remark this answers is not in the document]
kueda: I addressed this in https://github.com/inaturalist/iNaturalistReactNative/commit/09ed361d022c88e8333b0127db016e36b5e97728 by adding a polyfill, so I don't think this change is necessary anymore. However, `useSuspense` still seems to be necessary for Android.

kueda (changes requested):
- [x] Tests are not passing. Please fix whatever changes caused them to fail.
- [x] My chosen locale is now showing as selected, but when I change my Default Search Place and tap Save, the locale resets to Arabic. This seems to happen for a lot of attributes after I tap Save.
- [x] Bug: unchecking "Receive Email Notifications" seems to have the opposite effect. On the web, make sure Notify Me, Confirming IDs, and Receive Email Notifications are all turned on; in the app, go to the Notifications section, uncheck Receive Email Notifications; and tap "Save." The three checks now have the opposite of the values they had before I tapped "Save"
- [x] Bug: repeat saves seem to submit `undefined` values. Follow the steps for the bug above, then uncheck "Receive Email Notifications" again; you'll get a bunch of error responses from the API about submitting the wrong types, b/c the app submitted `undefined` as the value for a lot of stuff
- [ ] the `useRemoteSearchResults` hook isn't working with the new v2 response, which means you can't add blocked or muted users
- [ ] Relationships: my existing relationships aren't showing up

Regarding stopping the user from leaving settings if anything has changed, let's punt on that until we can discuss it with Abhas, b/c that's kind of an IA issue and indicative of the weirdness of having two separate top-level navigational elements (drawer AND tabs).

kueda (changes requested):
- [ ] I still can't mute or block users; I can search for them, but choosing the result does nothing, and I can't even scroll all the way down to the mute user search field
- [x] Existing blocked users show, but not with a username or name; they should at least show a username
- [ ] [Wrap all calls to inatjs with try/catch blocks](https://github.com/inaturalist/iNaturalistReactNative/pull/77#discussion_r878586370) and tell the user if the request fails for any reason
- [x] Remove `compatibilityJSON: "v3"` from src/i18n/index.js` per https://github.com/inaturalist/iNaturalistReactNative/pull/77#discussion_r878595006

budowski:
Fixed these issues in latest commit, however - the following APIs are missing from v2:
Creating/deleting muted and blocked users.

kueda (approved):
K, I think this is ok to merge once that conflict is resolved with the caveat that we still need to deal with some of the relationships functionality, modal editing for user profile attributes, and UI refinement, but I'm willing to kick those cans down the road for now.

PR #78, opened 2022-04-19
Save observations, photos, sounds in Realm for upload

albullington:
Thought it would be helpful to have a reviewer for this one because of the Realm changes.


src/providers/helpers/uploadObservation.js
```diff
@@ -0,0 +1,90 @@
+// @flow
+
+import inatjs from "inaturalistjs";
+
+import { getJWTToken } from "../../components/LoginSignUp/AuthenticationService";
+import { markUploaded, markPhotoUploaded, markSoundUploaded } from "./markUploaded";
```
kueda, line 6: This is a total style issue, but if I'm not trying to separate major functional groups like MVC components, I generally prefer to keep relevant code in the same file / module *unless* I need to reuse them. Here, these methods are only called from this module, so I think they can be included here.

src/providers/helpers/markUploaded.js
```diff
@@ -0,0 +1,49 @@
...
+    console.log( e, "couldn't mark photo uploaded in realm" );
+  }
+};
+
+const markSoundUploaded = async ( uuid: string, id: number ) => {
+  try {
+    const realm = await Realm.open( realmConfig );
+    const obsSound = realm.objectForPrimaryKey( "ObservationSound", uuid );
+    realm?.write( ( ) => {
+      obsSound.id = id;
```
kueda, line 38: Why don't sounds get a `timeSynced`?

src/providers/helpers/markUploaded.js
```diff
@@ -0,0 +1,49 @@
...
+  } catch ( e ) {
+    console.log( e, "couldn't mark photo uploaded in realm" );
+  }
+};
+
+export {
+  markUploaded,
+  markPhotoUploaded,
+  markSoundUploaded
+};
```
kueda, line 49: All these methods are really similar. Could they be combined into something like `markRecordUploaded( uuid: string, type: string, response: object )` and put the type-specific mapping work into other functions, maybe `attrsFromReponse( response: object, type: string )`?

src/providers/helpers/uploadObservation.js
```diff
@@ -0,0 +1,90 @@
...
+
+const uploadObservation = async ( obsToUpload: Object, localObs: Object ) => {
+  const FIELDS = { id: true };
+
+  try {
+    const apiToken = await getJWTToken( false );
+
+    const uploadParams = {
+      // TODO: decide how to format place_guess param
+      // right now it looks like street, city, state is preferred on the web
```
kueda, line 62: Formatting the `place_guess` should probably not happen during upload. That's user-editable, so any auto-population of that field should happen when the coordinates change.

src/providers/helpers/uploadObservation.js
```diff
@@ -0,0 +1,90 @@
...
+    // save id to realm and set time synced
+    await markUploaded( obsToUpload.uuid, id );
+    if ( localObs.observationPhotos ) {
+      createPhotoParams( id, apiToken, localObs ); // v2
+    }
+    if ( localObs.observationSound ) {
+      createSoundParams( id, apiToken, localObs ); // v2
+    }
+  } catch ( e ) {
+    console.log( JSON.stringify( e.response.status ), "couldn't upload observation: ", JSON.stringify( e.response ) );
```
kueda, line 86: Show errors to the user (or storing them) on the todo list, I take it?

src/providers/hooks/useObservations.js
```diff
@@ -17,10 +17,15 @@ const useObservations = ( ): Object => {
   const nextPageToFetch = observationList.length > 0 ? Math.ceil( observationList.length / perPage ) : 1;
   const [page, setPage] = useState( nextPageToFetch );
   const [userLogin, setUserLogin] = useState( null );
+  const [obsToUpload, setObsToUpload] = useState( [] );
 
-  const syncObservations = ( username ) => {
+  const syncObservations = ( username = null ) => {
```
kueda, line 22: Why does this take a `username` at all? We can't sync obs if there isn't an authenticated user.

src/providers/hooks/useObservations.js
```diff
@@ -34,12 +39,19 @@ const useObservations = ( ): Object => {
 
     // When querying a realm to find objects (e.g. realm.objects('Observation')) the result we get back
     // and the objects in it are "live" and will always reflect the latest state.
-    const localObservations = realm.objects( "Observation" );
+
+    const localObservations = realm.objects( "Observation" ).sorted( "timeObservedAt", true );
+    console.log( localObservations.length, "local obs in useObservations" );
+    const notUploadedObs = realm.objects( "Observation" ).filtered( "timeSynced == null && timeUpdatedLocally != null" );
```
kueda, line 45: Don't you just need `timeSynced == null` here? How can `timeUpdatedLocally` be null and if it can, why would you exclude those from upload?

kueda (commented):
These are mostly based on an initial read-through, not actually trying to run and use this, but I figured I'd start there and do another round based on actual use. Main overall question at this point is what's the intended scope here? Just to save these records to realm? Or to save and then provide UI to upload?

src/models/Observation.js
```diff
@@ -108,19 +147,34 @@ class Observation {
...
       created_at: { type: "string?", mapTo: "createdAt" },
       description: "string?",
+      geoprivacy: "string?",
+      id: "int?",
       identifications: "Identification[]",
       latitude: "double?",
       longitude: "double?",
       observationPhotos: "ObservationPhoto[]",
-      observationSounds: "ObservationSound[]",
+      observationSound: "ObservationSound?",
```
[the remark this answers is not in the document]
albullington: Oh, interesting. I think the UX confused me here -- I remember Abhas saying we only wanted to let users upload one sound at a time. And since we don't have anywhere to display existing sounds in the current UX, I didn't think there was a need for multiple sounds. I'll change this back.

src/models/Observation.js
```diff
@@ -108,19 +147,34 @@ class Observation {
...
+      observed_on_string: "string?",
+      owners_identification_from_vision_requested: "bool?",
+      species_guess: "string?",
       place_guess: { type: "string?", mapTo: "placeGuess" },
+      positional_accuracy: "double?",
       quality_grade: { type: "string?", mapTo: "qualityGrade" },
       taxon: "Taxon?",
       time_observed_at: { type: "string?", mapTo: "timeObservedAt" },
+      time_zone: "string?",
```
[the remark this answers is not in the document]
albullington: Oh great, I didn't realize we're setting time zone automatically now. I can remove this.

src/models/Observation.js
```diff
@@ -108,19 +147,34 @@ class Observation {
...
+      timeSynced: "date?",
```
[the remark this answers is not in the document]
albullington: I think I'm still getting mixed up on the difference between synced at, updated at, observed at, created at, and where we need to use each of them respectively.

I've been thinking of `time_observed_at` and `created_at` as the times coming back from the server, `timeSynced` as the time when data is downloaded from the server to the local device, and `timeUpdatedLocally` as the time when an observation gets saved locally for future upload. Does that sound right? Am I missing any other times that I need here? Or are there any I could combine? `created_at` and `timeUpdatedLocally` seem like basically the same thing to me, except that the source of truth is the server vs. the local device.

src/providers/helpers/markUploaded.js
albullington: Hmm, is `timeSynced` something that should be coming back from the server? And if so, is that from the `updated_at` attribute?

src/providers/hooks/useObservations.js
albullington: This was my funky way of making this work when a user logs in for the first time. Normally, we would fetch the user login from the `getUsername` function, but that didn't seem to be working the first time a user logs in -- I assumed there was a timing issue where the username didn't have time to be stored and retrieved from RNSInfo by the time a user lands on the ObsList. So this is a special case where I'm passing the username via navigation parameters from the Login screen.

I'm guessing there's a cleaner way to do this.

src/models/Observation.js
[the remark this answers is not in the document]
kueda: These are the relevant date or time fields from the server

* `created_at`: timestamp of when observation was created on the server; not editable
* `updated_at`: timestamp of when the observation was updated on the server; not editable
* `observed_on`: date when the observer observed the organism; user-editable, but only by changing `observed_on_string`
* `time_observed_at`: datetime when the observer observed the organism; user-editable, but only by changing `observed_on_string` (the reason we have both is because an observation can specify a date without specifying a time)

Locally, you probably want to track

* `_created_at`: datetime the observation was created on the device
* `_updated_at`: datetime the observation was updated on the device; you probably want to track this because you want to know what records were updated since the last sync, but you don't want to sort by this field, otherwise your list of observations will change around every time you edit an observation
* `_synced_at`: datetime the observation was last synced with the server, useful for determining which observations have never been uploaded and which ones have been edited locally and need to be updated on the server

src/providers/helpers/markUploaded.js
kueda: No, the server doesn't know anything about when a given client synced an observation. It can only tell you when the record was updated, which might have been b/c the client updated, or another client updated it, or some server process updated it, etc.

kueda (changes requested):
Ok, here's a review based on actually running the code.

* Doesn't need to be fixed now, but "Save" and "Delete" are unsettling options when backing out of editing an obs. My assumption is that I can just throw away any changes I made without saving them OR deleting the observation. That's usually what I want to do. In Android we record whether the user made any changes to the form and only show a confirmation alert when they back out if they made any changes. We don't show an option to delete.
* When I tap on an observation that hasn't been uploaded, I should see the obs detail view, not obs edit.
* When I tap on an observation that hasn't been uploaded, I immediately get the "cancel creating observations?" slideup for some reason.
* When I import a photo in the iOS simulator, it just shows as a black square on my obs list. After I upload it the photo shows.
* When I make an obs without evidence, the back button in the upper left does nothing
* `observation.description` is getting saved and submitted to the server, but never displayed in the obs edit form
* This happens when I try to set a date: `Observation.observed_on_string must be of type 'string?', got 'object' (Mon Apr 25 2022 15:12:15 GMT-0700 (PDT))] couldn't save observation locally`

src/models/Observation.js
```diff
@@ -108,19 +147,34 @@ class Observation {
...
+      geoprivacy: "string?",
+      id: "int?",
       identifications: "Identification[]",
       latitude: "double?",
       longitude: "double?",
       observationPhotos: "ObservationPhoto[]",
-      observationSounds: "ObservationSound[]",
+      observationSound: "ObservationSound?",
+      observed_on_string: "string?",
+      owners_identification_from_vision_requested: "bool?",
```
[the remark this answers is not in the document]
albullington: Is there actually an option to submit an `observation.taxon` at the time of upload? Right now, we're submitting a `species_guess` plus a `taxon_id`, similar to what we're doing in the current iOS app.

(Fwiw, I also see that we're submitting an `iconic_taxon_id` and an `id_please` attribute on iOS, and I'm not clear on when we would want to use either of those.)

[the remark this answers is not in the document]
kueda: > Is there actually an option to submit an observation.taxon at the time of upload?

Nope, but submitting `species_guess` and/or `taxon_id` can both result in an identification getting added. IMO, we should only send `species_guess` when no taxon has been chosen and there's just the placeholder text. And as I suggested above, submitting `taxon_id` might be less desirable than submitting an identification separately.

`iconic_taxon_id` is not user-editable, and `id_please` is obsolete, so please don't post either.

tests/unit/components/ObsEdit/ObsEdit.test.js
```diff
@@ -7,6 +7,8 @@ import ObsEdit from "../../../../src/components/ObsEdit/ObsEdit";
 import ObsEditProvider from "../../../../src/providers/ObsEditProvider";
 import { ObsEditContext } from "../../../../src/providers/contexts";
 
+jest.useFakeTimers( );
```
kueda, line 10: What does this do?

kueda (changes requested):
Blockers:

* I still can't edit an observation that I just made. Repro: make an obs without evidence and add a description; tap save; tap the obs in your obs list; your description does not show and there's no option to edit.
* Can't view an obs twice. Repro: make an obs w/o evidence and save; tap the obs and you land on the obs detail view; tap back arrow in upper right to return to your obs list; tap the observation again; nothing happens, get the following error:
  ```
  WARN  Possible Unhandled Promise Rejection (id: 0):
  Error: Exception in HostFunction: Accessing object of type Observation which has been invalidated or deleted
  @[native code]
  ```


Non-blocking issues:
* We should not show the upload UI when you're not signed in
* Why is `ObsEditProvider` necessary? My gut feeling is that the ObsEdit component and the Observation model can handle most of what it's doing, and that it is either causing or could cause problems b/c it circumvents the immutable, one-way-flow of data that makes React relatively easy to work with.
* the `useUserLocation` hook probably should not mess with permissions b/c again, it seems to result in permissions getting requested at odd times. Instead, we could extend `PermissionGate` to handle iOS permissions and wrap any components that need that permission.

Inline comments are not really blockers, mostly code style issues.

albullington:
Ok, the blocking changes & some of the non-blocking changes (particularly around moving way more code into the models) should be fixed here. A few things that I didn't update:

- I didn't globalize text, since I believe all the text is placeholders that either Abhas or I created & definitely not final. I've been globalizing text that is more unlikely to change, like a single word on a button or in a header.
- I didn't remove the `ObsEditProvider`. I like having a provider here, because it makes it easy to share data between the camera/gallery/sound recorder and the ObsEdit screen & have a single source of truth for the observations array. There are alternative ways we could load observations on ObsEdit, like passing an array of observations as react-navigation params, but I tend to avoid that when possible because it usually results in warnings like the one we're currently ignoring on the `ObsDetails` screen. If we were temporarily saving observations to realm, we could also save them and pass an array of uuids to the `ObsEdit` screen via navigation params, but it doesn't really make sense to me to save anything to realm before a user actually taps save or upload.
- I agree that we should use PermissionsGate for the `useUserLocation` hook, but that seems like it can be a separate PR
- I didn't move 100% of the `saveLocalObservation` code to the Observation model, because doing so results in a require-cycle error where realm is trying to open the realmConfig schema from `index.js`, while `index.js` is also requiring the Observation model.

So I think this is ready to merge as far as blocking changes, and we might want to change or revisit some of these non-blocking changes in future PRs.

src/providers/uploadHelpers/uploadObservation.js
```diff
@@ -0,0 +1,82 @@
...
+import realmConfig from "../../models/index";
+import { getJWTToken } from "../../components/LoginSignUp/AuthenticationService";
+import ObservationPhoto from "../../models/ObservationPhoto";
+import ObservationSound from "../../models/ObservationSound";
+
+const markRecordUploaded = async ( uuid: string, type: string, response: Object ) => {
+  const { id } = response.results[0];
+
+  try {
+    const realm = await Realm.open( realmConfig );
```
kueda, line 15: If you're talking to Realm this probably belongs in the model.

src/providers/uploadHelpers/uploadObservation.js
```diff
@@ -0,0 +1,82 @@
...
+const uploadToServer = async ( params, options, uuid, type, apiCall ) => {
+  try {
+    const response = await apiCall.create( params, options );
+    await markRecordUploaded( uuid, type, response );
+  } catch ( e ) {
+    console.log( JSON.stringify( e.response ), `couldn't upload ${type}` );
+  }
+};
+
+const createParams = ( response, options, evidence, mapLocalModelForUpload, type, apiCall ) => {
```
kueda, line 35: This doesn't create params, does it? What is it supposed to do?

src/providers/uploadHelpers/uploadObservation.js
```diff
@@ -0,0 +1,82 @@
...
+    const response = await inatjs.observations.create( uploadParams, options );
+    await markRecordUploaded( obsToUpload.uuid, "Observation", response );
+    if ( localObs.observationPhotos ) {
+      createParams(
+        response,
+        options,
+        localObs.observationPhotos,
+        ObservationPhoto.mapPhotoForUpload,
+        "ObservationPhoto",
+        inatjs.observation_photos
```
kueda, line 64: This isn't really an `apiCall`. It's an endpoint.

kueda (changes requested):
Blockers
* I still can't edit the description I added to an obs. If I make an obs with a description, save it, and edit it, the description I added is not there to be edited.
* Photos taken in Android generally don't appear in the app and I see a lot of `source.uri should not be an empty string` errors, which may be related. E.g if I use the camera, the image does not appear in the top preview area of the camera, nor does it appear in the group screen, nor does it appear on obs edit, nor does it appear in my obs list, obs detail, etc. If I import from the gallery, I can see the photos in the group screen, but no where else.

Not blocking (in that they don't seem directly related to the scope of work)
* I'm seeing a lot of repeated method names and logic in the models that I think imply the need for a parent class like INatModel that defines methods like `new` and `save` and requires subclasses to implement them
* Android: if you choose photos in the gallery and get to ObsEdit, then use the back button in the header, you end up in the camera after backing out of the PhotoGallery
* If you choose photos in the gallery, then back out to your ObsList, then choose to import photos from the gallery again, the same photos are selected; IMO none of the data in the ObsEditContext should stick around after you exit a flow for adding observations, and the context itself should be probably be destroyed.
* On the web, we upload photos and sounds before uploading an obs... which isn't currently supported in v2, but we will probably want to support that and do that in the app eventually.

src/providers/helpers/saveLocalObservation.js
```diff
@@ -0,0 +1,32 @@
...
+    } );
+    const newLocalObs = realm.objectForPrimaryKey( "Observation", currentObs.uuid );
+    return newLocalObs;
+  } catch ( e ) {
+    console.log( e, "couldn't save observation locally" );
+    return null;
+  }
+};
+
+export default saveLocalObservation;
```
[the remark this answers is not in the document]
kueda: @albullington you said

> I didn't move 100% of the saveLocalObservation code to the Observation model, because doing so results in a require-cycle error where realm is trying to open the realmConfig schema from index.js, while index.js is also requiring the Observation model.

I think there's a way to deal with this but I need to think about it, so let's consider this non-blocking.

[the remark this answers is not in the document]
kueda: I still don't think this needs to block, but here are the results of some further thinking:

* The [`realm-context` branch](https://github.com/inaturalist/iNaturalistReactNative/compare/save-observations...realm-context) shows how we could use a context to provide a single Realm instance to the entire app. This gets around the require cycle error and centralizes the opening and closing of Realm, which seems a bit DRYer to me, but it also involves passing this realm instance to the models or to helpers every time they need to read / write data, which feels very awkward.
* The [`realm-in-models` branch](https://github.com/inaturalist/iNaturalistReactNative/compare/realm-context...realm-in-models) is even messier, but it extends `realm-context` by storing a reference to realm as a class variable of a superclass to all models, allowing them all to interact with Realm and to abstract Realm away from the components (which may or may not be good, but it certainly allows the models to perform their duty as gatekeepers between the application and the data). This is very loosely based on ActiveRecord in Rails, and most of what's interesting is in the `Record` and `Observation` models.

While doing this, I also realized that [Realm actually provides its own `RealmProvider`](https://www.mongodb.com/docs/realm/sdk/react-native/use-realm-react/#using-the-realm-provider) and some associated helpers that we might want to use. The former is basically the same as my `RealmProvider`, I think.

I'm also still pretty suspicious of the `ObsEditContext` and practices like adding a new observation to the context before navigating to `ObsEdit` which then accesses that obs from the context. In my mind, navigation should just navigate, and `ObsEdit` should be responsible for making. new observation or loading an existing one. Again, doesn't need to block this PR, but we should think about separation of concerns after we merge.

src/providers/uploadHelpers/uploadObservation.js
albullington: Would it be more accurate to say this is flattening parameters with the FileUpload class? Maybe `flattenParams` is a better name here?

src/components/ObsDetails/ObsDetailsHeader.js
```diff
@@ -0,0 +1,54 @@
...
+      const isLocalObservation = await openSavedObservation( observationUUID );
+      setIsLocal( isLocalObservation );
+    };
+
+    navigation.addListener( "focus", ( ) => {
+      checkForLocalObservation( );
+    } );
+  }, [observationUUID, openSavedObservation, navigation] );
+
+  const goBack = ( ) => navigation.goBack( );
```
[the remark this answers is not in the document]
albullington: My understanding of React performance is that it's generally better to keep function definitions outside of render. Otherwise, the function will be recreated anytime the component is re-rendered. ([Here's](https://stackoverflow.com/questions/41369296/react-functions-inside-render) a convo on this on Stack Overflow). Maybe it doesn't impact performance much with a function this simple, but I've been thinking of it as best practice to keep all functions out of render (and a bit cleaner to read imo).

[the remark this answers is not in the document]
kueda: For a functional component like this, everything in the function is run on render, so even if you define it like this it will get redefined every time the component gets rendered... or am I missing something? If you want to avoid defining the function on render, you'd need to define it outside of the component (I guess it would take `navigation` as an arg), or switch to a class-based component and define it as a method in the class. Both of those approaches separate definition from use, which is what I was taking issue with here, but if there's a performance benefit, that seems worth it.

src/components/ObsDetails/ObsDetailsHeader.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <View style={viewStyles.headerRow}>
+      <HeaderBackButton onPress={goBack} />
+      <Headline>{t( "Observation" )}</Headline>
+      {isLocal ? <Button icon="pencil" onPress={navToObsEdit} color={colors.gray} /> : <View />}
```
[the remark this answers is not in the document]
albullington: I'm using the `<View />` here for styling purposes, since the back button, headline, and button are all in a row using `justifyContent: "space-between"`. This ensures that the headline doesn't move to the right side of the screen when there is no Button component. I think once we have final designs it will make more sense to use absolute positioning for each of these header components, I just didn't want to make random guesses on margins & positioning.

src/models/Observation.js
```diff
@@ -1,10 +1,101 @@
...
+  }
+
+  static async createObsFromNormalCamera( photos ) {
+    // take the observed_on_string time from the first photo in an observation
+    const observedOn = formatCameraDate( photos[0].metadata["{Exif}"].DateTimeOriginal );
+    const obsPhotos = await Observation.createObsPhotos( photos, ObservationPhoto.formatObsPhotoFromNormalCamera );
+
+    return await Observation.createObsWithPhotos( obsPhotos, observedOn );
+  }
+
```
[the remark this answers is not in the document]
albullington: I'm confused again. I thought the goal was to remove the logic of creating an observation from the NormalCamera and put it in the model instead. Are you saying that the NormalCamera *should* be creating an observation and passing that to ObsEdit? I'm unclear on where you want `const observation = Observation.new( );` to live.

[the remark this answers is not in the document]
kueda: K, sorry about that. I think I am over-scoping this PR so let's just ignore this issue for now. I'll just resolve this thread.

That said, here's where I'm coming from: I think it's good to have *reusable* CRUD code in the models, enough so that they provide tools to make it easier to work with that data in numerous contexts, but not so much that you're essentially dropping single-use methods in the model. Something like `createMutipleObsFromGalleryPhotos()` is very, very specific and should probably either not be a method or be defined closer to where it gets used, but something like `new()` that creates a new record with reasonable defaults or the methods you have that map between the Realm schema and the API schema are highly reusable and composable.

So, when I see stuff in, say, `NormalCamera` that's setting attributes like `_updated_at`, that seems like it should live in the model b/c we want to set that attribute *every* time we save a record, but when I see a method like `createMutipleObsFromGalleryPhotos()` in the model I wonder why it's there b/c I can't imagine it will get reused.

As to where a method like `new()` would be defined, that's part of what I was exploring in the `realm-in-models` branch, e.g. in the [Record](https://github.com/inaturalist/iNaturalistReactNative/blob/realm-in-models/src/models/Record.js) class, where there's a `new()` method that all model instances inherit and can override, like [Observation](https://github.com/inaturalist/iNaturalistReactNative/blob/realm-in-models/src/models/Observation.js#L36) does. `Record.new()` literally just makes a new instance with the properties in the arg it receives, but `Observation.new` also does things like set the default geoprivacy.

src/providers/uploadHelpers/uploadObservation.js
kueda: Just reading this method, it looks like it loops over an array of API response results and then uploads photos or sounds to the server with `uploadToUserver`, which doesn't seem like creating params or flattening params... but I'm guessing that's *not* what this method does, so back to my question: what does it do?

src/components/ObsDetails/ObsDetailsHeader.js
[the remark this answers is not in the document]
kueda: K. Definitely not a blocker so I'll resolve. FWIW, if we're going to use react-native-paper, we should consider using https://callstack.github.io/react-native-paper/appbar-header.html for all headers like this.

src/providers/uploadHelpers/uploadObservation.js
albullington: It's looping over either the local observation photos or observation sounds attached to an observation, then using either `mapPhotoForUpload` or `mapSoundForUpload` to return flattened parameters, then uploading the respective photo or sound.

kueda: Ok, if the ultimate action is to actually upload the evidence, maybe call it `uploadEvidence` and change the method signature to something like

```
uploadEvidence(
  evidence: array,
  type: string,
  apiSchemaMapper: function,
  baseParams: object,
  apiEndpoint: function
)
```

That would make this more readable to me, at least. Flattening params is a precondition for upload but not, ultimately, the reason you would call this function.

albullington:
Ok, my two cents is that we should work on the blocking issues in a separate PR since they're both potentially large pieces of work.

For editing the description, what we ultimately need is new functionality to allow a user to save a modified observation into Realm. I'd like to do this in a separate PR so I don't break the existing save functionality by mistake.

For Android photos not showing up on ObsEdit, what's happening is that the image resizer library is not able to recognize the file path for Android from either the camera or the photo gallery. This was working fine in Seek last I checked, so we need to dig in a bit to figure out how to save these images to a filepath that the image resizer can recognize. For the camera, this will likely involve saving the photo to a user's photo gallery (like we're doing in Seek), which will give back a filepath in a different format. Otherwise, we might try moving the photos into a different directory (instead of a temp directory) with RNFS.

I tried to clean up some of the code in the models, but once this is merged in we should explore using a Realm provider and/or a parent class like the INatModel to continue simplifying code.

kueda (approved):

PR #81, opened 2022-04-29
More permissions

kueda:
* Uses `PermissionGate` to manage Android permissions for the `SoundRecorder`
* Ensure the `PhotoGalleryProvider` doesn't load photos when it renders and instead only does so when a component decides it should, to prevent unexpected permission requests in iOS


src/navigation/cameraStackNavigation.js
```diff
@@ -36,6 +36,16 @@ const NormalCameraWithPermission = ( ) => (
   </PermissionGate>
 );
 
+const SoundRecorderWithPermission = ( ) => (
+  <PermissionGate permission={PermissionsAndroid.PERMISSIONS.RECORD_AUDIO}>
```
albullington, line 40: One question here: would it be better to have `PermissionGate` support [multiple permissions](https://reactnative.dev/docs/permissionsandroid#requestmultiple) instead of wrapping the SoundRecorder in three separate PermissionGate components?

That would (theoretically) pop up a single dialog instead of a user seeing audio, then storage as two separate pop ups in a row.

I guess the downside to supporting multiple permissions is that a user might deny permissions for storage by accident & then not be able to access the photo library.

albullington (commented):
Looks good to me and is working on Android -- one inline comment below.

src/navigation/cameraStackNavigation.js
kueda: Yeah, there's the denial of one and not another, and there's the issue of "priming," i.e. explaining to the user why we are asking for a permission before we do so. I haven't built that into PermissionGate yet but I think we should, and I think it makes more sense to prime one at a time. Also it seemed like the nested component approach worked fine here, so I didn't want to over-complicate the PermissionGate component.

PR #82, opened 2022-04-30
Force login screen to unmount when it loses focus, resetting state

### LINKED ISSUE #60: Reset state on login view when closed
Repro:

1. Go to the login view, enter incorrect credentials, and tap the "LOGIN" button. You should see a "Couldn't login" message
2. Tap "explore" in the tab bar
3. Tap "menu" in the tab bar and tap "login" again
4. Your previous credentials and the error message are still there; my expectation was that they would be reset

OS: Android 12
Device: Pixel 3
App version: 0.0.1

### LINKED ISSUE #61: Cannot log in after a failed attempt
Repro:

1. Go to the login view, enter incorrect credentials, and tap the "LOGIN" button. You should see a "Couldn't login" message
1. Enter the correct credentials... and nothing happens.

You have to force quit the app and enter your credentials correctly the first time in order to sign in. Probably related to #60

OS: Android 12
Device: Pixel 3
App version: 0.0.1

kueda:
Adds a <Mortal> component that only renders its children when it has focus in React Navigation, causing its children to unmount when that focus is lost, thereby resetting their state and cause all their effects to clean up after themselves.

Closes #60. Might close #61, need to test.


albullington (approved):
This looks good to me, and I learned something new. I thought using "blur" was totally valid and practical here.

kueda:
I think using the react-navigation focus and blur events are "valid" but should be used with caution. There's this issue with immortal components, but I've also noticed issues with `PermissionGate` where if two visible components listen to a react-navigation event, sometimes only one gets its callback triggered.

PR #97, opened 2022-05-17
Delete realm record on signout

albullington:
This does a few things:

- Adds a confirmation modal on the sign out button
- Deletes the realm database if a user chooses to sign out
- Wraps `myObservationsStackNavigation` in a Mortal component. This means a user can tap into Taxon Details and tap the back button to go back to their previous spot in the ObsList, instead of needing to scroll down from the top of the list each time they go back to the list. However, ObsList will be refreshed anytime a user leaves this navigation stack (i.e. going to login/signout, then revisiting the ObsList)


src/navigation/myObservationsStackNavigation.js
```diff
@@ -24,28 +25,30 @@ const hideHeader = {
...
+        component={UserProfile}
+        options={hideHeader}
+      />
+      <Stack.Screen
+        name="TaxonDetails"
+        component={TaxonDetails}
+        options={showBackButton}
+      />
+    </Stack.Navigator>
+  </Mortal>
```
kueda, line 51: I'm surprised wrapping an entire stack like this works, but ok!

kueda (approved):
Working well for me. Using `<Mortal>` around that stack already has some performance problems with rendering and maybe some network requests, but I suspect that's actually a good thing as it will force us to deal with the underlying causes.

PR #101, opened 2022-05-18
Consolidate API URL config into .env

kueda:
It was bugging me that we specified API URLs in multiple places. This will make it easier to find where those values are. Note that you have to update your `.env` file to get this branch to work correctly. See `env.example`.


albullington (approved):
Looks good. I can get this working with staging APIs.

PR #111, opened 2022-05-26
Show comments on ObsDetail

albullington:
The most important thing this does is displays comments on ObsDetail (fixed line 18 in `ActivityTab`).

A few other things it does:
1. Makes styling for ActivityItems (ids and comments) closer to what's in Figma
2. Adds a kebab menu which is currently only accessible on comments written by the current user
3. Kebab menu drops down a "Delete comment" option
4. Delete comment deletes the comment from the realm database, then deletes the comment remotely, then rerenders the ObsDetail screen (we'll probably need a more complex implementation for this, but that seemed out of scope)
5. Fetches comment uuids from the server


kueda (changes requested):
I wish the manual styling / fixed-width menu wasn't necessary, but I guess we can live with it for now.

Some issues in this screenshot:

Screen Shot 2022-05-31 at 3 10 03 PM

- [ ] Menu content should be above everything else, including comment text
- [ ] If possible, the menu should effectively be modal, i.e. any taps outside of the menu close that menu; that should prevent opening multiple menus

albullington:
I'm not seeing a great way to do either of these things. For the menu content, I moved the menu up so it's now on top of the header instead of the comment text. That makes it look better for the moment, but if we end up needing a menu with more than one item in it, it will still sit below the comment text.

There doesn't seem to be a way for the menu to be modal using the `react-native-paper` library, and I looked around for other popup menu libraries but they all seem to assume a user will only try to open one menu at a time. The way to dismiss these menus is by tapping a second time on the ellipses button used to open the menu, rather than the background screen. I think we'll ultimately need to build our own solution along [these lines](https://hartaniyassir.medium.com/how-to-create-a-popup-menu-in-react-native-d2fc8908e932), where we create a popup menu that is anchored to a particular element but is also a modal.

Maybe there's a way to adjust the design here so a user can see a single dialog in the middle of the screen, which they would need to dismiss before they can try to open any other menu options?

kueda (approved):
This works for now, but figuring out how to do a popover properly is definitely something we need to figure out. I don't think we need a different design, as popovers seem pretty conventional and as you indicated in your link, they are possible in RN.

Issue #99, opened 2022-05-17
Don't use a user JWT if the user is signed out


kueda:
Since we use `react-native-sensitive-info`, the user's JWT gets stored in shared preferences outside of the app, at least in iOS. That means that if the user signs in, deletes the app, and installs the app again, the app again has access to that old JWT even though the user is not signed in, and will use it to fetch observations and updates.

Instead, if the user is not signed in, the app should ignore any JWT it gets from the operating system. It should actually delete the JWT if it encounters one while not signed in.

In addition to not creeping out users, this will make developing a bit easier.

PR #112, opened 2022-05-31
Delete user JWT on 401 error

albullington:
This treats the user like they are signed out when encountering a 401 unauthorized error while fetching the JWT.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -536,9 +536,9 @@ SPEC CHECKSUMS:
   FBLazyVector: f7b0632c6437e312acf6349288d9aa4cb6d59030
   FBReactNativeSpec: 0f4e1f4cfeace095694436e7c7fcc5bf4b03a0ff
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 5337263514dd6f09803962437687240c5dc39aa4
+  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
   Permission-LocationWhenInUse: 006c85c8de0c05b5d8be8e8029e4f6b813270293
-  RCT-Folly: a21c126816d8025b547704b777a2ba552f3d9fa9
+  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
   RCTRequired: 0aa6c1c27e1d65920df35ceea5341a5fe76bdb79
   RCTTypeSafety: d76a59d00632891e11ed7522dba3fd1a995e573a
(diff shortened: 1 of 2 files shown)


kueda (changes requested):
This doesn't address the use case described in #99, namely that if you delete the app then reinstall the app, you're already signed in b/c the JWT is getting reused. I guess there are a few ways to deal with that, but my preference would be for the app to store some information about whether or not the user is signed in independent of whether it gets a JWT from outside the app.

albullington:
Ok. Right now, we're storing the info about whether the user is signed in or not in `react-native-sensitive-info`, so that is likewise not getting deleted when a user deletes the app & then reinstalls.

A couple ways I can think to solve this:
1) We could try to fetch a JWT token on app startup, so we can see if a user is actually signed in or not before showing them any content
2) We could store the username and/or accessToken in app storage that gets deleted when the app is deleted, so we won't keep a user signed-in on reinstall. Maybe a `User` object in Realm or a key-value pair in [react-native-mmkv-storage](https://github.com/ammarahm-ed/react-native-mmkv-storage)?

kueda:
IMO, the JWT is secure info so it seems reasonable to continue to use `react-native-sensitive-info` to store it. However, I think we need to store an additional piece of info that gets deleted with the app to indicate if the user is signed in, like the user ID of the currently-signed-in user. If the app starts and that value is blank, it should erase all the stuff we were storing with `react-native-sensitive-info`.

albullington:
Here's another pass. This stores the `userId` in `react-native-mmkv-storage` instead of `react-native-sensitive-info`, which means `userId` will be null on the first install of the app. When `userId` is null, we sign a user out, which includes deleting the realm file from any previous installs of the app. On ObsList, we check for a `userId` before opening realm -- otherwise a user might see realm observations from a previous install of the app before they are fully signed out & before their realm file is fully deleted.

Tests currently aren't passing. I tried [mocking `react-native-mmkv-storage` with jest](https://rnmmkv.vercel.app/#/mockjest?id=testing-with-jest), but this seemed to cause other errors with the Realm schema not being found. I believe the next step here would be to figure out how to mock two in-memory storage solutions simultaneously. Or, we could try using Realm instead of `react-native-mmkv-storage` to store the `userId`.

kueda (approved):
Aside from that one loading bug, I think this works and can be merged. I can get the tests to pass with a few additional mocks.

That said, I don't think adding an additional data store is worth the maintenance / dependency headache, so I'd like to explore doing this with Realm. I may look into that, but if you want to merge this after tests pass, that's fine with me too.

PR #117, opened 2022-06-25
Fix search results for Explore Dropdowns, Projects, Search

albullington:
Fixes crash that was happening due to /v2/search being updated to return type-specific records.

Also allows for only one dropdown to be displayed at a time on Explore.


kueda (approved):
Looks good, fixes the problems in the issue, at least.

PR #119, opened 2022-06-29
Logged out state for MyObservations

albullington:
- UI for login prompt in header and persistent bottom sheet. Both allow a user to navigate to the login screen
- Match the UI in the upload prompt to the UI in the login prompt bottom sheet
- Allow a user to save an observation locally, even when logged out (small bug fixes here)
- Check for first app launch and store this information in `AsyncStorage`. This is necessary so we can determine whether or not to delete any pre-existing realm files -- mostly in the case where a user previously installed the app, then deleted it, then opened MyObservations via a fresh app install and expected to see no data. With previous code, the realm file would be deleted every time a logged out user opened the app


kueda (changes requested):
Mostly looking good, just a few minor changes. Also, the "Welcome to iNat" text on the empty state of My Observations looks too tall on my iPhone 13 simulator. IMO, hard-coding the height like we're doing is pretty brittle, but actually styling the FlatList `ListEmptyComponent` also seems to cause trouble (still a deeply-nested component, hard to control flex context). Might be better to just not render the FlatList for the empty state.

albullington:
Ok, I think everything mentioned should be fixed. One thing I noticed when I went to fix `<Button color={...}>` is that we're not being consistent in whether we're using the Button component from `react-native` or the Button component from `react-native-paper`. I'm assuming we'll eventually want all our Buttons to be from `react-native-paper`, so that might be worth addressing in a separate issue.

kueda (changes requested):
Code looks fine, but now the logged out state doesn't show any text for me in the iPhone simulator, just the green banner at the top.

src/components/SharedComponents/BottomSheet.js
```diff
@@ -3,17 +3,17 @@
...
-const BottomModal = ( { children, height }: Props ): Node => (
-  <View style={[viewStyles.bottomModal, { height }]}>
+const BottomSheet = ( { children, height }: Props ): Node => (
+  <View style={[viewStyles.bottomSheet, { height }]}>
     {children}
   </View>
 );
 
-export default BottomModal;
+export default BottomSheet;
```
kueda, line 19: You might consider re-doing this with React Native Bottom Sheet, a 3rd-party dep we just added as a part of the Add ID component. At a mininum we could have consistent styling, but if the bottom sheet for the logged out state ever needs more interactivity that might make it easier to support.

albullington:
Updated to use React Native Bottom Sheet and fixed styling for EmptyList.

kueda (approved):
Looking good. I'm not sure if that bottom sheet is supposed to be modal in the signed out state, but since the designs don't make that clear, let's just go with what's here for now.

PR #120, opened 2022-06-29
Enforce globalized text with eslint plugin

albullington:
This uses the `eslint-plugin-i18next` plugin to enforce globalized text (i.e. no string literals in a `<Text>` component).

I set the eslint plugin to warning mode instead of error mode for the time being, mostly because we have a lot of text as placeholders for image assets that won't need to be translated, and we also have a lot of text in screens like CVSuggestions which have had complete design overhauls since we first built these components. I'm thinking we can slowly pick away at these over time, but maybe there's a better approach.

All the Settings components also have text that needs to be translated, and I'm assuming we can tackle that sooner since the text will likely be similar to what's on the web.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,10 +6,7 @@ module.exports = {
     "comma-dangle": [2, "never"],
     "space-in-parens": [2, "always"],
     "prettier/prettier": 0,
-    "no-var": 1,
-    // TODO: remove this line once we have image assets and no longer need
-    // placeholder text; better to enforce globalized text with errors, not warnings
-    "i18next/no-literal-string": 1
+    "no-var": 1
   },
   // need this so jest doesn't show as undefined in jest.setup.js
   env: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,12 +1,15 @@
 module.exports = {
   root: true,
-  extends: "@react-native-community",
+  extends: ["@react-native-community", "plugin:i18next/recommended"],
   rules: {
     quotes: [2, "double"],
     "comma-dangle": [2, "never"],
     "space-in-parens": [2, "always"],
     "prettier/prettier": 0,
-    "no-var": 1
+    "no-var": 1,
+    // TODO: remove this line once we have image assets and no longer need
+    // placeholder text; better to enforce globalized text with errors, not warnings
+    "i18next/no-literal-string": 1
   },
   // need this so jest doesn't show as undefined in jest.setup.js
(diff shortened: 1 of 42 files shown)


src/components/SharedComponents/ObservationViews/EmptyList.js
```diff
@@ -11,8 +12,7 @@ const EmptyList = ( ): Node => {
 
   return (
     <View style={viewStyles.center}>
-      <Text style={textStyles.text} testID="ObsList.emptyList">welcome to inaturalist!</Text>
-      <Text style={textStyles.text}>make sure you're logged in to fetch observations</Text>
+      <Text style={textStyles.text} testID="ObsList.emptyList">{t( "iNaturalist-is-a-community-of-naturalists." )}</Text>
```
kueda, line 15: This causes a crash when signed out, should be `iNaturalist-is-a-community-of-naturalists` (no period)

.eslintrc.js
```diff
@@ -1,12 +1,15 @@
...
   rules: {
     quotes: [2, "double"],
     "comma-dangle": [2, "never"],
     "space-in-parens": [2, "always"],
     "prettier/prettier": 0,
-    "no-var": 1
+    "no-var": 1,
+    // TODO: remove this line once we have image assets and no longer need
+    // placeholder text; better to enforce globalized text with errors, not warnings
+    "i18next/no-literal-string": 1
```
kueda, line 12: I'm concerned that we'll do this and never actually start enforcing this, so instead of making this a warning, why don't we make it an error and either remove all placeholder text replace all it all with something like `<PlaceholderText text="this is just a placeholder" />` and that component is just something like 

```javascript
const PlaceholderText = ( { text } ) => (
  // eslint-disable-next-line react-native/no-inline-styles
  <Text style={{ color: "red", textTransform: "uppercase", fontSize: 30 }}>
    { text }
  </Text>
);
```

That way if you *really* want to add placeholder text you can, but you have be more self-conscious about it, and we can find and remove it a little more easily. Very tempted to install [`react-blink-tag`](https://www.npmjs.com/package/react-blink-tag), but that's probably overkill, if amusing.

kueda (changes requested):

albullington:
Round 2: plugin enforces errors instead of warnings. All 100+ errors were updated either with the new PlaceholderText component, with globalized strings, or with an icon/avatar from `react-native-paper`.

kueda (approved):

Issue #105, opened 2022-05-19
Obs Detail add comment UI


kueda:
I realize there's some placeholder UI there now, but I'd like it to look and work a bit more like the wireframe:

![Obs Detail - Add a comment](https://user-images.githubusercontent.com/22691/169173996-c7afa084-7c7e-4a02-b0f8-f55dd4c72119.png)

When the user taps the Comment button on obs detail

- [x] open a modal bottom sheet with a text area for the comment and a button to submit the comment
- [x] in theory you should not need to disable the "Comment" button b/c the sheet will be modal (no interaction outside of the sheet)
- [x] dismissing the modal without tapping the submit button should open a confirmation alert with the text "Are you sure you want to discard this comment?"
- [x] submitting the comment closes the sheet and shows a spinner at the bottom of the activity section while the comment is submitted and the observation reloads

I'm not 100% sure using a bottom sheet will work correctly with the keyboard here, but I figured it was worth trying.

PR #127, opened 2022-07-10
#105 - obs details add comment UI

budowski:


kueda (changes requested):
Another reminder, please name your branches starting with the issue number, if one exists, so `105-obs-add-comment` instead of `obs_add_comment`. Doesn't need to change here but please keep it in mind for the future.

This mostly looks pretty good, particularly on Android. Here are some concrete todos:

- [ ] I'm getting a flicker every time I type a letter in the comment text box, as if the entire bottom sheet is re-rendering on every key press. Can we fix that? On the web I'd use an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html), though I'm not sure what the equivalent is in RN
- [ ] When I try to save a comment I get ` ERROR  TypeError: undefined is not an object (evaluating 'observation.observationPhotos.map')` at [ObsDetails.js:128](https://github.com/inaturalist/iNaturalistReactNative/pull/127/files#diff-f3017450aeb5c7dc7b74f3e729e38fce3460974a5a5344677a088d935d568e82R128)
- [ ] The keyboard covers the entire bottom sheet in iOS. You might need to follow the guidance at https://gorhom.github.io/react-native-bottom-sheet/keyboard-handling/

kueda (changes requested):
Ok, still some conflicts to resolve, but the layout for the comment box in iOS is now weird: too tall, and doesn't seem to have left and right padding. Here's Android on the left and iOS on the right for comparison:

Screen Shot 2022-08-01 at 6 20 05 PM

kueda (changes requested):
In addition to the need to merge from main and resolve conflicts,

1. The comment text input is still too tall, particularly in iOS. Given the mock at https://github.com/inaturalist/iNaturalistReactNative/issues/105, it should be 2-3 lines of text tall.
2. In iOS, tapping outside the bottom sheet shows an alert asking if you want to discard the comment, but if you say "Yes" the bottom sheet remains in place and the keyboard re-appears (you may have to uncheck I/O / Keyboard / Connect Hardware Keyboard)

kueda (approved):

PR #128, opened 2022-07-13
UI updates to bulk uploader

albullington:
This updates the UI for Camera Roll, Group Photos, and Obs Edit. It also adds some new features like deleting an observation from the bulk uploader flow on Obs Edit.


src/components/PhotoLibrary/GroupPhotoImage.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const unselectedIcon = ( ) => (
+    <Icon name="radiobox-blank" color={colors.white} size={35} style={viewStyles.selectionIcon} />
+  );
+  const selectedIcon = ( ) => (
+    <Icon name="check-circle" color={colors.inatGreen} size={35} style={viewStyles.selectionIcon} />
+  );
+  const numberOfPhotosIcon = ( ) => (
+    <FilterIcon
+      // $FlowFixMe
```
kueda, line 43: For situations like this where it's complaining about legitimate use of a library with a weird flow config, let's use `$FlowIgnore`, since we don't really have the ability to fix the flow issue.

kueda (changes requested):
Cool, this makes these screens look a lot nicer. Some todos:

- [ ] Could you center the items in the Obs Edit header on the vertical axis as well, please?
- [ ] `Yes-delete-observation` doesn't appear to be localized in English (you probably need to run the localization scripts... maybe that should be a pre-commit hook?)
- [ ] The Obs Edit header menu only seems to appear when there are multiple observations; let's add it for all observations, b/c presumably you might want to delete a new observation even if there's only one
- [ ] For me in the iPhone simulator, the keyboard pops up for the "Notes" field but it pushes just the field above the keyboard, as if both are floating above everything else, and sometimes it pushes it way above the keyboard. Maybe we should handle this in another issue, but something to consider at least.

PR #130, opened 2022-07-25
Upload progress bar

### LINKED ISSUE #92: Upload progress on My Observations
- [x] When you start a manual upload, the bottom sheet should display the progress of your upload using a progress bar while the upload is in progress
- [x] Tapping a button in the bottom sheet while the upload is in progress halts the upload.

Home Screen - Upload Observations

albullington:
Closes #92

### DIFF
--- a/src/providers/uploadHelpers/uploadObservation.js
+++ b/src/providers/uploadHelpers/uploadObservation.js
@@ -55,7 +55,10 @@ const uploadEvidence = (
   }
 };
 
-const uploadObservation = async ( obsToUpload: Object, localObs: Object ) => {
+const uploadObservation = async (
+  obsToUpload: Object,
+  localObs: Object
+): Promise<string> => {
   try {
     const apiToken = await getJWTToken( false );
     const options = { api_token: apiToken };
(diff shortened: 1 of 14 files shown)


src/providers/uploadHelpers/uploadObservation.js
```diff
@@ -86,8 +102,10 @@ const uploadObservation = async ( obsToUpload: Object, localObs: Object ) => {
         inatjs.observation_sounds
       );
     }
+    return "success";
   } catch ( e ) {
-    console.log( "couldn't upload observation: ", e );
+    console.log( "couldn't upload observation: ", JSON.stringify( e.response ) );
+    return "failure";
```
kueda, line 108: IMO, this `catch` block is catching too much. The part of the above where we're not really certain about whether an exception will occur is just `await inatjs.observations.create`. Maybe the network's spotty, maybe the server is having problems, maybe the server responds with an error message, etc., so we should try/catch around that because the app can't really control for those conditions. But we *can* control pretty much everything else, so we should not try/catch around that stuff b/c that just obscures exceptions from bugs we can and should fix.

PR #133, opened 2022-08-01
Camera shutter effect and thumbnails

albullington:
- Adds a thumbnail and loading wheel to camera preview when a user presses the shutter button
- Creates an animated view overlaying the camera, which fades in and out on shutter press


kueda (changes requested):
Haven't done a full read of this yet, but the first issue that occurs to me is that the camera doesn't fade back in until after the preview icon finishes loading. Can we separate those two processes? IMO, the camera view should just fade in and out very quickly, enough to provide an immediate response to the user's tap on the shutter. The preview generation can take as long as it needs to take, but it shouldn't hold up the process of taking more photos.

albullington:
The camera animation and the preview icon should be unrelated processes. Right now, the camera animation is being triggered by the shutter press. The fade out lasts for a certain duration in milliseconds, then when the fade out is complete, the fade back in is triggered for the same duration. I tried changing the duration to 5 milliseconds to see if that feels more like the duration you're expecting.

albullington:
I *think* this should be working better now.

src/components/Camera/FadeInOutView.js
```diff
@@ -0,0 +1,53 @@
...
+  useEffect( ( ) => {
+    if ( savingPhoto ) {
+      Animated.sequence( [
+        Animated.timing( fadeAnimation, fade( 1 ) ),
+        Animated.timing( fadeAnimation, fade( 0 ) )
+      ] ).start( );
+    }
+  }, [savingPhoto, fadeAnimation] );
+
+  const heightPhotoContainerCamera = 134;
```
kueda, line 36: Does this need to be hard-coded? It seems to get repeated [here](https://github.com/inaturalist/iNaturalistReactNative/pull/133/files#diff-ecca3df0533d935137cea989efddaefb2f009aeca3b7f14c8ffd4cf7ca1216ceR85) so maybe it should be extracted, or be some kind of constant in the style.

src/components/Camera/PhotoPreview.js
```diff
@@ -44,7 +48,11 @@ const PhotoPreview = ( { photoUris, setPhotoUris }: Props ): Node => {
   };
 
   const emptyDescription = ( ) => (
-    <Text style={textStyles.topPhotoText}>
+    <Text style={[
+      textStyles.topPhotoText, {
+        bottom: 16 + insets.top
```
kueda, line 53: Kind of the same thing here. You could just allow the style to set this value once with `bottom: textStyles.topPhotoText.bottom + insets.top`

src/components/Camera/PhotoPreview.js
```diff
@@ -62,7 +70,7 @@ const PhotoPreview = ( { photoUris, setPhotoUris }: Props ): Node => {
         <Modal
           visible={mediaViewerVisible}
           onDismiss={hideModal}
-          contentContainerStyle={viewStyles.container}
+          // contentContainerStyle={viewStyles.container}
```
kueda, line 73: Please delete this comment and maybe `viewStyles.container` unless we're actually going to use it.

src/components/MediaViewer/MediaViewer.js
```diff
@@ -4,15 +4,15 @@ import { HeaderBackButton } from "@react-navigation/elements";
...
+import { Dimensions, Image, View } from "react-native";
 import ImageZoom from "react-native-image-pan-zoom";
 import { Appbar, Button } from "react-native-paper";
 
 import Photo from "../../models/Photo";
 import { imageStyles, viewStyles } from "../../styles/mediaViewer/mediaViewer";
 import DeletePhotoDialog from "../SharedComponents/DeletePhotoDialog";
 import PhotoCarousel from "../SharedComponents/PhotoCarousel";
-import ViewNoFooter from "../SharedComponents/ViewNoFooter";
+// import ViewNoFooter from "../SharedComponents/ViewNoFooter";
```
kueda, line 15: Another comment to remove.

src/components/MediaViewer/MediaViewer.js
```diff
@@ -4,15 +4,15 @@ import { HeaderBackButton } from "@react-navigation/elements";
...
 
 const { width, height } = Dimensions.get( "screen" );
 const selectedImageHeight = height - 350;
```
kueda, line 18: I know this isn't a part of this PR, but this is another literal that could probably be calculated from styles or made a constant or something. Usually when I see something like this I think, "Why 350? What's this going to look like on a tablet?"

kueda (changes requested):
Mostly looking good, and thanks for all the back-and-forth about the shutter duration. These remaining requests are all pretty minor.

PR #136, opened 2022-08-01
Add swipe to media viewer

### LINKED ISSUE #43: Basic Media Viewer
The Media Viewer allows the viewer to view a collection of photos and/or sounds, generally associated with an observation. The user can enter the Media Viewer from the Normal Camera by tapping one of the photos taken, or from Obs Edit, by tapping a photo or a sound.

This issue addresses the basic functionality required for the [Observation Loop 1 milestone](https://github.com/inaturalist/iNaturalistReactNative/milestone/1), but we'll list the rest of the functionality for future reference

Functionality:
- [x] View a collection of photos with the selected media item in the main view and the others visible in a carousel below it
- [x] Carousel shows center-cropped thumbnails for photos; swipe left/right gestures show pan the carousel left and right
- [x] Tapping on items in the carousel shows them in the main area
- [x] Current item highlighted in the carousel with a differently-colored border
- [x] Swipe left/right gestures in the main area should navigate to the next/previous media item
- [x] View full sized image when selected (i.e. not a scaled down version), including zoom
- [x] Delete image or sound with confirmation alert
- [x] View and edit/delete other photos attached to the observation

[Design Figma file](https://www.figma.com/file/XmLaQzJ1McnsYQysB57hib/iNaturalist-Mobile-UX-Design?node-id=3%3A249) - look for the last screen in the multicapture camera section

Media Viewer - Photos

Future functionality:

- Show sounds as the selected media item in the main view and mixed in with other sounds and photos visible in the carousel below it
- Carousel shows sound icon for sounds
- Play a sound when in the main view
- Button to crop image using [react-native-image-crop-picker](https://www.npmjs.com/package/react-native-image-crop-picker) (or another, or custom, if preferred), which crops the image, saves it, and returns the user to the Media Viewer

albullington:
- Lets a user swipe between photos in the bottom carousel when the main photo is zoomed out
- Fixes some styling issues; makes the media viewer full screen on iPhone X devices

Closes #43


kueda (changes requested):
No problem with the code here but I'm pretty sure this is not the kind of swipe we're shooting for. I was thinking something more along the lines of https://www.npmjs.com/package/react-native-swiper, where you can see both the current photo and the one you're swiping to at the same time.

albullington:
@kueda I still have some work to do related to tapping a photo in the carousel, having the FlatList scroll to the correct spot, & updating the green selection box around the photo in the carousel (like, making this less jittery). But otherwise, does this look more like the type of swiper you were originally thinking of? Is there anything else we would need to do to make this work more like `react-native-swiper`? Want to make sure this is at least on the right track.

kueda (commented):
Totally on track in terms of functionality! I like.

albullington:
Ok, ready for re-review.

src/components/MediaViewer/HorizontalScroll.js
```diff
@@ -0,0 +1,100 @@
...
+        data={photoUris}
+        renderItem={renderImage}
+        initialScrollIndex={selectedPhotoIndex}
+        getItemLayout={getItemLayout}
+        horizontal
+        pagingEnabled
+        showsHorizontalScrollIndicator={false}
+      // $FlowIgnore
+        ref={horizontalScroll}
+        onScrollEndDrag={handleScrollEndDrag}
```
kueda, line 82: This approach is causing this bug: if you touch down, slide to the next image but don't release, then slide back to the original image, then release, this even seems to fire before the glide animation finishes, which means the `PhotoCarousel` highlights the next image while the `FlatList` is still showing the first.

I can sort of address this by using [`onMomentumScrollEnd`](https://reactnative.dev/docs/scrollview#onmomentumscrollend) instead ([thanks, SO](https://stackoverflow.com/questions/43370807/react-native-get-current-page-in-flatlist-when-using-pagingenabled)), but that seems to call the callback multiple times which causes jitter, and I'm not sure how to tell which event is the last without doing something unpleasant like `setTimeout`. Any ideas?

src/components/MediaViewer/CustomImageZoom.js
```diff
@@ -0,0 +1,52 @@
...
+  const onPanResponder = e => e.nativeEvent.touches.length === 2 || scaleValue.current > 1;
+
+  return (
+    <ImageZoom
+      cropWidth={width}
+      cropHeight={selectedImageHeight}
+      imageWidth={width}
+      imageHeight={selectedImageHeight}
+      minScale={1}
+      onStartShouldSetPanResponder={onPanResponder}
```
kueda, line 36: I'm nitpicking here, but I don't think there's a reason to define a method for this (unless there's some kind of performance savings?), and I might re-write it for readability like this:

```
onStartShouldSetPanResponder={e => {
  const pinching = e.nativeEvent.touches.length === 2;
  const alreadyZooming = scaleValue.current > 1;
  return pinching || alreadyZooming;
}}
```

kueda (changes requested):
This is looking really good. The only blocker is that fake-out scroll bug where the carousel and the flat list get out of sync.

albullington:
I think this should be looking better now -- changed the way `handleScrollEndDrag` determines which index to select next for the photo carousel.

kueda (changes requested):
Blocking:
- [ ] The back button in the media viewer should look the same in iOS and Android, right? It should at least not be dark gray in Android, which makes it almost invisible

Not blocking:
- [ ] If there are more photos than will fit in the PhotoPreview and you tap on the last one, there's no indication that there are more photos; you just have to try swiping. One way to address this would be size the photo previews so the last one is always half off-screen, but we should probably consult with Abhas

PR #141, opened 2022-08-04
Unviewed comments/ids in ObsList

### LINKED ISSUE #95: Show when observations have unviewed comments or IDs in My Observations
When an observation has unviewed comments and/or IDs, the comment and ID icons for that observation on My Observations should turn `#FF0000`.

- [x] Whenever observations are synced, fetch data about new updates
- [x] If an observation has new updates, the comment and ID icons should be `#FF0000` on my observations
- [x] When the user views that observation in Obs Detail, those icons should be the default color when the user views that observation again in My Observations
- [x] When the user views an observation in Obs Detail, it should tell the server that they viewed updates for that observation

I'm guessing this will require an implementation something like this:

* add an attribute to the Observation model to track whether or not the observation has been viewed; this attribute should only exist in the local database
* after the user syncs their observations, `GET /v2/observations/updates` to get a list of updates on the authenticated user's observations. More realistically this will probably look like this
  ```bash
  curl -XPOST "https://api.inaturalist.org/v2/observations/updates" \
    -H "Authorization: USERS-JWT" \
    -H 'X-HTTP-Method-Override: GET' \
    -d '{
      "per_page": 200,
      "observations_by": "owner",
      "fields": ["viewed", "resource_id"]
    }'
  ```
* for each of those observations identified by `resource_id`, set the local viewed attribute based on `viewed`

Alternatively, you could track the last time you fetched updates and use the `created_after` param to filter the updates to those created after the last fetch date, but that might be more complexity than required.

albullington:
- Code cleanup for observations hooks: split into a local observations listener hook and a remote fetching hook
- Fetch data in right order so a user can go to Explore -> back to ObsList and see the correct number and text/icon color for unviewed comments/ids
- Fixes related tests

Closes #95


src/components/Observations/TopCard.js
```diff
@@ -0,0 +1,27 @@
...
+import LoggedOutCard from "./LoggedOutCard";
+import UserCard from "./UserCard";
+
+type Props = {
+  numObsToUpload: number
+}
+
+const TopCard = ( { numObsToUpload }: Props ): Node => {
+  const isLoggedIn = useLoggedIn( );
+  if ( isLoggedIn === null ) {
```
kueda, line 18: What does it mean when `isLoggedIn` is null? I think this is definitely an improvement in that I'm not seeing the `LoggedOutCard` flash before seeing the `UserCard` when I'm logged in, but now there seem to be three potential views that could be rendered, which still results in some visible jitter in height. Maybe it would be better to render them all as children of as single component that always gets rendered and has a fixed height?

src/components/Observations/UserCard.js
```diff
@@ -9,16 +9,16 @@ import Icon from "react-native-vector-icons/MaterialCommunityIcons";
...
-}
-
-const UserCard = ( { userId, user }: Props ): Node => {
+const UserCard = ( ): Node => {
+  const userId = useCurrentUser( );
+  const { user } = useUser( userId );
   // TODO: this currently doesn't show up on initial login
   // because user id can't be fetched
   const navigation = useNavigation( );
+  if ( !user ) { return null; }
```
kueda, line 21: This might be part of the jitter. If there's no user maybe it should render a skeleton of some kind.

src/components/SharedComponents/ObservationViews/ObservationViews.js
```diff
@@ -35,22 +35,29 @@ const ObservationViews = ( {
...
 
   const navToObsDetails = async observation => {
     navigation.navigate( "ObsDetails", { observation } );
   };
 
-  const renderItem = ( { item } ) => <ObsCard item={item} handlePress={navToObsDetails} />;
+  const renderItem = ( { item, index } ) => {
+    if ( index === 0 ) {
+      // console.log( item.viewed, "first item viewed status" );
+    }
```
kueda, line 52: Does this need to be here?

kueda (approved):
Nothing blocking here as the stuff with the header is a bit out of scope for the issue, but something to consider.

Issue #102, opened 2022-05-18
Add photos to an existing observation


kueda:
- [x] Add a button in the evidence section to add new media
- [x] Tapping the button should open a bottom modal sheet with the title "Add Evidence" and with buttons to import photos or take a new photo
- [x] If the observation already has 20 photos, these buttons should be disabled and there should be a message stating "You can only add 20 photos and 20 sounds to an observation" (this will be the iNatRN version of https://github.com/inaturalist/inaturalist/issues/2624)
- [x] The bottom modal sheet should have a "Cancel" button that dismisses the sheet
- [x] Tapping outside of the bottom modal sheet should dismiss the sheet

Camera flow

- [x] Camera option should open the Standard Camera
- [x] Standard Camera should have a limit to how many photos it can take; when the limit is reached, disable the shutter button (don't hide it) and show the same message about only being able to add 20 photos (maybe use a [RN Paper Snackbar](https://callstack.github.io/react-native-paper/snackbar.html) for now)
- [x] When the user taps "Next" they land back on the same ObsEdit screen, all their existing changes intact, except with the new photos added in the evidence section _after_ the existing photos

Import flow

- [x] Camera roll screen of the Bulk Importer should open
- [x] Bulk Importer should have a limit to how many photos the user can select; when the limit is reached, additional taps do not select the photos and instead show the same message about only being able to add 20 photos (maybe use a [RN Paper Snackbar](https://callstack.github.io/react-native-paper/snackbar.html) for now)
- [x] Tapping "Import X Photos" just returns the user to ObsEdit with those photos added to the end of the evidence section, i.e. it skips the grouping step

Wireframes

![Obs Edit - Single - Sound](https://user-images.githubusercontent.com/22691/169116055-7f5c135b-9fbf-4301-94ee-c0d6e60c87cc.png)
![Screen Shot 2022-05-18 at 11 15 29 AM](https://user-images.githubusercontent.com/22691/169116151-bc61e4bc-8b21-43f7-9bb6-c66c3954ebb7.png)

PR #145, opened 2022-08-07
#102 - obs edit - take photo + import photos

budowski:


flow-typed/npm/react-native-image-picker_vx.x.x.js
```diff
@@ -0,0 +1,18 @@
...
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module "react-native-image-picker" {
+  declare module.exports: any;
+}
```
kueda, line 18: Were you getting flow errors without this?

src/i18n/strings.ftl
```diff
@@ -593,3 +593,6 @@ Yes-delete-photo = Yes, delete photo
 # Message shown when a permission is required to use a part of the app
 # (e.g. permission to access the camera) but the user denied the permission.
 You-denied-iNaturalist-permission-to-do-that = You denied iNaturalist permission to do that
+
+Add-evidence = Add evidence
+You-can-only-upload-images = You can only add 10 photos and 10 sounds to an observation
```
kueda, line 598: FWIW, Crowdin updates tend to work best when keys change, not values, so my preference is generally for keys that are fairly specific to their content, like `You-can-only-upload-20-media` or something.

src/components/ObsEdit/ObsEdit.js
```diff
@@ -131,10 +148,54 @@ const ObsEdit = ( ): Node => {
...
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...props}
+      pressBehavior="close"
+      appearsOnIndex={0}
+      disappearsOnIndex={-1}
+    />
+  );
+
+  const onImportPhoto = async () => {
+    const response = await launchImageLibrary( {
```
kueda, line 168: This should open the Camera Roll component that you see when you create an observation by importing photos, not `launchImageLibrary`

kueda (changes requested):
Looking mostly good. Here's a nice-to-have:

- [x] The limit is 20 photos, not 10

And these are the blockers:

- [x] Conflicts
- [ ] Importing new images should show exactly the same screen as the user sees when importing images to create an observation
- [ ] There's an extraneous [+] button on the Standard Camera on Android: Screenshot_20220808-205918

kueda (approved):

PR #147, opened 2022-08-09
Updated windowSoftInputMode for android

### LINKED ISSUE #109: Notes field keyboard disappears or covers text area
In Android, edit an existing observation and ensure that you are scrolled to the top of the screen. Tap the "Add optional notes" field. They keyboard appears and disappears quickly and the view scrolls down a bit. If you repeat without scrolling the same thing happens for a few times, until the keyboard stays visible, but the "Add optional notes" text area is either hidden or has a weird white background. Maybe we need https://reactnative.dev/docs/keyboardavoidingview?

meru20:
This PR resolves #109. Keyboard appears when you click on the TextInput.
Screen Shot 2022-08-09 at 2 19 46 PM


android/app/src/main/AndroidManifest.xml
```diff
@@ -21,7 +21,7 @@
         android:exported="true"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:launchMode="singleTask"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustPan">
```
kueda, line 24: This works, though it doesn't seem to be recommended on Android according to https://developer.android.com/guide/topics/manifest/activity-element. I don't think that needs to block merge, though. Why did you use this approach as opposed to, say, [`KeyboardAvoidingView`](https://reactnative.dev/docs/keyboardavoidingview)?

kueda (changes requested):
Definitely need to remove that file, but otherwise good to go.

android/app/src/main/AndroidManifest.xml
meru20: I tried using `KeyboardAvodingView `without changing any value in the `Android manifest `, but that didn't make the keyboard stay and it added some default padding to it that made the TextInput appear on the top. So in order to make  the keyboard pop open once the user clicks on the TextInput I had to change the value to `adjustPan`.

env.example.env
```diff
@@ -0,0 +1,12 @@
...
+
+# Base URL for the iNat API used for OAUTH
+OAUTH_API_URL=https://staging.inaturalist.org
+
+# Javascript Web Token for anonymous access to protected iNat API endpoints
+JWT_ANONYMOUS_API_SECRET=some-secret
+
+# iNaturalist OAuth client credentials (https://www.inaturalist.org/oauth/applications)
+OAUTH_CLIENT_ID=71daf266133c37c4ce6bee267a6b1388bb899b4c335233f86cd6eb762e79a415
+OAUTH_CLIENT_SECRET=5698d2751db3f9a49ad55808d150793bab5a65b06cc0e31b01fbbccbf0cf3ebf
```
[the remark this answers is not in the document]
meru20: OMG! my bad I accidentally pushed that file . I will be extra careful when I am pushing next time.

android/app/src/main/AndroidManifest.xml
kueda: Fair, I'm just worried about other consequences of this change. I'm not seeing any right now, just wanted to hear that you explored what seems like the more conventional approach.

kueda (approved):

Issue #134, opened 2022-08-01
Title missing from Obs Edit in Android


kueda:
But not in iOS, at least as of 65d67c2a64c500c89c9e9023c6b64e5b8cac503f.
<img width="917" alt="Screen Shot 2022-08-01 at 3 37 53 PM" src="https://user-images.githubusercontent.com/22691/182257357-62873518-4692-47cf-8948-0bf24350865f.png">

PR #148, opened 2022-08-10
Show title on Obs Edit

meru20:
This PR resolves issue #134. 
Applied Platform specific style to show the title header on Android
Screen Shot 2022-08-10 at 2 33 24 PM
.


kueda (approved):
This works! I suspect this style was originally there in an attempt to keep all the items in the header aligned, but it wasn't exactly doing that.

Issue #151, opened 2022-08-15
Make a wrapper for API calls


kueda:
The wrapper should fetch the JWT so we don't have to keep doing that every time we make an API call. This might be a custom hook for API calls that returns a loading status, error status, and response (similar to React Query), or maybe we should just use React Query, or maybe just a hook that returns a client object that will hold the API token and call the required endpoints with the token.

kueda:
Partially addressed in https://github.com/inaturalist/iNaturalistReactNative/pull/166, though I think Amanda's still working on mutation queries.

kueda:
I think Amanda has finished this in #206

PR #161, opened 2022-08-16
Show photo upload failures; upload unsynced photos without overwriting from remote data

### LINKED ISSUE #137: Evidence upload failures should not be silent
**Describe the bug**
If a photo fails to upload for some reason, the failure is currently silent, i.e. the user sees nothing but the photo never gets uploaded.

**To Reproduce**
Kind of impossible to reproduce reliably, but the code is at https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/providers/uploadHelpers/uploadObservation.js#L35-L40. If it fails, photo should still be marked as needing upload. If the upload process isn't already looking for photos that need upload (i.e. not just observations that need upload), it should start.

**Expected behavior**
I expect that if a photo fails to upload, My Observations should still say that there's an observation that needs to be uploaded, and when the user tries to upload it, the app tries to upload that photo to the server (not the obs, unless that needs to be uploaded too).

 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.1.1 (22)

albullington:
This addresses a few things in issue #137:

- Shows error messages specific to photo upload failure on UploadProgressBar (instead of just hanging & not progressing)
- Uses helper functions in Observation to look for observations with unsynced photos in addition to completely unsynced observations
- Fixes a bug where unsynced photos were overwritten when syncing data down from iNaturalist; instead, unsynced observations + observations with unsynced photos will not receive remote updates until they have been uploaded
- Displays a needs upload icon next to observations with unsynced photos in ObsList

Otherwise, I tried to simplify upload code a bit and move more functionality into the Observation model.

Closes #137


src/models/Observation.js
```diff
@@ -335,6 +348,89 @@ class Observation extends Realm.Object {
...
+    try {
+      response = await inatjs.observations.create( uploadParams, options );
+    } catch ( uploadError ) {
+      const body = JSON.parse( await uploadError.response.text( ) );
+      console.error( "[ERROR] Failed to upload observation: ", JSON.stringify( body ) );
+      return body;
+    }
+    return response;
+  }
+
```
kueda, line 433: This doesn't need to block, but in general I don't think network operations belong in the model. A model is a representation of data, so at a minimum it's a class with attributes corresponding to that kind of data, with methods for manipulating that data, and for persisting that data. So CRUD operations belong in the model, and maybe mapping model attributes into different schemas, but probably not things like performing network operations, and definitely not anything related to presentation logic (e.g. a model method would never return a component).

That said, I'm still not sure where this kind of stuff belongs. My gut says a hook, but maybe they belong in some kind of Uploader class that a hook or a component holds onto in state. In Rails this kind of thing would happen in a controller or maybe in some kind of utility class, and in Android this might be encapsulated in a repository of some kind.

src/models/Observation.js
```diff
@@ -313,19 +315,30 @@ class Observation extends Realm.Object {
...
+    const unsyncedObs = obs.filtered( `${unsyncedFilter} || ${photosUnsyncedFilter}` );
+    return unsyncedObs;
+  }
+
+  static isUnsyncedObservation = ( realm, obs ) => {
+    const obsList = Observation.filterUnsyncedObservations( realm );
+    const unsyncedObs = obsList.filtered( `uuid == "${obs.uuid}"` );
+    return unsyncedObs.length > 0;
+  }
+
```
kueda, line 332: These are the kinds of things I think belong in the model, though I don't think we need to specify the model name in the method name unless the method is returning some other kind of data you might not expect, so `unsynced()` and `isUnsynced()` would probably suffice. If you're trying to start methods names with verbs, maybe just `filterUnsynced()` or `getUnsynced()`.

`isUnsynced` feels more like an instance method, though. Maybe something like

```javascript
isUnsynced( ) {
  return (
    this._synced_at == null
    || this._synced_at <= this._updated_at
    || this.observationPhotos.filtered( "ANY _synced_at == null" ).length > 0
  );
}
```

kueda (approved):
This seems like it should address #137, so merge if you're ready. My concerns are mostly nitpicking.

src/components/Observations/UploadProgressBar.js
```diff
@@ -3,62 +3,58 @@
...
+  const progressFraction = calculateProgress( );
 
-    if ( !cancelUpload ) {
-      upload( allObsToUpload[currentUploadIndex] );
-    }
-  }, [cancelUpload, currentUploadIndex, allObsToUpload] );
+  const {
+    handleClosePress,
+    status
+  } = useUploadObservations( allObsToUpload );
```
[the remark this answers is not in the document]
albullington: Yes, that's correct. Once a user has already pressed the button on the UploadPrompt to be able to see the UploadProgressBar, the UploadProgressBar will load and automatically start uploading.

src/models/Observation.js
albullington: Ok, good feedback. Maybe we can take a closer look at where to put network operations while working on the wrapper API issue (#151). Hooks or shared hooks make sense to me.

src/models/Observation.js
albullington: Is this another thing we can audit as part of the tech check? I don't think we're using instance methods anywhere in the models, and I think I'd like to wrap my head around the difference between those and static functions more before making changes.

kueda: You mean like make an issue like, "use instance methods where possible"? I can do that. In general, object-oriented design says that if an operation only pertains to a single instance of an object, like `isUnsynced` does here, it should be an instance method, i.e. the instance is best-suited to perform that operation b/c it should know everything about itself. This kind of [encapsulation](https://en.wikipedia.org/wiki/Object-oriented_programming#Encapsulation) often makes code more reusable, but it also makes code more understandable and easier to write, e.g. if you know you have an instance of an Observation you can think to yourself, "whose job is it to know if this observation needs to be synced? Probably the observation itself." If you don't have that expectation code becomes much harder to write and manage, e.g. "Is there an ObservationSyncChecker that I need to get to check if this observation is synced?"

You might use a static / class method when you want to perform an operation that involves multiple instances of that class, or that operates on instances of many classes, e.g. a `getUnsynced()` method that gets all unsynced observations.

PR #162, opened 2022-08-18
Replace PhotoGalleryProvider with useBetween hook

### LINKED ISSUE #150: Try removing PhotoGalleryProvider
We use context / providers a lot, perhaps to excess. Sometimes using props might be a simpler approach, and in others [composition](https://reactjs.org/docs/composition-vs-inheritance.html) might be better. This issue is to explore other approaches by removing a single provider and thinking about how to live without it.

- [ ] Remove `PhotoGalleryProvider` while maintaining existing functionality.

Like other issues in this milestone, this is mostly exploratory. What are some ways of doing without the provider? Maybe it's better to just keep it?

albullington:
The main purpose of the photo gallery provider was to allow shared `selectedPhotos` state between PhotoGallery and GroupPhotos. I switched to using the `use-between` library which allows hooks to be shared between sibling components.

This closes #150.


src/navigation/rootNavigation.js
```diff
@@ -85,54 +84,52 @@ const App = ( ): React.Node => {
...
+                  name="explore stack"
+                  component={ExploreStackNavigator}
+                  options={hideHeader}
+                />
+                <Drawer.Screen
+                  name="network"
+                  component={NetworkLogging}
+                />
+              </Drawer.Navigator>
+            </ObsEditProvider>
```
kueda, line 132: So remind me, is one of the problems here is that we don't want to expose photo gallery state to the entire app, but we can't wrap a provider around a subset of screens? Like this won't work?

```javascript
<Stack.Navigator>
  <Stack.Screen
    name="ObsList"
    component={ObsList}
  />
  <PhotoGalleryProvider>
    <Stack.Screen
      name="PhotoGallery"
      component={PhotoGalleryWithPermission}
    />
    <Stack.Screen
      name="GroupPhotos"
      component={GroupPhotos}
    />
  </PhotoGalleryProvider>
</Stack.Navigator>
```

src/components/PhotoLibrary/hooks/useGalleryPermissions.js
```diff
@@ -0,0 +1,27 @@
...
+  useEffect( ( ) => {
+    if ( !canRequestPhotos ) {
+      setCanRequestPhotos( true );
+    }
+  }, [canRequestPhotos] );
+
+  return canRequestPhotos;
+};
+
+export default useGalleryPermissions;
```
kueda, line 27: Why is this in a hook if it is only called once and it basically just stores a boolean... that always gets set to true?

src/components/PhotoLibrary/PhotoGalleryImage.js
```diff
@@ -0,0 +1,85 @@
...
+  selectedAlbum,
+  getAllPhotos,
+  updatePhotoGallery
+}: Props ): Node => {
+  const useSharedSelectedPhotos = ( ) => useBetween( useSelectedPhotos );
+  const {
+    selectedPhotos,
+    unselectPhoto,
+    selectPhoto
+  } = useSharedSelectedPhotos( );
```
kueda, line 37: This seems like an excessive use of `useBetween`, or any way to share state. A PhotoGalleryImage component should presumably concern itself with a single image. `isSelected` makes sense as a prop here, but keeping track of which images are selected seems like a concern for whatever component contains all the images (which is probably `<PhotoGallery>` here).

kueda (commented):
Overall I'm not convinced `useBetween` is really superior to context. I guess it might isolate the shared state to the components that use a particular hook, which seems good, but why not just pass the photo IDs to GroupPhotos when you navigate to it? That loosens the coupling between those two components and doesn't add a dependency.

src/navigation/rootNavigation.js
albullington: Yes, as far as I know, we can't define a provider around a subset of screens. I think if we were able to break out certain screens into their own stack navigator, then we could wrap a provider around just those screens -- but that's probably not desirable for the photo gallery since we need to be able to navigate back from ObsEdit.

src/components/PhotoLibrary/hooks/useGalleryPermissions.js
albullington: Doesn't need to be in a hook. I mostly wanted to pull this out to make it easier to read the rest of what's happening in PhotoGallery, since there's a lot of unrelated functionality.

src/components/PhotoLibrary/PhotoGalleryImage.js
albullington: That's fair, I will change this.

albullington:
I thought about an option where we would pass `selectedPhotos` to `GroupPhotos` via navigation parameters. The thing I don't love about that option is that we would still need the ability to remove photos from `selectedPhotos` on the `GroupPhotos` screen and we would then need to pass those parameters back to `PhotoGallery` when a user taps the back button on `GroupPhotos`. It seems kind of awkward to me that both screens can manage the state of `selectedPhotos` & that we would need to pass `selectedPhotos` in both navigation directions without having a single source of truth, though I guess `react-navigation` does support [updating params](https://reactnavigation.org/docs/params#updating-params) and expect [data to be passed back to previous screens](https://reactnavigation.org/docs/params#passing-params-to-a-previous-screen).

Happy to try that implementation and see which we like better.

kueda:
> The thing I don't love about that option is that we would still need the ability to remove photos from selectedPhotos on the GroupPhotos screen and we would then need to pass those parameters back to PhotoGallery when a user taps the back button on GroupPhotos.

Do we? In my mind these are isolated screens: you make your selection in PhotoGallery, then do something with that selection in GroupPhotos, but if you return to PhotoGallery it should be in the same state as you left it no matter what happened in GroupPhotos. @abhasinat, any opinion here?

> Happy to try that implementation and see which we like better.

I'd certainly like to see that.

albullington:
Ok, if we don't need to sync states of selectedPhotos between GroupPhotos and PhotoGallery, everything is a lot simpler. I removed use-between and passed selectedPhotos in a single direction from PhotoGallery to GroupPhotos.

src/components/PhotoLibrary/PhotoGallery.js
```diff
@@ -61,18 +65,60 @@ const PhotoGallery = ( ): Node => {
...
+
+  useEffect( ( ) => {
+    if ( photosFetched ) {
+      if ( photoGallery[selectedAlbum]
+        && photoGallery[selectedAlbum].length === photosFetched.length ) {
+        return;
+      }
+
+      // store photo details in state so it's possible
+      // to select mutiple photos across albums
```
kueda, line 97: Is it even possible to choose photos from a particular album?

kueda (approved):
Nothing blocking, but I think there is something weird about needing to tell the `PhotoGallery` that it needs to reset its state. I also think that after consolidating the navigators, there's no real reason to wrap everything in `<MainStackNavigation>` in `<Mortal>`, since presumably it *isn't* mortal (there's always *something* from the `MainStack` that will be active). Maybe just the obs flow screens should be in their own stack and wrapped in `Mortal`? Or maybe you're current approach of resetting the stats with params is best. Thoughts?

src/components/PhotoLibrary/PhotoGallery.js
albullington: Yes, if you tap on the camera roll button at the top, you can select other albums.

albullington:
I'm not sure which of the screens in the `MainStackNavigator` would need to be reset with `Mortal`, so I think I'm going to leave this as is for now. React navigation also has a new way of [grouping components](https://reactnavigation.org/docs/group/), so maybe there's a way we could wrap certain screens in `Mortal` without needing a separate stack navigator.

PR #163, opened 2022-08-18
Consolidate stack navigators

### LINKED ISSUE #154: Explore consolidating stack navigators
React Navigation [does not recommend nesting navigators](https://reactnavigation.org/docs/nesting-navigators/#best-practices-when-nesting) in the way we're currently doing, so explore consolidating them into as few as possible (preferably one?).

albullington:
- Removes notifications and explore stack navigators
- Removes animation from landing screens in tab bar (screens were previously swiping in from the right)
- Renames observation stack navigator to main stack navigator
- Shows drawer items in the correct order (plus network/logging, which we're using for development)

I kept the projects and identify stack navigators for now, since I don't think those screens need to know the user's previous navigation history. For example, if a user goes to identify via the drawer, there's no back button to take a user back to the previous screen. Meanwhile, a user would want to preserve their navigation history while going back and forth between the uploader, observation list, and explore screens. Not totally sure if this is the right way to think about this, or if it's really better to have everything in one stack.

Closes #154.


kueda (approved):

kueda:
I think at the very least it simplifies things and focuses on making new stacks if and only if we really need them.

Issue #121, opened 2022-07-01
Button cleanup


kueda:
Note: designs for this might not quite be ready. Check with Ken-ichi or Abhas before starting.

Make a component called <Button> that can be configured to accommodate all of the variants of the plain (non-dropdown) buttons in the [style guide](https://www.figma.com/file/XmLaQzJ1McnsYQysB57hib/iNaturalist-Mobile-UX-Design?node-id=1516%3A36870), or rename and extent `<RoundGreenButton>`. IMO, this should be something like <Button level="primary"> for a ["filled button"](https://m3.material.io/components/buttons/guidelines#9ecffdb3-ef29-47e7-8d5d-f78b404fcafe) (in Material parlance, annoyingly called a "contained button" in react-native-paper) with an iNat green background, and <Button level="primary" risk="danger" /> for the same thing with a red background.

Then replace all of our `<Button>`s and `<RoundGreenButton>`s with this new component.

PR #165, opened 2022-08-19
Create Custom button

meru20:
This PR resolves issue #121 , 
 - Created custom Button component
 - Replaced buttons with custom component


kueda (changes requested):
Overall looking good. In addition to the todos in the comments, here's one more:

- [x] Remove RoundGreenButton, SecondaryButton, and all their associated styles

src/components/LoginSignUp/Login.js
```diff
@@ -171,13 +172,12 @@ const Login = ( ): Node => {
...
-        style={viewStyles.button}
-        buttonText="Log-in"
-        handlePress={login}
+      <Button
+        level="primary"
+        text="Log-in"
+        onPress={login}
         disabled={!email || !password}
         testID="Login.loginButton"
-        loading={loading}
```
meru20, line 180: I removed the `loading` prop since loading was passed as a prop for `Button` element from `react-native-paper`. Let me know if we still want to keep it.

src/styles/sharedComponents/buttons/buttonVariants.js
```diff
@@ -0,0 +1,51 @@
...
+  containerWarning: {
+    backgroundColor: WARNING,
+    borderColor: WARNING
+  },
+  containerNeutral: {
+    backgroundColor: NEUTRAL,
+    borderColor: NEUTRAL
+  },
+  containerDisabled: {
+    opacity: 0.65
```
kueda, line 34: I know this way is simple, but eventually you'll get into a situation where a button overlays something else and the background element will be partially visible, which is almost never what you want, so please explicitly style the disabled states of the different levels as they are in Figma:

Primary: `#C6DC98`
Neutral: `#D3D3D3`
Warning: [theoretically never disabled, but I would put something in here anyway, maybe just grab the hex code for whatever `#9B1111` is at 0.65 opacity.]

src/components/LoginSignUp/Login.js
kueda: We definitely want to keep it, otherwise the Login view cannot show that it is in a loading state. I guess I forgot to include this in the spec, so apologies for that, but the new `<Button>` component should have a `loading` prop that shows an activity indicator when true.

One simple way to do this would be to use the `react-native-paper` `<Button>` component inside our own `<Button>` component, so ours is basically a wrapper that accepts the props we want and configures the `react-native-paper` one appropriately.

src/components/SharedComponents/Buttons/Button.js
```diff
@@ -0,0 +1,44 @@
...
+      onPress={onPress}
+      style={buttonContainer}
+      disabled={disabled}
+      testID={testID}
+    >
+      <Text style={buttontext}>{text}</Text>
+    </TouchableOpacity>
+  );
+};
+
```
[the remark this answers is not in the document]
kueda: This approach is fine and doesn't need to block merge, but a more explicit way to define default props is the [`defaultProps` class property](https://reactjs.org/docs/react-component.html#defaultprops). Personally I think that convention makes code more readable, e.g. when I'm looking at a component I don't have to hunt through the code to figure out what the value of a prop will be when it's not explicitly set, I can just look at the `defaultProps` assignment. Less subjectively, it makes it easier for machines to understand your code, e.g. auto documentation software... which we're not using, yet.

kueda (changes requested):
Getting closer! I think once you address those remaining two issues we should be good to go here.

src/styles/sharedComponents/buttons/buttonVariants.js
```diff
@@ -0,0 +1,61 @@
...
+const WARNING_DISABLED = "#B95F5E";
+const NEUTRAL = "#979797";
+const NEUTRAL_DISABLED = "#D3D3D3";
+
+const viewStyles: { [string]: ViewStyleProp } = StyleSheet.create( {
+  containerDefault: {
+    alignItems: "center",
+    borderRadius: 40,
+    marginVertical: 15,
+    marginHorizontal: 10
```
kueda, line 20: Couple lingering issues here:

1. Using `alignItems: "center"` with react-native-paper's <Button> makes it so only the content of the button is tappable, not the entire background. I think it's safe to just remove that here.
2. I would remove the margins to maximize reusability. If a particular component needs a button with margins, it should be able to add them with a style prop... which means adding a style prop to our `<Button>`.

kueda (changes requested):

PR #166, opened 2022-08-23
Use React Query to simplify API requests

albullington:
This is related to issue #151. I wanted to explore using React Query to simplify our API calls and the way we handle error messages.

So far I'm liking this approach, because it makes it easier to handle errors instead of ignoring them or just logging them to console, and it encourages us to simply return data instead of doing a bunch of extra data processing or manipulation in the same fetch function.

I also like the idea of breaking out API calls into a separate `lib/` folder, since we haven't been sure where to put them (sometimes in the models, sometimes in hooks).

React Query has a lot of other functionality that will likely be useful for us, like `useInfiniteQuery` which should simplify the infinite scroll network requests on ObsList. We also might be able to let React Query handle refetching data instead of handling it ourselves like we're doing in ObsDetails.

This also includes an example of how to mock network data and test a component with `useQuery`.


src/components/TaxonDetails/TaxonDetails.js
```diff
@@ -10,22 +11,24 @@ import {
...
-import { useTaxonDetails } from "./hooks/useTaxonDetails";
 
 const TaxonDetails = ( ): React.Node => {
   const { params } = useRoute( );
   const { id } = params;
-  const { taxon, loading } = useTaxonDetails( id );
-  // const similarSpecies = useSimilarSpecies( id );
+  const {
+    data, isLoading, isError
+  } = useQuery( ["taxaFetch"], ( ) => fetchTaxa( id ) );
```
kueda, line 24: I think here you want something like `useQuery( ["taxon", id], ( ) => fetchTaxa( id ) )` so the query key is specific to this taxon. If you don't, when you visit the next taxon detail view, the current taxon appears for a moment before the cache for that query key gets refreshed.

src/lib/taxaFetchAPI.js
```diff
@@ -0,0 +1,43 @@
...
+    photo: PHOTO_FIELDS
+  },
+  wikipedia_summary: true,
+  wikipedia_url: true
+};
+
+const fetchTaxa = async ( id: number ): Promise<any> => {
+  const params = {
+    fields: FIELDS
+  };
```
kueda, line 33: `params` never changes so you could probably statically define it outside of the function like the rest of the fields stuff.

albullington:
Added `useAuthenticatedQuery` as a wrapper function for `useQuery` which fetches the jwt token and passes it as options to the queryFunction.

src/lib/taxaFetchAPI.js
```diff
@@ -0,0 +1,43 @@
...
+
+  try {
+    const { results } = await inatjs.taxa.fetch( id, params );
+    return results[0];
+  } catch ( e ) {
+    throw new Error( JSON.stringify( e.response ) );
```
kueda, line 39: Not sure this is adding anything. inatjs will throw an error if something goes wrong internally or if the request fails for some reason, which will satisfy React Query's need for an error. If we want to perform some standardized error handling this is just percolating up an error with a stringified object, and it's going to be tricky for whatever code is dealing with this error to parse out meaningful information.

Something like this might be preferable:

```javascript
if ( !e.response ) { throw e; }
const errorJson = await e.response.json( );
const error = new Error( errorJson.status );
throw error;
```

That way

1. If it's *not* an HTTP error we just throw it in its original state, and the consuming code can handle it as it sees fit
2. If it is an HTTP error we parse the meaningful JSON response and the HTTP status and return then as attributes of the error

I'm not convinced that's superior to just throwing the original error all the time, but it does help separate concerns. We'd want to extract it so it could be shared across these API files, though.

src/components/TaxonDetails/TaxonDetails.js
kueda, line 24: I think here you want something like `useQuery( ["taxon", id], ( ) => fetchTaxa( id ) )` so the query key is specific to this taxon. If you don't, when you visit the next taxon detail view, the current taxon appears for a moment before the cache for that query key gets refreshed.

src/lib/searchMessagesAPI.js
```diff
@@ -0,0 +1,21 @@
...
+
+  try {
+    const { results } = await inatjs.messages.search( params, options );
+    return results;
+  } catch ( e ) {
+    throw new Error( JSON.stringify( e.response ) );
+  }
+};
+
+export default searchMessages;
```
kueda, line 21: I don't think each API method needs to be its own file, and I've never liked calling a directory `lib` unless it contains actual libraries. Can we call this directory `api` and maybe this file should be `messages.js`?

src/navigation/rootNavigation.js
```diff
@@ -32,6 +36,10 @@ const hideHeader = {
   headerShown: false
 };
 
+const queryClient = new QueryClient( {
+  defaultOptions: { queries: { retry: 2 } }
```
kueda, line 40: Probably fine for now, but eventually we'd want retry to be a function that only retries errors that look like timeouts. Most HTTP errors should be idempotent.

src/sharedHooks/useAuthenticatedQuery.js
```diff
@@ -0,0 +1,17 @@
...
+  query: Function
+): any => useQuery( [queryKey], async ( ) => {
+  const apiToken = await getJWTToken( );
+  const options = {
+    api_token: apiToken
+  };
+  return query( options );
+} );
+
+export default useAuthenticatedQuery;
```
kueda, line 17: Interesting, but I think the query results will get cached and survive sign-out, no? We might want to make sure every query key array includes the user ID. You could also `useState` to store the JWT until it stops working and do the error handling for that here.

kueda (changes requested):
I like the separation of concerns that's happening here, and I think React Query is doing some good heavy lifting that will save us time down the road. I think we need to address some of the error handling at least before merging. I'm also rather curious to see how RQ mutations would work in our context.

kueda (approved):

Issue #144, opened 2022-08-06
Use EXIF data for imported photos


kueda:
Bit of a draft. If someone wants to take a stab at it go for it, but I will try and figure out what specific metadata tags we're using in the native apps.

- [x] When importing a photo, do not use the current date or location
- [ ] Extract date from the photo EXIF to populate `observed_on` and `time_observed_at`
- [x] Extract `latitude` and `longitude` from EXIF data
- [ ] Extract `positional_accuracy` from EXIF data
- [x] If an obs has multiple photos, use the first photo

kueda:
Will get more detailed fixture in the PR, but one more todo here:

- [ ] Add a test to make sure the coordinates, positional accuracy, date, and time all get parsed from this photo: ![moth-exif-location](https://user-images.githubusercontent.com/22691/187563722-4d2b4a4a-a9f4-45df-9ba4-595e36528e5a.jpg)

PR #168, opened 2022-08-27
144 import photo exif

budowski:


kueda (changes requested):
I like the hook approach, but this isn't quite working for me:

- [x] Date never seems to get parsed
- [x] `GPSHPositioningError` isn't getting parsed

IMO, it would help if you could add a unit test for `usePhotoExif` or an integration test for `ObsEdit` that uses the images in https://github.com/inaturalist/iNaturalistReactNative/issues/144#issuecomment-1232289198 to ensure that these values are getting successfully parsed.

kueda (changes requested):
- [ ] As far as I can tell nothing gets extracted in iOS
- [ ] I'm still not getting `GPSHPositioningError` from the moth photo. I can see it's there with exiftool, and I can see piexifjs is [supposed to recognize it](https://github.com/hMatoba/piexifjs/blob/2180d60b8cdf638e236e0a6703b7c378bb4f5785/piexif.js#L2131), so why doesn't it get read?
- [ ] When I import a photo in Android, the current datetime appears before it gets set from the EXIF; it should be blank by default when importing.

src/components/ObsEdit/hooks/usePhotoExif.js
```diff
@@ -0,0 +1,139 @@
...
+    let isCurrent = true;
+
+    const parseExif = async ( ): Promise<Object> => {
+      try {
+        // Don't read the entire file into memory - just read a subset, since we need only
+        // partial data in order to parse the EXIF metadata.
+        const data = await RNFS.read( photoUri, MAX_EXIF_READ_BYTES, 0, "base64" );
+        const rawExif = piexif.load( `data:image/jpeg;base64,${data}` );
+        console.log( "Raw EXIF", rawExif );
+        const newExif = await Exif.getExif( photoUri );
```
kueda, line 103: Does this actually get used?

tests/unit/hooks/usePhotoExif.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+test( "reads and parses EXIF data from photo", async ( ) => {
+  const { result } = renderHook( () => usePhotoExifWrapper() );
+
+  await waitFor( () => {
+    // This triggers the usePhotoExif hook
+    result.current.setUri( `${process.cwd()}/tests/resources/test_exif.jpg` );
+  } );
+
+  if ( result.current.exif ) {
```
kueda, line 55: This test only passes because this is never true.

kueda (changes requested):
Didn't checkout out iOS since I know you're still working on that, but in addition to inline comments, there's this in Android.

- [ ] Still not getting `GPSHPositioningError` in the Android emulator

tests/unit/hooks/usePhotoExif.test.js
kueda: Ok, I think the reason it's never true is that `Exif` from `react-native-exif` isn't defined in the test env, presumably b/c it needs to have those linked native libs working and the test env is pure js. Instead of the unspecific catch you have in the hook, maybe you could make the hook have an option so you can at least test the pure JS functionality? Maybe this is all a bit premature while you're figuring out if you're going to roll your own. Not sure how people write integration tests for hooks that depend on native code.

PR #169, opened 2022-08-28
Add flow types for `react-native-safe-area-context`; upgrade file to `@flow strict-local`.

frewsxcv:
Dipping my toes just a little bit into this React Native project...

### DIFF
--- a/flow-typed/npm/react-native-safe-area-context_vx.x.x.js
+++ b/flow-typed/npm/react-native-safe-area-context_vx.x.x.js
@@ -1,0 +1,31 @@
+declare module "react-native-safe-area-context" {
+  import type { Node } from "react";
+
+  declare interface EdgeInsets {
+    top: number;
+    right: number;
+    bottom: number;
+    left: number;
+  }
+
+  declare interface Rect {
+    x: number;
+    y: number;
+    width: number;
+    height: number;
+  }
+  declare interface Metrics {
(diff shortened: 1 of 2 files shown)


kueda:
@frewsxcv thanks for this and your other PRs. I think this one is a useful addition and helps us use flow more effectively, but I worry about maintenance, e.g. if something changes in `react-native-safe-area-context` I'm not sure I'd be able able to update the type in our repo to keep up. Is there a way to just use a maintained 3rd party implementation of the type via npm? If not, maybe you can explain your approach for adding these? Like, are you just porting the [typescript types](https://github.com/th3rdwave/react-native-safe-area-context/blob/main/src/SafeArea.types.ts) to flow?

frewsxcv:
> Is there a way to just use a maintained 3rd party implementation of the type via npm?

Good question! From what I've seen, most of the Flow community seems to have migrated to TypeScript. And there are TypeScript bindings for react-native-safe-area-context, and other libraries used in iNaturalistReactNative. So i think this is going to be a recurring theme going forward unfortunately. I'd be happy to help with a migration to TypeScript, if that was desired.

kueda:
Got it. In that case, I think I'll merge this, but please hold off on future flow type definition contributions unless we make issues for them. Moving to TypeScript might be worth it, but I'll need to discuss that with the team.

PR #170, opened 2022-08-28
Upgrade `flow-bin` dependecy to the latest version.

frewsxcv:


src/components/LoginSignUp/AuthenticationService.js
```diff
@@ -101,6 +102,7 @@ const getJWTToken = async ( allowAnonymousJWTToken: boolean = false ): Promise<?
 
   if (
     !jwtToken
+    || !jwtTokenExpiration
```
frewsxcv, line 105: flow wasn't happy with using an arithmetic operator with `jwtTokenExpiration` (which it thinks could be `void`) on the next line

src/components/LoginSignUp/AuthenticationService.js
```diff
@@ -82,10 +82,11 @@ const getAnonymousJWTToken = () => {
...
 const getJWTToken = async ( allowAnonymousJWTToken: boolean = false ): Promise<?string> => {
-  let jwtToken = await RNSInfo.getItem( "jwtToken", {} );
-  let jwtTokenExpiration = await RNSInfo.getItem( "jwtTokenExpiration", {} );
-  if ( jwtTokenExpiration ) {
-    jwtTokenExpiration = parseInt( jwtTokenExpiration, 10 );
+  let jwtToken: string | void = await RNSInfo.getItem( "jwtToken", {} );
+  const jwtTokenExpirationString: string | void = await RNSInfo.getItem( "jwtTokenExpiration", {} );
+  let jwtTokenExpiration: number | void;
+  if ( jwtTokenExpirationString ) {
+    jwtTokenExpiration = parseInt( jwtTokenExpirationString, 10 );
```
kueda, line 89: Just stylistic, but why not something like

```javascript
const jwtExpiraton: number | void = parseInt( await RNSInfo.getItem( "jwtTokenExpiration", {} ), 10 )
```

I *think* `parseInt` is pretty robust against various kinds of null-ish values it might receive.

package.json
```diff
@@ -97,7 +97,7 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "factoria": "^3.2.2",
     "faker": "^5.5.3",
-    "flow-bin": "^0.162.0",
+    "flow-bin": "^0.185.0",
```
kueda, line 100: No objection, but any advantage to this version bump?

src/components/LoginSignUp/AuthenticationService.js
frewsxcv: `parseInt(null)` returns `NaN`. So if `( Date.now() - NaN ) / 1000 > JWT_TOKEN_EXPIRATION_MINS * 60` is okay with you, then I can change it

package.json
frewsxcv: I was hitting a bug locally that was fixed in 0.185.0:

> Fix a bug that makes assignments in loop guards like x = some expression containing x have a type that's effectively any. If you turned on inference_mode=constrain_writes since the last release, there will be new errors.

https://github.com/facebook/flow/blob/main/Changelog.md

src/components/LoginSignUp/AuthenticationService.js
kueda: I'm pretty sure that would evaluate in the way we want it to, so yeah, I'd be fine with that. No strong feeling about it here, so feel free to leave it.

package.json
kueda: Cool, thanks. FWIW, it will help me review future PRs if you can mention things like this in your commit messages or the PR body.

kueda:
@frewsxcv were you planning on resolving that conflict? I think this is ready to merge otherwise.

kueda:
I'm going to close this for now since I'm going to assume it's abandoned. If you want to resolve those conflicts feel free to open another PR.

PR #179, opened 2022-09-08
Exploration of styling with RN Paper Themes

albullington:
This PR is an initial exploration of how we might start to style iNat mobile using RN paper themes. I focused on adding themes for Login screen input fields and buttons. A few initial thoughts:

1. This can definitely save us from repeating styles over and over in individual style sheets, which I believe is the main problem we're trying to solve. It will also make it easier to make app-wide changes to things like the font size of buttons
2. We'll want to move all colors into the theme instead of having a separate `colors.js` file
3. This will be particularly helpful if we're supporting [dark mode](https://callstack.github.io/react-native-paper/theming-with-react-navigation.html#using-context), since we can toggle themes on and off
4. For text, I think we'll want to use RN paper's `Text` component and customize font sizes using the `typescale` theme object

I can keep exploring what this would look like for text and other components in the Style Guide v1 in Figma (icons, lists, and sheets).


albullington:
Closing because we decided to use Tailwind for styling/theming instead.

PR #181, opened 2022-09-13
UI updates to bottom sheet

### LINKED ISSUE #177: Hide Bottom Sheet on My Observations when the user scrolls down
- [x] Make the Bottom Sheet on My Observations not modal
- [x] When the user scrolls down on My Observations, hide the Bottom Sheet, i.e. it should only be visible when you are scrolled all the way to the top

albullington:
This makes the bottom sheet a regular sheet, not a modal. It moves the bottom sheets into `ObservationViews` where the actual `FlatList` lives. It snaps the bottom sheet away when a user scrolls down and snaps it back up when a user scrolls to the top of the screen.

Need to get one more test passing.

Closes #177.

### DIFF
--- a/src/components/SharedComponents/BottomSheet.js
+++ b/src/components/SharedComponents/BottomSheet.js
@@ -2,20 +2,39 @@
 
 import BottomSheet, { BottomSheetView } from "@gorhom/bottom-sheet";
 import type { Node } from "react";
-import React, { useMemo, useRef } from "react";
+import React, {
+  useCallback, useEffect, useMemo, useRef
+} from "react";
 
 import { viewStyles } from "../../styles/sharedComponents/bottomSheet";
 
 type Props = {
-  children: any
+  children: any,
+  hasScrolled: boolean
 }
 
-const StandardBottomSheet = ( { children }: Props ): Node => {
+const StandardBottomSheet = ( { children, hasScrolled }: Props ): Node => {
   const sheetRef = useRef( null );
-  const snapPoints = useMemo( () => ["30%"], [] );
+  const snapPoints = useMemo( () => ["45%"], [] );
 
   // eslint-disable-next-line
   const noHandle = ( ) => <></>;
+
+  const handleClosePress = useCallback( () => {
+    sheetRef.current?.close();
+  }, [] );
+
+  const handleSnapPress = useCallback( ( ) => {
+    sheetRef.current?.snapToIndex( 0 );
+  }, [] );
+
+  useEffect( ( ) => {
+    if ( hasScrolled ) {
+      handleClosePress( );
+    } else {
+      handleSnapPress( );
+    }
+  }, [hasScrolled, handleClosePress, handleSnapPress] );
 
   return (
     <BottomSheet
--- a/src/components/SharedComponents/BottomSheet.js
+++ b/src/components/SharedComponents/BottomSheet.js
@@ -3,19 +3,20 @@
 import BottomSheet, { BottomSheetView } from "@gorhom/bottom-sheet";
 import type { Node } from "react";
 import React, {
-  useCallback, useEffect, useMemo, useRef
+  useCallback, useEffect, useRef
 } from "react";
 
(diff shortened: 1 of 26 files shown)


src/components/SharedComponents/BottomSheet.js
```diff
@@ -1,53 +1,53 @@
...
+  }, [] );
+
+  useEffect( ( ) => {
+    if ( hasScrolled ) {
+      handleClosePress( );
+    } else {
+      handleSnapPress( );
+    }
+  }, [hasScrolled, handleClosePress, handleSnapPress] );
+
```
kueda, line 38: IMO, this behavior should be optional, e.g. `<BottomSheet hideOnScroll />`. For example, Explore has a Bottom Sheet that might hide when you scroll in grid view, but not in map view.

src/components/SharedComponents/BottomSheet.js
```diff
@@ -1,53 +1,53 @@
...
+      snapPoints={snapPoints}
+      style={viewStyles.shadow}
+      handleComponent={noHandle}
+    >
+      <BottomSheetView style={viewStyles.bottomSheet}>
         {children}
-      </BottomSheetModal>
-    </BottomSheetModalProvider>
+      </BottomSheetView>
+    </BottomSheet>
```
kueda, line 49: You could argue that modal behavior should be optional too, e.g. `<BottomSheet modal />`. You could also make the argument that the behavior is so different that a separate component is justified, e.g. `<ModalBottomSheet />`. Not sure this requires a change in this PR, but something to consider: when you're making a component you know you're going to re-use, try to make the defaults generic but useful, and leave room for enabling expanded functionality. Definitely straddles the line between pre-optimization and prudent planning.

src/components/SharedComponents/BottomSheet.js
```diff
@@ -1,53 +1,53 @@
...
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...props}
-      pressBehavior="close"
-      appearsOnIndex={0}
-      disappearsOnIndex={-1}
-    />
-  );
+const StandardBottomSheet = ( { children, hasScrolled }: Props ): Node => {
+  const sheetRef = useRef( null );
+  const snapPoints = useMemo( () => ["45%"], [] );
```
kueda, line 18: Out of curiosity, I'm assuming your memoizing this so that `BottomSheet` doesn't receive a new array object of snap points every time this gets rendered. Could since the value here doesn't depend on props or context or anything, could you also just make it a constant in this file, e.g. `const SNAP_POINTS = ["45%"];` defined outside of the component? Is there an advantage to `useMemo` here?

kueda (changes requested):
Does the job, I just want to make sure we can use this `BottomSheet` component elsewhere.

src/components/SharedComponents/BottomSheet.js
albullington: I was definitely thinking of modal behavior as belonging in a separate component. It might make sense to create a `<ModalBottomSheet />` component in a separate github issue, since it looks like we have three separate bottom sheet modals on `ObsEdit`, `AddID`, and `ObsDetails`.

src/components/SharedComponents/BottomSheet.js
kueda: Ok, now it's optional, but as far as this component is concerned it has nothing to do with scrolling, right? It's just hiding the bottom sheet at the request of the parent component. Maybe the prop should just be `hide`?

src/components/SharedComponents/BottomSheet.js
kueda: Let's try to keep that in mind for when we get to those elements.

kueda (changes requested):
In addition to making the meaning of that prop more clear, the bottom sheet seems to have lost its drop shadow in Android:

Screen Shot 2022-09-15 at 2 57 48 PM

albullington:
Fixed the shadow on Android and changed the prop name to `hide`.

kueda (approved):

PR #182, opened 2022-09-15
Keep toolbar sticky & collapse header on ObsList scroll

albullington:


src/components/SharedComponents/ObservationViews/Toolbar.js
```diff
@@ -0,0 +1,60 @@
...
+  <>
+    {!isExplore && (
+      <View style={viewStyles.toggleButtons}>
+        {isLoggedIn && (
+        <Pressable onPress={syncObservations}>
+          <Icon name="sync" size={30} />
+        </Pressable>
+        )}
+      </View>
+    )}
```
kueda, line 32: All of this `isExplore` stuff is is probably a situation where you want to invert control: the parent component could pass in some extra buttons as props and Toolbar wouldn't have to worry about things like whether someone is logged in.

src/components/SharedComponents/ObservationViews/Toolbar.js
```diff
@@ -0,0 +1,60 @@
...
+    <View style={viewStyles.toggleButtons}>
+      {isExplore && (
+        <Pressable
+          onPress={( ) => setView( "map" )}
+          accessibilityRole="button"
+          testID="Explore.toggleMapView"
+        >
+          <Icon name="map-outline" size={30} />
+        </Pressable>
+      )}
```
kueda, line 42: For this one you might use a boolean prop like `showMapButton`.

kueda (changes requested):
Works, but I think we need a different approach to hiding the header.

src/components/SharedComponents/ObservationViews/Toolbar.js
albullington: I imagine we'll overhaul this code when we try to get the Explore screen working again. Right now, Explore seems broken enough that I'm not sure how the toolbar looks at all.

kueda (changes requested):
This is looking a lot better, but the green background of the header flashes white whenever you start or stop dragging.

albullington:
Added `useMemo` around the header to make sure it doesn't rerender on scroll. I think this should be ready to merge.

Issue #158, opened 2022-08-16
Refactor AuthenticationService as a collection of hooks


kueda:
Code like 
```javascript
const token = await getJWTToken( );
```
is synchronous and potentially holds up rendering, or at least conditions you to expect to hold up rendering as opposed to reacting to changes in state. Instead, I suspect we could refactor most of AuthenticationService as hooks, e.g.

Code like 
```javascript
const token = useJWTToken( );
```
In addition to refactoring that module, everything that depends on it would need to adjust to not having those data available immediately, e.g. code using `token` might need to go in an effect that depends on `token` instead of being used immediately.

kueda:
Assigning myself b/c my work on #149 started to require these kinds of hooks.

kueda:
https://github.com/inaturalist/iNaturalistReactNative/pull/183 added the `useApiToken` hook, which provides a reusable asynchronous way for components to get an API token. https://github.com/inaturalist/iNaturalistReactNative/pull/166 refactored a lot of API calls in a way that also performs that work asynchronously, so I think the original problem in this issue has been addressed. I'm not sure we really need to refactor that entire file yet, we just need to make sure it's synchronous behavior isn't holding up the UI.

Issue #149, opened 2022-08-15
Explore providing Realm with context and isolating data persistence in models


kueda:
Currently when we need to persist data in Realm, we open a new Realm connection in a component or a hook and either use it or pass it to a model instance where it gets used. IMO, the components and most of the hooks should not need to know that much detail about data persistence, and opening and closing multiple Realm connections sounds like a recipe for memory leaks and stale Realm handles.

The task here is to explore ways we might achieve two goals:

1. Provide a single Realm connection to all code in the app
2. Isolate data persistence functions from presentation functions like components

This work will be exploratory in nature so it may not necessarily result in a merge. I'm mostly interested in getting a feel for whether we should do something like this, and if so, what approach seems the most reasonable.

I put some time into [exploring these in a branch](https://github.com/inaturalist/iNaturalistReactNative/compare/bb4795d69c5fef05d761e0b93ace2608564b4ffb...realm-in-models) a while ago, by using Context and a Provider to provide a single Realm connection to the entire app, and by using a superclass of all models to store that Realm connection as a class constant (essentially a global). That might be one avenue worth pursuing. My caveat with that approach is that I'm not 100% sure a new Realm connection will get opened when needed (e.g. when the app moves from bacgkround to foreground), nor am I sure if using a class constant is a stable way to make the Realm connection available to code outside of components and hooks (which are the only places the useContext hook can be used to get access to the Realm connection).

[Realm itself provides a `RealmProvider`](https://www.mongodb.com/docs/realm/sdk/react-native/use-realm-react/#using-the-realm-provider) that we could use instead of writing our own, so that may be worth exploring.

Another approach to 2 might be use the [repository pattern](https://developer.android.com/codelabs/basic-android-kotlin-training-repository-pattern#3), e.g. each model might have a repository that performs tasks like Realm CRUD, but also deciding when and how to fetch remote data. Models would be reduced to schema declaration and schema mapping functions, and most of the work would live in the repository. I've mostly encountered this in Android development, so it might not be a great fit in an RN app, but in principle it seems like a reasonable way to isolate data from presentation.

kueda:
Here are some slightly more fleshed out thoughts I've had on the repository pattern in case anyone else is willing to take a stab at it. I started thinking about it by writing some component code that might use a repository:

```javascript
// Get a repository for Observation records using a hook. Note that we don't
// hand it a Realm connection or an API token; it should take care of those
// things
const repository = useRepository( "Observation" );

[...]

PR #183, opened 2022-09-19
Use RealmProvider

kueda:
Uses `<RealmProvider>` from Realm to provide a single Realm connection for most of the app as a step toward #149. Also does some work on #158 by adding `useApiToken` and `useCurrentUser` hooks. This approach necessitates passing a Realm connection from the component context to helper methods, which is generally pretty awkward, but hopefully not too disruptive.

Unfortunately, to get this to work I needed to update Realm slightly, which broken the image resizer, at least for me, so that had to be updated and our use of it changed a bit to get that to work.


src/components/SharedComponents/ObservationViews/hooks/useUploadObservations.js
```diff
@@ -2,12 +2,18 @@
```
albullington, line 2: Is uploading working consistently with `RealmProvider`? I tried uploading three local observations (two from the camera and one from photo library), and only one made it to staging, without a photo attached.

src/models/Observation.js
```diff
@@ -159,17 +152,19 @@ class Observation extends Realm.Object {
...
     return localObs;
   }
 
   static async saveLocalObservationForUpload( obs, realm ) {
     // make sure local observations have user details for ObsDetail
-    const id = await getUserId( );
-    const user = id && realm.objectForPrimaryKey( "User", Number( id ) );
-    if ( user ) {
-      obs.user = user;
+    const currentUser = realm.objects( "User" ).filtered( "signedIn == true" )[0];
```
albullington, line 165: We have so many different ways to check for a current user...

src/models/Observation.js
```diff
@@ -424,9 +420,9 @@ class Observation extends Realm.Object {
     try {
       response = await inatjs.observations.create( uploadParams, options );
     } catch ( uploadError ) {
-      const body = JSON.parse( await uploadError.response.text( ) );
-      console.error( "[ERROR] Failed to upload observation: ", JSON.stringify( body ) );
-      return body;
+      const errorText = await uploadError.response.text( );
+      uploadError.message = errorText;
+      throw uploadError;
```
albullington, line 425: I think this might be the reason uploads are no longer working for me. I'm getting an error thrown: `Error: {"status":"401","errors":[{"errorCode":"401","message":"JWT is missing"}]}`

src/models/Photo.js
```diff
@@ -23,22 +24,43 @@ class Photo extends Realm.Object {
...
 
-  static async resizeImageForUpload( path ) {
+  static async resizeImageForUpload( pathOrUri ) {
     const width = 2048;
     const { photoUploadPath } = Photo;
-
     await RNFS.mkdir( photoUploadPath );
+    let outFilename = pathOrUri.split( "/" ).slice( -1 ).pop( );
+
+    // If pathOrUri is an ios localIdentifier, make up a filename based on that
```
albullington, line 33: Do we need to create a new filename here? The way I thought this was working is that resizeImage creates a new filepath in a temp directory, which we immediately save to a permanent filepath within `/photoUploads`. Maybe this was working for iOS photos in the older version of `react-native-image-resizer`, but not the latest version?

albullington (changes requested):
Code definitely looks cleaner without needing to open realm everywhere. The main issues I'm running into in this branch are:

1) taking a photo from the camera (this works, but it throws errors)
2) deleting a photo from the camera photo preview or from obs edit photo preview. (does not work)
3) uploading photos (works inconsistently for some observations/photos but not others)

package.json
```diff
@@ -77,7 +77,7 @@
     "react-native-vision-camera": "^2.13.0",
     "react-spring": "^8.0.27",
     "react-tinder-card": "^1.4.5",
-    "realm": "^10.20.0-beta.1",
+    "realm": "^10.20.0-beta.3",
```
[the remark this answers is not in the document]
kueda: Probably, but I think I'd prefer to punt that to another issue in favor of maintaining stability. This version is working in this branch for now.

src/components/LoginSignUp/AuthenticationService.js
```diff
@@ -358,6 +382,11 @@ const getUserId = async (): Promise<string | null> => {
   return currentUserId;
 };
 
+const isCurrentUser = async ( username: string ): Promise<boolean> => {
```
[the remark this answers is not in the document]
kueda: Maybe. It's probably a little less of a performance / dependency liability b/c it just hits `react-native-sensitive-info`.

src/components/ObsEdit/ObsEdit.js
```diff
@@ -115,7 +115,7 @@ const ObsEdit = ( ): Node => {
 
   const setPhotos = uris => {
     const updatedObservations = observations;
-    const updatedObsPhotos = currentObs.observationPhotos.filter( obsPhoto => {
+    const updatedObsPhotos = Array.from( currentObs.observationPhotos ).filter( obsPhoto => {
```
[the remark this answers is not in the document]
kueda: Yeah, one of the major changes I've hit in this branch is that the `.map( o => o )` thing we've been doing to convert Realm objects to POJOs doesn't seem to work any more, but `Array.from` seems to work.

src/components/SharedComponents/ObservationViews/ObservationViews.js
```diff
@@ -87,7 +88,7 @@ const ObservationViews = ( {
     if ( name !== "Explore" && isLoggedIn === false ) {
       return <EmptyList />;
     }
-    return null;
+    return <PlaceholderText text="No observations yet" />;
```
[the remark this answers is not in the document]
kueda: Sure, I'll change it.

src/components/SharedComponents/ObservationViews/hooks/useUploadObservations.js
kueda: Definitely found and fixed some bugs in recent commits.

src/models/Observation.js
kueda: Agree, and I'm not making it better in this PR, but maybe the subject of a future issue?

src/models/Observation.js
kueda: This was intentional. I'm not a huge fan of catching all errors all the time. Some errors really should never happen, so if they do, I think they should bubble up and not be caught, so we learn about them. In this case, it's not this different error handling that's causing the problem, it's that for some reason attempts are being made to upload an obs with a blank API token.

src/models/Photo.js
kueda: It maybe a bug with `react-native-image-resizer`, but I was finding that it wasn't able to handle iOS localIdentifiers like old versions did, so I had to use `RNFS.copyAssetsFileIOS` instead, which does require you to specify and absolute output path.

albullington (changes requested):
Uploading from the camera + photo library is working for me now, and so is deleting a photo from the photo preview in camera.

I'm still getting a "Realm context not found" error when I try to access the MediaViewer and delete a photo from ObsEdit.

Otherwise, this looks ready to merge.

Issue #123, opened 2022-07-06
Fetch current location for new observations


kueda:
When the user lands on Obs Edit for a *new* observation

- [x] Request permission to access the user's location (probably good to use `<PermissionGate>`
- [x] Pull any existing location retrieval out of the models
- [ ] Fetch the user's current location if the observation's location is not already set, and continue fetching it until accuracy drops below 5 m
- [x] When the location updates, set the observation `latitude`, `longitude`, and `positional_accuracy` to match
- [x] When the latitude and longitude change, reverse geocode the coordinates to set the `place_guess`
- [x] Update the displayed location every time the observation gets updated

When *not* to fetch the location:

* for existing observations
* when the user has landed on Obs Edit by importing media (as opposed to capturing new media or making an obs w/o media)

PR #184, opened 2022-09-22
123 permission fetch location

meru20:
This PR resolves issue #123 

- [x] Request permission to access the user's location (probably good to use <PermissionGate>
- [x] Fetch the user's current location if the observation's location is not already set, and continue fetching it until accuracy drops below 15 m
- [x]  Pull any existing location retrieval out of the models
- [x] When the location updates, set the observation latitude, longitude, and positional_accuracy to match
- [x]  Update the displayed location every time the observation gets updated
- [x] Not fetch location for existing observation

### DIFF
--- a/src/components/ObsEdit/ObsEdit.js
+++ b/src/components/ObsEdit/ObsEdit.js
@@ -58,6 +58,7 @@ const ObsEdit = ( ): Node => {
   const [deleteDialogVisible, setDeleteDialogVisible] = useState( false );
   const [fetchedLocation, setFetchedLocation] = useState( false );
   const [positionalAccuracy, setPositionalAccuracy] = useState( 1000000 );
+  const [mounted, setMounted] = useState( true );
 
   const disableAddingMoreEvidence = photoUris.length >= MAX_PHOTOS_ALLOWED;
 
@@ -133,7 +134,7 @@ const ObsEdit = ( ): Node => {
         }
       };
 
-      if ( !fetchedLocation && !currentObs._created_at ) {
+      if ( !fetchedLocation && !currentObs._created_at && mounted ) {
         if ( positionalAccuracy >= 15 ) {
           fetchLocation();
         } else {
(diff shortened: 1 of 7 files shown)


src/models/Observation.js
```diff
@@ -60,8 +61,20 @@ class Observation extends Realm.Object {
...
+    return {
+      ...obs,
+      captive_flag: false,
+      geoprivacy: "open",
+      owners_identification_from_vision: false,
+      observed_on_string: createObservedOnStringForUpload( ),
+      quality_grade: "needs_id",
+      uuid: uuid.v4( )
+    };
+  }
```
kueda, line 74: You've made a new model method that makes a new observation without any location information... but that's not really *importing* an observation, since you're not getting info from elsewhere, you're just making an obs with different attributes.

My request in the issue was to completely remove any location fetching from the model, so IMO that doesn't require a different method like this, it just requires taking those lines out of the `new()` method and performing location fetching from elsewhere, like an effect in the component.

src/components/ObsEdit/ObsEdit.js
```diff
@@ -113,6 +118,34 @@ const ObsEdit = ( ): Node => {
...
+      const fetchLocation = async () => {
+        const location = await fetchUserLocation( );
+
+        updateObservationKeys( {
+          place_guess: location?.place_guess,
+          latitude: location?.latitude,
+          longitude: location?.longitude,
+          positional_accuracy: location?.positional_accuracy
+        } );
+        if ( location && mounted ) {
```
kueda, line 132: This works, though instead of `mounted` I think you could check for the presence of `currentObs`.

src/components/ObsEdit/ObsEdit.js
```diff
@@ -113,6 +118,34 @@ const ObsEdit = ( ): Node => {
...
+          setPositionalAccuracy( location.positional_accuracy );
+        }
+      };
+
+      if ( !fetchedLocation && !currentObs._created_at ) {
```
kueda, line 137: The `!fetchedLocation` means this only fetches the location when `fetchedLocation` is blank. That means if you fetch it and it sets the location but the accuracy is above threshold, it never fetches the location again.

kueda (changes requested):
Very close, just need to keep fetching the location until it's below threshold.

PR #185, opened 2022-09-27
Second exploration of Tailwind CSS

albullington:
I did a fresh install of Node, then installed Tailwind and Nativewind, and redid the styling for the login screen. So far, this seems to be working for me without dependency bugs. Creating a PR to make sure tests are passing outside of my local environment.


albullington:
@kueda curious if this is working in your dev environment (with Node 16) or if you run into bugs.

Using [`styled()`](https://www.nativewind.dev/api/styled) to create styled components looks like one way we might be able to avoid all the `FlowIgnore` statements. Flow isn't happy with using `className` as a prop on RN components.

kueda (commented):
Everything installed and built smoothly, and the login view is looking real nice.

albullington:
This is ready for a re-review. Using Tailwind for the ObsList screen (grid view and list view), as well as the footer and the generic view/scrollview screens. Updated realm to get jest running, and added the team id to Podfile to get the app running on the latest iOS OS with an M1 Mac.

We'll want to convert all of our existing stylesheets over to Tailwind at some point, and I think this PR provides a solid example for how to do so.

src/components/MediaViewer/CustomImageZoom.js
```diff
@@ -1,16 +1,16 @@
...
 import ImageZoom from "react-native-image-pan-zoom";
-import { imageStyles, viewStyles } from "styles/mediaViewer/mediaViewer";
 
 // lifted from this issue: https://github.com/ascoders/react-native-image-zoom/issues/42#issuecomment-734209924
 
-const { width } = Dimensions.get( "screen" );
+const { width, height } = Dimensions.get( "screen" );
 // $FlowIgnore
-const selectedImageHeight = imageStyles.selectedPhoto.height;
+const SELECTED_IMAGE_HEIGHT = height - 350;
```
kueda, line 13: Why is this hardcoded?

src/components/styledComponents.js
```diff
@@ -0,0 +1,41 @@
...
+import { styled } from "nativewind";
+import {
+  Image as StyledImage,
+  KeyboardAvoidingView as StyledKeyboardAvoidingView,
+  Pressable as StyledPressable,
+  SafeAreaView as StyledSafeAreaView,
+  ScrollView as StyledScrollView,
+  Text as StyledText,
+  TextInput as StyledTextInput,
+  View as StyledView
```
kueda, line 12: Not a blocker, but I would do this as `View as OriginalView`. The return value of the `styled()` function is presumably "styled", not the input.

kueda (approved):

PR #188, opened 2022-09-29
Use babel-plugin-module-resolver

### LINKED ISSUE #157: Use babel-plugin-module-resolver
The goal here would be to replace imports like
```javascript
import { closeButton } from "../../styles/login/login";
```
with
```javascript
import { closeButton } from "styles/login/login";
```

so you're not always wondering what the relative path to commonly-used files is. We might just want aliases for every child of the `src/` directory.

albullington:
Closes #157; makes it so we don't need to know the relative path of every imported file.


kueda (approved):
Looks good! My one comment shouldn't block merge.

PR #194, opened 2022-10-11
Move hooks into sharedHooks folder; reorganize Observations and models/

### LINKED ISSUE #155: Extract re-used hooks into shared folders
Review all hooks and where they are used. If a hook file lives in a component-specific directory but gets used by multiple components elsewhere in the tree, move it to `sharedHooks`. For example, `src/components/Observations/hooks/useCurrentUser.js` should probably be at `src/sharedHooks/useCurrentUser.js`

albullington:
Closes #155, which ideally makes it easier for us all to find and reuse hooks throughout the app.

Also moves Observation-related code into the Observations folder, instead of SharedComponents/ObservationViews where it previously lived.

And renames the models/ folder to realmModels/ so we can use the babel alias instead of relative paths.


kueda (approved):
Did some merge-from-main fun, but no real changes.

PR #195, opened 2022-10-18
Use authenticated queries for fetch/search requests

albullington:
This PR converts most of the inatjs fetch and search queries to `useAuthenticatedQuery` & moves actual API calls into the `/api` folder.

It upgrades realm to `11.0.0-rc.0` and node to `16.17.0`, which gets the jest test runner working, but creates some additional problems where tests are not passing due to the realm setup.

The ultimate goal here is to also add `useAuthenticatedMutation` and get all API calls using React Query, but it seems better to split that into a second PR since there is still a lot of work to be done on that.


src/api/observations.js
```diff
@@ -0,0 +1,91 @@
...
+    return handleError( e );
+  }
+};
+
+export {
+  faveObservation,
+  fetchRemoteObservation,
+  searchObservations,
+  unfaveObservation
+};
```
kueda, line 91: Not a blocker, but I worry that these wrappers are kind of just replicating inatjs. Are we going to make a separate method for every inatjs method here? Is there a way we can consolidate? Here's one way we do something similar on the web: https://github.com/inaturalist/inaturalist/blob/main/app/webpack/observations/show/ducks/observation.js#L547-L561

kueda (approved):
No blockers. I fixed a few bugs.

PR #206, opened 2022-10-25
Create / update / delete requests with useMutation

albullington:
Using React Query's `useMutation` to streamline the way we create, update, and delete server data. This should make it easier for devs to know where API code should live, handle errors, handle loading states, and refetch data.

This also makes it so we don't have API requests sprinkled throughout the UI.

With this PR, nearly all inatjs requests are in the `/api` directory. A few exceptions are the ones related to fetching observations with infinite scroll & uploading observations created in the mobile app. I left those alone for now since that code is more convoluted & involves coordinating realm updates, and my priority here is making it easy for devs to see consistency in how we make API requests throughout most of the app.


src/components/ObsDetails/ActivityItem.js
```diff
@@ -48,6 +52,18 @@ const ActivityItem = ( {
     isActiveUserTheCurrentUser( );
   }, [user] );
 
+  const handleSuccess = {
+    onSuccess: ( ) => {
+      queryClient.invalidateQueries( ["fetchRemoteObservation", item.uuid] );
+      refetchRemoteObservation( );
+    }
+  };
```
kueda, line 60: I know I harp on naming a lot, but a) this object literal doesn't really handle success, it's more like a configuration object for `useMutation()`, and b) is there a reason to define a separate constant here? Why not define the object literal when you call `useAuthenticatedMutation`?

src/components/ObsDetails/ActivityItem.js
```diff
@@ -48,6 +52,18 @@ const ActivityItem = ( {
...
+
+  const deleteCommentMutation = useAuthenticatedMutation(
+    ( uuid, optsWithAuth ) => deleteComments( uuid, optsWithAuth ),
+    handleSuccess
```
kueda, line 64: `useAuthenticatedMutation` takes three arguments: `queryFunction`, `handleCallback`, and `queryOptions`. As far as I can tell, the only reason this might be working is that you're basically treating the second arg (`handleCallback`) as if it were an object literal of options you pass to `useMutation`, so...

1. Please rename the second argument for `useAuthenticatedMutation` so it describes what it is, maybe something like `mutationOptions`
2. Why isn't flow complaining about this? The arg is declared as `handleCallback: Function`, and yet right here you're passing it an object literal. Shouldn't flow catch that and complain?

src/components/ObsDetails/ActivityItem.js
```diff
@@ -84,7 +100,10 @@ const ActivityItem = ( {
               <KebabMenu>
                 <Menu.Item
                   onPress={async ( ) => {
-                    Comment.deleteComment( item.uuid, realm, apiToken );
+                    // first delete locally
+                    Comment.deleteComment( item.uuid, realm );
+                    // then delete remotely
+                    deleteCommentMutation.mutate( item.uuid );
```
kueda, line 106: Is the lack of error handling intentional here? That's fine, just want to understand the scope of what you were going for.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -70,31 +71,84 @@ const ObsDetails = ( ): Node => {
   const bottomSheetModalRef = useRef( null );
   const [addingComment, setAddingComment] = useState( false );
   const [snapPoint, setSnapPoint] = useState( 100 );
-  const apiToken = useApiToken( );
-  const [localObservation, setLocalObservation] = useState( null );
+  const realm = useRealm( );
+  const localObservation = realm?.objectForPrimaryKey( "Observation", uuid );
```
kueda, line 75: This fetches the local observation on every render. Why do we need to do that?

src/api/comments.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+
+import inatjs from "inaturalistjs";
+
+import handleError from "./error";
+
+const createComments = async (
```
kueda, line 7: This creates a comment, not multiple comments, so I'd prefer `createComment`. Same elsewhere.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -70,31 +71,84 @@ const ObsDetails = ( ): Node => {
...
+      Alert.alert(
+        "Error",
+        error,
+        [{ text: t( "OK" ) }],
+        {
+          cancelable: true
+        }
+      );
+    }
+  };
```
kueda, line 128: Could you (or should you) refactor this to use `optimistic updates`? Maybe you could just get rid of `onIDAdded()`. Maybe a subject for later work.

kueda (changes requested):
There's a lot here so I basically only got through the observation-related stuff, but I figured that was probably enough feedback for one round.

src/components/ObsDetails/ActivityItem.js
albullington: I was trying to keep existing functionality and not add/change/break functionality. So in this case, there's no error handling because that wasn't there before, either.

src/components/ObsDetails/ObsDetails.js
albullington: We don't necessarily need to, but this fixes a bug where `ObsDetails` can't load because of the way state is set for `localObservation`. We're currently getting a `Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.` error and a white screen on observations made locally.

src/components/ObsDetails/ObsDetails.js
albullington: Optimistic updates as in show the user the regular ID result with black text instead of the ghost result with gray text? I'm down to do that, but was trying to keep functionality the same as it was before in this PR.

kueda: Meant to include a link. Was thinking about https://react-query-v3.tanstack.com/guides/optimistic-updates. Happy to just maintain prior functionality though.

albullington: This might be good to explore in a separate issue. I poked around at this for about an hour, but couldn't immediately get it working for updating a comment.

src/components/ObsDetails/ObsDetails.js
kueda: Was that coming from https://github.com/inaturalist/iNaturalistReactNative/pull/206/files#diff-f3017450aeb5c7dc7b74f3e729e38fce3460974a5a5344677a088d935d568e82L87-L89? I can't replicate that bug if I add something like that back, but wouldn't the solution there be to figure how why the items in the dependency array keep changing?

kueda (commented):
Sorry, been trying to re-review this and I'm stuck on the fact that adding a comment doesn't actually update the UI... which might not be related to the useMutation stuff, but might be a side-effect of the other changes in ObsDetail. As far as I can tell a new network copy of the obs *is* getting created, but it's not actually updating the local database, so nothing changes in the UI. I tried adding something simple to address that bit it seems like the remote copy of the observation always has a a value of `updated_at` that is out-of-date, and I think it might be related to some caching that react query is doing.

src/components/ObsDetails/ActivityItem.js
```diff
@@ -84,7 +98,10 @@ const ActivityItem = ( {
...
```
kueda, line 104: Not a blocker, but FWIW, I think this is actually the reverse of the order we probably want for comments. Ideally, everything would be deleted remotely first, and when we're sure that worked, deleted locally, because the server is the ultimate source of truth, and also because the network request might fail, so you probably only want to alter local state when you're sure the server has been updated.

For models that have to work offline (like observations) things are more complicated b/c we have to let the app get out of sync with the server and then resolve discrepancies, but comments can be simpler.

src/components/ObsDetails/ObsDetails.js
kueda: I guess this doesn't need to block, but this still seems like a problem to me, or rather a solution to a problem I can't replicate.

src/realmModels/Observation.js
```diff
@@ -408,13 +390,30 @@ class Observation extends Realm.Object {
...
+    if ( obs?.observationSounds?.length > 0 ) {
+      return Observation.uploadEvidence(
+        obs.observationSounds,
+        "ObservationSound",
+        ObservationSound.mapSoundForUpload,
+        id,
+        inatjs.observation_sounds,
+        realm,
+        options
+      );
```
kueda, line 416: Again, no need to block, but if we're trying to get network stuff out of the model, maybe this entire method belongs in `useUploadObservations`.

kueda (approved):

PR #207, opened 2022-10-25
Clear old observations when creating a new observation

### LINKED ISSUE #204: Chosen photos persist after cancelling new obs
**Describe the bug**
When you start a new obs by choosing from the gallery then back out and add one with no evidence, the photo you chose is still selected.

**To Reproduce**
1. Start a new obs by choosing a photo from the gallery
2. On Obs Edit, tap the back button in the upper left to return to the Camera Roll, then again to return to My Observations
3. Start a new obs by choosing no evidence
4. You land on obs edit, but the photo you chose in the first step has already been added to the obs

**Expected behavior**
When you start a new observation, the slate should be clean. Nothing should persist from previous observations you added or didn't add.

 - Device: iPhone 13 simulator
 - OS: iOS 15.5
 - App version: 0.1.1 (27)

albullington:
Closes #204 by clearing out whatever observations were on the previous ObsEdit screen before navigating to create a new observation.


kueda (changes requested):
This solves the problem specifically addressed by the repro steps, but not the more fundamental problem that the ObsEdit context persists even after you've left the components that use it. For example, here's a related bug:

1. Sign in to an account with some existing observations
2. View an observation with a photo and edit it to get to ObsEdit. The photo should appear.
3. Back out to My Observations and view another obs with a photo and edit it. On ObsEdit, you see the photo of the *previous* observation. Actually, you're just seeing the previous observation, b/c that's what's in context

One approach to that problem would be to make sure ObsEdit clears out the obs in context if it receives a UUID in the params, but maybe you could address all problems by making sure the context is only shared among the components that actually need it? Or ensuring that when you exit ObsEdit for any reason the context gets cleared out?

kueda (changes requested):
I think this is still treating the symptom. For example, if you use the Android back button instead of the app back button, `handleBackButtonPress` never fires and you can reproduce the "editing two observations" bug. Some ideas: since ObsEdit is wrapped in Mortal, it should *actually* dismount, so in theory you should be able to use a `useEffect` cleanup function to wipe the context when you leave ObsEdit for *any* reason. Another idea is to use the react-navigation `useFocusEffect` hook to wipe the context when ObsEdit blurs.

albullington:
The way `ObsEdit` is currently written, we don't want to clear context when exiting `ObsEdit` for any reason. If we do that, we lose the ability to navigate to the `AddID` screen and return to `ObsEdit` with an identification selection, as well as the ability to navigate to the `TaxonDetails` screen from `AddID` and then return to `AddID`, then `ObsEdit`.

We don't currently have a working location picker / location map screen, but I imagine this is another screen where we'll need to be able to navigate away from `ObsEdit` and come back without clearing context. (This could be a modal, rather than a separate navigation screen).

And we also need to be able to add evidence to an existing observation in `ObsEdit`, which means returning to the camera/gallery/sound recorder without wiping the existing observation context.

Some ways we could address the issue you raised:
1. We could add code specifically to clear context when the Android back button is pressed using the [BackHandler](https://reactnative.dev/docs/backhandler)
2. We could try turning AddID, location picker, and add evidence (camera, gallery, sound recorder) into modals from `ObsEdit` instead of allowing navigation to separate screens
3. We could clear context anytime a user leaves ObsEdit *except* for when they're navigating to the AddID screen, location picker, or tapping the add evidence button

Any preference or other ideas? 1 and 3 both seem pretty reasonable to me, and 1 would involve a lot less conditional logic.

kueda:
Hah, I was going to say I prefer 2 since all those views off of ObsEdit seem like modal experience to me (i.e. single-screen interactions that only serve to alter the state of the parent view). What's you're aversion to that approach? 1 and 3 both seem pretty conditional to me, but I guess they're preferable if 2 is massively complicated.

albullington:
2 seems pretty complicated to me if we're allowing a user to navigate through those screens (say Camera Options -> PhotoLibrary -> Group Photos -> ObsEdit) and also need to support those some of those screens in modal mode, with the same state as they had from the user navigating through (say, if a user returns to the PhotoLibrary and needs to be able to see what photos they already have selected for their existing observations)

kueda:
Fair enough. Let's just go with the BackHandler approach for now, then.

kueda (changes requested):
Still not working for me with these repro conditions:

1. Sign in to an account with some existing observations
1. View an observation with a photo and edit it to get to ObsEdit. The photo should appear.
1. Back out to My Observations and view another obs with a photo and edit it. On ObsEdit, you see the photo of the previous observation. Actually, you're just seeing the previous observation, b/c that's what's in context

It seems like after you back out and handleBackButtonPress fires, ObsEdit renders again and populates the context, basically undoing the effect of handleBackButtonPress.

albullington:
Fixed, I think.

kueda (approved):

PR #208, opened 2022-10-27
Realm 11

jtklein:


kueda (changes requested):
FWIW, I can get that one test to pass by upgrading react-native-paper to 5.0.0-rc.9.

Some other issues:

1. iOS app launch fails with what looks like a native error: "Urecognized font family 'Whitney-Medium'
2. In Android I get multiple errors about invalid hook calls when I go to ObsEdit

jtklein:
Yes, I can also get that one test to pass by upgrading react-native-paper to 5.0.0-rc.9.

The other two errors raised above should also be fixed now.

However, now there is a different error in npm run test: "Camera: Native Module 'CameraView' was null! Did you run pod install?" Does that sound familiar?

jtklein:
All tests are passing now, and CI exits without Error.

android/app/src/main/java/com/inaturalistreactnative/MainApplication.java
```diff
@@ -7,23 +7,25 @@
...
 import com.facebook.react.ReactPackage;
+import com.facebook.react.config.ReactFeatureFlags;
 import com.facebook.soloader.SoLoader;
+import com.inaturalistreactnative.newarchitecture.MainApplicationReactNativeHost;
 import java.lang.reflect.InvocationTargetException;
 import java.util.List;
 import com.lugg.ReactNativeConfig.ReactNativeConfigPackage;
 import com.facebook.react.bridge.JSIModulePackage;
-import com.swmansion.reanimated.ReanimatedJSIModulePackage;
+// import com.swmansion.reanimated.ReanimatedJSIModulePackage;
```
kueda, line 17: If we don't need the stuff you've commented out in this file, would you please just remove those lines?

android/gradle.properties
```diff
@@ -9,8 +9,12 @@
...
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx1024m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+# Default value: -Xmx512m -XX:MaxMetaspaceSize=256m
+org.gradle.jvmargs=-Xmx2048m -XX:MaxMetaspaceSize=512m
+
+# need this to avoid out of memory error on device
+# https://stackoverflow.com/questions/57606462/a-failure-occurred-while-executing-com-android-build-gradle-internal-tasks
+org.gradle.jvmargs=-Xmx4608m
```
kueda, line 17: Isn't this just overriding the value on line 13? Do we need both?

kueda (changes requested):
You did what I could not, sir! I salute you. Only minor changes needed. Are most of the Android files just stuff to support the new architecture?

android/app/src/main/java/com/inaturalistreactnative/MainApplication.java
jtklein: Have removed the commented out code

android/gradle.properties
jtklein: Have combined those two lines

jtklein:
Yes, the changes in the Android project files are mostly changes introduced into RN to accommodate the new architecture. All of the new files are for the new architecture. A few changes in the Android files are for updating gradle.

kueda (approved):

PR #209, opened 2022-10-28
Replacing styles/colors.js with tailwind theme colors. Closes #196.

### LINKED ISSUE #196: Replace styles/colors.js with Tailwind
Now that we've made the switch to Tailwind, let's make sure all colors are stored in `tailwind.config.js` and not in `styles/colors.js`

- [ ] Make sure all the named colors in `styles/colors.js` are in `tailwind.config.js`
- [ ] Replace all imports of `styles/colors.js` in the source files with `import colors from "tailwindcss/colors";`
- [ ] Delete `styles/colors.js`

angielt:
Created tailwind-colors.js which includes a resolveConfig to access tailwind values in javascript.


tailwind-colors.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import resolveConfig from "tailwindcss/resolveConfig";
+
+import tailwindConfig from "./tailwind.config";
+
+// $FlowIgnore
+const fullConfig = resolveConfig( tailwindConfig );
+
+export default fullConfig.theme.colors;
```
kueda, line 10: FYI, our convention is for camelcase file names, so if you were to use this approach, I would name this file `tailwindColors`, not `tailwind-colors`.

tailwind-colors.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+
+import resolveConfig from "tailwindcss/resolveConfig";
+
+import tailwindConfig from "./tailwind.config";
+
+// $FlowIgnore
+const fullConfig = resolveConfig( tailwindConfig );
```
kueda, line 8: What's the advantage of using `resolveConfig()` as opposed to importing from `tailwind.config.js` wherever we need it?

src/components/Camera/CameraOptionsModal.js
```diff
@@ -6,7 +6,8 @@ import { t } from "i18next";
 import { ObsEditContext } from "providers/contexts";
 import * as React from "react";
 import { IconButton } from "react-native-paper";
-import colors from "styles/colors";
+
+import colors from "../../../tailwind-colors";
```
kueda, line 10: Whether we use this file or import directly from `tailwind.config.js`, I think it would help a lot if we used a module alias in `babel.config.js` so we're not constantly juggling these relative paths.

kueda (changes requested):
This works and addresses the problem in the issue, and I can see why the `import colors from "tailwindcss/colors"` approach I recommended won't work, but can you explain why the `resolveConfig()` approach is necessary?

tailwind-colors.js
angielt: When just importing tailwind.config.js seems like the colors from the config file could not be accessed in the code, at least dynamically. It would try to find the default colors from tailwind that didn't exist. 
This link to the docs explains a bit more: 
https://tailwindcss.com/docs/configuration#referencing-in-java-script

kueda: Fair enough.

babel.config.js
```diff
@@ -17,7 +17,8 @@ module.exports = {
         realmModels: "./src/realmModels",
         sharedHelpers: "./src/sharedHelpers",
         sharedHooks: "./src/sharedHooks",
-        styles: "./src/styles"
+        styles: "./src/styles",
+        colors: "./tailwindColors"
```
kueda, line 21: Can we just put this in styles and refer to it as `styles/tailwindColors`? I think that helps emphasize that it's one of our files and not some mysterious node module called `colors`.

kueda (changes requested):

PR #211, opened 2022-11-02
201 restore profile pic and obs count to the my obs header

angielt:
Changed uri source to get icon data from production instead of staging.


src/components/Observations/UserCard.js
```diff
@@ -29,10 +29,15 @@ const UserCard = ( ): Node => {
   if ( !user ) { return <View className="flex-row mx-5 items-center" />; }
   const navToUserProfile = ( ) => navigation.navigate( "UserProfile", { userId: user.id } );
 
+  const productionIcon = remoteUser?.icon_url.replace( "staticdev", "static" );
```
kueda, line 32: You need to guard against `icon_url` being blank or null here, i.e. the user has no icon. In that situation, I think you probably just don't want to show an image at all.

kueda (changes requested):
Just the one change and I think this looks good.

kueda (approved):

PR #220, opened 2022-11-04
Fixes for ObsEdit & adding evidence to an existing observation

### LINKED ISSUE #217: Adding a photo on ObsEdit adds the original photo again
**Describe the bug**
When you add a photo and land on ObsEdit, when you try to add another photo the original photo gets duplicated. Bonus, if the GPS is actively retrieving coordinates, it constantly resets state so the UI blinks between the original state of 1 photo and the new state of 3 photos.

**To Reproduce**
1. Add a new observation by choosing from the gallery
2. When you land on ObsEdit, tap the "+" button to get to the gallery. Your original photo is still chosen, which seems wrong, but tap another photo and tap "Import 2 photos"
3. You land on ObsEdit, where your original photo from step 1 is duplicated and your new photo is there, making 3 photos. Also, if the GPS is retrieving coordinates, the UI is blinking between your original 1-photo state and your new 3-photo state

**Expected behavior**
When you import another photo,
1. The gallery should not allow the original photo to be unselected and its selected state should be different than any new photos (e.g. a "disabled" checkmark")
2. When you tap a new photo in the chooser the button should read "Import 1 photo" i.e. the count should be the count of the photos you're about to import, not the total number of photos for the obs
3. When you land back on ObsEdit, only the original photo and the new photo(s) you chose should appear
4. Location fetching should not affect any part of the UI other than the coordinates.

 - Device: Pixel 3 / iPhone 13 simulator
 - OS: Android 12 / iOS 15.5
 - App version: Build 28

### LINKED ISSUE #218: Existing obs photos appear in StandardCamera
**Describe the bug**
When you enter the StandardCamera from ObsEdit and ObsEdit was already showing some photos, those photos appear as thumbnails in the StandardCamera... even though you didn't just take them

**To Reproduce**
1. Start a new obs with the camera or from the gallery
2. When you land on ObsEdit, you should see your photo. Tap the "+" button and tap the camera button to enter the StandardCamera
3. The PhotoPreview at the top shows your existing observation photo

**Expected behavior**
The StandardCamera should never show photos from an existing observation, only photos the user just took in the Standard Camera.

 - Device: Pixel 3
 - OS: Android 12
 - App version: Build 28

albullington:
Closes #217 and closes #218.

Basically, this is an attempt to clean up the whole uploader flow. I renamed `ObsEditProvider` to `UploadProvider` to make it clear that this includes state for all the uploader screens, including the camera, photo gallery, and group photos screen.

I moved a lot of the logic related to creating an observation, obs photo, and photo out of the models/components and into the `UploadProvider` so everything is in one place.

I also fixed the logic related to adding evidence to an existing observation. Now when you return to the Camera or PhotoGallery, you should only see the images which were previously selected/shown on that screen, and you should be able to add evidence in the expected format, without any duplicate obsPhotos being created.

One other change is that this pulls the timestamp, location, and place_guess from the gallery photos, which we weren't doing before. When there are observations with multiple photos, we're taking that data from the first photo in the list.


kueda:
Lots here so I'll dig into this next week. Off the top though, why UploadProvider? Most of these views don't have anything to do with uploading, i.e. sending data from the client to the server. IMO ObsEditProvider is a more descriptive name, but maybe I'm missing something.

albullington:
Upload made sense to me because internally we keep referring to all the screens within this provider as the uploader flow. Not attached to it though.

src/providers/UploadProvider.js
```diff
@@ -0,0 +1,217 @@
...
+    const newObs = {
+      latitude,
+      longitude,
+      time_observed_at: formatDateAndTime( photo.timestamp ),
+      place_guess: placeGuess
+    };
+    return Observation.new( newObs );
+  }, [] );
+
+  const createObservationsFromGroupedPhotos = useCallback( async obs => {
```
kueda, line 71: Can we stop using "obs" as an abbreviation? I always get confused about whether it means a single observation or multiple observations.

src/providers/UploadProvider.js
```diff
@@ -0,0 +1,217 @@
...
+        setCurrentObsIndex( observations.length - 1 );
+        setObservations( observations );
+      } else {
+        observations.splice( currentObsIndex, 1 );
+        setCurrentObsIndex( currentObsIndex );
+        // this seems necessary for rerendering the ObsEdit screen
+        setObservations( [] );
+        setObservations( observations );
+      }
+    };
```
kueda, line 164: This method doesn't seem to get called. Can we delete it?

kueda (changes requested):
I think I like putting logic in the provider... even if I don't always love the way context adds the risk of state overstaying its welcome.

Two things that we should touch base with Abhas about:

1. I support moving components into folders that match the names we're using the Figma doc, but I think the "BulkUploader" should be called the "PhotoImporter" because it doesn't upload anything.
2. I think ObsEditContext is a more descriptive name than "UploadContext" because it doesn't store context for any kind of upload process, whereas it does store context for editing (and adding) observations.

Only other things are the very minor stuff I mentioned in comments.

kueda (approved):

PR #225, opened 2022-11-04
Replace react-native-localize with mock given by library

jtklein:


kueda (approved):
This is fine for now, though there might come a time when we want to test localization, and then we'll probably need a custom mock like the one we're removing here

PR #228, opened 2022-11-04
Fix obs detail comment input

### LINKED ISSUE #213: ObsDetail comment input over-writes what you're writing
**Describe the bug**
The comment input on ObsDetail seems to work, but while you're writing it tends to overwrite part of what you're writing with previous text, or place the cursor in weird positions

**To Reproduce**
1. Go to ObsDetail, any obs
2. Tap on "Add Comment
3. Start writing a comment. Write in full sentences and watch the the input; parts of your text will get deleted, others will get repeated or inserted in weird places

**Expected behavior**
The comment I write should remain the way I write it.

 - App version: [e.g. 1.0.0]

**Additional context**
I strongly suspect this has something to do with updating the state to much and triggering too many re-renders. Amanda suggested the `useDebounce` hook we use on the Settings screen. I sort of think the content of the input should be completely independent from component state until you hit the send button, but I'm up for whatever works.

albullington:
Closes #213. Fixes the input issue by using `defaultValue` instead of `value`, since `value` was pulling the text input field from state instead of letting TextInput handle its own text state.

Also moves `AddCommentModal` into a separate component, updates styling for react-native components to use Tailwind, and creates a standardized backdrop component which we can use for all bottom sheet modals.


src/components/ObsDetails/AddCommentModal.js
```diff
@@ -0,0 +1,132 @@
...
+    <BottomSheetStandardBackdrop props={props} />
+  );
+
+  const submitComment = async ( ) => {
+    setAddingComment( true );
+    clearComment( );
+    setShowCommentBox( false );
+    Keyboard.dismiss();
+    if ( comment.length > 0 ) {
+      console.log( comment, "comment in add comment modal" );
```
kueda, line 69: Probably don't need this. Need to a way to lint console.log but not console.error...

src/components/ObsDetails/AddCommentModal.js
```diff
@@ -0,0 +1,132 @@
...
+    // re-enable Add Comment button when backdrop is tapped to close modal
+    if ( index === -1 ) {
+      setShowCommentBox( false );
+    }
+  }, [setShowCommentBox] );
+
+  const renderTextInput = () => (
+    <BottomSheetTextInput
+      keyboardType="default"
+      style={[viewStyles.commentInput, viewStyles.commentInputText, textStyles.commentTextInput]}
```
kueda, line 85: No way to do this with Tailwind?

kueda (changes requested):
Mostly looking good, but for some reason "Add optional notes" on ObsEdit now pushes the input way up above the keyboard when you enter it. Seems like it might be related to these changes.

kueda (changes requested):
Now on my Pixel 3 there's a transparent gap between the soft keyboard and the input:

Screenshot_20221107-155203


and on my iPad the soft keyboard covers the input entirely:

IMG_0042

albullington:
I couldn't figure out how to use Tailwind to position an element above the keyboard -- any suggestions?

kueda:
@albullington I haven't quite figured it out myself yet, but a few thoughts:

1. You can set an arbitrary height with Tailwind like this: `bottom-[${keyboardOffset}px]`
2. I suspect the problem might stem from actually restyling the same element when the keyboard appears. When I use `bottom-[${keyboardOffset}px]` it pushes the keyboard really high... like it was ${keyboardOffset}px and not ${keyboardOffset}px from the bottom of the screen. So I'm wondering if the solution might be to render one component for displaying the text, but another for editing it, and the editable one only appears when the user taps the display text. The editable component would always have absolute positioning, so in theory there's no possility of changing the baseline for `bottom`.

albullington:
This seems to be happening in the main branch too and wasn't something I could figure out in the past 24 hours, so I made a [separate issue](https://github.com/inaturalist/iNaturalistReactNative/issues/239).

PR #234, opened 2022-11-08
Support for tablets. Closes #200.

### LINKED ISSUE #200: Support tablets
I just mean supporting full screen display on iOS and Android tablets, landscape and portrait, not all the layout changes that we'll need to make that actually look good. I'm expecting it to run on an iPad and an emulated Android tablet.

angielt:


kueda (changes requested):
Looks good to me aside from that one conflict.

kueda (approved):

PR #241, opened 2022-11-14
Enable hermes for iOS

jtklein:
I have tested on the simulator with Debug config to see if it compiles, and it did. I then tested it on a real iPhone with Release config and it compiled and opened the app. I am not aware of the current state of the app in terms of which bugs are known, so I did not test the app functions.

On the first opening of the app, directly after installation, the app crashed and closed. The second time opening was fine. This was also the behavior when compiling from main for me, so I don't see this as a regression.


kueda (approved):
Looks good to me, works in the simulator and on my iPad.

PR #242, opened 2022-11-15
Allow existing id to be edited in ObsEdit

### LINKED ISSUE #230: Remove iconic taxon buttons from ObsEdit for now
They don't currently do anything and the way they're supposed to work is still up in the air, so let's remove them for now. FWIW, I'm talking about the gray circles beneath the identification:

Screen Shot 2022-11-07 at 11 39 37 AM

### LINKED ISSUE #231: Can't edit an existing identification in ObsEdit
**Describe the bug**
When an observation has an identification, a pencil icon appears to its right that implies it can be edited, but tapping it does nothing.

**To Reproduce**
1. Edit an existing observation with an identification
2. Tap the pencil button to the right of the identification
3. Nothing happens

**Expected behavior**
The AddID view should appear when you tap that button, you should be able to choose a taxon, and change your identification.

**Screenshots**

**Additional context**

FWIW, this is actually a bit complicated because an observation can have many identifications by the same user, and you cannot change the taxon of an identification once it has been submitted, you can only withdraw or delete it. In the other native apps, the obs edit interface does not touch identifications. It just changes the `observation.taxon_id` property, submits that to the server, and the server deals with the complexities of adding a new ID and/or withdrawing an existing one. You could make the argument that the client should just manage that so the local UI updates correctly, but I'm not sure what approach is simpler here.

albullington:
- Removes iconic taxon buttons and some other unused components/stylesheets, which closes #230
 - Allows a user to tap the pencil icon and navigate to the AddID screen
 - Allows user to create a new id or update an existing id, which closes #231


src/components/ObsEdit/ObsEdit.js
```diff
@@ -49,11 +48,11 @@ const ObsEdit = ( ): Node => {
   useEffect( ( ) => {
     // when opening an observation from ObsDetails, fetch the local
     // observation from realm
-    if ( localObservation ) {
+    if ( localObservation && !currentObservation ) {
```
kueda, line 51: What was this for? I think it's making it so when you view obs detail, edit the obs, back out to My Obs, view another obs, edit that, you end up editing the original obs. In theory you can *only* edit local observations, right? so `localObservation` should always exist and if it does we should make 100% sure that's the current obs in context?

kueda (changes requested):

src/components/ObsEdit/ObsEdit.js
albullington: I added some more detailed comments here to clarify what this is for. Basically, we need to set the local observation in context only on the initial render, otherwise anytime a user tries to edit the identification, their identification will get overwritten by the local observation being set to context again.

This check means we only set the local observation into context once, and then the id can be edited as expected in ObsEdit.

kueda: FWIW, I *think* I addressed this in 28f4c2035f0c537ba83bd46ffe1647a28e100e4d

albullington:
Ok, I think this error should be fixed now. Before, we were only creating new observations on the server, so I added functionality to check whether an observation is already on the server. If it is, it will be updated with the `inatjs.observations.update` call. 

For the moment, we're skipping the `observation_photos` and `observation_sounds` updates for previously synced observations, since these don't seem to exist yet in API v2 ([#350](https://github.com/inaturalist/iNaturalistAPI/issues/350))

src/realmModels/Observation.js
```diff
@@ -285,6 +285,13 @@ class Observation extends Realm.Object {
     return unsyncedObs.length > 0;
   }
 
+  static wasPreviouslySyncedObservation = ( realm, obs ) => {
+    const previouslySyncedFilter = "_synced_at != null && _synced_at <= _updated_at";
+
+    const observationList = realm?.objects( "Observation" );
+    return observationList.filtered( `${previouslySyncedFilter} && uuid == "${obs.uuid}"` );
+  }
+
```
kueda, line 294: Why is this a static method? This seems like an instance method, like

```javascript
wasSynced( ) {
  return this._synced_at != null;
}
```

src/realmModels/Observation.js
```diff
@@ -371,27 +378,45 @@ class Observation extends Realm.Object {
...
+    // since these .update API calls aren't yet available in v2
+
+    if ( obs?.observationPhotos?.length > 0 && !wasPreviouslySynced ) {
       return Observation.uploadEvidence(
         obs.observationPhotos,
         "ObservationPhoto",
         ObservationPhoto.mapPhotoForUpload,
         id,
-        inatjs.observation_photos,
+        wasPreviouslySynced ? inatjs.observation_photos.update : inatjs.observation_photos.create,
```
kueda, line 408: This use of `wasPreviouslySynced` is redundant, right? You know it's false from line 402.

src/realmModels/Observation.js
```diff
@@ -371,27 +378,45 @@ class Observation extends Realm.Object {
...
+    }
+
     await Observation.markRecordUploaded( obs.uuid, "Observation", response, realm );
     const { id } = response.results[0];
-    if ( obs?.observationPhotos?.length > 0 ) {
+
+    // skip updating evidence for already synced observations
+    // since these .update API calls aren't yet available in v2
+
+    if ( obs?.observationPhotos?.length > 0 && !wasPreviouslySynced ) {
```
kueda, line 402: Did you find this was necessary? I'm pretty sure `POST /v2/observation_photos` behaves as an upsert. `POST www.inaturalist.org/observation_photos` does, specifically to accommodate the mobile apps.

kueda (changes requested):
Sorry, still a little confused about the changes to the Observation model.

src/realmModels/Observation.js
albullington: I don't think this is acting exactly as expected. I'm getting an undefined response back from `observation_photos` which have already been uploaded, which makes me think they're not upserting & also means we can't update the synced date on the client side.

For now, I decided to filter by evidence that hasn't been synced yet, so this can add new observation photos to an observation, but it can't update or delete observation photos. I think this is fine for now, since we also don't have a way to update observation photos in the client (like moving position) and we don't yet have an API v2 endpoint or working functionality in the client for deleting observation photos.

albullington:
Changed `wasSynced` to an instance method, simplified the code for uploading evidence, and for now only uploading evidence which has not previously been synced to the server.

kueda (approved):

src/realmModels/Observation.js
```diff
@@ -371,16 +390,31 @@ class Observation extends Realm.Object {
...
+      }, options );
+    } else {
+      response = await createObservation( uploadParams, options );
+    }
+
     await Observation.markRecordUploaded( obs.uuid, "Observation", response, realm );
     const { id } = response.results[0];
+
     if ( obs?.observationPhotos?.length > 0 ) {
       return Observation.uploadEvidence(
```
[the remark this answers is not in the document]
albullington: Changed these lines to await uploading photos and sounds instead of returning those responses.

PR #243, opened 2022-11-17
Standardize headers with react navigation

### LINKED ISSUE #226: Use React Navigation headers wherever possible
We have a lot of custom headers in the app at present, resulting in a lot of visual inconsistency and unnecessary code repetition. The somewhat open-ended task here is to work through the app screen by screen and try to make React Navigation headers work wherever possible... which I think should be almost everywhere given that the [buttons are customizable](https://reactnavigation.org/docs/header-buttons) and so is the [title](https://reactnavigation.org/docs/headers#replacing-the-title-with-a-custom-component). If a header *can't* be replaced, please explain that here in the issue so we can discuss. If there's really no way to do it except with a completely custom header, we need to explain why that's the case in the code.

albullington:
Closes #226 by removing custom headers and replacing them with react navigation headers.

One quirky thing about this is that left aligned titles are not supported on iOS, so we might eventually decide to replace the headers on some screens like Photo Importer and Group Photos with a customized header.


src/navigation/mainStackNavigation.js
```diff
@@ -100,6 +110,9 @@ const MainStackNavigation = ( ): React.Node => (
         <Stack.Screen
           name="PhotoGallery"
           component={PhotoGalleryWithPermission}
+          options={{
+            headerTitle: photoGalleryHeaderTitle
```
jtklein, line 90: I personally prefer when the header options are set in the screen component itself. So that the main navigation files are only about the logic of navigation. Then also if I look at a screen component I can see all of its layout including the header (e.g., including translated strings). So, if I look at the phone all of the screen I see I know is in one file (or subcomponents of course).

It was a bit easier when using class components. Then the header setting could be on top of the file, and always in the same location in the code. Now with functional components it would be in a useEffect hook in between the other hooks.

jtklein (approved):

src/navigation/mainStackNavigation.js
albullington: I guess I don't have a strong preference on this, but I kind of like that the headers live here. Since we're trusting react navigation to render the headers in a native way, it makes sense to me that they would be set in navigation, rather than where we're rendering the screen component.

I definitely don't love the idea of putting these in `useEffect` hooks except for the ones where we need an [interaction between the header and the screen component](https://reactnavigation.org/docs/header-buttons/#header-interaction-with-its-screen-component).

src/components/ObsDetails/ObsDetails.js
```diff
@@ -159,6 +158,19 @@ const ObsDetails = ( ): Node => {
...
+
+    const navToObsEdit = ( ) => navigation.navigate( "ObsEdit", { uuid: observation?.uuid } );
+    const editIcon = ( ) => ( obsCreatedLocally || obsOwnedByCurrentUser )
+    && <IconButton icon="pencil" onPress={navToObsEdit} textColor={colors.gray} />;
+
+    navigation.setOptions( {
+      headerRight: editIcon
+    } );
+  }, [navigation, observation, currentUser] );
+
```
kueda, line 173: Just so I understand, this is in an effect b/c `currentUser` gets fetched asynchronously? And possibly `observation` if it's remote?

src/navigation/mainStackNavigation.js
kueda: Definitely benefits to both approaches, but I'm inclined to err on the side of setting this config in the `Screen` definition and not the component because that seems to be what React Navigation folks do in their documentation. Let's at least try that for a while and if it's really annoying we can try something else.

kueda (changes requested):
Overall I think this looks good, and woo net decrease in code. Here are some things to address:

1. StandardCamera probably should not have a header
1. Taxon detail header has no title
1. Add evidence UI is only showing the camera and microphone options, which is making it hard to test photo import

src/components/ObsDetails/ObsDetails.js
albullington: Not exactly... this is in an effect because that's the [recommended way](https://reactnavigation.org/docs/header-buttons/#header-interaction-with-its-screen-component) to handle interactions between a component and a react navigation header.

In this case, the component knows there is an observation with a specific uuid, and if that observation belongs to the current user we want to show the edit button in the header. The header itself doesn't have a way of knowing which observation is being displayed in the component (& whether that observation belongs to the current user), so we need to use `navigation.setOptions` here to pass the correct right header icon from the component to the header.

albullington:
Ok, a few changes here, as requested:

1. Removed header from Standard Camera
2. Taxon Detail doesn't have a header in the Figma designs, unless I'm missing something?
3. Fixed camera options modal so it's now showing all options
4. Fixed flickering in Obs Edit title when observations have not yet loaded

kueda:
> Taxon Detail doesn't have a header in the Figma designs, unless I'm missing something?

You're [quite right](https://www.figma.com/file/XmLaQzJ1McnsYQysB57hib/iNaturalist-Mobile-UX-Design?node-id=1205%3A28121&t=rdOIKLu0Jy7Lt4M9-4)! @abhasinat was that an oversight? That seems inconsistent with most other screens.

abhasinat:
No header on Obs Detail, Taxon detail, and user profile is intentional.

kueda:
To be clear, these views have headers, they just don't have titles in those headers. And FWIW, in you designs on Figma, you have titles in the header for Obs Detail and User Profile, so that seems inconsistent with Taxon Detail.

Screen Shot 2022-11-21 at 2 55 26 PM

Screen Shot 2022-11-21 at 2 55 04 PM

abhasinat:
Ah right, here's what the updated UI version of obs detail, user profile, and taxon detail currently look like. You can see how I updated the header to still have buttons, but no longer be sticky to the top of the screen or have a title on any of these screens. No stress on these differences for now, it may just be helpful to know the direction these screens are going in.


image
image
image

kueda (approved):

Issue #178, opened 2022-09-08
PhotoGallery fetches the same page of results in Android


kueda:
**Describe the bug**
When you view the Photo Gallery while importing a photo, you can scroll but it just shows the same 28 photos over and over again.

**To Reproduce**
1. Make sure there are more than 28 photos available to choose from on your device
2. Start a new observation by importing photos
3. In the PhotoGallery, note the photo and start scrolling. You'll see it again at the start of the next page

**Expected behavior**
It should page through all the photos on the device.

**Screenshots**
![Screenshot_1662597454](https://user-images.githubusercontent.com/22691/189008479-599bb486-813b-435c-9e8f-8355f46272b3.png)

**Context (please complete the following information):**
 - Device: Pixel 3
 - OS: Android 12
 - App version: Build 26

**Additional context**
I looked into this a bit and it's possible @react-native-community/cameraroll just isn't working properly, like the `after` property works in iOS but not in Android. I tried screwing around with `toTime` and `fromTime` but wasn't able to get it to work.

PR #246, opened 2022-11-20
178 photo gallery android fetch

budowski:
#178 - Photo gallery screen - On Android, same page of results is shown again and again while scrolling


ios/Podfile
```diff
@@ -70,6 +70,8 @@ target 'iNaturalistReactNative' do
         config.build_settings['HEADER_SEARCH_PATHS'] << '"$(PODS_CONFIGURATION_BUILD_DIR)/React-bridging/react_bridging.framework/Headers" '
       end
     end
+    #react_native_post_install(installer)
+    `sed -i -e  $'s/__IPHONE_10_0/__IPHONE_12_0/' Pods/RCT-Folly/folly/portability/Time.h`
```
kueda, line 74: Is this necessary? If so can you explain why in a comment?

kueda (approved):
Working for me on my phone. The hack is annoying, but you gotta do what you gotta do.

ios/Podfile
budowski: It's explained [here](https://stackoverflow.com/questions/70357696/error-in-pods-rct-folly-folly-portability-time-h-and-typedef-redefinition-with-d) and [here](https://github.com/facebook/react-native/issues/31480):
But basically you get a compilation error for iOS when building it using latest XCode and iOS Build target, and this was the only workaround I found

PR #247, opened 2022-11-22
Skeleton UI and server results for adding a comment in ObsDetail

albullington:
This handles skeleton UI and displaying a new comment posted and successfully returned from the server. The code is pretty similar to what we're already doing for adding an identification. A comment is shown at 50% opacity, with a loading wheel underneath, until the server returns results. The API call now returns the correct fields to be able to display a new comment in the ActivityTab.

I also cleaned up styling using Tailwind & removed some unnecessary code from the ObsDetails tests.


kueda (approved):
Aside from the conflicts with main I think this looks good.

albullington:
Fixed merge conflicts.

PR #248, opened 2022-11-28
Fully delete photos from ObsEditContext when adding evidence

### LINKED ISSUE #240: Error thrown when returning to ObsEdit with no new evidence
**Describe the bug**
```
Possible Unhandled Promise Rejection (id: 0):
Error: Image uri invalid.
```

**To Reproduce**
1. Import a photo to ObsEdit via the photo gallery
2. Tap to add evidence from the StandardCamera
3. Take a couple photos, then delete both photos
4. Tap "Next" to return to ObsEdit screen with no new photos

**Expected behavior**
We should either:
1. Allow the user to proceed to ObsEdit with no new evidence and no errors
2. Or require that a user adds a photo before returning to ObsEdit - this is similar to how the PhotoImporter add evidence flow works currently

Either way, the app shouldn't throw an `ImageResizer` error.

**Screenshots**
IMG_FCC666A208BB-1

 - Device: iPhone XS
 - OS: 16.1

**Additional context**
N/A

albullington:
Closes #240, which throws an image resizer error for previously deleted photos when a user returns from the Standard Camera to the ObsEdit screen


src/components/SharedComponents/DeletePhotoDialog.js
```diff
@@ -18,28 +18,44 @@ type Props = {
...
     // spreading the array forces DeletePhotoDialog to rerender on each photo deletion
     setPhotoUris( [...updatedPhotos] );
 
+    // when deleting photo from StandardCamera while adding new evidence, remember to clear
+    // the list of new evidence to add
+    if ( evidenceToAdd && evidenceToAdd.length > 0 && setEvidenceToAdd ) {
+      const updatedEvidence = removePhotoFromList( evidenceToAdd, photoUriToDelete );
+      setEvidenceToAdd( [...updatedEvidence] );
+    }
+
```
kueda, line 58: Couple questions here:

1. Does this have to happen *every* time evidence gets removed, or *only* when it happens in the `DeletePhotoDialog` component? If needs to happen every time, then this should probably happen in the provider and this component should just receive something simple to do like `removeEvidence( photoUUID )` or something.
2. What is `evidenceToAdd`? There's the name issue, e.g. `mediaUrisToAdd` might be clearer, but also why is it necessary to have this part of state separate from the list of observations? If it's just an array of URIs, what if a single URI is associated with multiple observations? If you remove it from this array, is it removed from all those observations?

src/components/SharedComponents/DeletePhotoDialog.js
```diff
@@ -18,28 +18,44 @@ type Props = {
...
 
 const DeletePhotoDialog = ( {
   deleteDialogVisible,
   photoUriToDelete,
   photoUris,
   setPhotoUris,
-  hideDialog
+  hideDialog,
+  evidenceToAdd,
+  setEvidenceToAdd
```
kueda, line 33: This is the second level at which many of these props get passed down. That's good for modularity, but isn't that kind of the problem context/providers are supposed to solve?

kueda (changes requested):

src/components/SharedComponents/DeletePhotoDialog.js
albullington: 1. I'm not sure there's a way to remove evidence outside of the `DeletePhotoDialog`, is there? The `DeletePhotoDialog` component is used by the `MediaViewer`, which is accessible from the StandardCamera photo preview as well as ObsEdit. We currently also have the delete button on photos in the StandardCamera photo preview, which pops open the `DeletePhotoDialog`, but I believe we're planning to remove that functionality. Either way, I can't think of another place where evidence gets removed.
2. `evidenceToAdd` is an array I added in PR [#220](https://github.com/inaturalist/iNaturalistReactNative/pull/220/files). This keeps track of any additional camera photos or gallery photos a user is adding to their existing observation in `ObsEdit`. When a user navigates back to `ObsEdit`, we then use the `evidenceToAdd` array to create the respective realm objects and add them to the observation list. `ObsEditProvider` keeps track of a few distinct lists here -- one is for observations, which are already in realm object format, one is camera preview uris, which only keeps track of photos created from the camera, one is gallery uris, which likewise only keeps track of photos from the gallery, and one is evidenceToAdd, which keeps track of which objects still need to be converted into realm objects/observations. `evidenceToAdd` is cleared when the realm objects/observations are created (or in this case, when a photo is deleted before a user returns to `ObsEdit`).

kueda: Ok, this doesn't need to block, but this doesn't seem like view logic to me, in that it has nothing to do with presentation and more to do with manipulating state. Still not sure what the right pattern is in RN, but I feel like this belongs in the context, or some other controller-like piece.

kueda (approved):

albullington:
Ok, did some refactoring to move delete functionality into `ObsEditProvider`. `useContext` wasn't accessible using the Modal from `react-native-paper`, so I switched the media viewer modal to the react native modal component.

PR #255, opened 2022-12-01
Fetch user locale from server and change language

### LINKED ISSUE #254: Use user's locale preference
I wouldn't say this is a huge priority... except we have no other way to test if localizations are even working. The main requirement is just making sure that the user's locale attribute is used to set the language. I think that will require 

- [x] adding a `locale` attribute to the local User model
- [x] setting the local `locale` attribute when we get an updated user from the server
- [x] checking for the current user way up in the component hierarchy (rootNavigation?) and using [`changeLanguage`](https://www.i18next.com/overview/api#changelanguage) to use the user's locale

I looked into this a bit, and another small problem I have is that rootNavigation seems overloaded with a lot of stuff that has nothing to do with navigation. I also find it confusing that the navigator is the root component. My intuition is to look for an App component, so maybe that would be a better place to perform initialization tasks like setting the locale. It's also not clear to me why index.js exists, or what some setup happens there and others in rootNavigation.

albullington:
Closes #254.

This adds `App.js`, moves logic out of the root navigation drawer, and fetches the user locale from the server on app load. It changes the i18next language.

I manually tested changing my locale on staging to Russian, then checking the "Observations" string in UserProfile, and it's correctly displaying the localized string.


kueda (changes requested):
Had to add some translations to make sure it was working, which it does, but not when switching between users. Try this:

1. Make a fresh install of the app
1. Change your locale to Spanish on the web and sign in in the app; UserCard should show "observaciones"
1. Sign out and sign in as another user with locale set to English
1. UserCard still shows "observaciones"

kueda (approved):

kueda:
FWIW, these changes work, but the tests don't really test the issue at present: they just test that localization works, not that it changes when the user changes their locale on the web. That said, I've been banging my head against the tests here for way too long, so I'm going to merge for now and punt that test to an issue.

PR #257, opened 2022-12-02
216 remove excess space below the tab bar

angielt:
Android still has a small space at the bottom of the tab bar but if it is reduced to less the iOS tab bar buttons will be too close to the bar at the bottom swipe up.


kueda (changes requested):
I agree `h-16` looks a bit weird on iOS, but if it needs to be different on each platform, customize the className using `Platform.OS`.

kueda (approved):

PR #262, opened 2022-12-05
250 e2e init

### LINKED ISSUE #250: End-to-end test for adding an observation
I recently pushed a build in which the app crashes when you try to add a new observation in Android... which I thought I'd click tested, but I guess not, so I think it might be helpful to make our first [end-to-end test](https://reactnative.dev/docs/testing-overview#end-to-end-tests), at least as a way to learn how these kinds of tests. I basically want a test that

1. Boots the app (i.e. tests no crash on load)
2. Taps the "+" button in the footer to bring up the Add Evidence modal
3. Taps the "No Evidence" button
4. Asserts that the user has landed on Obs Edit with no crash

jtklein:
Fixes #250

In this PR there is only the iOS version setup for testing.

I also have explored letting the tests run on GitHubActions CI. I found that one test run would consume about 350 minutes (35 min runtime *10 for MacOS). Therefore, I have disabled the CI workflow. The longest time spent was on building the app, which might be outsourced to the actual physical machine of the developer and then uploaded and hosted somewhere to download in the action. Or we just have to stick to running the tests locally only.


kueda (changes requested):
I haven't take a deep dive yet, but I did get the tests to run locally. Regarding Github Actions, go ahead and enable it. We routinely run 30-40 mins on e2e tests for the Android app on Mac OS runners.

jtklein:
Okay, I did enable the e2e tests now also to run on GitHub Actions. I have enabled the workflow on "pull_request", so not on every push. When the UI is more stable later we might switch to "push" as well.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,76 @@
+##
+# Run e2e tests
+##
+name: e2e
+on:
+  pull_request:
+
+jobs:
+  pre_build:
```
kueda, line 9: Can we call this `setup`? I guess it technically precedes the `build` step, but that step both builds and tests, and it's really about getting us to the test.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: macos-latest
+    steps:
+      - uses: fkirc/skip-duplicate-actions@master
+        with:
+          github_token: ${{ github.token }}
+          cancel_others: true
+
+  build:
```
kueda, line 17: Let's call this `test` since that's the final outcome of this step.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: macos-latest
```
kueda, line 18: I think you're missing `needs` here, right? Without that, is `skip-duplicate-actions` going to do anything?

e2e/addObservationWithoutData.e2e.js
```diff
@@ -0,0 +1,36 @@
...
+    const loginText = element( by.text( "Log in to iNaturalist" ) );
+    await waitFor( loginText ).toBeVisible().withTimeout( 2000 );
+    await expect( loginText ).toBeVisible();
+  } );
+
+  it( "should navigate to observation add screen on add evidence button pressed", async () => {
+    await waitFor( element( by.text( "Log in to iNaturalist" ) ) )
+      .toBeVisible()
+      .withTimeout( 2000 );
+    await element( by.id( "camera-options-button" ) ).tap();
```
kueda, line 28: Let's use `by.label()` here and use accessibility labels for these buttons. In theory we won't be writing many e2e tests, but I think that's a good habit for us to have so we don't forget to add these labels.

kueda (changes requested):
Very cool, glad you were able to get this running so quickly.

Just so I understand your thinking, you've made this a separate workflow primarily because we don't want to run it on every push like we do for the `test` workflow, b/c it takes forever, right? Part of me doesn't want yet another test alert, but I see how separating these expensive tests into their own workflow makes sense.

Anyway, in addition to the inline comments, I think the only thing that's missing is Android. We can certainly merge this PR as is and do the Android stuff later, since I think you mentioned that might involve an expensive update to Jest, right?

e2e/addObservationWithoutData.e2e.js
jtklein: I was starting to use `by.label()` but had an error that it is matching multiple elements. It is checking against the native view hierarchy and not the RN side, so I am not sure there is a way around that problem. I checked the Detox [documentation](https://wix.github.io/Detox/docs/api/matchers/#bylabellabel) and they actually recommend to match elements by unique identifiers. So, for now I changed all the elements I matched by text to match by id as well. Then we can use the unit and integration tests to get into the habit of using accessibility labels.

.github/workflows/e2e.yml
jtklein: Have added `needs`

.github/workflows/e2e.yml
jtklein: Have renamed

.github/workflows/e2e.yml
jtklein: Have renamed

jtklein:
"Just so I understand your thinking, you've made this a separate workflow primarily because we don't want to run it on every push like we do for the test workflow, b/c it takes forever, right? Part of me doesn't want yet another test alert, but I see how separating these expensive tests into their own workflow makes sense."
Yes, to not have it running on every push, exactly. One reason is that it takes forever, but also because the UI might probably change many times between pushes in one branch, and be more finished when making a PR, I guess.

Having some time away from the task helped, and I have now also enabled Android e2e testing. One issue remaining though is the fact that our Android build setup produces a dynamically named .apk (including version number and code). So, for now, it works but I have hardcoded the .apk name into the e2e setup in `.detoxrc.json`. Would need to investigate if it is possible to get the dynamic name in this file, or if we can switch to a general filename.

Android is not enabled for the GitHub action workflow at the moment.

Also, one Android test is currently not passing because there is a crash when opening the "New observation" page. I don't know if that has been fixed for a different issue yet.

jtklein:
Have fixed the crash in Android app here: https://github.com/inaturalist/iNaturalistReactNative/pull/293
Should I update this PR to see if the one test in Android is passing now?

e2e/addObservationWithoutData.e2e.js
kueda: I think that's a fine approach for situations where there really are multiple elements in view with exactly the same accessibility text... but I think those cases are going to be rare, and this specific case isn't one of them. I've got this specific test working with `by.label( )` so I'll commit that as a proof-of-concept.

kueda (changes requested):
Looking good. I was able to get the Android tests to run (though not pass). I think if you can get them to pass in Github Actions we should be good to merge.

> One issue remaining though is the fact that our Android build setup produces a dynamically named .apk (including version number and code). So, for now, it works but I have hardcoded the .apk name into the e2e setup in .detoxrc.json. Would need to investigate if it is possible to get the dynamic name in this file, or if we can switch to a general filename.

Yeah, this is definitely a problem. That's coming from https://github.com/inaturalist/iNaturalistReactNative/blob/main/android/app/build.gradle#L161, and I suspect I did that so each release binary would have a unique, human-readable name... but that does seem kind of unconventional and causes other problems like this one. What's you're opinion? Just one consistent name?

README.md
```diff
@@ -34,12 +42,19 @@ npx jest MyObs
...
-1. `tests/unit`: Tests that only test specific modules, like a single component, or a hook.
+If you are running into some issues after the tests have been working for some time, try updating `applesimutils` with `brew update && brew upgrade applesimutils`.
 
-We're using [Jest](https://jestjs.io/) and [React Native Testing Library](https://callstack.github.io/react-native-testing-library/) for most of our tests, [factoria](https://github.com/phanan/factoria) and [faker.js](https://github.com/Marak/faker.js/) to generate test data for use with mocks. `Local*` model factories represent locally persisted data, while `Remote*` factories represent that kinds of records we get from an API or external package.
+If you want to run the Android tests make sure you have the correct [emulator setup](https://wix.github.io/Detox/docs/19.x/introduction/android-dev-env), and the emulator has the same name as in the .detoxrc.json file. You can then run `npm run e2e:build:android && npm run e2e:test:android`.
```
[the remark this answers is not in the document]
jtklein: Hm, I already had the system images installed, so I just created a new AVD in Android Studio. I tried to use the command line now to check the command line setup as described in the detox docs, but I get a Java error. So, should we just copy your commands into the Readme because they worked for you. (I don't remember if I installed the AOSP system image at one point or it came with the recent version of Android Studio).

For sure I would need to figure out the command line setup if we want to use it in GitHub Actions workflow as well.

.detoxrc.json
```diff
@@ -0,0 +1,51 @@
...
+    "ios.release": {
+      "type": "ios.app",
+      "binaryPath": "ios/build/Build/Products/Release-iphonesimulator/iNaturalistReactNative.app",
+      "build": "xcodebuild -workspace ios/iNaturalistReactNative.xcworkspace -scheme iNaturalistReactNative -configuration Release -sdk iphonesimulator -derivedDataPath ios/build"
+    },
+    "android.release": {
+      "type": "android.apk",
+      "binaryPath": "android/app/build/outputs/apk/release/com.inaturalistreactnative-v0.1.1+32-release.apk",
+      "testBinaryPath": "android/app/build/outputs/apk/androidTest/release/com.inaturalistreactnative-v0.1.1+32-release-androidTest.apk",
+      "build": "cd android && ./gradlew assembleRelease assembleAndroidTest -DtestBuildType=release && cd .."
```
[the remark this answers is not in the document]
jtklein: Really? It does not work for me without the assembleAndroidTest, and as far as I understood now it should not work. Because the test apk contains the actual UI tests (in Espresso = native Android e2e testing library) that interacts with the release build of our app. So, I think we need both commands and apks. What kind of error did you get when building with assembleAndroidTest?

jtklein:
Re apk name: I would use a generic name, but I have seen that in the fastlane config Fastfile we are also reading the apk path with dynamic version values as well. I don't know yet how much we rely on this dynamic name for fastlane.

android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <domain-config cleartextTrafficPermitted="true">
+        <domain includeSubdomains="true">10.0.2.2</domain>
+        <domain includeSubdomains="true">localhost</domain>
+    </domain-config>
```
kueda, line 6: Just so I understand, this is allowing cleartext traffic but only from localhost and 10.0.2.2, right?

e2e/addObservationWithoutData.e2e.js
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach( async () => {
+    // device.launchApp is preferred for an app of our complexity. It does work locally for both,
+    // but on CI for Android it does not work. So we use reloadReactNative for Android.
+    if ( device.getPlatform( ) === "android" ) {
+      await device.reloadReactNative( );
+    } else {
+      await device.launchApp( {
+        newInstance: true,
+        permissions: { location: "always" }
+      } );
```
kueda, line 26: Aren't we already doing this in the `beforeAll` above? Or do we need to do it twice?

kueda (approved):
I don't currently have the ability to actually run these tests locally, but just looking at the code I think this is looking good. I'd say merge when you've resolved the conflicts and we can deal with any minor issues like querying elements by accessibility labels later.

e2e/addObservationWithoutData.e2e.js
jtklein: In order to get passing tests on CI on Android, for some reason I had to change to use `device.reloadReactNative( )` even though the preferred `device.launchApp` was working fine on my local machine. So, for Android we need to launch the App in the beforeAll call. On iOS it is redundant with the first time it hits beforeEach, correct. But it does not take a lot of time, and I wanted to avoid another check for Platform.

android/app/src/main/res/xml/network_security_config.xml
jtklein: Yes, correct. The default for Android emulators is to not allow cleartext traffic at all. The domain-config is only for the specified domains.

PR #265, opened 2022-12-05
203 add placeholder images on my obs and obs detail when no media

angielt:


kueda (changes requested):
Looking good, the only thing missing is the grid view on My Observations.

kueda (approved):

PR #271, opened 2022-12-06
266 flash icon change

### LINKED ISSUE #266: Camera flash icon not changing
When changing the camera flash setting in the camera view, the icon is not changing according to what mode is set. Using flash or not is changing though.

jtklein:
Fixes #266


package.json
```diff
@@ -6,6 +6,7 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
+    "clean": "npx react-native clean-project-auto && npx pod-install && watchman watch-del-all && npm start",
```
kueda, line 9: Couple things here

* I think `clean-project-auto` already does `watchman watch-del-all` according to the [docs](https://github.com/pmadruga/react-native-clean-project#content)
* I would not expect a `clean` task to also start Metro. Maybe we should call this `clean-start`?
* `clean-project-auto` wipes all your cocoapods and the system and user pod caches, so you have to download and install them all ever time you clean. That makes this a bit extreme for my use, but no object to adding it if it's useful to you.

src/components/Camera/StandardCamera.js
```diff
@@ -90,13 +93,30 @@ const StandardCamera = ( ): Node => {
...
-      size={40}
-      icon={icon}
-      style={{ backgroundColor: colors.gray }}
-    />
-  );
+  const renderCameraOptionsButtons = icon => {
+    let accessibilityLabel = "";
+    switch ( icon ) {
+      case "flash":
+        accessibilityLabel = "The flash is turned on. Press here to disable it.";
```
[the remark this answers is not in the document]
kueda: These should just come from the translation files. I would add strings like `flash-button-label-flash = The flash is turned on. Press here to disable it.` I'd be fine with doing that in this PR, but also happy to do a more systemic task. Where else where you seeing string literals in the components like this?

tests/unit/components/Camera/StandardCamera.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+  test( "should change to flash enabled on button press", async () => {
+    renderStandardCamera();
+
+    const flashButton = await screen.findByLabelText(
+      "The flash is turned off. Press here to enable it."
+    );
+    fireEvent.press( flashButton );
+
+    await screen.findByLabelText(
```
kueda, line 48: For my own education, what's the advantage of using `screen` over the finder methods returned by `render`?

kueda (changes requested):
Looking good. The only thing I really want to change is the name of that npm task. Everything else could happen in separate issues, or here.

tests/unit/components/Camera/StandardCamera.test.js
jtklein: I was following this page by the creator of react-testing-library: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library
"The benefit of using screen is you no longer need to keep the render call destructure up-to-date as you add/remove the queries you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest." And for me also, that I don't have to destructure the render return in every test.

src/components/Camera/StandardCamera.js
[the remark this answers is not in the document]
jtklein: OK, I have added translations also to the other instances I found with accessibility labels. However, now my tests did not pass anymore because i18n is mocked in the tests. So, I switched to finding the button by id instead. But I think I would prefer to be able to find the buttons by label in English.

package.json
jtklein: Have removed `watchman watch-del-all` and renamed. I use it mainly when I am switching between branches or changing dependencies. But I do have some troubles from time to time that makes me want to clean everything.

tests/unit/components/Camera/StandardCamera.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+jest.mock( "components/Camera/FadeInOutView", () => ( {
+  __esModule: true,
+  default: () => mockView
+} ) );
+
+jest.mock( "components/Camera/PhotoPreview", () => ( {
+  __esModule: true,
+  default: () => mockView
+} ) );
```
kueda, line 40: Cool. Stuff. I wish there was a good way to do `jest.mock` in helper methods instead of only globally or in a test like this.

kueda (approved):

PR #273, opened 2022-12-07
Update obs list to use React Query, fetch observations via id_below

### LINKED ISSUE #244: Infinite obs fetch when you have 5 observations
**Describe the bug**
My Observations tries to fetch new observations forever if you have exactly 5 observations.

**To Reproduce**
1. Sign in to an account with 0 observations
2. Make 5 observations without evidence and return to My Observations
3. The app will lock up and you will probably see some jerky motion. In the log you'll see a repeating ` LOG  Couldn't fetch observations` message

**Expected behavior**
If the app tries to retrieve more of your observations from the server, it should stop the process if it gets no observations back, no matter how many observations you have

 - Device: iPhone 13 simulator
 - OS: iOS 15.5
 - App version: Build 31 (47e9916)

**Additional context**
I think there are at least two immediate problems here:

1. `handleEndReached` probably gets called repeatedly when the end of the list is visible, right? In theory it only needs to get fired once.
2. `fetchNextObservations()` in `useRemoteObservations` always fetches page 1 if the total number of local observations is exactly 5

FWIW, determining what page of results to fetch based on the number of local obs doesn't seem like a great approach, because they are sort of unrelated: you might have 57 local obs, none of which have been uploaded, but you would still want to fetch the first page of results from the server. Instead, `fetchNextObservations` might want to accept an observation (or a UUID, or an obs ID), and use that to ask for observations on the server that are older than that one, e.g. `id_below=${observation.id}`.

albullington:
This closes #244. A few changes here:

1. API calls in ObsList use React Query (observations.search and observations.updates)
2. `onEndReached` only gets triggered if the previous observations.search API call is not still loading
3. Adds some extra height to the FlatList so `onEndReached` can still be called and the FlatList is still scrollable if a user has less observations than fill the screen


src/components/Observations/ObservationViews.js
```diff
@@ -22,31 +27,64 @@ import ObsListHeader from "./ObsListHeader";
...
+    isLoading
+  } = useAuthenticatedQuery(
+    ["searchObservations", idBelow],
+    optsWithAuth => searchObservations( params, optsWithAuth ),
+    {
+      keepPreviousData: true
+    }
+  );
+
+  const handleEndReached = oldestId => setIdBelow( oldestId );
```
kueda, line 80: This is just getting called once, so is there a need for it to be its own function? IMO just doing this work in `onEndReached` below means one fewer jump when trying to understand what's happening.

kueda (changes requested):
Looks like you've got some failing tests from Explore that look legit. I think ObservationViews has become too specialized to My Observations, probably not starting in this PR but definitely exacerbated by it (Explore should not show local observations, Explore should not check for updates on the current user's observations, etc)..

To me this suggests that reused components like `ObservationViews` are doing too much, and stuff that has nothing to do with the view like fetching and updating obs should be moved elsewhere (hooks, providers, HOCs, not sure). So part of me wants to fix this now since Explore was sort of working, our tests succeeded in catching this problem, and re-using components between Explore and My Observations is a good thing.

Another part of me thinks Explore is not our focus right now and is inclined to delete it and its tests and focus on the observation creation loop. We can refactor to make things more reusable when we really need to reuse them.

Thoughts?

albullington:
I like the idea of removing Explore, since I think a lot of that code will have to be rewritten anyway. (The dropdowns weren't working, we don't necessarily want an Explore provider, and everything was using styles instead of Tailwind). I'd rather have less complicated code that works as expected for now, and we can refactor and add Explore back in when needed.

Tests are passing, though there are a couple warnings and open handles. I think these are related to `fetchRemoteUser` being called without a user id and `UserCard` rerendering in ways that are unnecessary -- this is the code I worked on simplifying yesterday in the [254-user-locale-preference](https://github.com/inaturalist/iNaturalistReactNative/tree/254-user-locale-preference)` branch, so maybe it's better to try to merge that one first.

kueda (approved):

PR #275, opened 2022-12-08
270 ensure button is visible above keyboard on log in

angielt:


kueda (changes requested):
This isn't showing the button above the keyboard for me on my Pixel 3 or in an iPhone 13 simulator. Assuming it's working for you, can you post a screenshot of what you're seeing?

angielt:
This PR makes the form scrollable so the button is accessible if you scroll up without clicking out of the keyboard but yes the button still requires a scroll to be seen even when you reach the password input box. I'm currently making changes. Is the 2nd video closer to the behavior we want? 

Current PR:
https://user-images.githubusercontent.com/13282161/206944051-6484c3c8-ff5c-4cd8-86b3-48e453a6242a.mp4


Making changes:
https://user-images.githubusercontent.com/13282161/206943626-75a7f944-b3fd-47e6-b896-7f8aa249381e.mp4

src/components/LoginSignUp/Login.js
```diff
@@ -103,7 +99,7 @@ const Login = ( ): Node => {
         autoCapitalize="none"
         keyboardType="email-address"
         selectionColor={colors.black}
-        onFocus={e => scrollToInput( findNodeHandle( e.target ) )}
+        onFocus={() => setExtraScrollHeight( 200 )}
```
kueda, line 102: I don't love this hardcoding, but I'm failing to find a better alternative so it works for now.

kueda (changes requested):
Only blocking problem is the fact that the X button no longer works.

PR #277, opened 2022-12-09
Delete observations from ObsEdit

### LINKED ISSUE #224: Delete observations
Currently we have a delete button on ObsEdit that does nothing but remove the observation from the ObsEditContext. Ideally this should *actually* remove the observation:

1. Delete the observation on the server
1. Delete the observation record from Realm

This would also have to work while offline, or at least deal with that situation. Maybe this is a good stopgap:

- [x] Remove the delete option for observations that haven't been persisted to Realm
- [x] Only allow delete while online
- [x] Delete the observation on the server in a blocking / synchronous process (i.e. don't let the user do anything else while this is happening)
- [x] Only delete the record from Realm when you're sure the server deletion was successful

albullington:
Closes #224 

This opens a kebab menu with a delete option for observations which have been persisted to realm.

Delete first deletes the observation on the server, then in the local realm database if the mutation is successful.

This also updates the UI so the kebab menu is closed when an item is tapped. Without this, a user would need to tap twice to confirm their deletion in the DeleteObservationDialog.


src/components/ObsDetails/ActivityItem.js
```diff
@@ -95,20 +96,27 @@ const ActivityItem = ( {
             )}
           {item.body && currentUser
             ? (
-              <KebabMenu>
+              <KebabMenu
+                visible={kebabMenuVisible}
+                setVisible={setKebabMenuVisible}
```
kueda, line 101: I can see why you're doing this and I don't think this needs to block, but this complicates the use of this component, in that every context that uses it now needs to control the state of whether the menu is open or not. I'm just not sure what a better approach would be. Best I can come up with is to have KebabMenu not accept children and instead accept a prop that's an array of Menu.Item props, and then KebabMenu wraps the onPress prop of each to execute the function and then set the internal visible state to false. Maybe a bit complicated.

src/components/ObsEdit/DeleteObservationDialog.js
```diff
@@ -19,25 +21,36 @@ const DeleteObservationDialog = ( {
...
   return (
     <Portal>
       <Dialog visible={deleteDialogVisible} onDismiss={hideDialog}>
         <Dialog.Content>
-          <Paragraph>{t( "Are-you-sure" )}</Paragraph>
+          <Text>{t( "Are-you-sure" )}</Text>
         </Dialog.Content>
         <Dialog.Actions>
-          <Button onPress={hideDialog} text={t( "Cancel" )} level="primary" />
+          <Button onPress={hideDialog} text={t( "Cancel" )} level="primary" className="m-0.5" />
```
kueda, line 48: For some reason the text in this button isn't centered. I looked into it a little bit and didn't see an obvious solution. The `style` prop seems to get populated with mysterious padding from... somewhere, maybe some react-native-paper magic.

kueda (approved):
Works! Would like a test, but... I'm pretty burnt out on tests at the moment.

Issue #259, opened 2022-12-04
Observed on datetime not parsed from EXIF in correct time zone


kueda:
**Describe the bug**
When I import a photo taken in Pacific Time (UTC-08:00), I end up with a datetime 8 hours in the future.

**To Reproduce**
1. Download ![PXL_20221203_191416189 MP](https://user-images.githubusercontent.com/22691/205468206-bee8f7fe-a95b-4bcd-b24b-78a9b9df5be1.jpg) to device
2. Import that photo as a new observation
3. Observe on date will appear as `Dec 3, 2022, 11:14:16 AM`, then switch to `Dec 3, 2022, 7:14:16 PM`

**Expected behavior**
The observed on date should appear as `Dec 3, 2022, 11:14:16 AM`.

**Screenshots**
![Screenshot_20221203-164315](https://user-images.githubusercontent.com/22691/205468417-b60a3575-27f9-49d2-b080-bfd31974e700.jpg)

**Context (please complete the following information):**
 - Device: Pixel 3
 - OS: Android 12
 - App version: build 32
 - Phone was in Pacific Time (UTC-08:00) when tested, photo was captured in Pacific Time (UTC-08:00) as well

**Additional context**
This is almost certainly a time zone issue given the 8 hour shift. The EXIF `date` seems to be `2022-12-03T19:14:16.000`... which looks like the UTC version of the actual date, and then the hook seems to be screwing it up somehow. Unless of course the EXIF date should not be the UTC date...

PR #286, opened 2022-12-09
Fix: observed on date in ObsEdit

albullington:
A few things here:

1. Refactored `parseExif` so it's no longer a hook, since it should really only be called once when a new observation is created in `createObservationFromGalleryPhoto`, not in the component itself. I also fixed the related tests.
2. It's worth noting that the test photo in [#259 ](https://github.com/inaturalist/iNaturalistReactNative/issues/259) doesn't actually return any exif data, so we're falling back to setting `observed_on_string` with the photo timestamp
3. I removed `time_observed_at` for now since I think that's supposed to come from the server, but lmk if this is something we should be setting when creating a new observation
4. New observations will either use the `observed_on_string` from the first photo (in the case of photo gallery) or use the current date if there is no `observed_on_string` from a gallery photo
5. The DatePicker will only display dates from `observed_on_string`, not `time_observed_at`, to keep things simple
6.  Separately, I made a small tweak that we should not fetch the current location if an observation already has a latitude & longitude


src/providers/ObsEditProvider.js
```diff
@@ -60,18 +61,27 @@ const ObsEditProvider = ( { children }: Props ): Node => {
   ), [] );
 
   const createObservationFromGalleryPhoto = useCallback( async photo => {
-    const latitude = photo?.location?.latitude || null;
-    const longitude = photo?.location?.longitude || null;
+    const originalPhotoUri = photo?.image?.uri;
+    const firstPhotoExif = await parseExif( originalPhotoUri );
```
kueda, line 65: When I use this branch on my Pixel 3, download the image in #259, insert

```
console.log( "firstPhotoExif: ", firstPhotoExif );
```

right after this line, and import that photo, I get
```
{"date": "2022-12-03T19:14:16.000", "latitude": 37.92129722222222, "longitude": -121.93714722222222}
```

Is that not what you're getting?

kueda (changes requested):
Haven't taken a deep dive on all the changes here, but I can still reproduce the bug described in #259, and I can definitely see EXIF in that linked file, and I see that EXIF getting parsed when I try to import it on this branch.

When I try to download the photo in an iPhone simulator and open it in Photos I get a crash, which is weird, but I don't see anything wrong with the photo. Opens fine in Preview on my Mac, for instance, and I can see the EXIF there.

So... what makes you believe there's no EXIF? And have you tried it in Android?

src/providers/ObsEditProvider.js
albullington: I'm getting the exif data on Android, but I'm seeing an empty exif object for the same photo on iOS.

I'm a little confused about what's happening in the native exif parser, because it seems like we're doing a lot of date parsing in the [Android native code](https://github.com/inaturalist/react-native-exif-reader/blob/d84d462f0f36fbf1a46b7159bbcb95dd5f5d7eef/android/src/main/java/com/reactnativeexifreader/ExifReaderModule.java#L145) and [iOS native code](https://github.com/inaturalist/react-native-exif-reader/blob/d84d462f0f36fbf1a46b7159bbcb95dd5f5d7eef/ios/ExifReader.swift#L33) to make sure the date is returned in UTC/GMT format, and then we're parsing the dates again in React Native. Wouldn't it be better to either 1) return exactly what we need from native code or 2) return an unedited datetime from the native library so we can handle parsing/formatting in the client only?

kueda: I agree that's pretty confusing and that the way the library should behave is to return the verbatim datetime in ISO8601 format, but for now, can we just take that UTC date, assume it's UTC, and convert to the local time zone?

albullington: I can see the EXIF date for other photos in iOS (specifically, the default photos in the iOS simulator) so maybe you're right that the EXIF data is getting stripped for this photo when we add it to a simulator/device.

albullington:
Ok, I think this is ready for re-review. Using `date-fns-tz` and `react-native-localize` to convert the UTC date to the user's timezone. I'm getting the expected date on Android (and iOS if I try hardcoding the date instead of using the empty EXIF object).

PR #289, opened 2022-12-10
Allow individual uploads from ObsList

### LINKED ISSUE #281: Feature: Upload button next to saved observation
A user should be able to tap the upload button next to a saved observation:

- [x] Upload button should be pressable and upload a single observation to the server
- [x] Upload button should be visible and pressable in grid view

albullington:
Closes #281 by adding upload functionality to the upload button in ObsList and making that button visible in gridview.


src/components/Observations/GridItem.js
```diff
@@ -79,7 +80,13 @@ const GridItem = ( {
             />
           </View>
         )}
-        <ObsCardStats item={item} view="grid" />
+        {!item._synced_at
```
kueda, line 83: Observations that were synced but have since been changed locally also need to be uploaded, so you might want to replace this with a model method. We have `wasSynced()`, but maybe we need `needsSync()`.

tests/unit/components/Observations/ObsList.test.js
```diff
@@ -80,7 +82,7 @@ it( "renders multiple observations", async ( ) => {
 
 it( "renders grid view on button press", async ( ) => {
   await waitFor( ( ) => {
-    const { getByTestId } = renderComponent( <ObsList /> );
+    const { getByTestId } = renderComponentWithObsEditProvider( <ObsList /> );
```
kueda, line 85: All of these tests pass for me with `renderComponent` What kind of errors were you seeing?

kueda (changes requested):

tests/unit/components/Observations/ObsList.test.js
albullington: Don't remember, but it seems to be working for me now too so I'll remove this.

PR #325, opened 2023-01-01
Add missing dependency

Chrischuck:
Looks like `@react-native-community/cli-platform-ios` doesn't get installed automatically, though `@react-native-community/cli-platform-android` does.

On a related note, this is my first PR and had some issues w/ node/npm versions. I just moved to latest: node v13.9.0 and npm v9.2.0

But for new devs is there a recommended version for both we can add to the readme or potentially w/ and nvmrc?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "@react-native-async-storage/async-storage": "^1.17.10",
     "@react-native-camera-roll/camera-roll": ">=5.0.4",
     "@react-native-community/checkbox": "^0.5.12",
+    "@react-native-community/cli-platform-ios": "10.0.0",
     "@react-native-community/datetimepicker": "^6.5.3",
     "@react-native-community/netinfo": "^9.3.6",
     "@react-native-picker/picker": "^2.4.8",


jtklein:
Hello Chris,
Thank you for your contribution. I am not sure I understand the reason you want to specify the `cli-platform-ios` version directly? It does get pulled in as dependency for `react-native`directly. See ["@react-native-community/cli-platform-ios": "9.2.1",](https://github.com/Chrischuck/iNaturalistReactNative/blob/f76869b9bce79cbbb566c77800bd06002414186d/package-lock.json#L31177-L31178)

Yes, we do have a `.nvmrc` file, currently we have specified Node v 16.17.0 for the project.

Chrischuck:
Thanks Johannes, that's my fault for not checking if there was an `.nvmrc` already. For me at least `@react-native-community/cli-platform-ios` wasn't getting pulled in automatically, though `@react-native-community/cli-platform-android` was and I was getting an error trying to run ios.

Can probably close this pr in that case?

jtklein:
You mentioned some issues locally with node. I would rather suspect the problem with the dependency to be local only as well. Please, check if you still have an issue when using node version as in .nvmrc.

Chrischuck:
No issues with .`nvmrc`, everything else is working as expected

jtklein:
So, the original issue with the missing dependency disappeared?

kueda:
If everything's working ok with the version of node specified in the `.nvmrc`, I think we can close this.

PR #327, opened 2023-01-02
314 no evidence icon not navigating

jtklein:


tests/unit/components/ObsDetails/ObsDetails.test.js
```diff
@@ -10,6 +11,12 @@ const mockObservation = factory( "LocalObservation", {
   created_at: "2022-11-27T19:07:41-08:00",
   time_observed_at: "2023-12-14T21:07:41-09:30"
 } );
+const mockNoEvidenceObservation = factory( "LocalObservation", {
```
jtklein, line 14: If not for this I would merge the PR myself. But I have not used factory before, so I am not sure if this is acceptable for making an observation without `observationPhotos`, or if there would be a better way to overwrite the default mocked "LocalObservation".

kueda: This is fine, though I think the best approach here is to update the factories so they make a minimally viable record. If tests need things like an observation with comments or photos, the test should construct them. Do you want to do that work here? I could also just do it in main.

kueda (approved):

PR #333, opened 2023-01-05
301 flag id functionality obsdetail

angielt:


src/components/ObsDetails/FlagItemModal.js
```diff
@@ -0,0 +1,178 @@
...
+  id:number,
+  showFlagItemModal: boolean,
+  closeFlagItemModal: Function
+}
+
+const FlagItemModal = ( {
+  id, showFlagItemModal, closeFlagItemModal
+}: Props ): Node => {
+  const keyboardScrollRef = useRef( null );
+  const [value, setValue] = useState( "none" );
```
albullington, line 32: What is `value` referring to here? Is this the type of flag or something else? Can we use more descriptive language with this and `toggleValue`?

src/components/ObsDetails/FlagItemModal.js
```diff
@@ -0,0 +1,178 @@
...
+                  className="text-sm"
+                  placeholder={t( "Flag-Item-Other-Input-Hint" )}
+                  value={explanation}
+                  onChangeText={text => setExplanation( text )}
+                  onFocus={e => scrollToInput( findNodeHandle( e.target ) )}
+                />
+                <Text>{`${explanation.length}/255`}</Text>
+              </>
+            )
+            : undefined}
```
albullington, line 158: Instead of using a tertiary statement and `undefined` here, you can use && to only show the component if the value equals "other".

src/components/ObsDetails/FlagItemModal.js
```diff
@@ -0,0 +1,178 @@
...
+
+  const scrollToInput = node => {
+    keyboardScrollRef?.current?.scrollToFocusedInput( node );
+  };
+
+  const toggleValue = checkbox => {
+    if ( value === checkbox ) { setValue( "none" ); } else { setValue( checkbox ); }
+  };
+
+  const clearForm = () => {
```
albullington, line 43: This isn't really a form in the HTML sense, so I think a description like `resetFlagModal` might make more sense here. What do you think?

albullington (changes requested):
Looking good so far -- requested some changes on naming and made a suggestion for what we should show on the UI after suggesting a flag.

I'm realizing belatedly that comments can also be flagged, so we might want to add this flagging functionality to comments as well. Let me know if that makes sense to tackle in this issue. If it feels like scope creep, we can make a separate issue for that.

src/components/ObsDetails/FlagItemModal.js
```diff
@@ -0,0 +1,178 @@
...
+    setValue( "none" );
+    setExplanation( "" );
+    closeFlagItemModal();
+  };
+
+  const createFlagMutation = useAuthenticatedMutation(
+    ( params, optsWithAuth ) => createFlag( params, optsWithAuth ),
+    {
+      onSuccess: ( ) => {
+        // waiting for final designs
```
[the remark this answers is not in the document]
angielt: Got it, when I flagged something on staging it presented a modal that showed "Object [object]" after I pressed save so I didn't know what to do. I just checked again and it doesn't do that anymore so no sure what I saw.

src/components/ObsDetails/FlagItemModal.js
angielt: I treated the checkboxes as a form with radio buttons but renaming to `resetFlagModal` could make sense as well. I don't have a strong preference so i'll change it to `resetFlagModal` for now.

src/components/ObsDetails/FlagItemModal.js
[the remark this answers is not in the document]
angielt: If the flag is for a comment, can I assume the` Flag.js `realm model will be linked to a Comment?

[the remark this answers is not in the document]
albullington: Yep, exactly. Similar to how a User realm model can be linked to a Comment, an Observation, or an Identification.

albullington (approved):

Issue #320, opened 2022-12-30
My Observations doesn't show any taxon name for observations of taxa without common names


kueda:
**Describe the bug**
In the list and grid views on My Observations, observations associated with a taxon that has no common name shows nothing in the name area.

**To Reproduce**
1. Make an observation of a taxon with no common name, like the [genus _Corbicula_](https://www.inaturalist.org/taxa/64009-Corbicula)
2. Save the observation and view it in list view on My Observations
3. The `place_guess` and date/time appear but it's just a blank space where the name should appear

**Expected behavior**
If an observation is associated with a taxon that has no common name, My Observations should still show the scientific name in both list and grid views.

**Screenshots**

List and grid views of the same observation of _Corbicula_:

![Screenshot_20221230-171602](https://user-images.githubusercontent.com/22691/210115195-6030f1e7-fa38-4870-927f-9969a00f8f7f.jpeg)
![Screenshot_20221230-171554](https://user-images.githubusercontent.com/22691/210115200-bbf39d32-3e44-479f-a722-4d976b4b542a.jpeg)

**Context (please complete the following information):**
 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.1.1 (32), d51685b48855fe469ffde50a28581c418fa2b0e8

**Additional context**
Taxon name display is complicated and I think we'll eventually need something similar to the [`<SplitTaxon>` component](https://github.com/inaturalist/inaturalist/blob/main/app/webpack/shared/components/split_taxon.jsx) we have on the web, so one way to address might be to get started on a version of that for RN. FWIW, "split" was supposed to mean "showing both common and scientific names whenever possible," but I've never been very happy with that. Open to better names if people have ideas. `DisplayTaxon`? `DisplayTaxonName`? `TaxonNameForUser`? `MoreComplicatedThanMaybeItNeedsToBeTaxonNameDisplay`?

Chrischuck:
Would love to take a stab at this @kueda

kueda:
Go for it, @Chrischuck.

Chrischuck:
<img width="341" alt="Screenshot 2023-01-02 at 8 30 39 PM" src="https://user-images.githubusercontent.com/13311268/210300653-64c8a69a-267a-4d68-af5e-9706db49d5cb.png">

Got these going for a couple of observations w/o common names.

[...]

PR #334, opened 2023-01-05
Create more robust taxon display name component - #320

Chrischuck:
Create new `TaxonDisplayName` component in regards to issue #320


src/components/Camera/StandardCamera.js
```diff
@@ -8,7 +8,7 @@ import { t } from "i18next";
 import { ObsEditContext } from "providers/contexts";
 import type { Node } from "react";
 import React, { useContext, useRef, useState } from "react";
-import { Platform, StatusBar } from "react-native";
+import { StatusBar, Platform } from "react-native";
```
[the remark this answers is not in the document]
albullington: Automatic lint changes are great.

src/components/DisplayTaxonName.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+  const taxonData = {};
+
+  taxonData.rank = taxon.rank;
+
+  if ( taxon.preferred_common_name ) {
+    taxonData.commonName = _.trim( taxon.preferred_common_name );
+  }
+
+  let { name: sciName } = taxon;
```
albullington, line 25: Could we add a comment here to remind ourselves that a lot of this logic mirrors the [SplitTaxon component](app/webpack/shared/components/split_taxon.jsx)? It might be helpful in case we need to make changes later.

src/components/DisplayTaxonName.js
```diff
@@ -0,0 +1,62 @@
...
+const DisplayTaxonName = ( { item: { user, taxon } }: Props ): Node => {
+  if ( !taxon ) {
+    const text = "unknown";
+    return <Text numberOfLines={1}>{text}</Text>;
+  }
+  const taxonData = {};
+
+  taxonData.rank = taxon.rank;
+
+  if ( taxon.preferred_common_name ) {
```
albullington, line 21: Could we add capitalization for the preferred common names? It could be pretty similar to what we're doing in [iNatModels.Taxon.titleCaseName](https://github.com/inaturalist/inaturalist/blob/1ff769e2df93b458e662200be080f55bd88f79ba/app/assets/javascripts/ang/models/taxon.js.erb#L169), and it probably makes sense for this to live in the Taxon model.

src/components/DisplayTaxonName.js
albullington, line 25: not important, but slight preference for us to spell out `scientificName` instead of `sciName` here

albullington (changes requested):
Hi @Chrischuck, this is looking good to me -- thanks for working on this! I left a few requested changes, mostly all pretty minor. Let me know if you have any questions.

src/components/DisplayTaxonName.js
```diff
@@ -0,0 +1,62 @@
...
+
+  let title = taxonData.sciName;
+
+  if ( user.prefers_scientific_name_first && taxonData.commonName ) {
+    title = `${title} (${taxonData.commonName})`;
+  } else if ( taxonData.commonName ) {
+    title = `${taxonData.commonName} (${title})`;
+  }
+
+  return <Text numberOfLines={1}>{title || "no name"}</Text>;
```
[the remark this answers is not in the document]
Chrischuck: I'm thinking not, I included it based on the previous fallback

src/components/DisplayTaxonName.js
Chrischuck: Went ahead and added the two functions. I made some small adjustments like using a set for faster look ups, but the core logic is the same.

My understanding is, if part of the common name is more of an adjective and contains dashes, only the first letter will be capitalized, like in `Crown-of-thorns Sea Star`. But if the noun part of the common name has dashes, it will all be capitalized, like `Blue Night-Heron`

Please let me know if I've misunderstood the naming schema and I can make adjustments. Added a test case for the above as well.

albullington:
This is looking good to me, so I'm going to go ahead and merge in. I think it's likely that we'll tweak the way names are displayed once we get closer to a final UI (maybe having only one name displayed instead of common name, scientific name, and rank), and having all the test cases should make this easy and straightforward to adjust. Thanks for your contribution @Chrischuck!

PR #341, opened 2023-01-06
Show localized error description from /oauth/token

### LINKED ISSUE #326: verifyCredentials failed when calling /oauth/token -  CLIENT_ERROR 400
Error thrown when logging in using my iNat account on an Android emulator. The app continues to work though and I can see my data when I dismiss the error.

Screenshot 2023-01-02 at 20 56 52

albullington:
Closes #326


src/components/LoginSignUp/AuthenticationService.js
```diff
@@ -215,14 +223,18 @@ const verifyCredentials = async (
   formData.append( "username", username );
 
   const api = createAPI();
-  let response = await api.post( "/oauth/token", formData );
+  let response = await api.post( "/oauth/token", { locale: i18next.language }, formData );
```
kueda, line 226: According to the [apisauce docs](https://github.com/infinitered/apisauce#calling-the-api), this just replaces the entire POST body with `{ locale: i18next.language }` and passes `formData` as a config object to axios. Instead I think you need to put the `locale` in the `formData`.

kueda (changes requested):

Issue #339, opened 2023-01-06
Obs count and profile picture don't populate on initial login


albullington:
**Describe the bug**
When a user first logs into the app, `observation_count` and `icon_url` have not yet been stored in realm, so a user sees missing data in the UserCard. Normally, we're fetching the remote user data in App.js, but that API call doesn't get triggered on the initial login.

**To Reproduce**
1. Log in
2. Look at the UserCard in ObsList

If you go to a different part of the app and return to ObsList, you'll see the correct profile picture and observation count.

**Expected behavior**
User data should be available before a user lands on ObsList (or UserCard should rerender with the correct data when it's available.)

**Screenshots**
![IMG_6540](https://user-images.githubusercontent.com/35536439/210902741-e49045a9-ff98-41f6-b25b-874c0489cd49.jpg)

PR #342, opened 2023-01-07
Fix missing obs count on login (#339)

Chrischuck:
#339 

Issue here was the memo was preventing children from receiving fresh props.

also includes removal of extra semi colon from my last pr.


src/components/Observations/ObservationViews.js
```diff
@@ -93,12 +93,12 @@ const ObservationViews = ( ): Node => {
     />
   );
 
-  const renderHeader = useMemo( ( ) => (
+  const renderHeader = ( ) => (
```
Chrischuck, line 96: I don't think we needed `useMemo` here since the props shouldn't change much given they're a ref and a setter from `useState`

It also prevented children from re-rendering and hence the bug.

tests/integration/MyObservations.test.js
```diff
@@ -11,57 +11,80 @@ import factory, { makeResponse } from "../factory";
...
+  describe( "when signed out", () => {
+    async function testApiMethodNotCalled( apiMethod, realm ) {
+      const signedInUsers = realm.objects( "User" );
       expect( signedInUsers.length ).toEqual( 0 );
+
       const { getByText } = renderAppWithComponent( <ObsList /> );
-      await waitFor( ( ) => {
-        expect( getByText( "Log in to iNaturalist" ) ).toBeTruthy( );
+      await waitFor( () => {
+        expect( getByText( "Log-in-to-iNaturalist" ) ).toBeTruthy();
```
Chrischuck, line 49: Test the key got returned. In the test suite there was a weird issue where the first test, the `t` func would return the key, but the second test would return the translated text. Seemed simpler to test that the correct key is getting fed to the component, rather than testing the i18next library.

tests/integration/MyObservations.test.js
```diff
@@ -11,57 +11,80 @@ import factory, { makeResponse } from "../factory";
...
+  return {
+    // Ideally we shouldn't have to test i18next itself
+    // just need to test that the correct keys/opts are being fed to i18next
+    t: ( key, options ) => {
+      if ( !options ) {
+        return key;
+      }
+      return originalModule.t( key, options );
+    }
+  };
```
Chrischuck, line 28: I feel this is a kind of hacky solution to ensure the localization tests still pass.

Issue #298, opened 2022-12-16
Top 2 observations in obs list are not tappable when scrolled down


albullington:
*To Reproduce**
1. Go to ObsList
2. Scroll down so the first observation is tucked slightly under the toolbar
3. First two observations are not pressable

**Expected behavior**
A user should be able to tap all observations and navigate to the appropriate screen (obs detail or obs edit)

**Screenshots**

**Context (please complete the following information):**
 - Device: iOS/Android

angielt:
I wasn't able to replicate this bug on iOS or android 

https://user-images.githubusercontent.com/13282161/210854424-a0c1e9f2-ea8f-490b-b0c8-cc7d5bbeb88e.mp4

albullington:
I can replicate. If I scroll up so the header is offscreen, I can't tap to navigate anywhere above the blue markup line.

<img width="521" alt="Screen Shot 2023-01-05 at 11 30 49 AM" src="https://user-images.githubusercontent.com/35536439/210864851-b6ec0953-8180-4f4b-ace9-d84d72162a0f.png">

Chrischuck:
Looks to be blocked by something around this component

<img width="1189" alt="Screenshot 2023-01-05 at 3 15 24 PM" src="https://user-images.githubusercontent.com/13311268/210881494-7bb0befb-d3e9-4e21-8f0a-a4cc7d15afd6.png">

PR #343, opened 2023-01-08
Fix unclickable observations bug (#298)

Chrischuck:
Fixes issue #298 and #339 

Ended up translating the entire list rather than just the header.

### DIFF
--- a/src/components/Observations/ObsListHeader.js
+++ b/src/components/Observations/ObsListHeader.js
@@ -3,46 +3,33 @@
 import { View } from "components/styledComponents";
 import type { Node } from "react";
 import React from "react";
-import { Animated } from "react-native";
 import useCurrentUser from "sharedHooks/useCurrentUser";
 
 import LoggedOutCard from "./LoggedOutCard";
 import Toolbar from "./Toolbar";
 import UserCard from "./UserCard";
 
-const { diffClamp } = Animated;
-
-const HEADER_HEIGHT = 101;
-
 type Props = {
-  scrollY: any,
   setView: Function
 }
 
 const ObsListHeader = ( {
-  scrollY, setView
+  setView
 }: Props ): Node => {
   const currentUser = useCurrentUser( );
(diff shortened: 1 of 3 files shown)


kueda (changes requested):
This fixes the relevant issues on iOS for me, but on Android the app crashes after sign in while loading My Observations. There's no stack trace, and the most useful info in logcat is this:

```
signal 6 (SIGABRT), code -1 (SI_QUEUE), fault addr --------
Abort message: 'terminating with uncaught exception of type facebook::jni::JniException: java.lang.NoClassDefFoundError: Class not found using the boot class loader; no stack trace available'
```

Can you replicate, and either way, any idea what's going on?

Chrischuck:
@kueda  i'm getting that issue even on the main branch. did a clean build and everything...guessing it's a deeper issue but i havent had much time to debug. will take a look later today or tomorrow

kueda:
Seems to have occurred in 516b1a487ddc140288eb5a7b35cb52e23b0f1027. I'm spending some time on it today, but holler if you have any ideas.

Chrischuck:
@kueda I got lucky and was able to get an error message. Pushed the fix along in this pr.

Screenshot 2023-01-10 at 2 24 50 PM

kueda:
Hm, we've been hitting different problems with the `capitalize` function in `DisplayTaxonName`, which probably has to do with unicode handling in Hermes for Android. Replacing the long unicode string at https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/components/DisplayTaxonName.js#L46 with a blank string fixes the problem, so we might have to disable that weird capitalization code for now. We're experimenting with it and may just push a fix to main.

Chrischuck:
Sounds good to me, appreciate the help on this one

kueda (approved):

kueda:
Ok, the Android issue should be fixed in #347, and yeah, definitely unrelated to this work, so I think this is good to go.

PR #345, opened 2023-01-09
Create reusable tab components

Chrischuck:
Screenshot 2023-01-21 at 5 01 58 PM
Screenshot 2023-01-21 at 5 02 37 PM

### DIFF
--- a/src/components/Messages/Messages.js
+++ b/src/components/Messages/Messages.js
@@ -1,18 +1,23 @@
 // @flow
 
 import searchMessages from "api/messages";
+import Tabs from "components/SharedComponents/Tabs";
 import ViewWithFooter from "components/SharedComponents/ViewWithFooter";
 import { Text } from "components/styledComponents";
 import { t } from "i18next";
 import type { Node } from "react";
-import React from "react";
+import React, { useState } from "react";
 import useAuthenticatedQuery from "sharedHooks/useAuthenticatedQuery";
 import useCurrentUser from "sharedHooks/useCurrentUser";
 
 import MessageList from "./MessageList";
 
+const NOTIFICATIONS_ID = "NOTIFICATIONS";
+const MESSAGES_ID = "MESSAGES";
+
 const Messages = ( ): Node => {
   const currentUser = useCurrentUser( );
+  const [activeTab, setActiveTab] = useState( NOTIFICATIONS_ID );
   const {
     data,
     isLoading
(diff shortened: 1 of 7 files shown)


Chrischuck:
@kueda Apologies, I intended for this to be apart of the messages screen for issue #34 

Wouldn't mind for this to be closed if it's unneeded

albullington (changes requested):
This seems useful to me, since we can reuse this component in a few different places throughout the app. The two places I know we have this same style of UI is in ObsDetail (activity and data tabs) and Projects (joined, featured, and nearby tabs).

It would be great (but not necessary for this PR) if we could update ObsDetail and Projects with these tab components.

The one change I would request here is to make sure the text for each of the tabs is internationalized.

src/components/Messages/Messages.js
```diff
@@ -24,14 +29,34 @@ const Messages = ( ): Node => {
     }
   );
 
+  const tabs = [
+    {
+      id: NOTIFICATIONS_ID,
+      text: "Notifications",
+      onClick: () => {
```
albullington, line 36: This is really minor, but I prefer to use `onPress` or `onTap` since a user is not actually clicking a keyboard mouse on mobile.

src/components/SharedComponents/Tabs.js
```diff
@@ -0,0 +1,42 @@
...
+          const active = activeId === id;
+          const borderClass = `${active ? "bg-primary" : "bg-white"} h-1 rounded-t-lg`;
+          return (
+            <View key={id} className="flex-1">
+              <Button
+                onPress={onClick}
+                title={text}
+                color={active ? colors.primary : colors.grayText}
+                accessibilityLabel={text}
+              />
```
jtklein, line 33: Please add an accessibility state that reads to the user if the tab is selected.
https://reactnative.dev/docs/accessibility#accessibilitystate

jtklein, line 33: There is the possibility to give [accessibilityRole](https://reactnative.dev/docs/accessibility#accessibilityrole) of "tab". But I could not find out yet if that is also best practice to use for tabs that are inside the screen and not at the bottom navigator. Anyway, a role of "button" or "link" would not make accessible the behaviour of the app when tapping this, so I would like to see a role added here that is best practice.

jtklein (changes requested):

jtklein:
Forgot to add earlier, but a unit test for this new component would be great as well.

Chrischuck:
sounds good! thanks for the feedback all, i'll hopefully be able to get this done this weekend

Chrischuck:
Switched to touchable opacity because buttons looked terrible on android. Now they look consistent on both platforms.

Also added `jest-native` so we can test things like accessibility states in the future.

lmk if any further changes are needed!

Chrischuck:
> 

@albullington could you make a couple issues to swap out tabs on the ObsDetail and Projects pages? would be happy to do so

jtklein:
Dear @Chrischuck thank you for your work on this.
I will go ahead and merge this one in as we have no more comments. I did, however, make a few changes myself on top of your work here. Most notably, if you need any clarification about those:
1) I used `screen` and byLabel to find the tab components in the unit test. Both is recommended by testing-library.
2) I have defined default prop (empty array) outside of the component, as defining in the destructuring call can lead to unwanted rerenders down the line.
3) I have also added a a11y role of tablist to the container view. This is best practice in React at least, could not find much information if this has any effect in RN, though.

PR #346, opened 2023-01-10
Message preview

Chrischuck:
#34 

Adding basic styling for the message preview. Would it be possible to break down what all of the different visual cues mean? Green circle, green background, etc. etc.

Screenshot 2023-01-09 at 10 14 28 PM

### DIFF
--- a/src/api/messages.js
+++ b/src/api/messages.js
@@ -8,6 +8,7 @@ import handleError from "./error";
 const MESSAGE_FIELDS = {
   subject: true,
   body: true,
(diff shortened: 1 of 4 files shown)


kueda:
Nothing wrong with this, but messages is pretty far in the future, so I actually just closed that issue and will review some of the older ones and close the ones that are currently out of scope. Sorry about that. The designs in that issue are pretty out of date at this point frankly, so the issue isn't really ready for developers any more.

PR #352, opened 2023-01-12
Check for network connectivity & update UI, ObsDetail

### LINKED ISSUE #317: Network error when viewing Obs Detail offline
**Describe the bug**
When you view an observation while offline, you get an error: `TypeError: Network request failed`

**To Reproduce**
1. Set your phone to airplane mode
2. Go to My Observations and tap on an observation
3. The observation will show, but you'll see the error message

**Expected behavior**
Obs Detail should just show you whatever local data you have for that observation. Any network requests that are not required to show that data not happen, fail silently, or indicate to the user that those parts of the view will only work with an internet connection.

 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.1.1 (32)

albullington:
Closes #317

- Checks for internet connectivity in the new `useIsOnline` hook
- Sets a no-internet image for data which isn't available without a network request (photos, user images, taxon images, & the map in the data tab)
- Adds tests to check that the UI changes as expected when a device is offline/online
- Makes sure that unviewed comments/ids are only marked as viewed locally if the server request is successful


src/components/ObsDetails/ActivityItem.js
```diff
@@ -63,6 +66,19 @@ const ActivityItem = ( {
     }
   );
 
+  const showNoInternetIcon = ( ) => (
+    <View className="mr-3">
+      <IconMaterial name="network-check" size={30} />
```
jtklein, line 71: I have not seen this icon used in the context of showing that no internet connection is available. Standard would be `wifi-off`.
Can you add an i18n'ed `accessibilityLabel` to this new component. That would read to the user what is not available.

Both comments also applicable in the other places of this icon use.

tests/unit/components/ObsDetails/ObsDetails.test.js
```diff
@@ -63,17 +64,27 @@ jest.mock( "../../../../src/components/LoginSignUp/AuthenticationService", ( ) =
 
 jest.mock( "components/ObsDetails/AddCommentModal" );
 
+jest.mock( "sharedHooks/useIsOnline" );
```
jtklein, line 67: I think you need to use fakeTimers for this test then, because netInfo uses timers when checking internet connection. But you would need to check if that is true.

tests/unit/components/ObsDetails/ObsDetails.test.js
```diff
@@ -63,17 +64,27 @@ jest.mock( "../../../../src/components/LoginSignUp/AuthenticationService", ( ) =
...
+
+const mockLatLng = {
+  latitude: "91",
+  longitude: "-121"
+};
+
+jest.mock( "sharedHooks/useUserLocation", ( ) => ( {
```
jtklein, line 74: This is a question: Why you don't use the user factory for this type of data?

tests/unit/components/ObsDetails/ObsDetails.test.js
```diff
@@ -99,41 +110,62 @@ describe( "Observation with no evidence", () => {
...
   } );
 } );
 
-test( "navigates to taxon details on button press", async ( ) => {
-  const { findByTestId } = renderComponent( <ObsDetails /> );
+describe( "data tab", ( ) => {
+  test( "renders data tab content when DATA tab is pressed", ( ) => {
+    const { getByText } = renderComponent( <ObsDetails /> );
+    const dataTab = screen.getByText( /DATA/ );
+    fireEvent.press( dataTab );
```
jtklein, line 156: Let's also check that the data tab is falsey before navigating to it.

jtklein (commented):
The failing tests might be related to having to use fake timers for the components that use the new hook. I will take another look after you made the tests pass.

However, using this PR I still get the error described in the issue when navigating to the ObsDetail page when turning off the wifi on simulator. I will test on actual device once the tests pass though.
```
 WARN  [TypeError: Network request failed]
 ERROR  Error while renewing JWT: NETWORK_ERROR - null
```

jtklein (changes requested):
Forgot to switch to Changes requested earlier.

src/sharedHooks/useIsOnline.js
```diff
@@ -0,0 +1,11 @@
+import { useNetInfo } from "@react-native-community/netinfo";
```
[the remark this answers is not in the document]
albullington: good call... would be nice if we could automate this in some way

tests/unit/components/ObsDetails/ObsDetails.test.js
albullington: We could, I guess. In this case, `useUserLocation` is fetching geolocation from the user's device, so in my mind this doesn't really have anything to do with the realm User object. I'd be more inclined to make a `DeviceLocation.js` factory to mock this.

tests/unit/components/ObsDetails/ObsDetails.test.js
albullington: You lost me here. Why would the data tab be falsy while it's visible in the DOM and available for pressing?

albullington:
Made requested changes except for the one unresolved convo.

PR #366, opened 2023-01-13
Fix crash with locale by using unicode instead of chars

Chrischuck:
Switch to use unicode for regex matching since it's supported in es6


src/components/DisplayTaxonName.js
```diff
@@ -34,21 +34,13 @@ const uncapitalized = new Set( [
...
...C}\u{1FE}\u{200}\u{202}\u{204}\u{206}\u{208}\u{20A}\u{20C}\u{20E}\u{210}\u{212}\u{214}\u{216}\u{218}\u{21A}\u{21E}\u{226}\u{228}\u{22A}\u{22C}\u{22E}\u{230}\u{232}\u{3A9}\u{1E00}\u{1E02}\u{1E04}\u{1E06}\u{1E08}\u{1E0A}\u{1E0C}\u{1E0E}\u{1E10}\u{1E12}\u{1E14}\u{1E16}\u{1E18}\u{1E1A}\u{1E1C}\u{1E1E}\u{1E20}\u{1E22}\u{1E24}\u{1E26}\u{1E28}\u{1E2A}\u{1E2C}\u{1E2E}\u{1E30}\u{1E32}\u{1E34}\u{1E36}\u{1E38}\u{1E3A}\u{1E3C}\u{1E3E}\u{1E40}\u{1E42}\u{1E44}\u{1E46}\u{1E48}\u{1E4A}\u{1E4C}\u{1E4E}\u{1E50}\u{1E52}\u{1E54}\u{1E56}\u{1E58}\u{1E5A}\u{1E5C}\u{1E5E}\u{1E60}\u{1E62}\u{1E64}\u{1E66}\u{1E68}\u{1E6A}\u{1E6C}\u{1E6E}\u{1E70}\u{1E72}\u{1E74}\u{1E76}\u{1E78}\u{1E7A}\u{1E7C}\u{1E7E}\u{1E80}\u{1E82}\u{1E84}\u{1E86}\u{1E88}\u{1E8A}\u{1E8C}\u{1E8E}\u{1E90}\u{1E92}\u{1E94}H\u{331}T\u{308}W\u{30A}Y\u{30A}\u{1E60}\u{1EA0}\u{1EA2}\u{1EA4}\u{1EA6}\u{1EA8}\u{1EAA}\u{1EAC}\u{1EAE}\u{1EB0}\u{1EB2}\u{1EB4}\u{1EB6}\u{1EB8}\u{1EBA}\u{1EBC}\u{1EBE}\u{1EC0}\u{1EC2}\u{1EC4}\u{1EC6}\u{1EC8}\u{1ECA}\u{1ECC}\u{1ECE}\u{1ED0}\u{1ED2}\u{1ED4}\u{1ED6}\u{1ED8}\u{1EDA}\u{1EDC}\u{1EDE}\u{1EE0}\u{1EE2}\u{1EE4}\u{1EE6}\u{1EE8}\u{1EEA}\u{1EEC}\u{1EEE}\u{1EF0}\u{1EF2}\u{1EF4}\u{1EF6}\u{1EF8}\u{2202}\u{2206}\u{2211}FIFL";
```
jtklein, line 43: I haven't tested this yet, but do we need to remove the "SS" an "FIFL" from this string? Or otherwise something that starts with any of those characters will not be replaced?

jtklein (changes requested):
Can you also remove the comments above those lines that are no longer needed.

src/components/DisplayTaxonName.js
Chrischuck: I'm unsure since those were in the original charset as well

jtklein (approved):

PR #370, opened 2023-01-17
Email debug logs from About screen; log sign in and sign out

### LINKED ISSUE #357: Local logging with email option
We should have some form of logging. We definitely want crash logging and reporting and some ability to collect this data somewhere central, but to start, we just need some basic logging and sharing, so logging to a local file that can be emailed seems like the simplest solution. Plus [we already do it in Seek,](https://github.com/inaturalist/SeekReactNative/blob/adfddb027faffb79eeb4abe08745bea35850d3c2/components/UIComponents/DebugEmailScreen.js) so we know it works.

- [x] Set up [react-native-logs](https://github.com/onubo/react-native-logs) to write to a single log file
- [x] Log something simple but regular, like every time we sign the user in and out
- [x] On the About or Network/Logging views, add a button that opens an email to `help+mobile@inaturalist.org` and adds the file as an attachment

albullington:
Closes #357
- adds `react-native-logs` library with configuration to write all logs to a single file
- adds `react-native-mail` so user can open the native email client from the About screen
- sets the email address to help@inat and fetches the log as an attachment when opening mailer
- creates new log lines when a user signs in or out of the app


src/components/LoginSignUp/AuthenticationService.js
```diff
@@ -13,6 +13,10 @@ import Realm from "realm";
 // eslint-disable-next-line import/extensions
 import realmConfig from "realmModels/index";
 
+import { log } from "../../../react-native-logs.config";
+
+const userLog = log.extend( "user" );
```
kueda, line 18: I find it helpful to use the name of the file or module when adding logging prefixes like this, so how about logger `const = log.extend( "AuthenticationService" )` instead? If there's a smart way to get the name of the file dynamically, even better, but I think it's ok to hardcode it.

kueda (changes requested):
Just the one change, otherwise looking good. I couldn't send email in iOS b/c I didn't have Mail installed, but in investigating alternatives I ended up just implementing one, so I'm going to commit that. I know, I'm a terrible code reviewer.

kueda (approved):

PR #371, opened 2023-01-18
Add JS and native exceptions

albullington:
I've been able to test that this is working for JS exceptions using `react-native-crash-tester`. For JS exceptions, the log will show an error in this format:

`4:07:46 PM | root | ERROR : JS Error: Fatal: Error testing JS exceptions with CrashTester! `

I haven't yet been able to test that this is working for native exceptions, but creating a PR now in case we want to change anything about the formatting of the error logs.


index.js
```diff
@@ -15,13 +15,30 @@ import RealmProvider from "providers/RealmProvider";
...
 import { SafeAreaProvider } from "react-native-safe-area-context";
 import colors from "styles/tailwindColors";
 
 import { name as appName } from "./app.json";
+import { log } from "./react-native-logs.config";
+
+const rootLog = log.extend( "root" );
+
+const jsErrorHandler = ( e, isFatal ) => {
+  rootLog.error( `JS Error: ${isFatal ? "Fatal:" : ""} ${e.name} ${e.message}` );
```
kueda, line 31: This is great, but when I trigger a simple syntax error, like adding a line that's just `f` in Obs Detail, the log looks like

```
6:17:50 PM | root | ERROR : JS Error:  undefined undefined 
6:17:50 PM | root | ERROR : JS Error: Fatal: ReferenceError Property 'f' doesn't exist 
```

So there are two errors, one has a `name` and `message`, but the other doesn't. If I add `rootLog.info( "e: ", e );`, the error without those properties prints a full stack trace.

So I guess two questions:

1. Why are there two errors?
2. If we can access the stack trace in the second error, can we please log that too? If not, is there some way to consolidate them so we're not logging twice every time we get an exception?

index.js
```diff
@@ -15,13 +15,30 @@ import RealmProvider from "providers/RealmProvider";
...
 import { GestureHandlerRootView } from "react-native-gesture-handler";
 import { startNetworkLogging } from "react-native-network-logger";
 import { MD3LightTheme as DefaultTheme, Provider as PaperProvider } from "react-native-paper";
 import { SafeAreaProvider } from "react-native-safe-area-context";
 import colors from "styles/tailwindColors";
 
 import { name as appName } from "./app.json";
+import { log } from "./react-native-logs.config";
+
+const rootLog = log.extend( "root" );
```
kueda, line 28: Like in the other PR, I'd do something like `const logger = log.extend( "index" );` or `const logger = log.extend( "index.js" );` so it's a little easier to find where these log statements are coming from.

kueda (changes requested):

index.js
albullington: I believe the 2nd error is an issue with the `react-native-exception-handler` library rather than our own code, so I added a line to skip error reporting if both the error name and message are undefined.

And added stack trace to the log file.

albullington:
Made requested changes, including adding stack trace to the JS error log.

Also tested that the native error is working by adding a native exception in `RNGestureHandler.m` (from `react-native-gesture-handler` library in node_modules.

The native error prints like this to the log:
```
4:43:11 PM | index.js | ERROR : Native Error: THIS IS A TEST EXCEPTION
(
    "4   libc++abi.dylib                     0x00000001ec03ef28 A0028FDF-20F4-3A76-A43D-F0FA725BEE9F + 69416",
    "5   libc++abi.dylib                     0x00000001ec041ecc __cxa_rethrow + 148",
    "6   libobjc.A.dylib                     0x00000001973496b4 objc_exception_rethrow + 44",
    "7   CoreFoundation                      0x000000019e0e8f84 CFRunLoopRunSpecific + 808",
    "8   GraphicsServices                    0x00000001d813f368 GSEventRunModal + 164"
) 
```

ios/iNaturalistReactNative.xcodeproj/xcshareddata/xcschemes/iNaturalistReactNative.xcscheme
```diff
@@ -41,7 +41,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
```
kueda, line 44: Is this something we're going to want to revert when actually releasing?

albullington: hmm, this file changes automatically for me when I build with XCode

kueda: Ok, I'm going to assume that's not a big deal then, but something to keep in mind if we have problems with the next build.

kueda (approved):

PR #372, opened 2023-01-20
Add double tap to zoom in Camera; update Gesture API

### LINKED ISSUE #355: Standard Camera needs double tap to zoom
**Describe the bug**
Double tap does not zoom in the Standard Camera, it just seems to tap to focus twice. Double tap to zoom is pretty standard across all cameras so I think we should support it.

**To Reproduce**
1. Start a new observation by opening the camera
2. Double tap in the camera area; the view does not zoom

**Expected behavior**
Double tapping in the camera area should zoom to an extent that resembles the devices native camera.

 - Device: any
 - OS: any
 - App version: any

**Additional context**

albullington:
Closes #355 by adding double tap to zoom.

Also updates code to use the newest `react-native-gesture-handler` API with Gesture & GestureDetector. This makes code cleaner to read & makes it easier to support multiple gestures without needing multiple components.

### DIFF
--- a/src/components/Camera/CameraView.js
+++ b/src/components/Camera/CameraView.js
@@ -52,6 +52,7 @@ const CameraView = ( { camera, device }: Props ): Node => {
     if ( scale.value < SCALE_MAX_ZOOM ) {
       scale.value = Math.min( SCALE_MAX_ZOOM, scale.value += 1 );
     }
+    savedScale.value = scale.value;
   };
 
   const singleTap = Gesture.Tap( )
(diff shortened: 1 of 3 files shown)


kueda (changes requested):
Basic functionality is there, but is there a way to make the zoom animated when you double tap? On Android at least the change in zoom level is immediate. I'd prefer it to be eased a little.

albullington:
Added animation on double tap.

src/components/Camera/CameraView.js
```diff
@@ -36,10 +36,15 @@ const CameraView = ( { camera, device }: Props ): Node => {
   const isActive = isFocused && isForeground;
 
   // used for pinch and double tap to zoom
+  const [tap, setTap] = useState( true );
```
kueda, line 39: Here I go again with the naming: can we make this `doubleTapping` since it's tracking the status of a double tap event and nothing else? That way there's no wondering if it's a verb-like function or it represents a single tap status.

kueda (changes requested):
This is great. One minor naming issue and we're good to go. Looking at the Google camera, they actually support double tap to zoom, but if you *are* zoomed, double tap zooms you all the way out, which might be something we want to support later.

src/components/Camera/CameraView.js
kueda: Ok, poor form on part, but I did this in the interest of merging the other PR that depends on this one and in getting a build otu.

PR #374, opened 2023-01-20
Catch Android camera tap to focus error

### LINKED ISSUE #363: Tapping to focus twice in the Standard Camera raises unhandled promise rejection in Android
**Describe the bug**
In the Standard Camera in Android, if you tap to focus twice you get a possible unhandled promise rejection.

**To Reproduce**
1. Open the Standard Camera
2. Tap the camera area to focus
3. When the focus box appears, tap somewhere else in the camera area; you should get the unhandled promise rejection warning

**Expected behavior**
You should be able to tap to focus to your heart's content without getting hassled.

**Screenshots**
Screenshot_20230112-182909

 - Device: Pixel 3
 - OS: Android 12
 - App version: [fbeedb2](https://github.com/inaturalist/iNaturalistReactNative/commit/fbeedb2f2b979656cf7d1aa812f8c1722fe10c50)

albullington:
Closes #363
- improves focus box by showing it before `await` statement and in correct location
- uses try/catch to catch promise rejections when focusing (which seem to be coming from the camera X library)

Merge after PR #372


src/components/Camera/CameraView.js
```diff
@@ -30,84 +28,87 @@ type Props = {
...
-  // #endregion
+  const singleTapToFocus = async ( { x, y } ) => {
+    try {
+      singleTapToFocusAnimation.setValue( 1 );
+      setTappedCoordinates( { x, y } );
+      await camera.current.focus( { x, y } );
+    } catch ( e ) {
+      // Android often catches the following error from the Camera X library
+      // but it doesn't seem to affect functionality
+      // [unknown/unknown] Cancelled by another startFocusAndMetering()] error focusing camera
```
kueda, line 53: How about we throw any error that is *not* that "another startForcusAndMetering" error? Otherwise we're squelching potential exceptions that could tell us about real bugs.

kueda (changes requested):
Fixes the bug on my device, but I'd prefer to be selective about what errors we're ignoring.

kueda (approved):

PR #377, opened 2023-01-23
Use different library for location fetching

### LINKED ISSUE #369: Location fetch almost never uses GPS on Android
**Describe the bug**
Location retrieval almost never uses GPS in Andoird, resulting in low accuracy coordinates

**To Reproduce**
Actual low accuracy can be hard to reproduce depending on how well your system determines location based on non-GPS factors, but in general,

1. Make a new observation without evidence
2. Watch the coordinates and watch the system status bar (the very top of the screen); if you are relative far from any wifi networks or cell towers, the accuracy takes a very long time to come down or it never gets below ~100m, and in the status bar, the GPS indicator almost never appears

Compare that with using the native iNaturalist Android app, where the status bar immediately shows the GPS symbol when you start a new observation and the accuracy falls below 10m in a few seconds.

**Expected behavior**
The app should force the system to use GPS to get the most accuracy coordinates possible, time and battery be damned. I expect to see the accuracy fall below 10m within a few seconds and for the GPS symbol to appear in the status bar while it's fetching location.

 - Device: Pixel 3
 - OS: Android 12
 - App version: [247d580](https://github.com/inaturalist/iNaturalistReactNative/commit/247d58022cc02c4a7784ab665131a710f27129a4)

**Additional context**
We're using https://github.com/Agontuk/react-native-geolocation-service/, which uses the [Fused Location Provider API](https://developers.google.com/location-context/fused-location-provider/), but even when we use the `forceLocationManager` option to disable it, the GPS doesn't seem to be engaged, so I'm wondering if there's something wrong with the implementation of `react-native-geolocation-service`. I've tried playing around with the JS settings of `react-native-geolocation-service` but nothing seems to consistently access real GPS. If there's a pure JS solution here, great, but my gut says you'll need to download your own copy of `react-native-geolocation-service`, add a bunch of logging in the Android implementation there, and load that locally to see what's happening.

kueda:
Uses https://github.com/michalchudziak/react-native-geolocation to ensure location fetch on Android prioritizes use of the GPS over cell and wifi location.

Closes #369


kueda:
FYI @budowski, I needed to make a few changes to get tests to pass, and I removed the `distanceFilter` option which didn't seem to be documented, but otherwise this seems to work a lot better. I suspect it might be better to use `watchPosition` for more accurate and energy efficient retrieval, but that can be a subject for another issue.

Issue #353, opened 2023-01-12
Ensure JWT gets refreshed


kueda:
We use JSON Web Tokens (JWTs) as tokens of authorization when accessing iNat on behalf of a user. Our JWTs expire after 30 minutes, after which the API should return a 401 response when you try to use them. We currently try to address this in `getJWTToken` (which should probably be renamed since we're getting a JSON Web Token, not a JSON Web Token Token) by storing the time when we retrieved the token, and the next time we ask for one, we check to see if the existing token is about to expire (or is past expiry) and if it is we fetch a new one.

However, occasionally the app reports problems about the JWT not working. I haven't found consistent repro conditions, but it seems to happen after the app has been idle for a while, which suggests the JWT is expiring and not getting refreshed.

I *believe* problems occur when components and hook hold on to the JWT in state, e.g. in the `useApiToken` hook, which seems to hold on to the token forever until the user logs out. So one solid TODO here is to make sure this doesn't happen, either by always using getJWTToken directly to take advantage of its expiration check, or to build the expiration check into hooks and components that hold on to the token.

You might also considering adding some defensive code to all authenticated API requests that responds to 401 and 403 responses by refreshing the JWT and trying again once.

kueda:
FWIW, I hit what might be a related issue today: I installed the app with `npm run android`, signed in while connected to Metro, disconnected by device, drove into the woods where I had spotty Internet connectivity, opened the app again and was immediately signed out.

jtklein:
We merged changes so that now we will fetch a new token and retry if a 401 or 403 is hit using react-query. And we fetch the token on every render in `useApiToken` hook.

PR #382, opened 2023-01-24
353 retry with jwt

jtklein:
I was never able to reliably reproduce errors with an invalid or expired JWT. So I am not sure if anything is fixed with this now. However, this PR addresses two TODOs in this issue #353 :
1) Adds a retry for react-query calls when 401 or 403 is hit, after fetching the JWT token again (if expired).
2) Now the token is not kept in state in the useApiToken hook.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ import {
 } from "@tanstack/react-query";
 import handleError from "api/error";
 import App from "components/App";
+import { getJWT } from "components/LoginSignUp/AuthenticationService";
 import inatjs from "inaturalistjs";
 import ObsEditProvider from "providers/ObsEditProvider";
 import RealmProvider from "providers/RealmProvider";
@@ -60,7 +61,7 @@ inatjs.setConfig( {
 const queryClient = new QueryClient( {
   defaultOptions: {
     queries: {
-      retry: ( failureCount, error ) => {
+      retry: async ( failureCount, error ) => {
         if (
           // If this is an actual 408 Request Timeout error, we probably want to
           // retry... but this will probably never happen
@@ -69,6 +70,12 @@ const queryClient = new QueryClient( {
           || ( error instanceof TypeError && error.message.match( "Network request failed" ) )
         ) return failureCount < 3;
         handleError( error, { throw: false } );
+        if ( error.status === 401 || error.status === 403 ) {
+          // If we get a 401 or 403, call getJWT
+          // which has a timestamp check if we need to refresh the token
+          const token = await getJWT( );
+          if ( token ) return failureCount < 2;
+        }
         return false;
       }
     }
--- a/src/api/error.js
+++ b/src/api/error.js
@@ -1,4 +1,7 @@
 // @flow
+import { log } from "../../react-native-logs.config";
+
+const logger = log.extend( "INatApiError" );
 
 class INatApiError extends Error {
   // Object literal of the JSON body returned by the server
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -69,6 +70,12 @@ const queryClient = new QueryClient( {
           || ( error instanceof TypeError && error.message.match( "Network request failed" ) )
         ) return failureCount < 3;
         handleError( error, { throw: false } );
+        if ( error.status === 401 || error.status === 403 ) {
+          // If we get a 401 or 403, call getJWT
+          // which has a timestamp check if we need to refresh the token
+          const token = await getJWT( );
+          if ( token ) return failureCount < 2;
+        }
```
kueda, line 78: Don't you want to call `handleError` after trying to renew the token?

src/api/error.js
```diff
@@ -22,7 +25,7 @@ const handleError = async ( e: Object, options: Object = {} ): Object => {
   if ( !e.response ) { throw e; }
   const errorText = await e.response.text( );
   const error = new INatApiError( errorText );
-  console.error(
+  logger.error(
```
kueda, line 28: If `options.throw` is true we'll end up logging this as an exception... but maybe it is a good idea to log all failed requests for now. In a production context I imagine that might become excessive, though.

src/sharedHooks/useApiToken.js
```diff
@@ -1,33 +1,32 @@
...
 import { useEffect, useState } from "react";
 import useCurrentUser from "sharedHooks/useCurrentUser";
 
 const useApiToken = ( ): string | null => {
   const [apiToken, setApiToken] = useState( null );
-  const [shouldFetchToken, setShouldFetchToken] = useState( true );
   const currentUser = useCurrentUser( );
 
+  // Without the dependency array, this hook will run on every render, then every time
+  // a new token is fetched, and if expired fetched from API (in getJWT).
```
kueda, line 12: If we call `getJWT()` on ever render, why use an effect at all?

kueda (commented):
Thank you for removing JWTToken. A small, weird part of my brain is now much more calm. Not requesting any changes yet, just comments.

index.js
jtklein: Yes, the idea behind that was to call `handleError` anyway because I added logging to the log file of all handled errors to that function so that we know when 401 and 403 happen. Otherwise, since they are not thrown from here we wouldn't know. All of this logging here I added because I could not reproduce errors with the JWT reliably.

src/api/error.js
jtklein: Yes, my thoughts there were to increase logging because I could not reproduce the JWT errors on my devices. Hoping, that more logging will increase our chances of finding the problem here. I can leave a TODO to remove this in the future.

src/sharedHooks/useApiToken.js
jtklein: Good question. When I made the change I was more in the mindset of how can I reduce what is already there to a point where it works as I want, and did not think about how where that point is when I would build it from the ground up if that makes sense. But I have no preference over either of those two solutions, should I change it to not use a `useEffect`. I am not sure whether there are any technical aspects of that decision that I don't know about yet.

kueda: Try it without the effect and see if it works. If it does, I'm not really sure useApiToken is necessary at all (I may have added because hooks seemed like the more React-y way of doing work like this, but maybe it was overkill in this case).

kueda (approved):
My one minor request aside, I think this is ready to merge, and I'd be ok with merging without further refactoring `useApiToken()` too if you're tired of working on this.

jtklein:
I did remove the `useEffect` from the hook, but I don't think I can refactor away the hook completely, because getting the JWT is an async function.

jtklein:
I tried without the `useEffect` but since `getJWT` is async I ended up with an error about setting state on every render (although we guarded against that with conditions). So I'll leave it as is for now.

Issue #380, opened 2023-01-24
Replace Tabs on ObsDetails with reusable tab component


albullington:
On the middle of the ObsDetails page, there are two tabs: activity and data. We should swap these for the reusable tab components from PR #345.

PR #385, opened 2023-01-25
380 Make obs detail page use reusable tabs component

Chrischuck:
#380 
Screenshot 2023-01-24 at 6 49 35 PM

### DIFF
--- a/src/components/ObsDetails/ObsDetails.js
+++ b/src/components/ObsDetails/ObsDetails.js
@@ -9,6 +9,7 @@ import {
 import PhotoScroll from "components/SharedComponents/PhotoScroll";
 import QualityBadge from "components/SharedComponents/QualityBadge";
 import ScrollWithFooter from "components/SharedComponents/ScrollWithFooter";
+import Tabs from "components/SharedComponents/Tabs";
 import UserIcon from "components/SharedComponents/UserIcon";
 import {
   Image, Pressable, Text, View
@@ -55,6 +56,9 @@ LogBox.ignoreLogs( [
   "Non-serializable values were found in the navigation state"
 ] );
 
+const ACTIVITY_TAB = "ACTIVITY";
+const DATA_TAB = "DATA";
+
 const ObsDetails = ( ): Node => {
   const isOnline = useIsConnected( );
   const currentUser = useCurrentUser( );
@@ -62,7 +66,7 @@ const ObsDetails = ( ): Node => {
   const [refetch, setRefetch] = useState( false );
   const { params } = useRoute( );
   const { uuid } = params;
-  const [tab, setTab] = useState( 0 );
+  const [tab, setTab] = useState( ACTIVITY_TAB );
   const navigation = useNavigation( );
   const realm = useRealm( );
   const localObservation = useLocalObservation( uuid );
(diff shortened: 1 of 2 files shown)


src/components/ObsDetails/ObsDetails.js
```diff
@@ -55,14 +56,17 @@ LogBox.ignoreLogs( [
...
+
 const ObsDetails = ( ): Node => {
   const isOnline = useIsConnected( );
   const currentUser = useCurrentUser( );
   const userId = currentUser?.id;
   const [refetch, setRefetch] = useState( false );
   const { params } = useRoute( );
   const { uuid } = params;
-  const [tab, setTab] = useState( 0 );
+  const [tab, setTab] = useState( ACTIVITY_TAB );
```
jtklein, line 69: Would change to currentTabID

src/components/ObsDetails/ObsDetails.js
```diff
@@ -55,14 +56,17 @@ LogBox.ignoreLogs( [
   "Non-serializable values were found in the navigation state"
 ] );
 
+const ACTIVITY_TAB = "ACTIVITY";
```
jtklein, line 59: Also here, maybe better ACTIVITY_TAB_ID

src/components/ObsDetails/ObsDetails.js
```diff
@@ -241,27 +243,21 @@ const ObsDetails = ( ): Node => {
...
-    let textClassName = "color-gray text-xl font-bold";
-
-    if ( active ) {
-      textClassName += " color-inatGreen";
+  const tabs = [
+    {
+      id: ACTIVITY_TAB,
+      testID: "ObsDetails.ActivityTab",
+      onPress: ( ) => setTab( ACTIVITY_TAB ),
+      text: "ACTIVITY"
```
jtklein, line 251: Also here, like in other PR, we would use the translated strings here: `t( "ACTIVITY" )`

jtklein: Ah, I just see now that you have added the tranlation in the Tabs component directly. I haven't really realized then but I think I would prefer the Tabs component not to have the translation function call, and instead pass in the string already translated.
Also, the default testID should probably use the id string instead of title.

jtklein (changes requested):
Hi @Chrischuck thank you for your contribution!
I have requested added a few minor changes.

I have tested on Android and iOS and the tab functions do work there.
However, I have seen that everytime the Data tab is opened. Even when it was already opened before in the same page view, the whole tab get's initialized again, so, e.g., the map in the tab starts with the loading indicator. I think we should try to have the tab initialise on the first render only, and then keep it available in the page for re-access.
Also, on Android the map does not appear. But I will have to check if that is a regression from main or not.

Chrischuck:
@jtklein Thanks for the review! I went ahead and made those changes. Regarding the map reloading, looks like the ternary causes the components to come off the DOM every switch, and thus would trigger re-initialization like you said. Because that's how it previously was, do you think it would be out of scope of this PR to address? If so, I can totally take a look at it this weekend.

jtklein:
Thanks again @Chrischuck all good for me now. I have tested with a release build on both platforms, so I will merge it in now.

Issue #360, opened 2023-01-13
Standard Camera layout doesn't adapt when rotating to landscape orientation


kueda:
**Describe the bug**
When you rotate the device from portrait to landscape orientation, the layout does adjust and the bar at the bottom remains the same width

**To Reproduce**
1. With the device in portrait orientation, open the Standard Camera
2. Rotate the device to landscape orientation; the bar including the shutter and back buttons remains the same width it was in portrait, flush left

**Expected behavior**
That bar should fill the width of the screen, OR the Standard Camera should simply not allow rotation on small screens (this seems to about the behavior of the Google Camera on my Pixel, though I feel like rotation should be allowed on large devices like tablets where there's room for it.

**Screenshots**
Here's what this looks like on my Pixel in landscape:
![Screenshot_20230112-140426](https://user-images.githubusercontent.com/22691/212221261-b75024ae-335a-4876-8998-49d5165ddb32.jpg)

**Context (please complete the following information):**
 - Device: Pixel 3
 - OS: Android 12
 - App version: [fbeedb2](https://github.com/inaturalist/iNaturalistReactNative/commit/fbeedb2f2b979656cf7d1aa812f8c1722fe10c50)

**Additional context**
Possibly the same solution as #350. One additional requirement: when the device is in landscape orientation, the pictures taken should also be in landscape orientation, even if the UI doesn't change!

albullington:
@angielt are you still working on this one? It looks like most of this is done, but we still need:

- [ ] Show landscape mode on tablets, but not phones (landscape is working for me on iPhone at the moment)
- [ ] Saved photos to be taken & displayed (in the media viewer, for example) in landscape mode, not portrait mode

angielt:
This issue was probably one I worked on in a branch for another issue since they were similar and I forgot to close it at the time. I remember disabling landscape orientation for smaller devices but it seems to not be working currently. I'll work on this while i'm working on other issues with the Standard Camera.

Issue #350, opened 2023-01-11
Can't tap "Next" from Standard Camera on iPad in portrait orientation


kueda:
**Describe the bug**
When you open the Standard Camera on an iPad (or presumably another tablet) and the device is in portrait orientation, the shutter button is off-center and the "Next" button is off screen. You have to rotate to landscape to tap "Next"

**To Reproduce**
1. Open the app in a tablet
1. Start a new observation with the Standard Camera
1. Take a picture
1. Rotate the device to portrait orientation; the "Next" button is off screen and the shutter button is off-center

**Expected behavior**
Standard Camera buttons should adapt to any orientation the device supports on any device.

**Screenshots**

![IMG_0059](https://user-images.githubusercontent.com/22691/211702062-17e0e7bb-bfc4-422f-ab87-7f0daa061f9a.PNG)

**Context (please complete the following information):**
 - Device: iPad Air, 5th gen
 - OS: iOS 15.7
 - App version: [8c13f0b](https://github.com/inaturalist/iNaturalistReactNative/commit/8c13f0ba1cb5f41e932cdab5bc6d8d865442730c)

PR #386, opened 2023-01-25
350 cant tap next from standard camera on ipad in portrait orientation

angielt:
Combined issues 350 and 360. Currently has some npm dependency warnings.


src/components/Camera/PhotoPreview.js
```diff
@@ -12,13 +12,14 @@ import React, { useContext, useState } from "react";
...
-  savingPhoto: boolean
+  savingPhoto: boolean,
+  deviceOrientation: string
 }
 
 const PhotoPreview = ( {
   photoUris,
   setPhotoUris,
-  savingPhoto
+  savingPhoto, deviceOrientation
```
kueda, line 22: New line for this, please, and basically any time you see a newline-separated list of object keys.

src/components/Camera/PhotoPreview.js
```diff
@@ -57,6 +58,22 @@ const PhotoPreview = ( {
...
+  //       setRotation( true );
+  //       console.log( "rotation true" );
+  //     } else if ( !isTablet && width < height ) {
+  //       // change image rotation
+  //       setRotation( false );
+  //       console.log( "rotation false" );
+  //     }
+  //   } );
+  // }, [] );
+
```
kueda, line 76: If we're not using this any more, please delete it.

src/components/SharedComponents/PhotoCarousel.js
```diff
@@ -88,6 +94,22 @@ const PhotoCarousel = ( {
...
+      if ( deviceOrientation && !isTablet ) {
+        if ( deviceOrientation === "portrait" ) {
+          className += "rotate-0";
+        }
+        if ( deviceOrientation === "landscapeLeft" ) {
+          className += "-rotate-90";
+        }
+        if ( deviceOrientation === "landscapeRight" ) {
+          className += "rotate-90";
+        }
```
kueda, line 108: These are all mutually exclusive, so please use `else` or a `switch`.

tests/unit/components/Camera/StandardCamera.test.js
```diff
@@ -17,6 +17,17 @@ jest.mock( "@react-navigation/native", () => {
...
+  addEventListener: jest.fn(),
+  addDeviceOrientationListener: jest.fn(),
+  removeEventListener: jest.fn(),
+  lockToPortrait: jest.fn(),
+  lockToLandscapeLeft: jest.fn(),
+  lockToLandscapeRight: jest.fn(),
+  unlockAllOrientations: jest.fn(),
+  removeOrientationListener: jest.fn()
+} ) );
+
```
kueda, line 30: I can't think of a reason that we'd be able to ever use a non-mocked version of this library, so I think this belongs in `jest.setup.js`, or another global test setup area.

kueda (changes requested):
Mostly looking great, and that is REALLY weird about the different orientation definitions in different libraries. #350 is definitely fixed here, and #360 is mostly addressed, except for the final requirement:

> when the device is in landscape orientation, the pictures taken should also be in landscape orientation, even if the UI doesn't change!

That actually does work on my iPad (yay), but not on my Pixel 3, and I suspect not on any non-tablet. Fixing that might be a whole different and more difficult problem. Did you look into it at all? If not, maybe we should carve that out into its own issue, since the work in this PR doesn't make that problem any worse.

In addition to the comments, some more minor cleanup:

- [x] Remove commented code
- [x] Remove `console.log` statements

src/navigation/mainStackNavigation.js
```diff
@@ -82,7 +85,7 @@ const MainStackNavigation = ( ): React.Node => (
       <Stack.Screen
         name="StandardCamera"
         component={StandardCameraWithPermission}
-        options={hideHeader}
+        options={{ ...hideHeader, orientation: isTablet ? "all" : "portrait" }}
```
kueda, line 88: I think this is reasonable given the info you were working with, so this is just a note to myself to make an issue to disallow rotation on non-tablets everywhere *except* the AR and Standard Cameras.

angielt: For the react-native-orientation-locker there was an active issue with iOS and `Orientation.lockToPortrait()` so this line was also for that case but if orientation is locked for all non-tablet devices this line and `Orientation.lockToPortrait()`  may not be needed. I should have asked earlier since i was wondering if it was best to just keep orientation locked to portrait on smaller devices.

kueda: [Per Abhas](https://inaturalist.slack.com/archives/C02G5AATVGA/p1674755429544969?thread_ts=1674699000.505279&cid=C02G5AATVGA), "I think in the AR camera & regular camera yes, but otherwise I don’t think it’s necessary", so I think we'll eventually not want to disable rotation for smaller devices, but disable it on most screens except the cameras. Again, I don't think we need to deal with that here, though.

src/components/Camera/StandardCamera.js
```diff
@@ -154,15 +178,16 @@ const StandardCamera = ( ): Node => {
...
         photoUris={cameraPreviewUris}
         setPhotoUris={setCameraPreviewUris}
         savingPhoto={savingPhoto}
+        deviceOrientation={imageOrientation}
       />
       <FadeInOutView savingPhoto={savingPhoto} />
-      <View className="absolute bottom-0">
-        <View className="flex-row justify-between w-screen mb-4 px-4">
+      <View className="absolute bottom-0 w-full">
+        <View className={`flex-row justify-between w-${footerWidth} mb-4 px-4 `}>
```
kueda, line 190: Extraneous space at the end here.

src/components/Camera/StandardCamera.js
```diff
@@ -63,10 +53,47 @@ const StandardCamera = ( ): Node => {
...
+      if ( isTablet ) setFooterWidth( width );
+    } );
+
+    return () => {
+      Orientation.removeOrientationListener( onDeviceRotation );
+    };
+  }, [isTablet, footerWidth] );
+
   const takePhoto = async ( ) => {
+    console.log( "pressed" );
```
kueda, line 96: There are still a few more console.log statements.

kueda (changes requested):
Just really minor stuff. Otherwise good to merge.

kueda (approved):

PR #394, opened 2023-01-26
E2e test UI library

jtklein:
I have added an e2e test for the UI library page.
At the moment it just taps on every element on the screen. We can update this test as long as we push new components to this screen to check the integrity of the base components. Then we might use this test also for copying interactions with standard components into other test scenarios.


kueda:
I think this is an over-application of testing. The UI Library is just a development tool we can consult while building other things. Users should never see it and thus I don't think it warrants the time spent writing and maintaining tests. I agree each individual UI component merits UI testing, and they merit integration testing on the user-facing views they're used in, and they will implicitly be tested by a few e2e tests of actual functionality, but I don't think we need to test our temporary development tools.

PR #395, opened 2023-01-26
110 upload sounds and obs separately

budowski:


kueda (commented):
I get a 422 when I try to post an observation with a photo because POST /v2/observation_photos requires a file param. This is partially on me because I thought it would just accept a photo ID and an observation ID and be good to go, but I guess not. Let me consult with @pleary about how we want the API to support making observations *after* posting media. I guess this paused until then.

budowski:
We basically did an update to the `inatjs` library to call that endpoint in a specific way that doesn't require the `file` param. Maybe `npm install` latest version?

kueda:
@budowski, I think you didn't update package-lock.json to reflect that new version of inaturalistjs, so I just did that and merged from main. Will merge if/when tests pass.

PR #402, opened 2023-01-27
UI redesign: Typography components

### LINKED ISSUE #383: Typography components
We need a collection of typography components for rendering text at specific sizes and font faces. These are available in the "Style Guide" section of the UI designs, but I'll try and list their current state and their properties here, and we can change them if necessary. IMO, we should making these as separate components and not, say, named styles, for clarity and for ease of dev/design communication.

Note that the UI designs specify the font "Novel Display" for some styles, but we don't have a license for that yet and our attempts to learn about one have met with silence, so we're going to use Whitney for now. Also, the designs specify sizes in pixels, but I have converted them to `rem` units here based on an 18px base size.

| Component | Font Face | Font Size | Line Height | Letter Spacing | Tailwind Font Size |
| --------- | --------- | --------- | ----------- | -------------- | ------------------ |
| `<Heading1 />` | Whitney-Medium | 1.44rem | 1.733rem | default | `3xl` |
| `<Heading2 />` | Whitney-Medium | 1.22rem | 1.466rem | default | `2xl` |
| `<Heading3 />` | Whitney-Medium | 1.055rem | 1.266rem | default | `lg` |
| `<Heading4 />` | Whitney-Medium | 0.888rem | 1.066rem | 0.11rem | |
| `<Subheading1 />` | Whitney-Light | 1.166rem | 1.4rem | default | `xl` |
| `<Body1 />` | Whitney-Light | 1rem | 1.2rem | default | `base` |
| `<Body2 />` | Whitney-Light | 0.888rem | 1rem | default | |
| `<Body3 />` | Whitney-Light | 0.777rem | 1rem | default | `sm` |
| `<Body4 />` | Whitney-Light | 0.666rem | 0.8rem | default | `xs` |
| `<List1 />` | Whitney-Light | 1rem | 1.2rem | default | `base` |
| `<List2 />` | Whitney-Light | 0.777rem | 0.933rem | default | `sm` |

For each of those, make a component that

- [x] Accepts children, including text
- [x] Can be further styled with a `className` prop
- [x] Can be further styled with a `style` prop
- [x] Accepts a `testID` prop
- [x] Show how to use it in the UI Library
- [x] Replace equivalent uses of `<Text>` in the UI Library with these typography components

IMO, these should go somewhere like `src/components/SharedComponents/typography`, but I'm open to suggestions. I strongly recommend using the Tailwind Font Size values I've listed to customize the tailwind theme, and then use tailwind to size the components, e.g. `Heading1` would be something like `<Text className="text-2xl">Foo</Text>`.

If any of the above is confusing

albullington:
Closes #383


src/components/UiLibrary.js
```diff
@@ -1,32 +1,39 @@
...
         component, please put it here first and try to show what it looks
         like with different property configurations.
-      </Text>
-
-      <Text className="text-xl">Buttons</Text>
-
-      <Text className="text-lg">Button</Text>
+      </Body1>
+      <Heading1>Buttons</Heading1>
+      <Subheading1>Button</Subheading1>
```
kueda, line 36: I don't *think* the intent of `Subheading1` is to show hierarchy, so let's use Heading2 here and everywhere else you've used `Subheading1`.

kueda (changes requested):
Looks good, but a few requests:

- [ ] If we can't use rem units, please use the px values in Figma, so Heading1 is 26px, line height 31.2px
- [ ] There's a lot of repeated logic in these components. There are a number of ways we could address that, like inheritance (i.e. classes and subclasses) or just not making these separate components at all and make an `<INatText>` component with different types, or maybe there's a way to do this with the custom Tailwind theme. Thoughts?

kueda (changes requested):
As we discussed elsewhere, the text sizes aren't quite in line with the spec, but otherwise this looks good. I wish there was a way to reduce duplication *and* abide by the `react/jsx-props-no-spreading` rule, but I don't really see a way around that right now.

PR #406, opened 2023-01-27
Fix nearby projects query for android

### LINKED ISSUE #404: Nearby tab shows no data in projects on Android
**Describe the bug**
On Android, on the Projects screen the Nearby tab does not display any projects, whereas on iOS data is displayed. This was brought up in PR https://github.com/inaturalist/iNaturalistReactNative/pull/389

**To Reproduce**
1. Go to 'Projects' screen
2. Tap on 'Nearby' tab
3. Data for projects does not change

**Expected behavior**
Data should appear in the Nearby tab, as it does on iOS.

**Additional context**
An error is thrown in the console: `Error: Fatal: TypeError: Cannot read property 'latitude' of null`

Chrischuck:
Mostly fixes #404 

Turns out it was just slow on Android. Updated the `useUserLocation` hook to optionally make the place guess which was causing an unnecessary  delay (both places which use `useUserLocation` do not use the place guess currently.

### DIFF
--- a/src/components/Projects/ProjectTabs.js
+++ b/src/components/Projects/ProjectTabs.js
@@ -37,10 +37,10 @@ const ProjectTabs = ( ): Node => {
     } else if ( currentTabId === FEATURED_TAB_ID ) {
       setApiParams( { features: true } );
     } else if ( currentTabId === NEARBY_TAB_ID && latLng ) {
-        setApiParams( {
-          lat: latLng.latitude,
-          lng: latLng.longitude
-        } );
+      setApiParams( {
+        lat: latLng.latitude,
+        lng: latLng.longitude
+      } );
     }
   }, [
     memberId,
--- a/src/components/Projects/ProjectTabs.js
+++ b/src/components/Projects/ProjectTabs.js
@@ -36,13 +36,11 @@ const ProjectTabs = ( ): Node => {
       setApiParams( { member_id: memberId } );
     } else if ( currentTabId === FEATURED_TAB_ID ) {
       setApiParams( { features: true } );
-    } else if ( currentTabId === NEARBY_TAB_ID ) {
-      if ( latLng ) {
+    } else if ( currentTabId === NEARBY_TAB_ID && latLng ) {
         setApiParams( {
           lat: latLng.latitude,
           lng: latLng.longitude
         } );
-      }
     }
   }, [
     memberId,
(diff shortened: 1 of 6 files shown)


src/components/Projects/ProjectTabs.js
```diff
@@ -29,47 +29,45 @@ const ProjectTabs = ( ): Node => {
...
-      fetchUserJoinedProjects( );
+    if ( memberId && currentTabId === JOINED_TAB_ID ) {
+      setApiParams( { member_id: memberId } );
+    } else if ( currentTabId === FEATURED_TAB_ID ) {
+      setApiParams( { features: true } );
+    } else if ( currentTabId === NEARBY_TAB_ID ) {
+      setApiParams( {
+        lat: latLng.latitude,
+        lng: latLng.longitude
+      } );
```
Chrischuck, line 41: Moved these to the `useEffect` because I found in some instances that you can be on one tab, but different data shows. Usually happened when navigating away from the projects page and back.

Now the query is a reflection of the current tab.

src/realmModels/Observation.js
```diff
@@ -192,7 +192,7 @@ class Observation extends Realm.Object {
   }
 
   static projectUri = obs => {
-    const photo = obs.observation_photos[0];
+    const photo = obs?.observation_photos?.[0];
```
Chrischuck, line 195: unrelated bug, but this was giving me undefined errors when viewing nearby projects on android

kueda:
@Chrischuck if you're still working on this, can you mark it as a draft PR? That way we know not to review it yet.

Chrischuck:
@kueda should be ready to review now!

src/components/Projects/ProjectTabs.js
```diff
@@ -20,64 +21,65 @@ const ProjectTabs = ( ): Node => {
...
-
   useEffect( ( ) => {
-    if ( memberId ) {
-      fetchUserJoinedProjects( );
+    if ( memberId && currentTabId === JOINED_TAB_ID ) {
+      setApiParams( { member_id: memberId } );
+    } else if ( currentTabId === FEATURED_TAB_ID ) {
+      setApiParams( { features: true } );
+    } else if ( currentTabId === NEARBY_TAB_ID ) {
+      if ( latLng ) {
```
kueda, line 40: I think you could just add this to the `else if` above, e.g. `} else if ( currentTabId === NEARBY_TAB_ID && latLng ) {`

kueda (approved):
Looks good to me. My one request is cosmetic, so if you don't have time to get to that let me know and I'll just merge.

Issue #399, opened 2023-01-27
NavBar component


kueda:
This is the element at the bottom of every screen with drawer, explore, add evidence, my obs, and notifications. In designs we've been calling it the Tab Bar, and in code we're calling it the Footer, but I want to unify around the term NavBar (nothing's perfect, but I think that's a little less-likely to be confused with other elements... maybe).

![NavBar](https://user-images.githubusercontent.com/22691/214983702-c185a9a3-de57-4555-a9ae-6f3936450cba.png)

- [ ] Rename Footer to NavBar
- [ ] Replace My Obs button image with the current user's profile pic when signed in, use the same generic user icon when signed out
- [ ] Make icon green when the user is viewing that part of the app (for My Obs while signed in the profile pic should have a green border
- [ ] Add those drop shadows on the whole bar and the add evidence button
- [ ] Adjust dimensions to match the [spec](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=333%3A723&t=C1renIwFPVp304V0-1)

I don't think this needs to be in the UI library since it's only used in one, really obvious place.

Chrischuck:
I can make some incremental changes on this. I can have a PR up for renaming, user's profile pic, and making icons green.

jtklein:
This PR https://github.com/inaturalist/iNaturalistReactNative/pull/407 addresses the bullet points as mentioned above.

PR #407, opened 2023-01-28
Incremental changes to navbar component

Chrischuck:
#399 
Screenshot 2023-01-28 at 12 57 57 PM


jtklein (commented):
I have not looked at everything in detail yet. But I can send you more about the design specs if you want to continue working on this:
The outermost two icons are both 32x32 px.
The two elements around the center button are both 40x40px
And the center AddObsButton is 69x69px.

jtklein:
@Chrischuck we have now included our own icons in a font file. Could you update this PR from main and resolve the conflicts?

src/components/SharedComponents/NavBar/NavButton.js
```diff
@@ -0,0 +1,75 @@
...
+  size: number,
+};
+
+const imgStyle = ( size, active ) => StyleSheet.create( {
+  img: {
+    borderWidth: active ? 3 : 0,
+    height: size,
+    width: size,
+    borderColor: colors.inatGreen
+  }
```
Chrischuck, line 26: I'd rather have used tailwind here...

Unfortunately setting the border width using class `border-[3px]` didn't work on android

setting height via classes didnt work on either platform

and you can't interpolate class's in tailwind

jtklein:
Thank you @Chrischuck! I did add some more changes on top. I have removed all of your additions to testing the unrelated screens for this PR. To keep them more granular as unit tests I have added a mock for the route instead of adding a layer of elements in the testing tree. The navbar test you added is also interacting with the drawer, so it is more an integration test than unit test. I have moved it. So, probably good to add a unit test at one point.

Issue #398, opened 2023-01-26
InlineUser component


kueda:
`<InlineUser user={user} />` corresponds to ["Username & Profile Picture"](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=1492%3A12073&t=C1renIwFPVp304V0-0) in the UI designs:

<img width="191" alt="Screen Shot 2023-01-26 at 3 22 28 PM" src="https://user-images.githubusercontent.com/22691/214972803-b44113ac-810a-401d-9fa2-091dbd191051.png">

So that's

- [x] a 22px / 1.222rem `<UserIcon />`
- [x] 7px / 0.388rem space
- [x] `@` and the username (login) in Body3 style (depends on #383)

I guess for now we don't worry about truncation.

kueda:
@Chrischuck would you mind commenting on the issue before working on it? The work is much appreciated, but I don't think I can assign you until you comment, and it also helps others know that the issue is being worked on already.

Chrischuck:
@kueda apologies, forgot to do so.

jtklein:
This #410 addresses some of the TODOs only. The further required change from the above TODO is the Text after the new Text components is available.

However, I would like to add:

- [ ] Replace the fallback images (no user image & offline) with icons from our own icon font after those two icons are included
- [ ] Add unit test for the UserIcon component
- [ ] Lastly (after all UI is finished): Add snapshot test for the UserIcon and InlineUser component

PR #388, opened 2023-01-25
Update Button component with new styling

albullington:
Closes #384

PR #410, opened 2023-01-28
Create base inline user component

Chrischuck:
#398 

Base component for inline user. Simple user icon for fallback. Not totally complete because text is not up to spec
Screenshot 2023-01-29 at 6 47 30 PM

### DIFF
--- a/src/components/SharedComponents/InlineUser.js
+++ b/src/components/SharedComponents/InlineUser.js
@@ -4,6 +4,7 @@ import { useNavigation } from "@react-navigation/native";
 import {
   Image, Pressable, Text, View
 } from "components/styledComponents";
+import { t } from "i18next";
 import type { Node } from "react";
 import React from "react";
 import IconMaterial from "react-native-vector-icons/MaterialIcons";
@@ -22,8 +23,11 @@ const InlineUser = ( { user }: Props ): Node => {
   const userHandle = User.userHandle( user );
   return (
     <Pressable
+      testID="InlineUser"
       className="flex flex-row items-center"
       accessibilityRole="link"
+      accessibilityLabel={t( "Navigate-to-user-profile" )}
+      accessibilityValue={{ text: userHandle }}
       onPress={() => {
         navigation.navigate( "UserProfile", { userId: user.id } );
       }}
--- a/src/components/SharedComponents/InlineUser.js
+++ b/src/components/SharedComponents/InlineUser.js
@@ -19,7 +19,7 @@ const InlineUser = ( { user }: Props ): Node => {
         <Image
           testID="InlineUser.ProfilePicture"
           className="w-6 h-6 rounded-full mr-1.5"
-          accessibilityRole="img"
+          accessibilityRole="image"
           source={userImgUri}
         />
       ) : (
@@ -27,7 +27,7 @@ const InlineUser = ( { user }: Props ): Node => {
           <IconMaterial
             testID="InlineUser.FallbackPicture"
             name="person"
-            role="img"
+            accessibilityRole="image"
             size={24}
             color={colors.logInGray}
           />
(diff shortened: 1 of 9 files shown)


src/components/SharedComponents/InlineUser.js
```diff
@@ -0,0 +1,29 @@
...
+  user: any,
+};
+
+const InlineUser = ( { user }: Props ): Node => {
+  const userImgUri = User.uri( user );
+  const username = `@${user.login}`;
+  return (
+    <View className="flex flex-row items-center">
+      {userImgUri ? (
+        <Image className="w-6 h-6 rounded-full mr-1.5" source={userImgUri} />
```
Chrischuck, line 20: `w-6` and `h-6` both correlate to `width: 1.5rem; /* 24px */` on the tailwinds docs. Closest to the 22px spec.

jtklein (commented):
I haven't looked at the full code yet, however:
1) The image is not rounded when I run this branch.
2) The distance between default image (= no user photo) and text is too small.
3) I would like to see a unit test for all those new base components.
4) I would like to see default accessibility props.

If 3, 4 does not interest you, let me know and I can take a look.

Chrischuck:
@jtklein Very odd...I'm thinking tailwinds is somehow off and causing the square image? I'll look into it.

Can I get more guidance for the sizing and spacing. I went off of the following: 
`22px / 1.222rem <UserIcon />` ->  `w-6/h-6` in tailwinds -> `width: 1.5rem; /* 24px */`
`7px / 0.388rem space` -> `mr-1.5` in tailwinds -> `margin-right: 0.375rem; /* 6px */`

As for Text size, I just had that as a place holder until the typography components are in.

I for sure can get unit tests and a11y changes in. Would you be able to link what exactly to add? I'm not well versed in that area and would love to learn!

Chrischuck:
@jtklein 

I think [this](https://github.com/inaturalist/iNaturalistReactNative/pull/388/files#diff-ddc1067f080930bc48865f4660316a14b21bbc9f9ba85975ce8479342b983d2eR26-R29) might've broken it. After taking those lines out, cleaning project, and rebuilding border radius has come back.

jtklein (changes requested):
Thanks @Chrischuck that was a lot of work!
I have had some more time now to look at this in detail. I was trying to run the branch but I got an error message about using an a11y role of "img" being not available, so I changed it to work for me, hope you don't mind. I also updated from main to get the fix for rounded elements in. This also works now.

One general thing I have noticed, but it is also not in the issue text, is that the component has no press handler. I think that the default behaviour of this component should be to navigate to the shown user profile. Can you wrap the entire component in a Pressable and add the navigation functions. You can copy the code from example 1) see below.

The accessibility props you can also copy over from example 1).

Another good thing to include I think would be a fallback image for when the component is trying to fetch an image from server for a user not in the local db, and the internet fails. You can copy over the baseline code for that from example 2).

In general I would like to ask you to move over the code from those two examples and replace them with the new component as well.

We have two places where we use a version of this currently:
1) [Pressable](https://github.com/Chrischuck/iNaturalistReactNative/blob/cd7514a2bf8e1eb660f9e5cdb7929c91ebdf71a8/src/components/ObsDetails/ObsDetails.js#L323)
2) [Pressable](https://github.com/Chrischuck/iNaturalistReactNative/blob/cd7514a2bf8e1eb660f9e5cdb7929c91ebdf71a8/src/components/ObsDetails/ActivityItem.js#L91)

jtklein:
Screenshot 2023-01-30 at 14 03 15

Chrischuck:
@jtklein Thanks for the review! Totally don't mind pushed on top of this. I was getting that `img` error. Turns out `accessibilityRole` must be `image` and role is `img`.

jtklein:
Hey @Chrischuck, I did some more refactoring and added a different fallback image for when the device is offline. I moved some bits in the tests around to not repeat ourselves. The approach with the beforeEach after Each for the offline test is too much for just one test probably but I did it that way as a more exemplary test to follow for the basic components we will be adding soon.

Issue #405, opened 2023-01-27
ObsListToolbar component


kueda:
The `<ObsListToolbar>` sits at the bottom of the header on My Observations and remains stuck to the top of the screen as you scroll down. The intent here is to make a simple component that just responds to state, not one that actually uploads observations. I guess this isn't reusable, but it is something we'll need on My Observations, and it has a lot of moving parts.

There is a component for this in the Components sheet of the UI designs, but most of the actual states are on the [designs for My Observations](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=315%3A481&t=V7n3F2CGYm2TIfQO-1).

Props:
* `navigateToExplore`: navigates the user to Explore with contextually appropriate filters (i.e. only showing the current user's observations
* `startUpload`: function starts uploading all observations that need it
* `stopUpload`: function that stops upload in progress
* `uploadInProgress`: boolean indicating whether an upload is in progress
* `numToUpload`: number of records that need to be uploaded
* `numUploaded`: number of records in the current upload batch that have been uploaded
* `uploadError`: string containing one
* `layout`: string describing current layout, `grid` or `list` (default: `list)
* `toggleLayout`: function to change the layout to the opposite of what it is now

Here's an attempt at describing all the functionality, but you'll need to consult the designs for all the visual details.

- [ ] Explore button (compass rose icon): fires navigateToExplore when tapped, only visible when signed in
- [ ] Upload status (sync and related icons): combination button and status indicator,
  - [x] should be `inatGreen` when there are observations to upload or an upload is in progress, `darkGray` otherwise
  - [x] when `uploadInProgress` is false, tapping it should fire `startUpload()`
  - [ ] when `uploadInProgress` is false and numToUpload is 0, text should not appear
  - [x] when `uploadInProgress` is false and numToUpload is > 0, text should read "Upload {$numToUpload} observations"
  - [ ] - [ ] when `uploadInProgress` is false and numToUpload is > 0, icon should be an exclamation point inside of a sync icon (might just need to be a separate icon)
  - [ ] when `uploadInProgress` is false and numToUpload is > 0 and the device is the smallest size, text should read "Upload"
  - [x] when `uploadInProgress` is true, text should read "Uploading {$numUploaded} of {$numToUpload} observations"
  - [ ] when `uploadInProgress` is true and the device is the smallest size, text should read "{$numUploaded} of {$numToUpload}"
[...]

PR #411, opened 2023-01-28
Intro work to obs list toolbar redesign

Chrischuck:
#405 
Will need future changes as Icon and typography components enter the mix.

Will also need to change locale tests to look for a different copy, probably "welcome" instead of observations as it was previously looking for.

Screenshot 2023-01-28 at 8 06 56 PM
Screenshot 2023-01-28 at 8 21 01 PM
Screenshot 2023-01-28 at 8 23 59 PM


src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
 // @flow
```
albullington, line 1: Can we rename this component `ObsListToolbar`?

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
-        </Pressable>
-        <Pressable
-          onPress={( ) => setView( "grid" )}
-          testID="ObsList.toggleGridView"
+    <View className="bg-white border-b border-[#e8e8e8]">
+      <View className="py-5 flex flex-row px-[15px]">
+        <View className="flex justify-between items-center flex-row grow">
+          <View className="flex items-center flex-row">
+            {currentUser && (
+              <TouchableOpacity className="mr-3" accessibilityRole="button">
```
albullington, line 80: Can we add an `onPress` prop here to navigate the user to the Explore screen? It's currently a placeholder screen.

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
+
+  const cleanProgress = ( progressToClean: number ) => {
+    if ( Number.isNaN( progressToClean ) || progressToClean < 0 ) {
+      return 0;
+    }
+
+    return progressToClean;
+  };
+
+  const getStatusText = ( ) => {
```
albullington, line 54: When I upload an observation successfully, the status text changes to "UPLOAD 0 OBSERVATIONS". This text should disappear if there are no more observations to upload.

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
+    if ( !uploadInProgress && totalObsToUpload > 0 ) {
+      return t( "UPLOAD-X-OBSERVATIONS", { count: numUnuploadedObs } );
```
albullington, line 56: Let's use sentence case text instead of uppercase text here.

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
+          <View className="flex items-center flex-row">
+            {currentUser && (
+              <TouchableOpacity className="mr-3" accessibilityRole="button">
+                <IconMaterial name="language" size={30} />
+              </TouchableOpacity>
+            )}
+            <TouchableOpacity
+              onPress={getSyncClick( )}
+              accessibilityRole="button"
+              disabled={loading || uploadInProgress}
```
albullington, line 87: For now let's also disable the sync button when uploadInProgress is false and numToUpload is 0. Otherwise, it will create an error and a promise rejection when trying to make an API call.

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
-        <Pressable
-          onPress={( ) => setView( "list" )}
-          accessibilityRole="button"
-        >
-          <IconMaterial name="format-list-bulleted" size={30} />
-        </Pressable>
-        <Pressable
-          onPress={( ) => setView( "grid" )}
-          testID="ObsList.toggleGridView"
+    <View className="bg-white border-b border-[#e8e8e8]">
```
albullington, line 75: Is there a way to style everything here without having 4 nested Views? I find this a bit difficult to read.

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
+
+    return progressToClean;
+  };
+
+  const getStatusText = ( ) => {
+    if ( !uploadInProgress && totalObsToUpload > 0 ) {
+      return t( "UPLOAD-X-OBSERVATIONS", { count: numUnuploadedObs } );
+    }
+    if ( totalObsToUpload > 0 ) {
+      return t( "Uploading-X-Observations", { count: numUnuploadedObs } );
```
albullington, line 59: There are a few other text states mentioned in the issue, where there should be different text for smaller screen sizes. I believe the breakpoint we're looking at for the smallest screen size is 240px of width. Can we add those states in?

src/components/Observations/Toolbar.js
```diff
@@ -1,53 +1,139 @@
...
+                <IconMaterial name="language" size={30} />
+              </TouchableOpacity>
+            )}
+            <TouchableOpacity
+              onPress={getSyncClick( )}
+              accessibilityRole="button"
+              disabled={loading || uploadInProgress}
+              accessibilityState={{ disabled: loading || uploadInProgress }}
+            >
+              <IconMaterial name="sync" size={26} color={getSyncIconColor( )} />
```
albullington, line 90: Can we add the clockwise rotation animation when an upload is in progress? We'll swap out the icon later, but it would be nice to have the animation now.

albullington (changes requested):
Looking pretty good -- I left some comments to try to address some of the outstanding functionality described in issue #405.

Also noting that if we're going to skip the localization test because there's currently no Spanish text, it would be helpful to create a separate issue for that.

src/components/Observations/UserCard.js
```diff
@@ -1,41 +1,22 @@
...
   return (
-    <View className="flex-row px-5 items-center rounded-bl-3xl rounded-br-3xl bg-primary h-24">
-      {uri && <UserIcon uri={uri} />}
+    <View className="flex-row px-5 items-center h-24 bg-white">
       <View className="ml-3">
-        <Text className="color-white my-1">{User.userHandle( currentUser )}</Text>
-        <Text className="color-white my-1">
-          { t( "X-Observations", { count: currentUser.observations_count || 0 } )}
-        </Text>
+        <Text className="mt-1 text-3xl font-light">{t( "Welcome back," )}</Text>
```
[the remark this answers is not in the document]
kueda: Yup, that translation string will need to accept the username as a variable, though if the username needs to be styled differently as it seems to here, I'm not yet sure how we do that in RN. On the web we could just use CSS, or we might use `ReactDOMServer.renderToString()` to render the formatted username to a string and pass that as a variable to the translation, but I doubt there's an analog in RN. Any ideas, anyone? If not, we can punt on it in this PR and just render the username the same as we do "Welcome back".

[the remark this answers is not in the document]
jtklein: I did create a substring-different-styled text component in Seek: [SubstringStyledText](https://github.com/inaturalist/SeekReactNative/blob/adfddb027faffb79eeb4abe08745bea35850d3c2/components/SeekYearInReview/SeekYearInReviewScreen.js#L45)

It works well for the UI side, but if I remember correctly it breaks the possibility to find this entire component in the unit test by text. The result looks like this:
Screenshot 2023-01-31 at 13 28 45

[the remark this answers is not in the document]
Chrischuck: `react-i18next` can handle this, at least on web. Not sure about native though https://react.i18next.com/latest/trans-component#usage-with-simple-html-elements-like-less-than-br-greater-than-and-others-v10.4.0

I can give it a try

src/components/Observations/Toolbar.js
Chrischuck: I saw this as well, it seems that after update `const { allObsToUpload } = useLocalObservations( );` doesn't update. Not too sure why. Is it safe to just go off of `const numUnuploadedObs = useNumUnuploadedObservations( );`?

src/components/Observations/Toolbar.js
Chrischuck: Should this button still run the `syncObservations` button when there is nothing to upload?

src/components/styledComponents.js
```diff
@@ -64,5 +67,6 @@ export {
   ScrollView,
   Text,
   TextInput,
+  TouchableOpacity,
```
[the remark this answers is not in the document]
Chrischuck: Just going to use Pressable for now, if that's OK

src/components/Observations/UserCard.js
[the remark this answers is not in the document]
albullington: I sed an HTML renderer to handle this kind of thing in Seek, where the internationalized string would include HTML. Here's an example: [agree_to_terms](https://github.com/inaturalist/SeekReactNative/blob/adfddb027faffb79eeb4abe08745bea35850d3c2/translations/en.json#L125) is the string, and [here](https://github.com/inaturalist/SeekReactNative/blob/347397c30a073ea0b48ed1cf2bf6ffc19e8f70a1/components/Auth/Signup/LicensePhotosScreen.js#L171) is where it gets rendered in the code. In this case, I was using it to render links within a text string, but this can also be used to style specific html tags within a text string.

src/components/Observations/Toolbar.js
albullington: I'm unclear on this. @kueda might have a better understanding of how this is supposed to work. I think for now it's fine to disable this button when there's nothing to upload.

kueda: @Chrischuck, no, all it should do is fire the `startUpload()` function it received in props.

PR #414, opened 2023-01-30
Framework for adding custom icons to iNatRN

### LINKED ISSUE #387: Custom Icon Set
New UI designs have a lot of icons so we need a way to support them. I think I know how we should go about this, but I haven't actually done this before, so maybe view this as a suggested approach, but if you find a better one, make a case for it. Generally, we are relying pretty heavily on `react-native-paper`, which uses `react-native-vector-icons`. The latter supports custom icons when you give it a custom font, which is great because we already have a custom font of iNat-specific icons that we can expand on.

- [x] Export a version of the Explore icon from the designs as SVG and name the file `compass-rose.svg`
- [x] In [Fontastic,](https://app.fontastic.me/#select/jtHCj56TFzYnrjdgit8yeb) under the "INaturalist" icon set, choose "Add icons to this set" from the menu and choose the SVG file (note that fontastic should be smart about removing colors, BUT you may need to edit the SVG to get the spacing to work well; we can always revise that later)
- [x] Select your new icon, click the "Customize" tab and make sure, and make sure it's named `compass-rose`
- [x] On the "Publish" tab, click "DOWNLOAD" to get a zip including the font files and a CSS document that maps class names to characters in the font
- [x] Follow the `react-native-vector-icons` instructions for [creating a glyph file from a CSS file](https://github.com/oblador/react-native-vector-icons#generating-your-own-icon-set-from-a-css-file), and for [creating a custom glyphset / component](https://github.com/oblador/react-native-vector-icons#createiconsetglyphmap-fontfamily-fontfile) using the mappings you made from the fontastic CSS file (this is the part that I'm not sure will work)
- [x] Configure react-native-paper to [use our custom icon class](https://callstack.github.io/react-native-paper/icons.html)
- [x] Show how to use your icon in the UI Library

If that actually worked... you probably broke all the other icons by using a custom icon class that lacks all the defaults from `MaterialCommunityIcons`. So the todos continue:

- [ ] Try to add all the icons you can extract from the UI components to the font. You don't have to add color variants, but you will need to add solid / outline variants. You generally don't need to add a circle or button border if one surrounds an icon, with some exceptions like the `ObsEdit-SectionIndicator`. Try to name icons based on what they depict, not what they should mean in the UI (we'll name components semantically, but icons descriptively).
- [ ] Audit all existing icons and try to replace them with the named icons in our custom font

I imagine those two steps can happen in multiple iterations, e.g. add a few icons, replace existing icons with them, go back and add more icons, etc.

Also, **DO NOT change the names of the existing iNat icons**. We use those names on the web and probably elsewhere, so if you change the names something will break.

albullington:
Closes #387 

- Imports iNaturalist icons from Fontastic
- Configures `react-native-paper` to use iNat icon class instead of default icons
- Displays icons in the UI library using IconButton to display the custom INatIcon component
- Adds instructions to the README on how to add more custom icons
- Replaces a few keys icons in Footer and AddObsModal screens 

Note: I can't figure out how to get rid of the console.warning in jest which expects the icon to use MaterialCommunityIcons. I expect we'll either need to mock the PaperProvider or IconButton in `jest.setup.js`, or we'll need to wrap the renderComponent helper function with PaperProvider. Neither of those seemed to work for me when I tried, though.


README.md
```diff
@@ -123,6 +123,24 @@ git commit -a -m "Updated translations"
...
+
+1. Download custom icon from Figma as an SVG file.
+2. Add new icon to the iNaturalist icon set in Fontastic. Select all relevant iNaturalist icons, tap the Publish tab, and download the zip of icons.
+3. Create a glpyh file from the CSS file you just downloaded, using the following command (be sure to replace /path/to/styles with your path):
+
+```
+./node_modules/.bin/generate-icon '/path/to/styles.css' --componentName=INatIcon --fontFamily=inaturalisticons > 'src/components/INatIcon.js'
+```
+
+4. Go to `src/components/INatIcon.js` and change the name of the .ttf file to `inaturalisticons.ttf` to match the name of the .ttf file you downloaded
```
kueda, line 136: I can't tell if this is saying to edit the line in INatIcon.js to refer to a different font file, or to change the name of the font file to `inaturalisticons.ttf`. The latter seems like it would result in less churn.

index.js
```diff
@@ -85,22 +86,32 @@ const queryClient = new QueryClient( {
...
   <QueryClientProvider client={queryClient}>
     <RealmProvider>
       <SafeAreaProvider>
-        <PaperProvider theme={theme}>
+        <PaperProvider
+          settings={{
+            icon: renderCustomIcon
+          }}
+          theme={theme}
+        >
```
kueda, line 114: Ignore if you think this is out of scope, but perhaps we could move all this RN Paper setup into a custom `<InatPaperProvider>` component and file?

src/components/UiLibrary.js
```diff
@@ -14,94 +15,136 @@ import ViewWithFooter from "./SharedComponents/ViewWithFooter";
...
-  </ViewWithFooter>
-);
+        <Text className="text-lg">SecondaryCTAButton</Text>
+        <SecondaryCTAButton>
+          <Text>SecondaryCTAButton</Text>
+        </SecondaryCTAButton>
+        <SecondaryCTAButton disabled>
+          <Text>Disabled SecondaryCTAButton</Text>
+        </SecondaryCTAButton>
+        <Text className="text-lg">Custom iNaturalist Icons</Text>
```
kueda, line 101: This section is more about how to use IconButton, which is definitely useful, but maybe call it "IconButton w/ Custom iNaturalist Icons"?

Also, it would be cool to see all our icons here. If you export `glyphMap` from INatIcon.js, you could map over that to render an `<INatIcon />` for every icon.

src/components/UiLibrary.js
```diff
@@ -14,94 +15,136 @@ import ViewWithFooter from "./SharedComponents/ViewWithFooter";
...
+              )}
+              mode="contained"
+              containerColor={theme.colors.secondary}
+              iconColor={theme.colors.onSecondary}
+            />
+          </View>
+          <View>
+            <Text className="text-center">Warning</Text>
+            <IconButton
+              icon="tab-notifications"
```
kueda, line 131: This should be `notifications-bell`, right?

src/components/UiLibrary.js
```diff
@@ -14,94 +15,136 @@ import ViewWithFooter from "./SharedComponents/ViewWithFooter";
...
+              onPress={( ) => Alert.alert(
+                "",
+                "You tapped!"
+              )}
+            />
+          </View>
+          <View>
+            <Text className="text-center">Focused</Text>
+            <IconButton
+              icon="icon-createobservation"
```
kueda, line 117: This one doesn't seem to exist yet.

kueda (changes requested):
Looking good, and I especially appreciate the instructions in the README. Some requests and bugs:

- [ ] Let's not include the word `icon` in icon names, so `kebabmenu` instead of `icon-kebabmenu` (this is often easiest if you rename the file before uploading to fontastic)
- [ ] I'm getting a warning when viewing the UILibrary: Screen Shot 2023-01-30 at 6 15 36 PM
- [ ] The NavBar is cut off on the bottom in Android: Screenshot_20230130-181642
- [ ] Getting an error on Android when I try to take a picture: Screenshot_20230130-182246

src/components/UiLibrary.js
albullington: Noting that we'll need to tweak either the `./node_modules/.bin/generate-icon` script, the `createIconSet` script from `react-native-vector-icons` script, or both to be able to export `glyphMap` automatically, instead of adding this every time a new INatIcon.js component is generated. In the spirit of timeboxing, I'm inclined to make a separate issue for that, as well as renaming the `.ttf` file, which is the other manual change we need to make in this component.

kueda (approved):

PR #427, opened 2023-02-02
418 create an index.js file for importing shared components

jtklein:
This PR mainly does one thing: adds an index.js file to the folder “SharedComponents”. Then we can import the components listed in this file in a single line with named imports. As best exemplified in UiLib now.

As I was going through the components which ones would be good to include in this file as basic building blocks, I cleaned up some unused components, added some more accessibility according to the recently added eslint-plugins for this, and refactored and added some testing for some base components.


tests/unit/components/SharedComponents/InlineUser/InlineUser.test.js
```diff
@@ -22,16 +22,32 @@ jest.mock( "@react-navigation/native", ( ) => {
...
+  } );
+
   it( "displays user handle and image correctly", async ( ) => {
     render( <InlineUser user={mockUser} /> );
     // Check for user name text
     expect( screen.getByText( `@${mockUser.login}` ) ).toBeTruthy( );
     // This image appears after useIsConnected returns true
     // so we have to use await and findByTestId
-    const profilePicture = await screen.findByTestId( "UserIcon.photo" );
+    const profilePicture = await screen.findByTestId( "mockUserIcon" );
```
albullington, line 49: What's the reason for querying by testID throughout this component? It's still my understanding that the best way to test for accessibility is by looking for elements the way the user will see and interact with them on screen.

src/components/UiLibrary.js
```diff
@@ -182,22 +201,63 @@ const UiLibrary = ( ) => {
...
+          activeId="TAB1"
         />
-        <Body2 className="my-2">InlineUser component for a user that has no icon set</Body2>
-        <InlineUser user={{ login: "frogfinder23" }} />
 
         <Heading2 className="my-2">Quality Grade Status</Heading2>
         <View className="flex flex-row justify-between">
           <View>
             <Body2 className="text-center">Research</Body2>
+            { /* TODO: refactor to not have color prop because we only need black and white */}
```
albullington, line 260: It's also strange to me here that we're using colors that are not part of the tailwind theme. I would think every color we use, even black, should be accessed as a variable. But not really part of this PR.

albullington (approved):
This looks good to me. Left a couple questions/comments not really related to changes in this PR.

tests/unit/components/SharedComponents/InlineUser/InlineUser.test.js
jtklein: Yes, it was my understanding that when I want to have a container take all the accessibility props, I should not give the children any accessibility props (but I can look into that again). So, there was nothing else really available to querying for for the images, following https://testing-library.com/docs/queries/about/#priority until the end to byTestId

Issue #258, opened 2022-12-03
Selecting photos kinda slow


kueda:
In the Photo Gallery when you tap a photo to select it, the green check mark takes almost a second to appear. Why isn't that immediate? Sorry, kind of vague, this might be more an exploratory issue.

PR #428, opened 2023-02-02
#258 - use FlashList instead of FlatList

budowski:


jtklein (changes requested):
When I am trying to run `npm install` in your branch I get a warning that the package-lock file was created with a different npm version. Did you do `nvm use` before you added flash-list? Also, the complete re-write of package-lock.json would re-introduce the error we have with react-native-picker that we fixed by overriding the peerDependency in package-lock.json directly. Therefore, I could not test your branch on Android.

On, iOS, the branch did compile unto device. When using the gallery, in the recycler list of FlashList I did see some state persisting between different items, so after scrolling the old image was visible until the new image was loaded from CameraRoll.

kueda:
Superseded by https://github.com/inaturalist/iNaturalistReactNative/issues/431

PR #429, opened 2023-02-02
Remove warnings about custom icons / Material icons in jest

### LINKED ISSUE #421: Mock react-native-paper with custom icons in Jest
Now that we added custom icons into iNatRN and `react-native-paper`, we're getting a warning message in most of our jest tests:

`Warning: Failed prop type: Invalid prop name of value notifications-bell supplied to Icon, expected one of ["ab-testing","abacus"`

This warning indicates that jest does not know about the custom icons and expects `PaperProvider` to be using the default icons from Material Community Icons.

To fix, we'll either need to make some adjustments to the mock of `react-native-paper` in `jest.setup.js`, or maybe we can use the new `INatPaperProvider` component in tests, since this includes settings for the custom icon.

albullington:
Closes #421

Taking a combination approach here to get rid of all warnings:

For warnings about the wrong INatIcon name being supplied: 
- Replaced the names of Material Icons with placeholders that exist in our current custom icon set. We'll need to swap out these names as more custom icons are added. The two placeholders we're using are "camera" and "pencil"

For warnings about the wrong Material Community Icon name being supplied:
- Wrapped `renderComponent` function in `INatPaperProvider`
- Wrapped any components with footers that are being rendered in test suite without `renderComponent` function in `INatPaperProvider`
- Making sure we're not mocking wrapper components like `ScrollWithFooter` or `ViewNoFooter`, since this seems unnecessary with our `renderComponent` function and also meant that the footer showed icon-related warnings

### DIFF
--- a/tests/unit/components/ObsDetails/ObsDetails.test.js
+++ b/tests/unit/components/ObsDetails/ObsDetails.test.js
@@ -73,15 +73,6 @@ jest.mock( "components/ObsDetails/DataTab", () => ( {
   default: () => mockDataTab
 } ) );
 
-jest.mock(
-  "components/SharedComponents/ScrollWithFooter",
-  () => function MockContainer( props ) {
-    const MockName = "mock-scroll-with-footer";
-    // eslint-disable-next-line react/jsx-props-no-spreading
-    return <MockName {...props}>{props.children}</MockName>;
-  }
-);
-
 jest.mock( "sharedHooks/useIsConnected" );
 
 const mockLatLng = factory( "DeviceLocation" );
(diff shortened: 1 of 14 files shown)


tests/unit/components/ObsDetails/ObsDetails.test.js
```diff
@@ -73,15 +73,6 @@ jest.mock( "components/ObsDetails/DataTab", () => ( {
   default: () => mockDataTab
 } ) );
 
-jest.mock(
```
jtklein, line 76: I have added those mocks to make the unit tests of these components more granular, that they don't care what happens in the container and only in the component we want to test.

I am not sure how these mocks can be the source of the footer showing icon-related warnings. I put them in place extra to not have the footer in the element tree. Curious, how did it help you to remove those mocks?

albullington: Whoops good catch -- just me going too quickly trying to eliminate causes. I added these back and will merge.

PR #430, opened 2023-02-03
Update copy for observation uploads

Chrischuck:
- [x] when uploadInProgress is true, text should read "Uploading {$numUploaded} of {$numToUpload} observations"
- [x] when uploadInProgress is true and the device is the smallest size, text should read "{$numUploaded} of {$numToUpload}"
- [x] Explore button (compass rose icon): fires navigateToExplore when tapped, only visible when signed in

Screenshot 2023-02-02 at 7 07 38 PM


src/i18n/l10n/en.ftl
```diff
@@ -576,6 +576,18 @@ Uploading-X-Observations = Uploading {$count ->
...
 }
 
+# Shows the number of observations a user is currently uploading out of total on my observations page
+Uploading-x-of-y-observations = Uploading {$uploadedCount} of {$total ->
+    [one] 1 observation
+    *[other] {$total} observations
+}
+
+# Shows the progress of uploads for X of Y
+Uploading-x-of-y = Uploading {$uploadedCount} of {$total ->
```
jtklein, line 586: You don't need to handle different cases if both counts are displayed as integers.

src/components/Observations/ObsListToolbar.js
```diff
@@ -80,15 +97,17 @@ const Toolbar = ( { setLayout, layout }: Props ): Node => {
...
-          <Pressable
-            className="mr-3"
-            accessibilityRole="button"
+          <IconButton
+            icon="compass-rose"
             onPress={( ) => navigation.navigate( "MainStack", { screen: "Explore" } )}
-          >
-            <IconMaterial name="language" size={30} />
-          </Pressable>
+            accessibilityLabel={t( "Explore" )}
```
jtklein, line 105: I have changed the accessibility label here into a hint. What we used to write as labels, the longer half sentence like strings, are actually more to be used as hints according to the iOS documentation at least. The labels are supposed to be succinct, one or two words. We need to refactor those into hints everywhere we used them.
I added some more explanation in the string.ftl file: [704](https://github.com/Chrischuck/iNaturalistReactNative/blob/b0b25e4dd4b0cbd9277e0d47703f894f14e37fb5/src/i18n/strings.ftl#L704-L705)

jtklein (changes requested):
Only minor change.

src/i18n/l10n/en.ftl
Chrischuck: good catch, will fix!

src/components/Observations/ObsListToolbar.js
Chrischuck: Sounds good! will make that a note for the future

PR #442, opened 2023-02-04
398 inline user component

### LINKED ISSUE #398: InlineUser component

jtklein:
When finished this closes #398 .
Currently, we are still missing the icons to be used as fallback images for when a user has no icon set and when the device is offline. I can finish this up with the icons later.

### DIFF
--- a/src/components/SharedComponents/InlineUser/InlineUser.js
+++ b/src/components/SharedComponents/InlineUser/InlineUser.js
@@ -1,7 +1,9 @@
 // @flow
 
 import { useNavigation } from "@react-navigation/native";
+// Directly imported, not from index.js to avoid circular dependency
 import Body3 from "components/SharedComponents/Typography/Body3";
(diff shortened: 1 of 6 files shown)


jtklein:
Everything here is ready with the exception of the fallback icons for "no internet" and "no user icon" state. Those are currently not in the design specs yet.

Issue #434, opened 2023-02-03
ObsStatus component


albullington:
<img width="516" alt="Screen Shot 2023-02-03 at 10 32 03 AM" src="https://user-images.githubusercontent.com/35536439/216680239-859964ad-fcf4-411c-8fc6-c7d80c4a0693.png">
<img width="394" alt="Screen Shot 2023-02-03 at 10 32 27 AM" src="https://user-images.githubusercontent.com/35536439/216680296-9ee73aae-dc80-4930-9964-56a3b5b1845d.png">

This component should have:
- [ ]  ActivityCount component for ids
- [x]  ActivityCount component for comments
- [x]  QualityGradeStatus component

Default color for these components should be:
- [x] #454545 (darkGray)

This component should accept a `layout` prop. If the layout is "list", the subcomponents should display in a column & use the default subcomponent colors.

If the layout is "grid", subcomponents should display in a row and the color of subcomponents should be white.

In each layout, subcomponents should have a margin of 10px between them.

Chrischuck:
I can work on this. Not sure how much different it is from the current component?

I will put up a PR that can be discussed.

PR #443, opened 2023-02-05
434 obs status component

Chrischuck:
#434 Not too sure what really changed based off the spec, other than colors + margins. I did see a chance to simplify up some logic though.
Screenshot 2023-02-05 at 6 23 05 PM

Screenshot 2023-02-05 at 6 23 24 PM


src/components/Observations/ObsCardStats.js
```diff
@@ -1,74 +1,72 @@
...
+        />
+        */
+      }
+      <View className={classnames( margin, "flex flex-row items-center" )}>
+        <Image
+          style={{ height: 15, width: 15, tintColor: iconColor }}
+          source={require( "images/ic_id.png" )}
+        />
+        <Body3 className="ml-1" style={{ color: iconColor }}>{item.identifications.length}</Body3>
       </View>
```
kueda, line 59: Why not use <ActivityCount /> here? If it's just lack of an icon, maybe use `cv-sparklylabel` or just the speech bubble again.

src/components/Observations/ObsCardStats.js
```diff
@@ -1,74 +1,72 @@
...
-      {renderQualityGrade( )}
+      <ActivityCount
+        marginClass={margin}
+        count={item.comments?.length}
+        color={iconColor}
+        accessibilityLabel={t( "Number-of-comments" )}
```
kueda, line 64: Nothing needs to change here right now, but I think we need to change the way accessibility labels for for ActivityCount (I've taken a stab at it in #447). I *think* `accessibilityValue` is mostly for interactive controls and that `accessibilityLabel` should be used for informational elements like this instead, and that `accessibilityLabel` shoudl convey the same thing that's conveyed visually, so something like "12 comments" instead of "Number of comments".

src/components/Observations/ObsCardStats.js
```diff
@@ -1,74 +1,72 @@
...
-    </View>
-  );
-
-  const renderRow = ( ) => (
-    <View className="flex-row absolute bottom-1 justify-between w-full px-2">
-      <View className="flex-row">
-        {renderIdRow( )}
-        {renderCommentRow( )}
+  const iconColor = getIconColor( );
+  const margin = layout === "list" ? "mb-2.5" : "mr-2.5";
```
kueda, line 33: Let's change the margin to `1`. I know that might not be what the issue said, but it looks closer to spec to me with `1`.

src/components/SharedComponents/ActivityCount/ActivityCount.js
```diff
@@ -13,18 +14,19 @@ type Props = {
...
   icon?: string,
   accessibilityLabel: string,
-  testID?: string
+  testID?: string,
+  marginClass?: string
 }
 
 const ActivityCount = ( {
-  count, color, icon, accessibilityLabel, testID
+  count, color, icon, accessibilityLabel, testID, marginClass = ""
```
kueda, line 22: Any reason to use a blank string over null here?

src/components/Observations/ObsCardStats.js
```diff
@@ -1,74 +1,72 @@
...
-import { View } from "components/styledComponents";
+import { ActivityCount, Body3, QualityGradeStatus } from "components/SharedComponents";
+import { Image, View } from "components/styledComponents";
 import type { Node } from "react";
 import React from "react";
 import { useTranslation } from "react-i18next";
 import { useTheme } from "react-native-paper";
 
 type Props = {
   item: Object,
```
kueda, line 13: I know this is a bit out of scope, but let's change this prop to `observation` since `item` is always an observation here.

kueda (changes requested):
Couple inline requests, plus these:

- [x] Rename `ObsCardStats` component to `ObsStatus`
- [ ] Add to the UI Library showing at least the two `layout` options

src/components/SharedComponents/ActivityCount/ActivityCount.js
Chrischuck: No reason other than keeping the types the same. `classnames` ignores falsey values.

src/components/Observations/ObsCardStats.js
Chrischuck: I would have preferred as well, just did not know what place holder icon to use and wanted to remain in spec w/ the ui.

src/components/Observations/ObsStatus.js
```diff
@@ -0,0 +1,61 @@
...
+} from "components/SharedComponents";
+import { View } from "components/styledComponents";
+import type { Node } from "react";
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { useTheme } from "react-native-paper";
+import Observation from "realmModels/Observation";
+
+type Props = {
+  item: typeof Observation,
```
Chrischuck, line 17: Is this correct @kueda ? Couldn't find any usage of specific types in the repo

src/components/SharedComponents/ActivityCount/ActivityCount.js
kueda: Fair enough.

src/components/Observations/ObsStatus.js
kueda: That's great if flow doesn't complain about it. However, I wanted to change the *name* of the prop to `observation` so it's clear to developers what it is, not just IDEs and linters.

src/components/Observations/ObsStatus.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const getIconColor = () => {
+    if ( layout === "horizontal" ) {
+      return theme.colors.onSecondary;
+    }
+    if ( item.viewed === false ) {
+      return theme.colors.error;
+    }
+    return theme.colors.primary;
+  };
```
kueda, line 34: Sorry I didn't catch this earlier, but let's specify color in a prop and not in the component. Again, this component doesn't need to know if it's a list of a grid view, just how it needs to lay itself out.

kueda (changes requested):
One more change, plus there are a few conflicts to resolve (mea culpa, I merged in that change to ActivityCount).

src/components/Observations/ObsStatus.js
Chrischuck: Ohhh, I see. Will do

Issue #435, opened 2023-02-03
TaxonName component


albullington:
<img width="309" alt="Screen Shot 2023-02-03 at 10 39 34 AM" src="https://user-images.githubusercontent.com/35536439/216681572-4869b76b-5df4-4670-963e-5862adaa4e21.png">
<img width="156" alt="Screen Shot 2023-02-03 at 10 49 03 AM" src="https://user-images.githubusercontent.com/35536439/216683245-9b781be5-a6a8-4504-84a4-2c348423f461.png">
<img width="515" alt="Screen Shot 2023-02-03 at 10 49 26 AM" src="https://user-images.githubusercontent.com/35536439/216683306-c3b62492-2566-4541-b35e-ccd5210da2fb.png">

This component should build off of the existing [DisplayTaxonName](https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/components/DisplayTaxonName.js) component. It should display:

- [ ] Optional Common Name (Body1 sizing) - default to scientific name if there is no common name available
- [ ] Optional Scientific Name (Body3 sizing, italic)

Default color should be:
- [ ] #454545 (darkGray)

- [ ] Since names can be long, all text should be able to wrap onto multiple lines for components like ObsListItem
- [ ] We'll also need a prop to limit Common Name to a max of 3 lines for ObsGridItem

See comments below for more implementation details.

kueda:
I'd prefer to a) build off of https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/components/DisplayTaxonName.js, and b) restrict the scope here to deal only with the taxon name. Name display is complicated and we don't want to repeat all the logic involved in deciding what name to show first, whether to show a common name, what common name to show, etc. The additional details only pertain to an observation and pretty much only occur in the ObsListItem so I'm not sure they need to be their own components.

As to *how* we can build off of `<DisplayTaxonName>`, I guess that's up to the developer. One option would be to add a prop to that component so renders two text components. Another might be to abstract some of the name choice logic out and make two components, one for the name in a single line and another for a two-line display.

albullington:
Ok, updated these issues to move date/location/confidence interval into ObsListItem to-dos and to clarify that this should build off the DisplayTaxonName component.

Chrischuck:
I can get something going on this.

PR #444, opened 2023-02-06
435 - Create taxon name component

Chrischuck:
#435 

list view common name first
Screenshot 2023-02-05 at 7 46 46 PM

list view sci name first
Screenshot 2023-02-05 at 7 46 29 PM

grid view sci name first
Screenshot 2023-02-05 at 7 46 22 PM

grid view common name first
Screenshot 2023-02-05 at 7 46 54 PM

### DIFF
--- a/src/components/DisplayTaxonName.js
+++ b/src/components/DisplayTaxonName.js
@@ -14,10 +14,10 @@ type Props = {
 };
 
 const styles = StyleSheet.create( {
-  maxWidth: {
+  commonName: {
     maxWidth: "100%"
   },
-  italic: {
+  scientificName: {
     fontStyle: "italic"
   }
 } );
@@ -39,22 +39,23 @@ const DisplayTaxonName = ( {
 
   return (
     <View
-      className={classnames( "flex flex-wrap", {
-        "flex-row-reverse": isListView && scientificNameFirst,
-        "justify-end": isListView && scientificNameFirst,
-        "flex-row items-end": isListView,
+      testID="display-taxon-name"
+      className={classnames( "flex", {
+        "flex-row-reverse flex-wrap-reverse justify-end items-start":
+          scientificNameFirst && isListView,
+        "flex-row items-end flex-wrap": isListView && !scientificNameFirst,
         "w-11/12": isListView,
         "flex-col-reverse": !isListView && scientificNameFirst
       } )}
     >
       {taxonData.commonName && (
-        <Body1 numberOfLines={3} style={styles.maxWidth}>
+        <Body1 numberOfLines={3} style={styles.commonName}>
           {`${taxonData.commonName}${
             !scientificNameFirst && isListView ? " " : ""
           }`}
         </Body1>
       )}
-      <Body3 style={styles.italic}>
+      <Body3 style={styles.scientificName}>
         {`${taxonData.scientificName}${
           scientificNameFirst && isListView ? " " : ""
         }`}
--- a/src/components/DisplayTaxonName.js
+++ b/src/components/DisplayTaxonName.js
@@ -9,11 +9,11 @@ import { generateTaxonPieces } from "sharedHelpers/taxon";
 
 type Props = {
   item: Object,
-  layout?: string,
+  layout?: "horizontal" | "vertical",
 };
 
 const DisplayTaxonName = ( {
-  layout = "list",
+  layout = "horizontal",
   item: { user, taxon }
 }: Props ): Node => {
   const { t } = useTranslation( );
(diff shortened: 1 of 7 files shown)


tests/unit/components/DisplayTaxonName.test.js
```diff
@@ -36,106 +36,109 @@ const uncapitalizedTaxon = factory( "LocalTaxon", {
...
     render(
-      <DisplayTaxonName item={{ taxon: subspeciesTaxon, user }} />
+      <DisplayTaxonName layout="grid" item={{ taxon: subspeciesTaxon, user }} />
     );
 
-    expect(
-      screen.getByText( "Lupinus albifrons var. collinus (Silver Lupine)" )
-    ).toBeTruthy();
+    expect( screen.getByTestId( "display-taxon-name" ) ).toHaveTextContent(
+      "Lupinus albifrons var. collinusSilver Lupine"
```
Chrischuck, line 141: Struggling to fix these tests. Not really sure how to get matchers to "understand" css.

Even though in actuality it is "Lupinus albifrons var. collinus Silver Lupine" on screen, it thinks it's "Silver Lupine Lupinus albifrons var. collinus".

Chrischuck: For future reference, It's better to change the document flow because screen readers won't pick up on the visual changes.

src/components/DisplayTaxonName.js
```diff
@@ -1,159 +1,67 @@
...
-
-const capitalizeCommonName = name => {
-  if ( !name ) {
-    return name;
+const styles = StyleSheet.create( {
+  commonName: {
+    maxWidth: "100%"
+  },
+  scientificName: {
+    fontStyle: "italic"
```
Chrischuck, line 21: italic fonts need to be added

tests/unit/components/DisplayTaxonName.test.js
```diff
@@ -36,106 +36,110 @@ const uncapitalizedTaxon = factory( "LocalTaxon", {
...
+      `${highRankTaxon.rank} ${highRankTaxon.name}`
+    );
   } );
 
-  test( "renders correct taxon for a subspecies", () => {
-    render(
-      <DisplayTaxonName item={{ taxon: highRankTaxon, user }} />
-    );
+  test( "renders correct taxon for a subspecies", ( ) => {
+    render( <DisplayTaxonName item={{ taxon: highRankTaxon, user }} /> );
```
kueda, line 67: `highRankTaxon` has `rank_level: 27` in your test, which makes it a subfamily, not a subspecies, so you should change the name of this test.

src/components/DisplayTaxonName.js
```diff
@@ -1,159 +1,57 @@
...
+import { Body1, Body3 } from "components/SharedComponents";
+import { View } from "components/styledComponents";
 import type { Node } from "react";
 import React from "react";
+import { useTranslation } from "react-i18next";
+import { generateTaxonPieces } from "sharedHelpers/taxon";
 
 type Props = {
   item: Object,
+  layout?: string,
```
kueda, line 12: If I'm understanding correctly, this can have the value of `list` when the name should be displayed as a single line as it is in the observation list... but this component will be used in a lot of places, not just in list vs grid scenarios, so maybe what we need here is `layout?: "horizontal" | "vertical"`

kueda (changes requested):
A few inline requests, but the larger issue is that I think there was probably some slight confusion in the issue: the first name should always have the same typography regardless of whether it's a common or scientific name, and same with the second. If there's only one name, it should be styled as the first name. Here are some examples from the web where the styling is different but the overall approach of one style for the first and another for the second is the same:

Common name first:
Screen Shot 2023-02-07 at 12 25 16 PM

Scientific name first:
Screen Shot 2023-02-07 at 12 25 36 PM

Scientific name only:
Screen Shot 2023-02-07 at 12 25 57 PM

Also, scientific names at genus and below are *always* italic, except for an interstitial rank like "var.", so like "*Vulpes vulpes* ssp. *vulpes*" for a subspecies but "Juncaceae" for a family. That's regardless of where they appear in this display. It's a bit maddening.

Chrischuck:
@kueda I appreciate the breakdown, did not know any of those rules prior. Since we don't have fonts added in italics, they still don't show up. However the logic to display italics should work for when they are added.

I tried to add as many examples in the images below i.e. no common name, ssp./var., etc.

common name first - horizontal
Screenshot 2023-02-07 at 1 47 39 PM
Screenshot 2023-02-07 at 1 47 59 PM
common name first - vertical

Screenshot 2023-02-07 at 1 49 32 PM

Sci name first - horizontal
Screenshot 2023-02-07 at 1 48 19 PM

Sci name first - vertical
Screenshot 2023-02-07 at 1 49 01 PM

Screenshot 2023-02-07 at 1 49 32 PM

src/components/DisplayTaxonName.js
kueda: https://github.com/inaturalist/iNaturalistReactNative/issues/449

kueda (approved):

Issue #437, opened 2023-02-03
Date component


albullington:
This is a tiny component that is reused in a number of places:

- [ ] Clock icon (size 15) with 5px margin to the right
- [ ] Date string in 2/2/21 6:42 PM format (Body4 text size)
- [ ] Default color: #454545 (darkGray)

Chrischuck:
I think I can get this done

PR #445, opened 2023-02-07
437 date display component

Chrischuck:
#437


package.json
```diff
@@ -7,7 +7,7 @@
     "ios": "react-native run-ios",
     "start": "react-native start",
     "clean-start": "npx react-native clean-project-auto && npx pod-install && npm start",
-    "test": "jest",
+    "test": "TZ=UTC jest",
```
kueda, line 10: Cool, didn't know you could do this.

src/components/SharedComponents/DateDisplay.js
```diff
@@ -0,0 +1,35 @@
...
+  dateTime
+}: Props ): React.Node => {
+  const { t } = useTranslation( );
+  const displayTime = ( ) => {
+    if ( dateTime ) {
+      return formatObsListTime( dateTime );
+    }
+    return t( "no time given" );
+  };
+  console.log( displayTime() );
```
kueda, line 24: Please remove.

kueda (changes requested):
Looking good, only minor changes needed. Date formatting needs to be handled a little differently but that wasn't in the issue so I punted that to #448. FWIW, I'm also getting some `Warning: An update to DateDisplay inside a test was not wrapped in act(...).` errors when running tests, though I'm not 100% sure why. Would be nice to fix that, but I don't think it needs to block.

Chrischuck:
@kueda Looks like it has to do with the `useTranslation` hook. After removing it and running tests it seems to disappear. I can take a look into it as another issue.

kueda (approved):

Issue #438, opened 2023-02-03
Location component


albullington:
This is a tiny component that is reused in a number of places:

- [ ] Location pin icon (13px) with 5px margin to the right
- [ ] Place name from `useLocationName` hook (Body4 text size)
- [ ] Default color: #454545 (darkGray)

Chrischuck:
I can also do this

PR #446, opened 2023-02-07
438 location display component

Chrischuck:
#438


src/components/SharedComponents/LocationDisplay.js
```diff
@@ -0,0 +1,32 @@
...
+  longitude
+}: Props ): React.Node => {
+  const { t } = useTranslation( );
+  const locationName = useLocationName( latitude, longitude );
+
+  return (
+    <View className="flex flex-row items-center">
+      <IconMaterial name="location-pin" size={15} />
+      <Body4 className="text-darkGray ml-[5px]">
+        {locationName || t( "no place guess" )}
```
kueda, line 26: Please change "no place guess" to "Missing Location" and [add it to the translations file](https://github.com/inaturalist/iNaturalistReactNative#adding-new-text) (FYI, the capitalization of the key should match the text if possible, so here it would be `Missing-Location`).

src/components/SharedComponents/LocationDisplay.js
```diff
@@ -0,0 +1,32 @@
...
+  latitude: number,
+  longitude: number
+};
+
+const LocationDisplay = ( {
+  latitude,
+  longitude
+}: Props ): React.Node => {
+  const { t } = useTranslation( );
+  const locationName = useLocationName( latitude, longitude );
```
kueda, line 20: Sorry, I'm going to change the requirements here: this is for displaying observations, potentially a lot of them, so we don't want to potentially reverse geocode coordinates 100s of times if you're perusing your observations. Instead, this component should take an `observation` as a prop and do something like this pseudocode:

```
if placeGuess is not blank:
  show placeGuess
else if latitude and longitude are not blank
  show `${latitude}, ${longitude}`
else
  show "Missing Location"
```

Let me know if that's confusing at all.

src/components/SharedComponents/LocationDisplay.js
```diff
@@ -0,0 +1,32 @@
...
+    <View className="flex flex-row items-center">
+      <IconMaterial name="location-pin" size={15} />
+      <Body4 className="text-darkGray ml-[5px]">
+        {locationName || t( "no place guess" )}
+      </Body4>
+    </View>
+  );
+};
+
+export default LocationDisplay;
```
kueda, line 32: Since this does need to be specific to showing the location of an observation, let's rename it to `ObservationLocation` or `ObsLocationDisplay` or something that makes it clear this is just for observations.

kueda (changes requested):
Sorry for the switcheroo, but #438 needed a little re-work (I was busy with other stuff last week so I didn't have a chance to take a look at it before you got to it).

src/components/SharedComponents/LocationDisplay.js
Chrischuck: Makes sense to me.

will a missing longitude/latitude be `null`? Thinking about the one case where an observation is 0,0 (if there's even been one at that location)

Chrischuck:
@kueda No worries, simple fixes anyway.

src/components/SharedComponents/LocationDisplay.js
kueda: In theory a missing coordinate should be null, and 0,0 should be interpreted as literally at 0,0 (the only obs marooned on [Null Island](https://en.wikipedia.org/wiki/Null_Island) are due to bugs, but they do exist and should be treated as if they were really there).

kueda (approved):

Issue #436, opened 2023-02-03
ObsGridItem


albullington:
<img width="154" alt="Screen Shot 2023-02-03 at 10 59 12 AM" src="https://user-images.githubusercontent.com/35536439/216685089-bb1a9a73-bbb0-4248-abdc-d3ac9331b863.png">
<img width="281" alt="Screen Shot 2023-02-03 at 11 02 41 AM" src="https://user-images.githubusercontent.com/35536439/216685716-e3bb6c41-3efd-4a27-8488-79ba8efd51fe.png">

This component should display:
- [ ] Taxon photo as the background, at a width/height of 172px and border radius of 15px

Overlaying the taxon photo, this component should display:
- [ ] Top Left: sound icon if the observation has an associated sound
- [ ] Top Right: icon displaying the number of photos in the observation
- [ ] Bottom Left: either the ObsStatus component (#434) or the UploadIcon component (#433) if observation needs to be uploaded to the server
- [ ] Bottom Left: TaxonName component (#435), with color set to white, only the Common Name showing, a maximum of three lines of text, and a maximum text width of 145px. There should be 5px of space above this component (between it and UploadIcon or ObsStatus)

The color for all elements overlaying the taxon photo should be:
- [ ] #FFFFFF (white)

Chrischuck:
i can take a stab at this

Chrischuck:
Is there an example w/o a taxon image?

kueda:
@Chrischuck, this component should never show a taxon image, but it should show an observation photo. Without a photo it should just show a gradient like this:

```css
background: linear-gradient(180deg, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.5) 100%), #454545;
background-blend-mode: multiply, normal;
border-radius: 15px;
```

PR #450, opened 2023-02-08
436 obs grid item

Chrischuck:
#436 
Did not add spacing or margins to the list yet. Just built to the issue spec. I do think 172x172 is little small
Screenshot 2023-02-08 at 7 00 59 AM

### DIFF
--- a/src/components/DisplayTaxonName.js
+++ b/src/components/DisplayTaxonName.js
@@ -8,18 +8,26 @@ import { useTranslation } from "react-i18next";
 import { generateTaxonPieces } from "sharedHelpers/taxon";
 
 type Props = {
-  item: Object,
+  observation: Object,
   layout?: "horizontal" | "vertical",
+  color?: string
 };
 
 const DisplayTaxonName = ( {
   layout = "horizontal",
-  item: { user, taxon }
+  observation: { user, taxon },
+  color
 }: Props ): Node => {
   const { t } = useTranslation( );
 
+  const textColorClass = color || "text-darkGray";
+
   if ( !taxon ) {
-    return <Body1 numberOfLines={1}>{t( "unknown" )}</Body1>;
+    return (
+      <Body1 className={textColorClass} numberOfLines={1}>
+        {t( "unknown" )}
+      </Body1>
+    );
   }
 
   const {
@@ -64,6 +72,7 @@ const DisplayTaxonName = ( {
       } )}
     >
       <Body1
+        className={textColorClass}
         numberOfLines={scientificNameFirst ? 1 : 3}
       >
         {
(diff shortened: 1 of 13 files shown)


src/components/DisplayTaxonName.js
```diff
@@ -8,18 +8,26 @@ import { useTranslation } from "react-i18next";
 import { generateTaxonPieces } from "sharedHelpers/taxon";
 
 type Props = {
-  item: Object,
+  observation: Object,
```
kueda, line 11: Hah, this is like the opposite of my previous feedback: `DisplayTaxonName` is something we're going to use all over the place, not just in the context of an observation, so it would be more appropriate for it to accept a `taxon` prop and a `user` prop, since those will always be relevant to displaying a taxon name.

src/components/Observations/GridItem.js
```diff
@@ -1,118 +1,114 @@
...
-  index?: number,
-  item: Object,
+  observation: Object,
   handlePress: Function,
-  // Number of columns in the grid; we need this to set the margins correctly
-  numColumns?: number,
   uri?: string,
 };
 
 const GridItem = ( {
```
kueda, line 24: Let's rename this component and file to `ObsGridItem` since it is really only for rendering observations in a grid.

src/components/Observations/GridItem.js
```diff
@@ -1,118 +1,114 @@
...
+          </ImageBackground>
         ) : (
-          <View className="grow aspect-square justify-center items-center">
-            <IconMaterial name="image-not-supported" size={150} />
-          </View>
+          <LinearGradient
+            colors={["rgba(0, 0, 0, 0)", "rgba(0, 0, 0, 0.5) 100%)"]}
+          >
+            <View className="grow aspect-square justify-center items-center">
+              <IconMaterial name="image-not-supported" size={150} />
```
kueda, line 80: If the observation has no photos, don't show this `image-not-supported` icon. We just want that gray gradient.

kueda (changes requested):
Looking good. I'm about to step out the door but here are a few things I've noticed so far. Will try and take a more thorough look when I have time, maybe later today or tomorrow.

src/components/DisplayTaxonName.js
Chrischuck: No worries lol, small change. Will fix that.

Chrischuck:
Screenshot 2023-02-08 at 11 51 38 AM
@kueda made those dimension changes. A little oblong, but margins will fix that eventually...

updated `DisplayTaxonName` as well.

kueda (approved):

PR #451, opened 2023-02-08
Add italic Whitney font

### LINKED ISSUE #449: Support italic in Whitney
The `italic` class doesn't currently work in Tailwind, and we need it to properly render scientific names. I'm not entirely sure what the correct approach is, but maybe we need to follow the [nativewind advice](https://www.nativewind.dev/tailwind/typography/font-family) and model fonts usage off of https://github.com/jsamr/react-native-font-demo.

albullington:
Closes #449

Worth noting that commit [#44f641](https://github.com/inaturalist/iNaturalistReactNative/commit/44f641009704a7ddf4d3df76ad862fb96d3015c9) made it so that we can no longer add fonts with `npx react-native-asset` and instead have to add them the manual way.

I didn't make a custom typography component for italic font & instead updated the className directly in `DisplayTaxonName`. If we think we'll reuse italic font outside of this component, it might be worth having a separate typography component.


kueda:
I'm not committed to the gradle approach to keeping the font icon correctly linked and updated in Android, it just seemed like the easiest solution to that problem and I didn't know about `react-native-asset`. Is there a way to achieve the same result with `react-native-asset`? I probably read this line in the README and assumed there wasn't: "you shouldn't need to do this unless you're adding a .ttf file with a different name".

FWIW, we *do* need to support italics in UserText. Not sure how we're going to do that yet though.

albullington:
I'm not clear on what problem was happening with fonts and Android linking. Using `npx react-native-asset` was working fine for me for keeping our custom icon font linked on both iOS and Android (as long as I remembered to clean the build folder), but I'm not sure if that was the problem?

albullington:
It looks like we can use `className="italic"` on iOS and Android as long as there is an italic font linked from the assets folder.

kueda (approved):
It works! We agreed to look into going back to `react-native-asset` separately.

PR #458, opened 2023-02-10
415 snapshot testing for UI buttons

### LINKED ISSUE #415: Snapshot testing for UI buttons
We added redesigned UI buttons in PR #388, but we didn't add any testing. Ideally, all redesigned UI components will have test coverage.

I looked into unit tests, but I couldn't find a way to access `className`, the prop used for styling with Tailwind CSS, via jest. Instead, @jtklein suggested we should write snapshot tests for the buttons. That way we can see if any of the styling changes unintentionally in the future.

These will be our first snapshot tests in iNatRN.

jtklein:
I actually misread the issue as add "snapshots for base components" but it was only about UI buttons I just realized.
So, this PR adds snapshot testing for Typography, Buttons, Tabs, Divider, and INatIcon.

For all components except INatIcon, I also updated or created the UI to be like the design specs. 

Closes #415


jtklein:
Tests are failing because of translation issues with accessibility props. This should be addressed by https://github.com/inaturalist/iNaturalistReactNative/tree/448-handle-timeless-dates-i18n-init

tests/unit/components/SharedComponents/Buttons/Button.test.js
```diff
@@ -0,0 +1,121 @@
...
+
+    it( "has no accessibility errors", async () => {
+      const button = (
+        <Button level="warning" text="WARNING DISABLED" disabled />
+      );
+      // Snapshot test
+      expect( button ).toBeAccessible();
+    } );
+  } );
+} );
```
kueda, line 121: Lots of boilerplate here. Any disadvantage to dynamically making the tests? I was thinking something like:

```javascript
["primary", "netural"].forEach( level => {
  describe( level, ( ) => {
    it( "renders correctly", async () => {
      render( <Button level={level} text={`${level.toUpperCase( )} BUTTON`} /> );
      expect( screen ).toMatchSnapshot();
    } );

    it( "has no accessibility errors", async () => {
      const button = <Button level={level} text={`${level.toUpperCase( )} BUTTON`} />;
      expect( button ).toBeAccessible();
    } );
  } );
} );
```

Maybe a bit less readable, but less to maintain.

kueda (approved):
Looks good to me. Inline comments are mostly just questions and suggestions, nothing that needs to block merge.

tests/unit/components/INatIcon.test.js
```diff
@@ -0,0 +1,28 @@
+import { render, screen } from "@testing-library/react-native";
+import INatIcon, { glyphMap } from "components/INatIcon";
+import React from "react";
+
+const iconName = "camera";
+
+describe( "INatIcon", () => {
+  it( "renders correctly", async () => {
```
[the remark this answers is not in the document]
jtklein: Correct.

tests/unit/components/SharedComponents/Buttons/Button.test.js
jtklein: Have added a loop with `describe.each`.

PR #463, opened 2023-02-10
Add photo gallery icon

### LINKED ISSUE #461: Add gallery icon to AddEvidenceModal
Currently it's referring to an icon that doesn't exist.

- [ ] Add the icon to the font
- [ ] Use in the `AddEvidenceModal`

Here's the icon:
gallery

albullington:
Closes #461 

- noting that we will eventually need to replace this icon with an SVG that shows detail instead of a filled in rectangle


kueda:
I updated the font in fontastic. Could you try downloading and linking up the font again?

For my own reference, an SVG should have no paths with the `fill-rule:evenodd` style. If there are holes being achieved in that way, they need to be replaced by setting the `fill-rule` to `nonzero` and subtracting the relevant shape.

Issue #423, opened 2023-02-01
Display internationalized strings with variables that have different HTML formatting


albullington:
One thing that came up in comments in PR #411 is the need to be able to use translated strings with variables that have different types of styling. This is specifically to address strings like "Welcome back, @username" (displayed below), though it's likely to come up in other places throughout the app.

![](https://user-images.githubusercontent.com/13311268/215300884-afa723a6-1229-4f99-b3fa-37681b22732e.png)

A few ideas we discussed:

1. Creating a [SubstringStyledText](https://github.com/inaturalist/SeekReactNative/blob/adfddb027faffb79eeb4abe08745bea35850d3c2/components/SeekYearInReview/SeekYearInReviewScreen.js#L45) component like in Seek
2. Using [react-i18next with HTML elements](https://react.i18next.com/latest/trans-component#usage-with-simple-html-elements-like-less-than-br-greater-than-and-others-v10.4.0)
3. Using an HTML renderer, like [this one](https://github.com/inaturalist/SeekReactNative/blob/347397c30a073ea0b48ed1cf2bf6ffc19e8f70a1/components/Auth/Signup/LicensePhotosScreen.js#L171) in Seek

Chrischuck:
I can take a look at this

PR #476, opened 2023-02-11
423 translation interpolations

Chrischuck:
#423

### DIFF
--- a/src/i18n/l10n/en.ftl
+++ b/src/i18n/l10n/en.ftl
@@ -623,13 +623,14 @@ Username-or-Email = Username or Email
 
 Visually-search-iNaturalist-data = Visually search iNaturalist’s wealth of data. Search by a taxon in a location
 
-Welcome-back = Welcome back,
-
 Whenever-you-get-internet-connection-you-can-upload = Whenever you get internet connection, you can upload your observations to iNaturalist.
 
 Which-traditional-projects-can-add-your-observations = Which traditional projects can add your observations?
 
 Who-can-add-observation-fields-to-my-observations = Who can add observation fields to my observations?
+
+# Welcome user back to app
+Welcome-user = Welcome back, {$userHandle}
 
 # Banner above Explore Map showing total number of results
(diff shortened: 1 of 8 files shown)


src/i18n/l10n/en.ftl
```diff
@@ -615,14 +615,15 @@ Username-or-Email = Username or Email
...
-Welcome-back = Welcome back,
-
 Whenever-you-get-internet-connection-you-can-upload = Whenever you get internet connection, you can upload your observations to iNaturalist.
 
 Which-traditional-projects-can-add-your-observations = Which traditional projects can add your observations?
 
 Who-can-add-observation-fields-to-my-observations = Who can add observation fields to my observations?
 
+# Welcome user back to app
+Welcome-user = <0>Welcome back,</0><1>{$userHandle}</1>
```
Chrischuck, line 625: Not sure if this is correct syntax to indicate different elements in fluent.

kueda: Well it works, so that's good. Looks like you can also add some semantic specificity, e.g. 

fluent:
```fluent
Welcome-user = <plain>Welcome back,</plain><bold>{$userHandle}</bold>
```

javascript:
```javascript
<Trans
  i18nKey="Welcome-user"
  parent={View}
  values={{ userHandle: User.userHandle( currentUser ) }}
  components={{
    plain: <Body1 className="mt-1 text-3xl" />,
    bold: <Body1 className="mb-1 text-3xl font-semibold" />
  }}
/>
```

But I'm not sure that's necessary here with only two interpolations. It might help translators understand what those interpolations will do, but it could just as easily mislead them into translating the variable names (I've seen the latter happen a lot).

kueda (approved):
Very cool, didn't know react-i18next had already solved this problem for us.

Chrischuck:
Trans component doesn't seem to be doing its job in jest...Not sure how to fix it atm, might get a couple hours today but this week is busy for me :(

Issue #462, opened 2023-02-10
Allow adding an observation from anywhere the NavBar is visible


kueda:
Currently you get a navigation error any time you do this from outside My Observations. If the NavBar is visible, tapping all the buttons in the AddEvidenceModal should work.

This will probably be pretty disruptive since it will involve rethinking how we've organized our navigators (it's not clear to me why we have multiple in the first place).

Chrischuck:
I can take a look at this later today. Will probably get a proposal first in this issue thread before doing any real coding since, like you said, looks pretty involved.

Chrischuck:
Briefly looking, I think it's reasonable and a pretty common pattern to have two navigators (one drawer, one stack)

I think one possible solution might look like the following in a sub stack (mainStackNavigation, projectStackNavigation, etc.)
```
import SharedStackScreens from './SharedStackScreens'
    <Stack.Navigator screenOptions={showHeader}>
      <Stack.Screen
        name="StandardCamera"
        component={StandardCameraWithPermission}
        options={{ ...hideHeader, orientation: isTablet ? "all" : "portrait" }}
      />
    {SharedStackScreens}
    ...
```

I haven't written any code yet so I'm not sure if there will be any mismatched state issues...

Chrischuck:
Above did not work...Looks like the lowest barrier to entry is simply specifying which stack you want to navigate to. Like: `navigation.navigate( "MainStack", { screen, ...params } );` Though I'm trying to find a clean way to do this

Chrischuck:
Looks like an issue many have had with react navigation. See [here](https://github.com/react-navigation/react-navigation/issues/6434)

implemented a solution like this [one](https://github.com/react-navigation/react-navigation/issues/6434#issuecomment-604523082)

It seems to work fairly well. Able to go from `ProjectDetails` to `ObsEdit` and back to the project you were viewing (if you discard the observation you are editing)

albullington:
Closed via #507

PR #477, opened 2023-02-12
462 create wrapper for react nav

Chrischuck:
#462

### DIFF
--- a/src/components/AddObsModal.js
+++ b/src/components/AddObsModal.js
@@ -1,11 +1,11 @@
 // @flow
 
-import { useNavigation } from "@react-navigation/native";
 import { Text, View } from "components/styledComponents";
 import { t } from "i18next";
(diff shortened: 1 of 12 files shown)


jtklein (changes requested):
Hi @Chrischuck I haven't looked at your changes in detail here, but I think a better approach would be to move the bottom tab navigator on top of the entire navigation stack, and nest the drawer and other stacks inside. That way this issue will also be taken care of as well as another issue with the navbar sliding in when navigating (which we haven't formulated as an issue yet). Could you change this PR to draft for now, I think we should try the above mentioned solution and compare with this PR first.

Chrischuck:
@jtklein  Sure, so move everything to the main stack and have the drawer nested inside the mainstack?

jtklein:
Have added a description of how I think to go forward here: #499

Issue #453, opened 2023-02-09
My Observations UI


kueda:
Integrate our new UI components into an updated MyObservations. Vague goals are to use "MyObservations" naming, try to pull up as much state as possible, maybe experiment with container components as places to put hooks and tie into context.

PR #478, opened 2023-02-12
453 Integrate reusable components into obs details

Chrischuck:
Small contribution to #453 
Screenshot 2023-02-12 at 12 08 09 PM


src/components/ObsDetails/DataTab.js
```diff
@@ -55,25 +52,18 @@ const DataTab = ( { observation }: Props ): Node => {
...
-          <Text className="ml-2">
-            {`${t( "Date-uploaded-colon" )} ${displayTime( "createdAt" )}`}
-          </Text>
-        </View>
+        <DateDisplay
+          pretext={t( "Date-observed-colon" )}
+          dateString={checkCamelAndSnakeCase( observation, "timeObservedAt" )}
+        />
+        <DateDisplay
+          pretext={t( "Date-observed-colon" )}
```
jtklein, line 64: Should be `Date-uploaded-colon` here.

src/components/SharedComponents/DateDisplay.js
```diff
@@ -8,16 +8,19 @@ import IconMaterial from "react-native-vector-icons/MaterialIcons";
 import { formatApiDatetime } from "sharedHelpers/dateAndTime";
 
 type Props = {
+  pretext?: string,
```
jtklein, line 11: "pretext" is also a noun, I was confused at first, but I see what you mean. Maybe "label"

src/components/SharedComponents/DateDisplay.js
```diff
@@ -8,16 +8,19 @@ import IconMaterial from "react-native-vector-icons/MaterialIcons";
...
 const DateDisplay = ( {
-  dateString
+  dateString,
+  pretext
 }: Props ): React.Node => {
   const { t } = useTranslation( );
   return (
     <View className="flex flex-row items-center">
       <IconMaterial name="watch-later" size={15} />
+      {pretext && <Body4 className="ml-[5px]">{ pretext }</Body4>}
```
jtklein, line 23: I would rather combine the strings with a whitespace in the lower Body4.

jtklein: Also text-darkGray is redundant now in the lower one, because it is default in Body4.

src/components/SharedComponents/index.js
```diff
@@ -5,6 +5,7 @@ export { default as EvidenceButton } from "./Buttons/EvidenceButton";
 export { default as DateDisplay } from "./DateDisplay";
 export { default as Divider } from "./Divider/Divider";
 export { default as InlineUser } from "./InlineUser/InlineUser";
+export { default as Map } from "./Map";
```
jtklein, line 8: I don't think that we will need this Map here, I'd rather suspect we will end up with different interfaces in all the instances we show a map. The designs for all the different screens where we will end up having a map are not final yet, so I would not add this here for now.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -334,28 +334,11 @@ const ObsDetails = ( ): Node => {
...
-            />
-            <QualityGradeStatus
-              qualityGrade={checkCamelAndSnakeCase( observation, "qualityGrade" )}
-              color={colors.darkGray}
-            />
-          </View>
+          <ObsStatus
+            layout="vertical"
+            observation={observation}
+            color={colors.darkGray}
```
jtklein, line 340: I would rather make this the default color inside ObsStatus (and all its children) and override only with a "white" boolean. So, that there is no possibility for now to use another color.

jtklein (changes requested):
Only some minor changes for the implementation, however, some tests are also failing now.

Issue #473, opened 2023-02-11
Hold to show delete button in PhotoPreview


kueda:
Currently the PhotoPreview in the StandardCamera has delete buttons visible all the time. Instead, we'd like to show them on hold / long-press.

- [x] Remove existing delete buttons from PhotoPreview
- [x] Add the `trash` icon to the icon font
- [x] When the user long-presses on a thumbnail, show delete buttons over all the thumbnails; buttons should be translucent to show the photo behind them but they should cover the whole thumbnail so the entire area is tappable
- [x] When the user taps a delete button, delete the photo immediately
- [x] Tapping anywhere other than one of the delete buttons should exit deletion mode, i.e. the buttons disappear and the thumbnails appear as they did before

Here's what the thumbnails look like in delete mode:

<img width="300" alt="Screen Shot 2023-02-10 at 5 47 34 PM" src="https://user-images.githubusercontent.com/22691/218232265-74c1a4cd-e151-438c-b789-c06b50754572.png">

Chrischuck:
I can work on this

PR #479, opened 2023-02-13
473 Create hold to delete photo interaction

Chrischuck:
#473 - put wrong issue # in branch by accident

- [x] Remove existing delete buttons from PhotoPreview
- [ ] Add the trash icon to the icon font
- [x] When the user long-presses on a thumbnail, show delete buttons over all the thumbnails; buttons should be translucent to show the photo behind them but they should cover the whole thumbnail so the entire area is tappable
- [x] When the user taps a delete button, delete the photo immediately
- [x] Tapping anywhere other than one of the delete buttons should exit deletion mode, i.e. the buttons disappear and the thumbnails appear as they did before

https://user-images.githubusercontent.com/13311268/218353123-709f4f5c-c0cd-4bc9-a6fd-3f3c7c04de11.mov

Other usages of `PhotoCarousel` persist
Screenshot 2023-02-12 at 6 46 58 PM
Screenshot 2023-02-12 at 6 48 31 PM


jtklein (changes requested):
Hi @Chrischuck layout-wise everything works as expected. Implementation also seems fine, however, could you revert back to the previous API of PhotoCarousel in a sense that the logic of deleting the photo is in PhotoPreview with a handleDelete prop in PhotoCarousel. The PhotoCarousel component should stay presentational only.

Chrischuck:
@jtklein can do, originally had to do that because of the modal but realized taking out the modal was part of the spec. Can revert this.

PR #486, opened 2023-02-15
432 obs list item 2

Chrischuck:
changes from pr #456


src/components/SharedComponents/ObservationLocation.js
```diff
@@ -14,25 +15,18 @@ type Props = {
...
-    !locationName
-    // Check for undefined or null, Not 0
-    && observation?.latitude != null
-    && observation?.longitude != null
-  ) {
-    locationName = `${observation.latitude}, ${observation.longitude}`;
-  } else if ( !locationName ) {
-    locationName = t( "Missing-Location" );
+  let displayLocation = checkCamelAndSnakeCase( observation, "placeGuess" );
+  if ( !displayLocation && observation.latitude ) {
```
Chrischuck, line 19: I think this was brought up in another comment, but this if statement wouldn't run on any location on the equator since `0` would be falsey.  `foo != null` would be `false` for `null` or `undefined` but would allow `0` to pass

Chrischuck: Made a test case for the above.

src/components/Observations/ObsPreviewImage/index.js
```diff
@@ -0,0 +1,82 @@
...
+import IconMaterial from "react-native-vector-icons/MaterialIcons";
+
+import Background from "./Background";
+
+type URI = {
+  uri: string,
+};
+
+type Props = {
+  uri?: URI,
```
jtklein, line 16: Traditionally this would be called "source" ( = { uri } )

src/components/Observations/ObsGridItem.js
```diff
@@ -2,124 +2,76 @@
...
-        layout="horizontal"
-        color={colors.white}
-      />
-    );
-  };
-
   return (
     <Pressable
-      onPress={onPress}
+      onPress={( ) => onPress( observation )}
```
jtklein, line 41: Refactor Pressable out of this file, see ObservationViews.js

src/components/Observations/ObsGridItem.js
```diff
@@ -2,124 +2,76 @@
...
+import ObsPreviewImage from "./ObsPreviewImage";
 import ObsStatus from "./ObsStatus";
 import UploadButton from "./UploadButton";
 
 type Props = {
   observation: Object,
-  handlePress: Function,
-  uri?: string,
+  onPress: Function,
+  displayType?: "project" | "observation",
```
jtklein, line 20: If there is a default value, this could also be a boolean.

jtklein (changes requested):
Hello @Chrischuck this was quite an effort, thanks!

I was comparing the app layout in the simulator to the Figma designs we have, and there are some layout props to be changed, in addition to the detailed review with inline comments. I will attach some screenshots as well:
- The margins left and right for one entire list item should be 15px to the end of the screen.
- The center component is overflowing into the ObsStatus when the location name for example is too long. So it needs to get a limit on its width, and also needs margin right of 10px. (Currently there is no truncation logic in the children though, and adding it there is out-of-scope here).
- The margin in top and bottom of one entire list item should be 11px.
- The element on the right is centered along the height of the list item but it actually should be top-aligned (with fixed spaces between its children).


The ObsGridItem is not working in the project screen, have to check on main if it was working there.

jtklein:
IMG_7665

This shows the overflow of the center element.

jtklein:
Screenshot 2023-02-16 at 14 36 35

When there is a lot of text you can see the right element to be top aligned

src/components/SharedComponents/ObservationLocation.js
```diff
@@ -1,37 +1,35 @@
...
-  let locationName = observation?.place_guess;
-
+  let displayLocation = checkCamelAndSnakeCase( observation, "placeGuess" );
   if (
-    !locationName
-    // Check for undefined or null, Not 0
-    && observation?.latitude != null
-    && observation?.longitude != null
+    !displayLocation
+    && observation.latitude != null
```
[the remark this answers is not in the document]
kueda: FWIW, an observation can definitely have blank `latitude` and `longitude` values. Regarding `!=` vs `!==`, we have an eslint rule to enforce that (`no-constant-condition`) so I'm not sure why that didn't catch this.

Chrischuck:
@jtklein I think I got everything, lmk if I missed something. Was also able to add truncation + elipsis when the wording is too long.

src/components/Projects/ProjectObservations.js
```diff
@@ -25,12 +28,20 @@ const ProjectObservations = ( { id }: Props ): React.Node => {
...
-      handlePress={navToObsDetails}
-      uri="project"
-      width="w-6/12"
-    />
+    <Pressable
+      className="w-6/12"
+      onPress={() => navToObsDetails( item )}
+      accessibilityRole="link"
+      accessibilityHint={t( "Navigate-to-observation-details" )}
+      accessibilityLabel={t( "Observation-details", {
```
jtklein, line 36: Only Observation

jtklein (approved):

jtklein:
Thanks a lot, @Chrischuck. I went ahead and merged your branch in. There were a few things left to do, but I think because you force-pushed within your branch my local history was with wrong commits and I had a lot of merge conflicts. So I just added the missing changes after I merged your branch.

PR #488, opened 2023-02-16
400 uploadstatus component

angielt:


src/components/SharedComponents/QualityGradeStatus/QualityGradeStatus.js
```diff
@@ -26,24 +24,31 @@ const qualityGradeSVG = ( qualityGrade, color ) => {
...
 
-const QualityGradeStatus = ( { qualityGrade, color }: Props ): React.Node => (
-  <View>{qualityGradeSVG( qualityGrade, color )}</View>
-);
+const QualityGradeStatus = ( { qualityGrade, color }: Props ): React.Node => {
+  const setAccessibilityLabel = () => (
+    ( qualityGrade ) ? `Quality-Grade-${qualityGrade}` : "No grade" );
+
+  return (
+    <Pressable
```
jtklein, line 44: Why did you change to Pressable here, if this button has no function when pressed it can be a plain View?

src/components/SharedComponents/QualityGradeStatus/QualityGradeStatus.js
```diff
@@ -26,24 +24,31 @@ const qualityGradeSVG = ( qualityGrade, color ) => {
...
     />
   );
 };
 
-const QualityGradeStatus = ( { qualityGrade, color }: Props ): React.Node => (
-  <View>{qualityGradeSVG( qualityGrade, color )}</View>
-);
+const QualityGradeStatus = ( { qualityGrade, color }: Props ): React.Node => {
+  const setAccessibilityLabel = () => (
+    ( qualityGrade ) ? `Quality-Grade-${qualityGrade}` : "No grade" );
```
jtklein, line 41: "No grade" is not added to the strings

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -0,0 +1,37 @@
...
+import CircularProgressBase from "react-native-circular-progress-indicator";
+import colors from "styles/tailwindColors";
+
+type Props = {
+  color: string,
+  completeColor: string,
+  progress: number
+}
+
+const UploadStatus = ( { color, completeColor, progress }: Props ): React.Node => (
```
jtklein, line 15: Please also give the color and completeColor a default value of darkGray.

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -0,0 +1,37 @@
...
+  <View className="relative justify-center items-center">
+    {( progress < 0.05 )
+      ? <INatIcon className="m-1" name="status-saved" color={colors.darkGray} size={50} />
```
jtklein, line 18: After you added fixed dimensions to the View and the icon is centered in it, no margin needed here I guess.

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -0,0 +1,37 @@
...
+import colors from "styles/tailwindColors";
+
+type Props = {
+  color: string,
+  completeColor: string,
+  progress: number
+}
+
+const UploadStatus = ( { color, completeColor, progress }: Props ): React.Node => (
+  <View className="relative justify-center items-center">
```
jtklein, line 16: Can you add a default width and height 49 x 67 here. In the design specs every one of the different status icons has a box of these dimensions around it. I guess so that it fills a similar space than ObsStatus in a list item.

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -0,0 +1,37 @@
...
+    {( progress < 0.05 )
+      ? <INatIcon className="m-1" name="status-saved" color={colors.darkGray} size={50} />
+      : (
+        <>
+          {( progress < 1 )
+            ? <INatIcon className="absolute" name="upload-arrow" color={color} size={25} />
```
jtklein, line 22: Can you change the icons so that the size of the circle is 33px by default. The design specs have 31x31 when using these status icons in the observation list screen. (The save icon is acually 33x33 in the specs, but we would need to check if the layout jumps when the progress is moving.)

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -0,0 +1,37 @@
...
+            : <INatIcon className="absolute" name="checkmark" color={completeColor} size={25} />}
+          <CircularProgressBase
+            value={progress}
+            radius={29}
+            activeStrokeColor={( progress < 1 ) ? color : completeColor}
+            progressValueColor={colors.transparent}
+            maxValue={1}
+            inActiveStrokeOpacity={0}
+            activeStrokeWidth={4}
```
jtklein, line 31: The stroke width in the design specs is 2px (when the icon is 31x31).

jtklein (changes requested):
Implementation seems mostly fine. I left some comments mainly about layout. There are some differences from what I have seen in the UiLib compared to the Figma designs. I think I left a comment for every difference I could find but maybe good to double check in the end. I would prefer to have the component close to design specs before merging.

It is not part of the issue, but could you also add a unit test for the component?

It would be nice to see the full animation cycle in the UiLib but not needed.

jtklein:
Screenshot 2023-02-17 at 18 31 45
Icon dimensions when used in ObsListItem

jtklein:
Screenshot 2023-02-17 at 18 31 27
Dimensions of the container when used in ObsListItem

src/components/SharedComponents/UploadStatus/UploadStatus.js
angielt: Should the default value for completeColor be inatGreen instead?

src/components/SharedComponents/UploadStatus/UploadStatus.js
angielt: Yes, you are right

tests/unit/components/SharedComponents/UploadStatus/UploadStatus.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+jest.mock( "react-native/Libraries/Animated/NativeAnimatedHelper" );
+describe( "UploadStatus", () => {
+  it( "does not show progress bar when progress is less than 5%", async () => {
+    render( <UploadStatus progress={0.01} color="#454545" completeColor="#77b300" /> );
+
+    // eslint-disable-next-line testing-library/no-debugging-utils
+    screen.debug();
+    expect( screen.queryByTestId( "UploadStatus.CircularProgress" ) ).toBeFalsy();
+    expect( screen ).toMatchSnapshot();
```
jtklein, line 15: I would separate those assertions into different `it`s, so that one is a unit test and one a snapshot test.

tests/unit/components/SharedComponents/UploadStatus/UploadStatus.test.js
```diff
@@ -0,0 +1,26 @@
...
+  } );
+
+  it( "shows progress bar when progress is greater than 5%", async () => {
+    render( <UploadStatus progress={0.5} color="#454545" completeColor="#77b300" /> );
+
+    // eslint-disable-next-line testing-library/no-debugging-utils
+    screen.debug();
+    expect( screen.getByTestId( "UploadStatus.CircularProgress" ) ).toBeTruthy();
+    expect( screen ).toMatchSnapshot();
```
jtklein, line 24: See above

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -1,37 +1,101 @@
...
+  const rotate = {
+    transform: [
+      {
+        rotate: interpolateRotating
+      }
+    ]
+  };
+
+  return (
+    <View className="relative items-center justify-center w-[49px] h-[67px]">
```
jtklein, line 48: You would also need to add accessibility props here: `accessible` to make sure this View acts as the accessible element over it's children. Then a good label that also reads out aloud the current progress while uploading and the status in the other case: waiting for upload, finished uploading; maybe some more.

tests/unit/components/SharedComponents/UploadStatus/UploadStatus.test.js
```diff
@@ -0,0 +1,26 @@
+import { render, screen } from "@testing-library/react-native";
+import { UploadStatus } from "components/SharedComponents";
+import React from "react";
+
+jest.mock( "react-native-circular-progress-indicator", () => "CircularProgress" );
+
+jest.mock( "react-native/Libraries/Animated/NativeAnimatedHelper" );
+describe( "UploadStatus", () => {
```
jtklein, line 8: Can you also add the accessibility matcher test for the component here as well.

jtklein (changes requested):
Have checked everything again now. Everything seems mostly fine now and good to go with the exception of a few inline comments. Otherwise, some of the snapshot tests were failing and I pushed some missing updates to the snapshots, however, there is still the INatIcon unit test failing, if you can fix this test passing, I think this is good to go.

One question I had was: Abhas mentioned in the Figma designs that the whole UploadStatus should fade out when the progress has reached 1, and the ObsStatus then fade in. Would you add that fading animation to the component itself or to the container of both components? That is not to add here, but depending on your answer you could make a new issue for that.

src/components/SharedComponents/UploadStatus/UploadStatus.js
jtklein: Yes, probably better so.

Issue #494, opened 2023-02-21
ActivityCount component


jtklein:
Some more specs for this component:

- [x] Default color darkGray, boolean prop for white version
- [x] Add comments icon filled to font
- [x] Add comments icon not filled to font, after #518 
- [x] Add IDs icon filled to font, after #518 
- [x] Add IDs icon not filled to font, after #518 
- [x] (Potentially add components for comments and IDs with logic to show filled or not, and a11y props)
- [x] Unit test
- [x] Snapshot test

albullington:
Closed by #495

Issue #487, opened 2023-02-15
Rename and re-upload icons


kueda:
We're accumulating icons with names that have extraneous stuff like `icon-`, overly specified names like `noevidence`, nonsensical names like `vector-stroke`, and most need to be re-uploaded on a square 24x24x document.

Each icon should
- [ ] Have its own 24x24 px SVG file
- [ ] Be named with just its name, e.g. `pencil.svg`, not `icn-pencil.svg` or `pencil-icon.svg`
- [ ] Have no `fill-rule: evenodd` styles
- [ ] Be checked into this repo (under assets/fonts?)

On fontastic
- [ ] Delete all entries with bad names and re-upload the new SVG files

...orrr maybe write a script to generate a font from our icons in this repo so anyone can do it without relying on fontastic?

In the code
- [ ] Refactor to use the new icon names
- [ ] Ensure there are no broken icons

kueda:
Now I'm kind of wondering if fonts should be checked into their own repo...

abhasinat:
super helpful, thanks for this guidance ken-ichi.

PR #495, opened 2023-02-21
494 activity count

jtklein:
Some refactoring of the ActivityCount component.
Added some basic unit tests but TBC after #487


jtklein:
Closes #494

src/components/Observations/ObsStatus.js
```diff
@@ -37,21 +36,17 @@ const ObsStatus = ( {
 
   return (
     <View className={classNames( "flex", flexDirection, classNameMargin )}>
-      <ActivityCount
-        icon="label"
-        margin={margin}
+      <IdentificationsCount
+        classNameMargin={margin}
```
albullington, line 40: shouldn't this and `CommentsCount` have a filled prop which will be filled if the Id or Comment is unread?

jtklein: Yes, both those components do have a `filled` boolean prop ready to be used. I have only concerned myself with the UI for this PR because the issue was about UI only. In the UiLib there is an example of the filled usage. Do you want me to add the logic part to check whether to use the filled UI for an observation here or we make a separate issue for that?

albullington: I think it either makes sense to leave this component in the UI library & not try to integrate it into `ObsStatus` at all, or it makes sense to have a fully working component in `ObsStatus`.

jtklein: So, I was doing some digging in the git history of the ObsStatus component. From what I can see the boolean used for both comments and identifications together was `observation.viewed` to indicate if it was viewed or not. This is only local I think (not coming from the API). In our case here we would need two values separately for the id and comment.
I also tried to make an API call to `v2/observations/updates?viewed=false&observation_by=owner` which differs in the result values from API v1, returning currently an array like so: `[{"id": 371531736}, {"id": 371518022}]` which is neither the id of the identification nor the comment which was not viewed in this case.

I want to try to not leave unfinished work until the end of the week, so I have removed the usage of the components from ObsStatus and will punt the logic part to a separate issue.

jtklein: But I also think the logic part will not end up in ObsStatus because, e.g., in ObsDetail we are using also ObsStatus and it will always have all comments and identification viewed there.

Issue #483, opened 2023-02-14
PhotoCount component


kueda:
Small component to represent a number of photos with an icon.

Props:
- [ ] count: integer, number of photos being represented, 0-99
- [ ] shadow: boolean, whether to show a drop shadow or not

Visual:
- [ ] Don't render anything if count === 0
- [ ] Show count in the empty area of a `photos` icon: ![photos](https://user-images.githubusercontent.com/22691/218613783-20f09369-e2eb-4181-be9c-58e16704da8d.svg)
- [ ] Icon should be white
- [ ] Give it a shadow if `shadow` is true, something like `drop-shadow(0px 1px 2px rgba(0, 0, 0, 0.25));`

<img width="188" alt="Screen Shot 2023-02-13 at 5 14 12 PM" src="https://user-images.githubusercontent.com/22691/218612747-83f7f022-86e9-46a7-876c-5a3627099d8d.png">

Also:
- [ ] Add it to the UI Library showing counts of 1 and 99

YahiaJabeur:
I created a [PR](https://github.com/inaturalist/iNaturalistReactNative/pull/497) for this .

PR #497, opened 2023-02-21
483 create PhotoCount component

YahiaJabeur:


src/components/SharedComponents/PhotoCount.js
```diff
@@ -0,0 +1,58 @@
...
+import Svg, { Path, Text } from "react-native-svg";
+
+type Props = {
+  count: number,
+  size?: number,
+  shadow: boolean,
+};
+
+const PhotoCount = ( { count, size, shadow }: Props ): React.Node => {
+  if ( count <= 0 || count > 99 ) {
```
jtklein, line 14: Count should not be negative, so you don't need to account for negative values here. So, please change to `=== 0` .

jtklein, line 14: Also, we should still show the icon if count is > 99.

src/components/SharedComponents/PhotoCount.js
```diff
@@ -0,0 +1,58 @@
+// @flow
+
+import { View } from "components/styledComponents";
+import * as React from "react";
+import Svg, { Path, Text } from "react-native-svg";
+
+type Props = {
+  count: number,
+  size?: number,
+  shadow: boolean,
```
jtklein, line 10: Shadow could be optional and have a default value in this component of false.

src/components/SharedComponents/PhotoCount.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+const PhotoCount = ( { count, size, shadow }: Props ): React.Node => {
+  if ( count <= 0 || count > 99 ) {
+    return null;
+  }
+
+  return (
+    <View
+      className={`drop-shadow-md w-[${String( size )}px] h-[${String( size )}px] ${String(
```
jtklein, line 20: We are using https://www.npmjs.com/package/classnames for more optional combinations of classNames. So, e.g. you can refactor to add `shadow && "shadow"` as one argument to the classNames fct.

src/components/SharedComponents/PhotoCount.js
```diff
@@ -0,0 +1,58 @@
...
+        shadow && "shadow"
+      )}`}
+    >
+      <Svg
+        height={size}
+        width={size}
+        viewBox="0 0 24 24"
```
jtklein, line 27: If the viewBox is 24x24 and the icon is smaller it will be aligned at 0,0 (so top left) correct? If that is true we have to change the viewBox by size I guess.

src/components/SharedComponents/PhotoCount.js
```diff
@@ -0,0 +1,58 @@
...
+        />
+        <Path
+          // eslint-disable-next-line
+          d="M15.364 3.636h-9.53A4 4 0 019.818 0H20a4 4 0 014 4v10.182a4 4 0 01-3.636 3.984v-9.53a5 5 0 00-5-5z"
+          fill="#fff"
+          clipRule="evenodd"
+          fillRule="nonzero"
+          filter="url(#drop-shadow)"
+        />
+        <Text x="38%" y="77%" textAnchor="middle" fontSize="10" fill="gray">
```
jtklein, line 46: Is it possible to use our already styled component Body3 here?

src/components/UiLibrary.js
```diff
@@ -362,7 +362,18 @@ const UiLibrary = () => {
             identifications: [1, 2, 3, 4, 5, 6]
           }}
         />
-
+        <View className="my-2 bg-lightGray p-2 rounded-lg flex-row justify-evenly">
+          <PhotoCount
```
jtklein, line 366: Can you add a heading here that names `PhotoCount` explicitly.

jtklein, line 366: Can you add here one version where count is 0.

src/components/UiLibrary.js
```diff
@@ -362,7 +362,18 @@ const UiLibrary = () => {
...
+            count={9}
```
jtklein, line 367: Could you also change one into a version with only count and the rest default values.

jtklein (changes requested):
Hello @YahiaJabeur thank you so much for your contribution. Have tested on real devices and looks good.
I have left a few inline comments. In addition, could you also setup a basic unit test that checks for the correct size and values of the component.

src/components/SharedComponents/PhotoCount.js
jtklein: Also, shadow should have 0.25 instead of 0.1 for alpha channel

src/components/SharedComponents/PhotoCount.js
YahiaJabeur: No need to viewBox, I tried and it works fine.
changing viewBox by size make ui issue

src/components/SharedComponents/PhotoCount.js
YahiaJabeur: Yes it s possible to use Body3 but :
```
When you use the Text component from react-native inside an Svg component created with react-native-svg, the text is rendered as a bitmap image, which means that it doesn't have the same level of scalability and interactivity as other vector elements created with react-native-svg. If you need to use scalable text in your SVG graphics, it's better to use the Text component from react-native-svg
```

jtklein: I would like to avoid having to add any styles here to this Text, and instead, use our already existing styled Text component that fits the design specs here. Then we have a central place to change if needed.
Do I understand you correctly that as long as we do not change the size of the icon or count number during runtime we can use Body3 because the bitmap would not need to change because we would not scale or interact with it? Is there a performance impact in terms of having the bitmap?

src/components/SharedComponents/PhotoCount.js
YahiaJabeur: I think that [NativeWind](https://www.nativewind.dev/)  dont support all drop shadow as in CSS
I used for the shadow react native styles

src/components/SharedComponents/PhotoCount.js
YahiaJabeur: by using Text from the svg dep the font size for example it s managed by the size of the svg component.
I tried using react native Text component and I did't find a way to make the required stype.
Please feel free to try and commit in the branch

src/components/SharedComponents/PhotoCount.js
```diff
@@ -3,23 +3,29 @@
 import { View } from "components/styledComponents";
 import * as React from "react";
 import Svg, { Path, Text } from "react-native-svg";
+import { viewStyles } from "styles/sharedComponents/photoCount";
+import colors from "styles/tailwindColors";
```
jtklein, line 8: We are in the process of unifiying our color usage through the app. In the meantime could you please switch to import colors using the `useTheme` hook from react-native-paper. E.g. here: https://github.com/YahiaJabeur/iNaturalistReactNative/blob/33633c06f110878c0c9b117ea21f4fa6428bb9fb/src/components/UiLibrary.js#L47

src/components/SharedComponents/PhotoCount.js
jtklein: Okay, thanks for checking. We will have to change back for everything else here though, see new inline comment.

src/components/SharedComponents/PhotoCount.js
```diff
@@ -3,23 +3,29 @@
...
+  if ( count > 99 ) {
+    photoCount = 99;
+  }
+
   return (
     <View
-      className={`drop-shadow-md w-[${String( size )}px] h-[${String( size )}px] ${String(
-        shadow && "shadow"
-      )}`}
+      style={[{ height: size, width: size }, shadow && viewStyles.shadow]}
```
jtklein, line 28: We want to rely mostly on Tailwind/Nativewind for styling of our app. So, since you checked that drop-shadow is not possible to add with class names, you can leave that here in style, but change everything back to class name. Also, maybe change the size probe to `classNameSize`, then you don't need to create a string from the size number here, just concatenate it.

src/components/UiLibrary.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
jtklein, line 1: Thanks, good catch.

src/components/SharedComponents/PhotoCount.js
jtklein: I don't understand. The viewBox prop is still there. Is it required? Expo documents mention to best have it included for Android. Currently, it is still present in the component. Will there be a problem with diferent sizes?

src/components/SharedComponents/PhotoCount.js
jtklein: Can you please position the new Body3 element correctly in the svg.

src/components/SharedComponents/PhotoCount.js
```diff
@@ -32,19 +38,23 @@ const PhotoCount = ( { count, size, shadow }: Props ): React.Node => {
           fillRule="nonzero"
           clipRule="evenodd"
           d="M4 5.818a4 4 0 00-4 4V20a4 4 0 004 4h10.182a4 4 0 004-4V9.818a4 4 0 00-4-4z"
-          fill="#fff"
-          filter="url(#drop-shadow)"
+          fill={colors.white}
```
jtklein, line 42: See above

src/components/SharedComponents/PhotoCount.js
```diff
@@ -32,19 +38,23 @@ const PhotoCount = ( { count, size, shadow }: Props ): React.Node => {
...
         />
         <Path
           // eslint-disable-next-line
           d="M15.364 3.636h-9.53A4 4 0 019.818 0H20a4 4 0 014 4v10.182a4 4 0 01-3.636 3.984v-9.53a5 5 0 00-5-5z"
-          fill="#fff"
+          fill={colors.white}
```
jtklein, line 47: See above

src/components/SharedComponents/PhotoCount.js
```diff
@@ -32,19 +38,23 @@ const PhotoCount = ( { count, size, shadow }: Props ): React.Node => {
           fillRule="nonzero"
           clipRule="evenodd"
           d="M4 5.818a4 4 0 00-4 4V20a4 4 0 004 4h10.182a4 4 0 004-4V9.818a4 4 0 00-4-4z"
-          fill="#fff"
-          filter="url(#drop-shadow)"
+          fill={colors.white}
         />
         <Path
           // eslint-disable-next-line
```
jtklein, line 45: Haven't let a comment here before. But if you disable eslint, please be specific about which rule.

tests/unit/components/SharedComponents/PhotoCount/PhotoCount.test.js
```diff
@@ -0,0 +1,33 @@
+import { render, screen } from "@testing-library/react-native";
+import { PhotoCount } from "components/SharedComponents";
+import React from "react";
+
+describe( "PhotosCount", () => {
+  it( "should render correctly", () => {
+    render( <PhotoCount count={10} /> );
+
+    const photoCount = screen.queryByTestId( "photo-count" );
```
jtklein, line 9: `queryByTestId` is to be used in situations when you do not expect the element to be present.

If you expect the element to be present, you have to use `getByTestId` (or `findByTestId` if you expect the element to appear after some async changes).

tests/unit/components/SharedComponents/PhotoCount/PhotoCount.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+    expect( photoCount ).toBeNull();
+  } );
+
+  it( "should show the correct value", () => {
+    render( <PhotoCount count={14} /> );
+
+    expect( screen ).toMatchSnapshot();
```
jtklein, line 25: You can change to only have one snapshot test with the default values only.

tests/unit/components/SharedComponents/PhotoCount/PhotoCount.test.js
```diff
@@ -0,0 +1,33 @@
...
+import { PhotoCount } from "components/SharedComponents";
+import React from "react";
+
+describe( "PhotosCount", () => {
+  it( "should render correctly", () => {
+    render( <PhotoCount count={10} /> );
+
+    const photoCount = screen.queryByTestId( "photo-count" );
+
+    expect( photoCount ).toBeTruthy();
```
jtklein, line 11: Here it would be nicer, to assert for the correct text to be shown, in this case, 10.

tests/unit/components/SharedComponents/PhotoCount/PhotoCount.test.js
jtklein: Also, check for correct text here instead.

tests/unit/components/SharedComponents/PhotoCount/PhotoCount.test.js
```diff
@@ -0,0 +1,33 @@
...
+  it( "should show the correct value", () => {
+    render( <PhotoCount count={14} /> );
+
+    expect( screen ).toMatchSnapshot();
+  } );
+
+  it( "should show photo count value equal to 99", () => {
+    render( <PhotoCount count={200} /> );
+
+    expect( screen ).toMatchSnapshot();
```
jtklein, line 31: Also, check for correct text here instead.

jtklein (changes requested):
Thanks again, requested some more changes in inline comments.

src/components/SharedComponents/PhotoCount.js
jtklein: Hi, we actually have added a default style for drop-shadow in a different pull request. Please import the `dropShadow` style from here `styles/global.js`.

src/components/SharedComponents/PhotoCount.js
YahiaJabeur: Yes, the viewBox is required.
There are no issues if the icon size is smaller or bigger than 24

src/components/SharedComponents/PhotoCount.js
YahiaJabeur: That was possible only by adding a react-native component inside ForeignObject react-native-svg component

tests/unit/components/SharedComponents/PhotoCount/PhotoCount.test.js
YahiaJabeur: Thanks for the explanation !

jtklein:
Closes #483

Issue #500, opened 2023-02-22
AddObsButton styling


jtklein:
According to the design specs [here](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=333%3A541&t=ktTRo97Q9YUxm9i3-4) the AddObsButton component should be updated with the following changes:

- [x] Default size of 69x69
- [x] Replace the current "plus" icon with the correct version after #487 
- [x] The plus icon in the centre should have a size of 31x31 without any padding or margin. (Thereby leaving a 19px distance from the edge of icon to edge of button).
- [x] Add a drop shadow to the button: `box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.25)`
- [x] Add a linear gradient to the button's background color: `background: linear-gradient(156.95deg, #77B300 0%, #297F87 194.7%, #297F87 194.7%);`

Chrischuck:
I can put something together for this

PR #504, opened 2023-02-23
500 add obs btn styling

Chrischuck:
#500 

A couple of issues:

1. 69x69 looks pretty big for the navbar, not sure if the dimensions of it should be adjusted
2. Drop shadow is hard to see
3. The icon doesn't take up the full width of its actual element, makes knowing its actual size difficult

Screenshot 2023-02-22 at 6 26 25 PM


jtklein:
The "plus" icon will need to be reimported from our design files, and then the icon will be in a square just as big as the strokes, so the size should be the actual size. For now you can assume the 31x31.
For the drop shadow, I think we might want to add a padding on phones with stuff in the bottom, but for now it is not in the design specs.

src/components/SharedComponents/Buttons/AddObsButton.js
```diff
@@ -1,17 +1,34 @@
...
 import { t } from "i18next";
 import * as React from "react";
-import { IconButton, useTheme } from "react-native-paper";
+import { Pressable, StyleSheet } from "react-native";
+import LinearGradient from "react-native-linear-gradient";
+import colors from "styles/tailwindColors";
 
-const AddObsButton = ( ): React.Node => {
-  const theme = useTheme( );
+const shadow = StyleSheet.create( {
```
jtklein, line 13: Is it possible to use the nativewind shadow classname here instead. The nativewind shadow adds a box-shadow though, I don't know if that is any different than a "drop-shadow".
If not, at least the shadowColor should be black here.

src/components/SharedComponents/Buttons/AddObsButton.js
```diff
@@ -20,19 +37,30 @@ const AddObsButton = ( ): React.Node => {
...
-        disabled={false}
-        testID="add-obs-button"
-      />
+        accessibilityRole="button"
+        accessibilityState={{
+          disabled: false
+        }}
+        style={shadow}
+      >
+        <View className="w-[69px] h-[69px] rounded-full overflow-hidden">
```
jtklein, line 52: Just curious, is there a need for this View here?

jtklein (approved):
Excellent job! Only two questions inline.

jtklein (approved):
Excellent job! Only two questions inline.

src/components/SharedComponents/Buttons/AddObsButton.js
Chrischuck: I think it could be omitted if `Pressable` is wrapped as a styled component so it can consume the `className` prop.

src/components/SharedComponents/Buttons/AddObsButton.js
Chrischuck: Unfortunately no, I think react-native does something special under the hood to apply the shadow effect. In any case, android specifically uses `elevation` style to determine the drop shadow effect.

jtklein: Great idea to have a central place for the drop-shadow style then. I have just moved it into the styles folder, that makes more sense to me.

PR #505, opened 2023-02-23
455 update colors and unify usage in the app

angielt:


albullington (approved):
This looks good to me. Not a blocker, but I do have one question: is there any reason we're still accessing any colors with `import colors from "styles/tailwindColors";`? Can we use `useTheme` to access these colors instead?

angielt:
There were a few instances where the color of the item does not change with the change in theme(like the white checkmark inside the circle on the StandardCamera screen or the black screen of the camera). For the most part we can access colors from useTheme instead of colors. Some screens are also in the process of a redesign so I used`import colors from "styles/tailwindColors";` for now.

Issue #499, opened 2023-02-22
NavBar component should stay in place


jtklein:
Currently, when navigating to a screen inside one of the bottom tab's stack the navbar component is replaced by a new instance animated into the view together with the screen component.
Expected behaviour is for the bottom NavBar to stay fixed in place while the screen is animated into view.

Screen recording of the wrong behaviour:

https://user-images.githubusercontent.com/17345891/220632930-3b2b1ac1-c65b-4a42-9b68-a61d4e62f3a9.mp4

jtklein:
I think that this issue here, as well as #462 can be both solved simultaneously by applying the following changes to our react-navigation setup:

1. Create a [Bottom Tab Navigator](https://reactnavigation.org/docs/bottom-tab-navigator/) as the outermost navigator of the app.
2. Create a stack navigator for each tab, as described in the react-navigation docs about [Tab navigation](https://reactnavigation.org/docs/tab-based-navigation#a-stack-navigator-for-each-tab)
3. Refactor the NavBar component
4. Use the current drawer navigator as the stack for the left-most tab. I have not combined a drawer plus bottom tabs myself, so this part would be to try out and see if it works.

Chrischuck:
If nobody picks this up by the weekend I can take a look. Looks manageable given what you provided above.

Chrischuck:
Got a basic tab component going...This is a little tricky, because I think a stack for each tab would run into the same issue described in #462, with screens siloed away from each other.

But I'm thinking we follow a pattern shown [here](https://reactnavigation.org/docs/hiding-tabbar-in-screens) where there is a root stack component and all screens w/ the tab bar (including projects, settings, etc.) will live in the nested tab bar component, their tabs just won't be rendered.

From here, instead of using drawer navigator, we can use `react-paper`'s [side bar component](https://callstack.github.io/react-native-paper/docs/components/Drawer/DrawerSection).

I __think__ this will work but I'm only a couple hours in.

In any case, the tab bar I made should get us somewhere...

<img width="348" alt="Screenshot 2023-02-25 at 7 11 28 AM" src="https://user-images.githubusercontent.com/13311268/221359411-2f394f83-df1d-467c-a31a-ef7e23d87a7e.png">

Chrischuck:
#507 is working, only issue is login/logout are a state behind

albullington:
I believe this was closed via #507

PR #507, opened 2023-02-25
Create tab navigator

Chrischuck:
#499 & #462

https://user-images.githubusercontent.com/13311268/221372286-f8920da1-0c9c-4428-8844-097254f5c23a.mov


src/components/ObsDetails/ObsDetails.js
```diff
@@ -202,8 +203,16 @@ const ObsDetails = ( ): Node => {
...
+      <HeaderBackButton
+        tintColor={colors.black}
+        onPress={navigation.goBack}
+      />
+    )
+
     navigation.setOptions( {
-      headerRight: editIcon
+      headerRight: editIcon,
+      headerLeft: backButton
```
Chrischuck, line 215: I'm not sure how this page had a back button w/o this code previously. But for what it's worth, I think that this should be defined w/ the screen definition in the navigator.

src/realmModels/User.js
```diff
@@ -23,6 +23,8 @@ class User extends Realm.Object {
 
   static currentUser = realm => realm.objects( "User" ).filtered( "signedIn == true" )[0]
 
+  static firstSignedInUser = realmObjects => realmObjects.filtered( "signedIn == true" )[0]
```
jtklein, line 26: This deos not seem to be used somewhere.

jtklein (changes requested):
Hi @Chrischuck, have now managed to look through all of this PR. For the most part it looks good to me. Some issues I have seen in the apps:

On Android the back button is not visible, you can tap it though. I haven't checked if it is transparent or same color as the background, or a text with no string or ...

There is no more back button in the Taxon details screen.

Now that the footer is no longer there, these two pairs of components can be combined and renamed, as they are identical now:
ScrollNoFooter & ScrollWithFooter
ViewNoFooter & ViewWithFooter

src/components/ObsDetails/ObsDetails.js
Chrischuck: Can do, for future reference, in a Stack navigator, the back button gets automatically added if it is possible to go back. But when we switched to tab navigation, the behavior did not follow

Issue #431, opened 2023-02-03
Import photos using native photo chooser


albullington:
Instead of using `react-native-camera-roll`, we'd like to switch to using the native photo selection system on iOS and Android. An example of an app that works like this is Gmail -- when you go to attach a photo to an email, the native photo selector pops up on top of the app. The user can choose a photo and then will be returned to the app with their selected photo.

At first glance it looks like [react-native-image-picker](https://github.com/react-native-image-picker/react-native-image-picker) is a library that would work for this.

Todos:

- [x] Get native photo picker working on Android
- [x] Get native photo picker working on iOS
- [x] Ensure native photo picker is used for importing new photos to Group Photos *and* selecting photos for an existing observation (i.e. on Obs Edit)
- [x] Ensure native photo picker allows selection of multiple photos
- [ ] Check for compatibility issues on older devices. (For example, I believe iOS introduced a new[ Photos Picker](https://developer.apple.com/documentation/photokit/selecting_photos_and_videos_in_ios) that only works on iOS 15+, so we may need a different solution for phones with an older OS)
- [x] Make sure we can still get coordinates from exif data

Chrischuck:
@albullington I tried out `react-native-image-picker` but exif data doesn't work for ios. `expo-image-picker` might work?

Chrischuck:
I played around w/ this for a few hours and got to a point where I found something that should work moving forward. There's a ton of moving pieces in the `ObsEditProvider`, which I still haven't figured out.

But I threw up a draft PR #509 which I got working to some extent.

`react-native-image-picker` wasn't playing nice with `@bam.tech/react-native-image-resizer` and I couldn't figure out why for the life of me. I tried `expo-image-picker` and couldn't even get that working to begin with. 

Lastly tried out [react-native-image-crop-picker](https://github.com/ivpusic/react-native-image-crop-picker) which seems to work with `@bam.tech/react-native-image-resizer` just fine and can pull `exif` data from the get-go which is nice. I also think it can resize images from the start so `@bam.tech/react-native-image-resizer` might not even be needed.

I think for me, a lot of the code can be removed since we're now using a native image picker - but on my end I'm lost in the flow of the current image picker and how to translate it over.

https://user-images.githubusercontent.com/13311268/221426866-fab46138-e1b0-4d74-a6a8-72cbd2830cdd.mov

Chrischuck:
[...]

PR #509, opened 2023-02-26
Exploratory pr for native image picker

Chrischuck:
#431

- [x] Get native photo picker working on Android
- [x] Get native photo picker working on iOS
- [x] Ensure native photo picker is used for importing new photos to Group Photos and selecting photos for an existing observation (i.e. on Obs Edit)
- [x] Ensure native photo picker allows selection of multiple photos
- [x] Check for compatibility issues on older devices. (For example, I believe iOS introduced a new[ Photos Picker](https://developer.apple.com/documentation/photokit/selecting_photos_and_videos_in_ios) that only works on iOS 15+, so we may need a different solution for phones with an older OS)
- [x] Make sure we can still get coordinates from exif data

## android
https://user-images.githubusercontent.com/13311268/221446206-eaba9e27-26ed-414e-b9db-5fa8664a8438.mov
## ios 14.5
https://user-images.githubusercontent.com/13311268/221471649-c38718c5-6f71-414b-b0fe-678fe98d5c85.mov
## ios
https://user-images.githubusercontent.com/13311268/221446207-40c11c0a-b081-486b-a098-3a93629d5623.mov

### DIFF
--- a/src/components/AddObsModal.js
+++ b/src/components/AddObsModal.js
@@ -5,8 +5,8 @@ import { Text, View } from "components/styledComponents";
 import { t } from "i18next";
 import { ObsEditContext } from "providers/contexts";
 import * as React from "react";
+import ImagePicker from "react-native-image-crop-picker";
 import { IconButton, useTheme } from "react-native-paper";
-import ImagePicker from 'react-native-image-crop-picker';
 
 type Props = {
   closeModal: ( ) => void
@@ -33,35 +33,59 @@ const AddObsModal = ( { closeModal }: Props ): React.Node => {
     closeModal( );
   };
 
-  const navToPhotoGallery = async ( ) => {
-    const result = await ImagePicker.openPicker({
+  const navToPhotoGallery = async () => {
+    const selectedImages = await ImagePicker.openPicker({
       mediaType: "photo",
       multiple: true,
       maxFiles: 20,
       includeExif: true,
-      loadingLabelText: ""
-    }).then(images => images.map(image => ({
-        timestamp: parseInt(image.creationDate),
-        type: "image",
-        location: {
-          altitude: image.exif["{GPS}"].Altitude || 0,
-          longitude: image.exif["{GPS}"].Longitude || null,
-          latitude: image.exif["{GPS}"].Latitude || null,
-          heading: image.exif["{GPS}"].ImgDirection || 0,
-          speed: image.exif["{GPS}"].Speed || 0
-        },
-        image: {
-          width: image.width,
-          height: image.height,
-          filename: image.filename,
-          fileSize: image.size,
-          playableDuration: null,
-          uri: `file://${image.path}`
-        }
-    })));
+      loadingLabelText: "",
+    }).then((images) =>
+      images.map((image) => {
+        const gpsData = image.exif["{GPS}"];
+        const latitudeNegative = gpsData.LatitudeRef === "S";
+        const longitudeNegative = gpsData.LongitudeRef === "W";
 
-    const foo = await obsEditContext.createObservationFromCleanGallery(result[0])
-    navigation.navigate( "ObsEdit", { lastScreen: "PhotoGallery" } );
+        let latitude = gpsData.Latitude ?? null;
+        let longitude = gpsData.Longitude ?? null;
+
+        if (latitude && latitudeNegative) {
+          latitude *= -1;
+        }
+
+        if (longitude && longitudeNegative) {
+          longitude *= -1;
+        }
+
+        return {
+          timestamp: parseInt(image.creationDate, 10),
+          type: "image",
+          location: {
+            altitude: gpsData.Altitude || 0,
+            longitude,
+            latitude,
+            heading: gpsData.ImgDirection || 0,
+            speed: gpsData.Speed || 0,
+          },
+          image: {
+            width: image.width,
+            height: image.height,
+            filename: image.filename,
+            fileSize: image.size,
+            playableDuration: null,
+            uri: `file://${image.path}`,
+          },
+        };
+      })
+    );
+
+    if (selectedImages.length === 1) {
+      obsEditContext.createObservationFromGallery(selectedImages[0]);
+      navigation.navigate("ObsEdit");
+      closeModal();
+    }
+
+    
   };
 
   const navToSoundRecorder = ( ) => navAndCloseModal( "SoundRecorder" );
--- a/src/components/AddObsModal.js
+++ b/src/components/AddObsModal.js
@@ -6,6 +6,7 @@ import { t } from "i18next";
 import { ObsEditContext } from "providers/contexts";
 import * as React from "react";
 import { IconButton, useTheme } from "react-native-paper";
+import ImagePicker from 'react-native-image-crop-picker';
 
 type Props = {
   closeModal: ( ) => void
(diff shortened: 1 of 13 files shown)


src/components/PhotoImporter/PhotoGallery.js
```diff
@@ -185,6 +185,7 @@ const PhotoGallery = ( ): Node => {
       navToObsEdit( );
       return;
     }
+
     navigation.navigate( "GroupPhotos", { selectedPhotos } );
```
albullington, line 189: not sure if this is an issue specific to this PR, but I noticed that the number of observations never updates on the GroupPhotos screen when combining, separating, or deleting photos. And number of photos is also incorrect.

src/providers/ObsEditProvider.js
```diff
@@ -64,30 +63,20 @@ const ObsEditProvider = ( { children }: Props ): Node => {
   };
 
   const createObsPhotos = useCallback( async photos => Promise.all(
-    photos.map( async photo => ObservationPhoto.new( photo?.image?.uri ) )
+    photos.map( async photo => ObservationPhoto.new( photo?.image?.uri, undefined, true ) )
   ), [] );
 
   const createObservationFromGalleryPhoto = useCallback( async photo => {
-    const originalPhotoUri = photo?.image?.uri;
```
albullington, line 71: this code should not be removed -- we want to get exif data using these methods.

src/components/AddObsModal.js
```diff
@@ -32,7 +33,17 @@ const AddObsModal = ( { closeModal }: Props ): React.Node => {
     closeModal( );
   };
 
-  const navToPhotoGallery = ( ) => navAndCloseModal( "PhotoGallery" );
+  const navToPhotoGallery = async () => {
```
albullington, line 36: Is there a way to make the navigation experience smoother after closing the image picker? It seems a little glitchy to me that the modal closes and the user sees MyObservations with the AddObsModal still pulled up before navigating to the next screen.

albullington (changes requested):
This is looking pretty good to me. Main thing is that we need to make sure we're getting exif data natively, rather than through the image picker's JS code.

If we can make the navigation feel a little smoother when closing the modal, that would also be a nice improvement.

src/components/AddObsModal.js
Chrischuck: Not really sure on this one unless there's an intermediary page that's behind the image picker when it's up.

Also if `AddObsModal`, the call to `selectImagesFromGallery` gets destroyed. 

currently in inat ios, the behavior is similar but when the image picker closes, there is a loading state

src/sharedHelpers/selectImagesFromGallery.js
```diff
@@ -0,0 +1,84 @@
+import { Platform } from "react-native";
+import ImagePicker from "react-native-image-crop-picker";
+
+const handleIosImage = image => {
```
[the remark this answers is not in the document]
Chrischuck: Hmm, could explain more on the reliable/easy part?

`react-native-image-crop-picker` pulls the exif data natively, but there is no normalization layer, so the shape/attributes for ios/android simply come back differently. This js code just normalizes data  so the app can use it effectively.

I ask this because `native-exif-reader` isn't able to use the files from `react-native-image-crop-picker`, not exactly sure why

src/components/PhotoImporter/PhotoGallery.js
Chrischuck: Yes I noticed that too when I first worked on this. I wasn't exactly sure and didnt look into how they were calculated

src/sharedHelpers/selectImagesFromGallery.js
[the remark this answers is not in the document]
albullington: @kueda, any thoughts on this? I know Yaron spent a while making sure we can get exif data natively, but I'm not 100% certain why we need that or if that's a dealbreaker here for using `react-native-image-crop-picker`.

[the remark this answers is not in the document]
kueda: @budowski did this work as a part of #144. I believe we wrote our own EXIF library because we were unable to consistently extract `GPSHPositioningError` using existing react native libraries. @Chrischuck, [this image](https://user-images.githubusercontent.com/22691/187563722-4d2b4a4a-a9f4-45df-9ba4-595e36528e5a.jpg) should have that tag. Are you seeing it getting extracted with your code in this branch on *both* iOS and Android?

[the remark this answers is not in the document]
Chrischuck: @kueda looks like it isn't able to get that tag. If I recall, `react-native-image-picker` worked fine w/ exif reader but the resizer didn't like it. Might be easier to go that route and just find a different resizer library

Chrischuck:
@kueda @albullington

I gave `react-native-image-picker` a go and that did not work with the exif-reader. What I think is happening is `react-native-image-picker` creates a tmp file copy and loses the exif data. Unfortunately, it doesn't give you direct access to the original image. `react-native-camera-roll` on the other hand, points you to the original file.

I tried `expo-image-picker` which is able to get the exif data on ios for the ceanothus? moth. However it does not work on android. So far all of the image pickers that use native make a copy of the asset and lose exif data for our own reader.

src/sharedHelpers/selectImagesFromGallery.js
```diff
@@ -0,0 +1,84 @@
...
+    includeExif: true,
+    loadingLabelText: ""
+  };
+
+  if ( resize ) {
+    pickerParams.compressImageMaxWidth = 2048;
+    pickerParams.compressImageMaxHeight = 2048;
+  }
+
+  const selectedImages = await ImagePicker.openPicker( pickerParams );
```
kueda, line 79: I think this raises an exception that we need to handle if the user backs out of the image picker in Android.

src/components/AddObsModal.js
kueda: I agree this isn't ideal, but I think it's fine for now. We could have some empty intermediate as Chris suggested, but I think this is more a question for Abhas when this is ready for him to look at.

kueda (changes requested):
One minor requested fix, but the bigger issue is lack of access to the original so we can use our own exif parser and get at `GPSHPositioningError`. @Chrischuck do you want to using https://github.com/IceOfSummer/react-native-image-picker as instead of the main repo and seeing if you can get https://github.com/inaturalist/react-native-exif-reader working with the origina file path from that? If that works and that PR never gets merged, maybe it's worth maintaining our own fork.

Chrischuck:
@kueda I gave another forked repo a try that was supposed to give the original filepath, I forgot which one... But anyway I can give the one IceOfSummer fixed a shot. 

I'm also thinking it might be better to fork one of those repos and maintain under inaturalist, unfortunately, all of the maintenance on the photo pickers seems dead.

kueda:
> @kueda I gave another forked repo a try that was supposed to give the original filepath, I forgot which one... But anyway I can give the one IceOfSummer fixed a shot.

Cool.

> I'm also thinking it might be better to fork one of those repos and maintain under inaturalist, unfortunately, all of the maintenance on the photo pickers seems dead.

Maybe. https://github.com/react-native-image-picker/react-native-image-picker seems somewhat alive, though they don't seem to be reviewing PRs. If you can get that IceOfSummer fork to work with our exif parser, I think that's probably worth maintaining a fork for.

Chrischuck:
@kueda Will probably open up another pr for this. But I'll get to it saturday or sunday

Chrischuck:
@kueda unfortunately, IceOfSummer's fix only works for android. I tried to updating the code to get the original file path for ios, but I don't know enough on that side to get anything going...

I ended up having to fork IceOfSummer's repo and publishing to npm. Trying to install directly from github got some build issues because of TS.

kueda:
I'm going to close this b/c @budowski is working on a custom fork of that package and will do the work in this repo in a separate branch. Definitely appreciate the work here, though, @Chrischuck.

PR #511, opened 2023-03-01
Reverted icon names in Fontastic to match main (mostly)

kueda:
Somehow many of the icon names in Fontastic were changed. This changes them back to match what we use on the web and what we use in the main branch of this repo, with a few exceptions to attempt to make names more consistent.

Also updated the README in the hopes of avoiding situations like this in the future.

This only partially addresses #487 by addressing the naming problem. The sizing problem is pending the square icon files.


jtklein (approved):
I have added one not yet included icon from figma, just to test the readme steps, and this time for me everything worked as expected. Only the new icon was added to the glyphMap, no renaming occurred. So for me this is good to go with a minor comment where I was first confused about the wording.

PR #515, opened 2023-03-05
514 bandaid fix

### LINKED ISSUE #514: Android crashes on start up
**Describe the bug**
Not sure if it is only me, but Android always crashes unless I run `npm run clean-start`

After it runs once, if I even so much as look at the app wrong, it will crash. 

I narrowed it down to the `t` function in `ObsStatus` but I am perplexed on what is causing the issue.

Error from the app:
Screenshot 2023-03-04 at 8 36 34 PM

**To Reproduce**
1. `npm run android`
2. ???

 - Device: Pixel 6
 - OS: Android API 33

Chrischuck:
#514 

I definitely do not think this is the best solution, but I really don't know what's causing this crash.

### DIFF
--- a/src/components/Observations/ObsStatus.js
+++ b/src/components/Observations/ObsStatus.js
@@ -35,15 +35,15 @@ const ObsStatus = ( {
   const numIdents = observation.identifications?.length || 0;
   const numComments = observation.comments?.length || 0;
 
-  let identificationA11yLabel = ""
-  let commentA11yLabel = ""
+  let identificationA11yLabel = "";
+  let commentA11yLabel = "";
   try {
     // not exactly sure why this causes a consistent error every time you run android
     // for the first time...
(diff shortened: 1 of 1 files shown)


albullington:
Hmm. This does stop the crash in Android for me, but I'm also not sure why it's happening. Maybe certain Android devices are having trouble parsing numbers into a translation string immediately?

Chrischuck:
@albullington not sure exactly the issue, the error message is pretty vague

albullington:
I'm going to close without merging since we already applied this [bandaid fix](https://github.com/inaturalist/iNaturalistReactNative/blob/1a2b042aa1119aa3c9b9e317f0589e73aca2dacc/src/components/MyObservations/ObsStatus.js#L38) to the updated MyObservations code in PR #515

PR #518, opened 2023-03-10
Add icons to repo & use script to generate font

kueda:
This takes #487 in a different direction by ditching Fontastic all together in favor of the fantasticon npm package. This way all the fonts SVGs are checked into the repo and changes to fonts in branches don't have to interfere with other branches. Plus anyone with an SVG can update the fonts.


jtklein (approved):
There were a few usages for which there could be no icon found. I could replace most of them from the design specs, but for some I was not sure, those were:

`dotted-outline`: I have changed it to use `upload-saved`, however, @angielt  was splitting up the `arrow-up` part from the `dotted-circle` part into two icons because only the dotted outline should be animated and the centre arrow stay in place. The `arrow-up` icon now covers the centre arrow, so for adding this icon (`upload-saved` = `arrow-up-circle-dots.svg`) it could only be the dotted circle, I think.
`upload`: I have changed it to use `upload-arrow`, but it still did not show the icon. There was a mismatch in the alias value and the filename. I wasn't sure what the source of truth should be. `arrow-up.svg` is the file name (= glyph map), but in aliases was used `arrow-up-plain`. Have changed it in aliases to be like the filename.

I was taking a quick glance over the icons added and didn’t find any major things for now. Only, the trash.svg icon should be outlined according to the design specs.

I tested the scripts to generate the font by adding the “filter” icon from the design specs, because I could see that it is missing. When I ran the create icon script I got an error for missing gem on first try but `bundle install` installed the required dependencies and after that it worked well. It only gave me an error that the svg is not 24x24 size (is 20x20 in Figma and I could not change it's size), which I assume is supposed to happen that way. (I changed the icon to have a bigger box around to see if the scripts finish and they did, but did not commit because the size of paths in the icon were wrong then).

Have tested wit release variant on Android and iOS and compiled okay.

README.md
```diff
@@ -127,33 +127,13 @@ git commit -a -m "Updated translations"
...
-1. Add `inaturalisticons.ttf`, to `assets/fonts`.
-1. Clean build folders for iOS/Android with `rm -rf ios/build && rm -rf android/app/build`.
-1. Run `npx react-native-asset` to link the new assets.
-1. Add your icon somewhere in the app using `<INatIcon name="compass-rose" />` where name is set to the name of your new custom icon.
-1. Build the app on either platform, and you should see your custom icons.
-
-If anything goes wrong, check the [troubleshooting steps](https://github.com/oblador/react-native-vector-icons#troubleshooting) from `react-native-vector-icons` instructions on creating a glyph map.
+1. Add / edit SVGs to / in `src/images/icons/` (`git add` any new icons)
+1. `npm run icons`
+1. Rebuild the app (you'll have newly-linked assets that won't hot reload)
```
albullington, line 136: should we add some guidance here about how to use icons?

it still bothers me a bit that we're using icons within Pressable components in some places instead of using IconButtons everywhere, but maybe that's better for a separate cleanup issue.

albullington (approved):

README.md
kueda: I think I'd prefer a separate issue.

Issue #522, opened 2023-03-11
Tab bar icons not centered


Chrischuck:
**Describe the bug**
AddObs button is not centered

when user icon loads, there is a layout shift

**To Reproduce**
1. Log on + view tab bar

**Expected behavior**
Navbar icons should be evenly spaced + centered

**Screenshots**

<img width="727" alt="Screenshot 2023-03-11 at 5 46 47 PM" src="https://user-images.githubusercontent.com/13311268/224516621-3090297a-cf53-4ea0-809a-0a8a57c42566.png">

PR #523, opened 2023-03-11
Center tab bar icons

Chrischuck:
#522
Screenshot 2023-03-11 at 5 45 31 PM


jtklein (approved):
Looks good.
I have noticed that the bell icon in the nav bar is too big. It should be 32x32 as the hamburger menu is. Would you be able to update this change also here? If not I can merge in this one.

Chrischuck:
@jtklein Yes, I remember that from the spec, can do

Issue #517, opened 2023-03-09
Create box shadow on Android for custom header w/ ScrollView


albullington:
In PR [#508](https://github.com/inaturalist/iNaturalistReactNative/pull/508), we decided to punt on getting a working box shadow on Android. The shadow is currently working on iOS, but not Android.

We tried `react-native-fast-shadow` as a solution, but no luck. So this needs more exploration.

I suspect we might need a somewhat hacky solution like what we did for the footer in Seek -- using an image of a shadow, instead of trying to create a shadow with CSS.

Chrischuck:
Tested out `react-native-shadow-2` which I was pretty confident would work - but unfortunately it does not. So this might be a pretty rough bug

Chrischuck:
Figured out that it's because the context header is getting rendered before the screen, so the screen covers up the shadow. Similar to this [issue](https://stackoverflow.com/questions/49078735/react-native-zindex-not-taken-seriously-on-android) in an unrelated repo.

Unfortunately, we can't really re-arrange the dom tree since react-navigation takes care of it.

`elevation` and `zIndex` are being completely ignored here for some aggravating reason

Worst case scenario would be to just include the context header in the screen itself and have react-navigation render no header. That should work.

albullington:
Closed in #528

PR #528, opened 2023-03-12
Fix shadow on android context header

Chrischuck:
What a bug... #517 

Screenshot 2023-03-12 at 11 14 32 AM

See comments in code on what was happening + the solution. I believe the issue is more on react-navigation's side and seems very edge case.

This solution works but I think ideally I think it'd be better to remove the stack navigation all together, place all routes in the tab navigator, and then extend the options prop that the custom tab bar can consume to show/hide. It is a pain to get the the `options` prop, so that's the only real issue with that.


albullington:
Cool that you found a solution for this! I'm assuming we will eventually not need this on the `GroupPhotos` screen since we're wanting to replace that screen with the native picker, but fine to have that there for now.

Issue #530, opened 2023-03-14
ObsList loading spinner freezes on sync


Chrischuck:
**Describe the bug**
Green loading spinner freezes on state change. Should be a simple fix by switching animation from native to reanimated

**To Reproduce**
1. Create a few unsynced observations
2. sync them
3. loading spinner will freeze as upload progresses

**Expected behavior**
Spinner should continue to spin until `uploadInProgress` is false

**Screenshots**
<img width="322" alt="Screenshot 2023-03-14 at 1 25 22 PM" src="https://user-images.githubusercontent.com/13311268/225102215-8a02f453-d4c7-47d5-a36b-083ddcf827b4.png">

**Context (please complete the following information):**
 - App version: [e.g. 1.0.0]

**Additional context**

PR #531, opened 2023-03-14
530 fix freezing loading spinner

Chrischuck:
#530


jtklein (changes requested):
Thanks @Chrischuck, by design this icon should: "While observations are uploading, icon-sync rotates 180 degrees clockwise every 1 sec, with .5 sec of pause in between" could you update the animation? Otherwise, I have checked and it works.

Chrischuck:
@jtklein can do! was curious if the spinner arrows should face the other way? because it looks a little odd when the spin is opposite of the direction of the arrows

jtklein:
Yes, I have not seen that the first pass. The animation should be the other way. Counter-clockwise, in the direction the arrows are pointing. I think it was like this before as well.

Issue #534, opened 2023-03-14
UploadStatus progress circle is not synced to actual upload progress


albullington:
**Describe the bug**
The `UploadStatus` progress circle should display real upload progress, instead of an approximation. This means we'll probably need a way for the realm `Observation.uploadObservation` method to update the rest of the app on individual upload status. Right now, how we want to display progress is open to interpretation, but it might be something like:

1. Observation uploaded without evidence: 0.33 percent progress
2. Observation photos uploaded: 0.5 percent progress
3. Observation sounds uploaded: 0.66 progress

**Screenshots**
<img width="276" alt="Screen Shot 2023-03-14 at 1 56 14 PM" src="https://user-images.githubusercontent.com/35536439/225134251-d5c41406-c7c1-46b9-bfe7-5841c920caa4.png">

Chrischuck:
I think some of the logic from #529 can be transferred over. I also think the progress circle is out of sync w/ the sync button.

I can get a spike pr going that we can colab from there.

albullington:
Closed in #538

PR #538, opened 2023-03-15
Sync upload button w/ progress

Chrischuck:
#534

I figured using an event emitter was probably the easiest path forward here.

https://user-images.githubusercontent.com/13311268/225186226-35db86ba-b097-41cf-b481-239ed00fb2ae.mov

### DIFF
--- a/src/providers/ObsEditProvider.js
+++ b/src/providers/ObsEditProvider.js
@@ -55,7 +55,7 @@ const ObsEditProvider = ( { children }: Props ): Node => {
       increments => {
         setUploadProgress( currentProgress => {
           increments.forEach( ( [uuid, increment] ) => {
-            currentProgress[uuid] ??= increment;
+            currentProgress[uuid] ??= 0;
             currentProgress[uuid] += increment;
           } );
 
--- a/src/providers/ObsEditProvider.js
+++ b/src/providers/ObsEditProvider.js
@@ -4,9 +4,10 @@ import { activateKeepAwake, deactivateKeepAwake } from "@sayem314/react-native-k
 import { searchObservations } from "api/observations";
 import type { Node } from "react";
 import React, {
-  useCallback, useMemo, useState,
-   useEffect
+  useCallback, useEffect,
+  useMemo, useState
 } from "react";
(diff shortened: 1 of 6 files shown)


albullington (changes requested):
Neat, I like this approach. One thing I'm seeing is that the `cleanup( )` function in `useUploadObservations` clears the final status text that we want to display on the Toolbar (see screenshot below). I don't think we need that cleanup function when there are no more uploads -- instead, I added [a listener](https://github.com/inaturalist/iNaturalistReactNative/blob/c89fc837da5831f373696528b49f7bc3343de563/src/components/MyObservations/ToolbarContainer.js#L104) to the ToolbarContainer that should reset the Toolbar when a user leaves this screen.

But, I'm not picky about what approach we take as long as we display the correct status text to show a user that their uploads completed.

Screen Shot 2023-03-15 at 11 19 37 AM

albullington:
The completed progress bar is still there in the design, so I don't think we need to reset it until after a user leaves the screen. Here's the link to the [Figma screens](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=2964%3A48949&t=zOjVkWzeRWtKTXLr-0) for what the toolbar should look like when an upload is completed.

Chrischuck:
@albullington cool, reverted any changes on that side of things

src/providers/ObsEditProvider.js
```diff
@@ -47,6 +49,25 @@ const ObsEditProvider = ( { children }: Props ): Node => {
     setUnsavedChanges( false );
   }, [] );
 
+  useEffect( () => {
+    const progressListener = EventRegister.addEventListener(
+      "INCREMENT_OBSERVATIONS_PROGRESS",
+      increments => {
+        setUploadProgress( currentProgress => {
+          increments.forEach( ( [uuid, increment] ) => {
+            currentProgress[uuid] ??= 0;
```
albullington, line 58: @Chrischuck what is `currentProgress[uuid] ??= 0;` doing here? The `??=` operation seems to be causing e2e tests to fail, and I'm not sure why this line is necessary.

PR #552, opened 2023-03-28
Reverse geocode only when internet is connected

### LINKED ISSUE #390: Don't show warnings when reverse geocoding fails
**Describe the bug**
When reverse geocoding fails for some reason (e.g. you're offline), we should not show any kind of warning or error message.

**To Reproduce**
1. Put your device in airplane mode
2. Start a new observation without evidence and wait for the coordinates to fetch from GPS
3. You should see a Console Warning like `efjr: UNAVAILABLE: Unable to resolve host`

**Expected behavior**
Failure to reverse geocode is expected behavior while offline and should just happen silently. No need to imply to the user that something's wrong when it isn't. `place_guess` will be populated automatically when the observation is posted to the server.

**Screenshots**
Screenshot_20230124-175829

Screenshot_20230124-175837

 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.1.1 (37)

**Additional context**
If this is the kind of thing that wouldn't show up in a production build, I guess we can ignore it, but I'd rather not.

albullington:
Closes #390


src/sharedHelpers/fetchPlaceName.js
```diff
@@ -32,15 +33,12 @@ const setPlaceName = ( results: Array<Object> ) => {
...
-  try {
-    const results = await Geocoder.geocodePosition( { lat, lng } );
-    if ( results.length === 0 ) { return null; }
-    return setPlaceName( results );
-  } catch ( e ) {
-    console.warn( e, "couldn't fetch geocoded position with coordinates: ", lat, lng );
-    return null;
-  }
+const fetchPlaceName = async ( lat: ?number, lng: ?number ): any => {
+  const { isConnected } = await NetInfo.fetch( );
```
kueda, line 37: I think you want `isInternetReachable` here. The problem is that we need to know if we can connect to Google, which requires a network connection (`isConnected`), but that network has to be connected to the Internet (`isInternetReachable`).

kueda (changes requested):
Bug isn't fixed in Android for me, and for some reason it's hard to replicate on iOS, maybe due to location caching.

kueda (approved):

Issue #535, opened 2023-03-14
Improve performance of MyObservations


albullington:
This one is a bit experimental. There are some things on MyObservations that seem particularly slow to render, like switching between gridview and listview. The goal of this issue is to track down anything hindering performance (i.e. unnecessary rerenders, using FlatList instead of a more optimized list, etc.) and make everything feel less laggy.

Chrischuck:
I can take a look at adding flashlist to MyObservations

PR #553, opened 2023-03-29
Use flash list obs page

Chrischuck:
#535 

It seems to be snappier switching between grid/list view.

Only real issue I've had was on querying for more observations. I believe that is now working but something to check when testing this pr.


kueda (approved):

kueda:
Many thanks, @Chrischuck. Definitely a vast performance improvement.

PR #554, opened 2023-03-29
Navigation fixes / remove stack navigator

### LINKED ISSUE #541: Back button on species info does not go back to AddId
With the following navigation:

AddObsButton -> Camera -> Take a photo -> ObsEdit -> AddID -> any taxon info

If you then press the back button the app goes back to the observation list, and not back to observation flow (AddId) and the observation seems to be discarded as well (at least not reachable anymore from the observation list).

albullington:
Closes #541

- Removes stack navigator
- Moves project, identify, obs edit, add ID, standard camera, and sound recorder into tab navigator with tab bar hidden (on obs edit, add ID, standard camera, and sound)
- Allows a user to navigate from AddID to TaxonDetails and back


kueda (approved):
Bug fixed!

PR #559, opened 2023-03-30
474 adjust photopreview sizes in standardcamera for different breakpoints

angielt:


src/components/SharedComponents/PhotoCarousel.js
```diff
@@ -88,22 +96,40 @@ const PhotoCarousel = ( {
...
+            },
+            {
+              "m-[3px]": screenBreakpoint === ( "sm" || "md" ),
+              "m-[8.5px]": screenBreakpoint === ( "lg" || "xl" || "2xl" )
+            }
+          )}
         >
-          <View className="rounded-lg overflow-hidden">
+          <View
+            testID="ObsEdit.photo"
```
albullington, line 114: can we rename this testID to `PhotoCarousel.photo` here and in the tests?

albullington (changes requested):
a few small changes requested

albullington (changes requested):
@angielt, were you able to test this out on a real tablet? I built this on my iPad, and I'm seeing the breakpoint is returned as `sm` no matter whether I'm on a large device or a small one. It looks like `getBreakpoint` is expecting an initialWidth to be passed in, and we're not passing a width in from the `StandardCamera` screen. So that's one issue.

To fix that, I tried passing in initialWidth, and that still isn't working. Your `getBreakpoint` function is comparing a number to a string (like, the number 1180 for initial width is being compared with the string `1366px`), which will always return `sm`.

PR #563, opened 2023-03-31
561 fix hidden bottom sheet

Chrischuck:
Screenshot 2023-03-30 at 8 14 51 PM

**Have not tested the other usages of the bottom sheet component yet.**


albullington (changes requested):
Looks good, just need to resolve a couple conflicts with `main` and make sure that a user can tap the backdrop of LoginSheet, then see the LoginSheet pop up again when tapping the sync button or upload button. Right now, the LoginSheet disappears for good.

I'm also not sure if we should be using the `hide` prop here. I removed this in the DiscardChangesSheet in StandardCamera, since it seems to keep the bottom sheet rendered and pop it up when a user changes from portrait to landscape orientation or vice versa. Fine to leave it if this doesn't end up being an issue on the MyObservations screen -- maybe `BottomSheetModalProvider` helps with this.

Chrischuck:
@albullington Thanks for the review and apologies for getting to this a little late! But I went ahead and merged any conflicts, fixed tests, and made sure you could click on the modal back drop to close. I also checked all the other instances that either use our wrapped modal component or the BottomSheetModal in general to ensure they 1) still work and 2) appear over the tab bar if applicable.

Issue #465, opened 2023-02-11
StandardCamera buttons


kueda:
- [x] Add the `flash` ![flash](https://user-images.githubusercontent.com/22691/218228671-fd5f9fcc-25ae-4cbb-9d71-06587dfafedc.svg) and `flashoff` ![flashoff](https://user-images.githubusercontent.com/22691/218228673-5527fbc9-c4f3-4493-bbbc-a3f78001dd6f.svg) icons to the icon font and the flash button
- [x] Add the flip icon but call it `rotate` ![rotate](https://user-images.githubusercontent.com/22691/218228831-10c9f69e-fc21-4af8-9204-69da565626ac.svg) so it's more reusable
- [x] Style and positions all buttons according to [spec](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=552%3A4920&t=nNPVG46Bv29RJHaR-1) (sizing, opacity for buttons over camera area, shutter style, confirmation button position, etc.)

albullington:
@jtklein added these in #518

albullington:
@angielt are you actively working on this one? trying to figure out what's in progress

PR #567, opened 2023-04-07
472 standardcamera large screen layout

angielt:
includes #465


src/components/Camera/StandardCamera.js
```diff
@@ -167,105 +174,195 @@ const StandardCamera = ( ): Node => {
...
+    </>
+  );
+
+  const checkmarkClass = "bg-inatGreen rounded-full h-[40px] w-[40px] justify-center items-center";
+
+  const renderLargeScreenCameraOptions = () => {
+    const buttonSpacing = ( imageOrientation.toLowerCase() === "portrait" )
+      ? "space-y-[40px]" : "space-y-[32px]";
+    return (
+      // eslint-disable-next-line max-len
```
kueda, line 245: Instead of disabling the rule, you could do

```javascript
const cameraOptionsClassName = "bg-black/50 w-[40px] h-[40px] justify-center "
    + "items-center rounded-full";
```

or even better,

```javascript
const cameraOptionsClassName = [
    "bg-black/50",
    "h-[40px]",
    "items-center",
    "justify-center",
    "rounded-full",
    "w-[40px]"
  ].join( " " );
```

In general, try work with the linting rules instead of disabling them. They're almost all there for a reason, in this case to maintain the sanity of your fellow developers so we can all quickly and easily see all the values in this list of class names.

src/components/Camera/StandardCamera.js
```diff
@@ -167,105 +174,195 @@ const StandardCamera = ( ): Node => {
...
+      <PhotoPreview
+        photoUris={cameraPreviewUris}
+        setPhotoUris={setCameraPreviewUris}
+        savingPhoto={savingPhoto}
+        screenBreakpoint={breakpoint}
+      />
+      <View className="relative flex-1">
+        {device && <CameraView device={device} camera={camera} orientation={imageOrientation} />}
+        <FadeInOutView savingPhoto={savingPhoto} />
+        {breakpoint === "sm" || breakpoint === "md" ? renderSmallScreenCameraOptions()
```
kueda, line 314: The `breakpoint === "sm" || breakpoint === "md"` is repeated and its meaning is not immediately apparent, so you might want to pull that out into a variable so you're not repeating yourself and it's more readable, e.g. `const isSmallScreen = ["sm", "md"].includes( breakpoint );`.

Also, my simulated iPhone and my Pixel 3 both seem to have the `lg` breakpoint in portrait orientation, so they always get the large screen layout. Shouldn't we consider `lg` to be a small screen layout?

src/components/Camera/StandardCamera.js
```diff
@@ -167,105 +174,195 @@ const StandardCamera = ( ): Node => {
...
+            className={checkmarkClass}
+            onPress={navToObsEdit}
+            accessibilityLabel={t( "Navigate-to-observation-edit-screen" )}
             accessibilityRole="button"
+            accessibilityState={{ disabled: false }}
+            disabled={false}
           >
-            <Avatar.Icon
+            <INatIcon
+              name="checkmark"
```
kueda, line 285: This is causing a warning because the icon is named `check` not `checkmark`.

src/components/Camera/StandardCamera.js
```diff
@@ -167,105 +174,195 @@ const StandardCamera = ( ): Node => {
...
       testID = "flash-button-label-flash";
       accessibilityLabel = t( "Flash-button-label-flash" );
       break;
-    case "flash-off":
+    case "off":
+      name = "flash-off";
       testID = "flash-button-label-flash-off";
       accessibilityLabel = t( "Flash-button-label-flash-off" );
       break;
     default:
```
kueda, line 195: We're not going to render the flash button if the device doesn't have a flash, right? If that's the case, I think the `off` state is the more sensible default rather then potentially rendering a button with a bunch of empty props.

There's also a bug here causing the flash button not to appear on my Pixel 3. I think it's because `const devices = useCameraDevices( "wide-angle-camera" );` is asynchronous, so when these lines run for the first time:

```javascript
const device = devices[cameraPosition];
const hasFlash = device?.hasFlash;
const initialPhotoOptions = hasFlash ? { flash: "off" } : { };
const [takePhotoOptions, setTakePhotoOptions] = useState( initialPhotoOptions );
```

`takePhotoOptions` gets initialized as an empty object literal because `devices` hasn't been populated yet.

src/components/Camera/StandardCamera.js
```diff
@@ -167,105 +174,195 @@ const StandardCamera = ( ): Node => {
...
-      />
+        accessibilityState={{ disabled: false }}
+      >
+        <INatIcon
+          name={name}
+          color={colors.white}
+          size={20}
+        />
+
+      </Pressable>
```
kueda, line 213: Why not use `<IconButton>` here? In general I think we should prefer it so we can rely on the button functionality from react-native-paper instead of trying to recreate it. I'm sure there are some situations where we can't use that, though, so I'm curious why this might be one of them.

src/components/Camera/StandardCamera.js
```diff
@@ -44,7 +48,7 @@ const StandardCamera = ( ): Node => {
     evidenceToAdd,
     setEvidenceToAdd
   } = useContext( ObsEditContext );
-  const theme = useTheme( );
+  // const theme = useTheme( );
```
kueda, line 51: Let's just delete this instead of leaving it commented out.

kueda (changes requested):
Not quite a comprehensive review, but still plenty to work on.

In addition to the inline comments, the photo preview bar is still at the top of the screen when my iPad is in landscape orientation. This happens for me in the simulator as well as on a device, so you should be able to replicate it with

```bash
npx react-native run-ios --simulator "iPad Air (5th generation)"
```

kueda (approved):

PR #570, opened 2023-04-12
ObsEdit  - bottom sheets

albullington:
This PR adds the bottom sheets from Figma and updates some of the styling in Obs Edit.

There are currently a few failing tests that I haven't been able to figure out. It looks like a combo of a jest timeout plus a need for i18n translation in the tests.

Seeing this error: `Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout`


src/components/ObsEdit/Sheets/GeoprivacySheet.js
```diff
@@ -0,0 +1,80 @@
+// @flow
+
+import CheckBox from "@react-native-community/checkbox";
```
kueda, line 3: This looks nice, but it's off spec and I don't think it's necessary given https://callstack.github.io/react-native-paper/docs/components/RadioButton/ (which is also closer to the designs). Did you try the RN paper version?

kueda (changes requested):
Looking a lot better, but still some stuff to do.

Functionality:

- [ ] Getting ` WARN  Please report: Excessive number of pending callbacks: 501. Some pending callbacks that might have leaked by never being called from native code` when I edit an existing observation downloaded from the server, e.g. sign in as a user with existing observations, tap on one, tap the pencil button in the upper right, the app locks up for a second or two, and you eventually get that warning when ObsEdit loads. Weirdly this might only happen the first time I try this.
- [x] Warning triangle icon appears next to "EVIDENCE" header when I edit any existing remote observation that has a photo
- [x] When I make a new obs by importing a photo, I get the red warning triangle next to EVIDENCE
- [x] When I close a bottom sheet by tapping the faded background, that bottom sheet appears again when I open another bottom sheet. E.g. tap geoprivacy to bring up that bottom sheet, tap the faded background to close, then tap "Organism is wild." The "WILD STATUS" bottom sheet appears, but the "GEOPRIVACY" bottom sheet appears behind it
- [x] I can't seem to close a bottom sheet by tapping on the background in Android
- [x] I get the "LOCATION TOO IMPRECISE" bottom sheet when a try to save a remote observation where `positional_accuracy` is null
- [x] If I make an obs without evidence, the EVIDENCE header has a green check next to it. Shouldn't it have the red warning triangle? Not sure how to test the missing evidence bottom sheet without this
- [x] Need some more whitespace between the text and the bottom in the LOCATION TO IMPRECISE bottom sheet
- [x] When I create a new obs with a photo and tap the back button in the header, I don't get the "discard" warning bottom sheet

Implementation (I wasn't too thorough but did notice these):

- [x] `ImreciseLocationSheet` and `MissingEvidenceSheet` are very similar. Could their shared code be consolidated into a wrapper or a superclass?
- [x] Same with `GeoprivacySheet` and `WildStatusSheet` (and maybe `MissingEvidenceSheet`, though I didn't figure out how to view that)

src/components/SharedComponents/Sheets/TextInputSheet.js
```diff
@@ -0,0 +1,84 @@
...
+      }}
+    >
+      <View className="p-5">
+        <TextInput
+          accessibilityLabel="Text input field"
+          keyboardType="default"
+          multiline
+          mode="flat"
+          onChangeText={text => setInput( text )}
+          value={input}
```
kueda, line 52: Should this be `initialValue`? Not sure if the previous notes UI was working very well, but this controlled value approach is resulting in my text getting overwritten a lot.

src/components/ObsEdit/Sheets/NotesSheet.js
```diff
@@ -0,0 +1,41 @@
...
+      initialInput={description}
+      confirm={textInput => {
+        addNotes( textInput );
+        handleClose( );
+      }}
+    />
+  );
+};
+
+export default NotesSheet;
```
kueda, line 41: Does this component need to exist? I feel like you've done such a good job abstracting things into the reusable `<TextInputSheet>` that you could just use that directly inside `<OtherDataSection>`... unless you think we're going to reuse `NotesSheet` with exactly the same configuration somewhere.

src/i18n/l10n/en.ftl
```diff
@@ -879,5 +924,19 @@ Failed-to-log-in = Failed to log in
 # Generic error message
 Something-went-wrong = Something went wrong.
 
+#Geoprivacy sheet descriptions
+Anyone-using-iNaturalist-can-see = Anyone using iNaturalist can see where this species was observed, and scientists can most easily use it for research.
+The-exact-location-will-be-hidden = The exact location will be hidden publicly, and instead generalized to a larger area. (Threatened and endangered species are automatically obscured).
+The-location-will-not-be-visible = The location will not be visible to others, which means it may be difficult to identify.
+
+#Wild status sheet descriptions
```
kueda, line 932: FYI, comments in FTL refer only to the following line and are meant to provide context to translators just for the following string, so try to make sure comments a) refer to the following string only, and b) help translators translate that string.

As tempting as it is to try and group translations logically, there is no grouping that is logical to everyone, so I'd just add new translations to the end of the file for now. Ultimately I think we should alphabetize them, but I think it makes more sense to automate that in a pre-commit hook (will require parsing the FTL so we can keep the comments in the right places).

jtklein: There is a mention of `##` for block comments in the fluent docs.

Issue #571, opened 2023-04-14
UploadStatus: excessive callbacks


albullington:
**Describe the bug**
When navigating from MyObservations to a new ObsEdit screen with no evidence, there is sometimes a console warning. According to the stack trace, it looks like this has to do with the animation in UploadStatus, which is likely still rendering after a user leaves the screen.

`WARN  Please report: Excessive number of pending callbacks: 501. Some pending callbacks that might have leaked by never being called from native code`

**Screenshots**
<img width="367" alt="Screenshot 2023-04-14 at 3 08 47 PM" src="https://user-images.githubusercontent.com/35536439/232163016-e5eb19fa-649c-422f-804d-b98342c317cb.png">
<img width="371" alt="Screenshot 2023-04-14 at 3 09 12 PM" src="https://user-images.githubusercontent.com/35536439/232163063-3b05b611-71df-45a8-a40c-06240c15f50a.png">

**Additional context**
I believe Ken-ichi experienced the same bug with [slightly different repro steps](https://github.com/inaturalist/iNaturalistReactNative/pull/570#pullrequestreview-1384426546). Fwiw, we're both trying this from the `obs-edit-bottom-sheets` branch, but I think it's worth an issue since it seems to be coming from `UploadStatus`, not any of the new code in this branch.

Chrischuck:
I can take a look at this

albullington:
Haven't been able to replicate this in [v0.19.0+67](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.19.0%2B67)

PR #573, opened 2023-04-18
543 upload status UI is not updating while uploading

angielt:


jtklein (changes requested):
I have started the review of this PR but have seen that now the uploading animation (= the dots rotating around the arrow pointing up) is animated even while the observation is not yet uploaded. In other words, I have saved a new observation, navigate back to the list, and the animation is already running before I hit upload. The animation should only be running while the observations are uploading.

src/components/SharedComponents/UploadStatus/UploadStatus.js
```diff
@@ -37,18 +37,19 @@ const UploadStatus = ( {
...
+  } );
 
-  // const rotate = {
-  //   transform: [
-  //     {
-  //       rotate: interpolateRotating
-  //     }
-  //   ]
-  // };
+  const rotate = {
```
Chrischuck, line 45: Hi @angielt, I did not mean to step on your toes on this one, but I was checking out #571 and I _think_ it was caused by the following:

```
  Animated.loop(
    Animated.timing( rotateAnimation, {
      toValue: 1,
      duration: 10000,
      easing: Easing.linear,
      useNativeDriver: true
    } )
  ).start( () => {
    rotateAnimation.setValue( 0 );
  } );
  ```
  
 Not entirely sure though, as I couldn't repro the issue. In any case I make a draft PR switching out from React Native's animation library to reanimated. Reason being the spinning dots animation will most likely freeze once uploads start and the JS thread clogs. The other being to hopefully fix #571 

But you can take a look at that PR and if you decide to go with it, I don't mind and can close mine!

angielt: No worries, I had thoughts of potentially switching to use the reanimated library. In this case I think #571 was because I forgot to add a stop animation call.

Chrischuck: @angielt ahh that makes sense

PR #575, opened 2023-04-18
Swap loading animations to reanimated

Chrischuck:
https://user-images.githubusercontent.com/13311268/232908870-4d27ecbf-a1f9-4eb0-8fd6-03cb03828399.mov

### DIFF
--- a/src/components/MyObservations/ObsUploadStatus.js
+++ b/src/components/MyObservations/ObsUploadStatus.js
@@ -12,7 +12,6 @@ import UploadCompleteAnimation from "./UploadIcons/UploadCompleteAnimation";
 
 type Props = {
   observation: Object,
-  uploadStatus: Object,
   layout?: "horizontal" | "vertical",
   white?: boolean,
   classNameMargin?: string,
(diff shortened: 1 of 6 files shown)


src/components/MyObservations/ObsUploadStatus.js
```diff
@@ -33,11 +31,10 @@ const ObsUploadStatus = ( {
   const startSingleUpload = obsEditContext?.startSingleUpload;
   const uploadProgress = obsEditContext?.uploadProgress;
   const wasSynced = observation.wasSynced( );
-  const { allObsToUpload } = uploadStatus;
   const whiteColor = white && theme.colors.onPrimary;
 
   const displayUploadStatus = ( ) => {
-    if ( allObsToUpload?.find( upload => upload.uuid === observation.uuid ) ) {
+    if ( !observation.id ) {
```
Chrischuck, line 37: Unrelated to #571 but there is a moment when `ObsStatus` shows right when you add an unuploaded observation.

I believe unuploaded observations do not have an ID (but have a uuid). So it's faster to check for id truthiness and thus gets rid of the flash of ObsStatus.

However, if this isn't a reliable way to detect if an Observation has been uploaded, then I will revert.

Issue #396, opened 2023-01-26
Map does not show in Android on ObsDetail page


Chrischuck:
**Describe the bug**

On android, it appears the map does not show in the ObsDetail page. This was brought up in PR #385 

**To Reproduce**

1. Select an observation
2. Click on "Data" tab
3. See that the map shows on ios but not android

**Expected behavior**
The map should show

**Screenshots**
<img width="775" alt="Screenshot 2023-01-26 at 3 46 56 PM" src="https://user-images.githubusercontent.com/13311268/214957833-c350bdf2-73e3-42ee-bef2-2949a603d8cc.png">

**Context (please complete the following information):**
 - Device: Android Pixel 3
 - App version: [e.g. 1.0.0]

**Additional context**

Poking around, I think this [line](https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/components/SharedComponents/Map.js#L30)

should be 
```
  if ( !initialLatitude || !initialLongitude ) {
```

I don't have a google maps api key, but when I make that change an error about not having an api key comes up. So I think it's going in the right direction.

However, I think there's a deeper issue in that the `useUserLocation` hook only gets the location for iOS, which is why `latLng` is null in android.

kueda:
Not happening in [v0.17.1+64](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.17.1%2B64)

PR #593, opened 2023-04-25
Add location picker to ObsEdit

albullington:
- Also gets GoogleMaps working, as long as you have the API key, which is needed for #396


albullington:
@jtklein any idea why the e2e-Android test isn't passing here? I tried a couple different ways of exposing the GMAPS_API_KEY secret to AndroidManifest.xml, because that seemed to be the original problem.

jtklein:
I updated the "Create .env file" step to use printf to print the Google Maps key into the newly created .env file in the GitHub Action runner. It is using the same key as in env.example. It is building now, but I did not check if the actual key is used correctly. Since it is building we know that a value is present at not null, but it could be an empty string. Anyway, the emulator we are using in the runner has no Play Services enabled, so it will not be able to show the map as it is currently because it requires Play Services to be present on the device. At least that is what I see when I try to run the app on the same kind of emulator locally.

PR #604, opened 2023-05-04
Use native zoom drivers for vision-camera

### LINKED ISSUE #467: Tap to set focus/exposure should work even when zoomed in
Currently when you zoom in and tap the camera area to set the focus, the box that appears is not where you tapped, and I'm not really sure what point is being used to set focus and exposure. When you tap the camera view, it should set the focus and exposure to the point that you tapped, and the box that appears should have its center at the point where you tapped, regardless of whether your zoomed in or not.

### LINKED ISSUE #555: GestureDetector not working in camera on iOS
The GestureDetector should allow pinching and tapping gestures. This is currently working on Android, but not iOS.

We'll probably need to fix this before tackling [#467](https://github.com/inaturalist/iNaturalistReactNative/issues/467)

jtklein:
I was starting to work on this thinking that it will fix #555, however, it did not.
The work on the native drivers let me to find the underlying reason why #555 exists though, so I fixed it as well here.
So, this fixes #555.
This also fixes #467.


kueda:
Is double tap to zoom working for you here? It's not for me in Android, at least.

jtklein:
No, I have removed double-tap to zoom because it is neither in the Figma designs (only pinch to zoom and tap to focus) nor in the native apps (Android and iOS).

kueda:
> No, I have removed double-tap to zoom because it is neither in the Figma designs (only pinch to zoom and tap to focus) nor in the native apps (Android and iOS).

Thoughts, @abhasinat and @budowski? I guess it's not a convention in iOS, but I suspect it is in Android. I've certainly always relied on it when taking pictures one-handed.

kueda (approved):
Definitely fixes both bugs and I guess the need to for double tap to zoom is up for debate, so I think this is good to merge.

abhasinat:
I'm neutral about having double tap to zoom since I don't have it/don't use it on iOS. If you think it's important, @kueda, I'm fine with including it.

PR #605, opened 2023-05-04
Obs detail: Details Tab and Activity Tab styling 

angielt:
- Styling of Details tab and Activity tab of ObsDetail to match design
- PhotoScroll button icons added and back button 

(activity and details tab bar sticky header not in this PR)


src/components/ObsDetails/ActivityItem.js
```diff
@@ -11,6 +11,7 @@ import { t } from "i18next";
 import _ from "lodash";
 import type { Node } from "react";
 import React from "react";
+import { Divider } from "react-native-paper";
```
jtklein, line 14: There is a custom-styled Divider component in SharedComponents.

angielt: Oh I didn't realize. Thanks

src/components/ObsDetails/Attribution.js
```diff
@@ -13,25 +14,25 @@ type Props = {
...
 
   const renderLicenseCode = ( ) => {
     if ( !licenseCode ) {
       return t( "all-rights-reserved" );
     } if ( licenseCode === "cc0" ) {
-      return t( "no-rights-reserved" ) + licenseCode;
+      return `${t( "no-rights-reserved" )} ${licenseCode}`;
     }
-    return t( "some-rights-reserved" ) + licenseCode;
+    return `${t( "some-rights-reserved" )} ${licenseCode}`;
```
kueda, line 26: This entire string needs to be localized because the ordering might different in different locales, so it should look something like `t( "attribution_cc_by_nc" )`.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+  const [locationKebabMenuVisible, setLocationKebabMenuVisible] = useState( false );
+
+  const displayQualityGradeOption = option => {
+    const qualityGrade = observation?.quality_grade;
+    const labelClassName = ( qualityGrade === option ) ? "font-bold" : "";
+
+    return (
+      <View className="flex-col space-y-[8px]">
+        <QualityGradeStatus qualityGrade={option} />
+        <Body4 className={labelClassName}>{_.startCase( _.camelCase( t( option ) ) )}</Body4>
```
kueda, line 46: This needs to be under translator control. In general, if you find yourself using `startCase()` or `toUpperCase()` on text the user sees, you probably should be globalizing it instead, so here you probably want something like

```javascript
switch ( option ) {
  case "research":
    return t( "quality_grade_Research" );
  case "needs_id":
    return t( "quality_grade_Needs_ID" );
  default:
    return t( "quality_grade_Casual" );
}
```

Note that the keys provide context (we're talking about "research" the quality grade, not "research" the second person imperative verb), and that they specify the desired case, which translators can override if they want.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+    );
+  };
+
+  return (
+    <>
+      {observation.description && (
+        <>
+          <View className={sectionClass}>
+            <Heading4 className={headingClass}>{t( "NOTES" )}</Heading4>
+            <Body2>{observation.description}</Body2>
```
kueda, line 57: Use the `<UserText>` component here. Users can put all kinds of markup in observation descriptions.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+      <View className="flex-row justify-between items-center mt-[8px] mx-[15px]">
+        <Heading4>{t( "LOCATION" )}</Heading4>
+        <KebabMenu
+          visible={locationKebabMenuVisible}
+          setVisible={setLocationKebabMenuVisible}
+        >
+          <Menu.Item
+            onPress={async ( ) => {
+
+            }}
```
kueda, line 72: Please don't add methods that do nothing, especially interactive ones. At least add a toast or an alert that says "TODO" so people testing the app know it isn't broken.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+          />
+        </KebabMenu>
+      </View>
+      {isOnline
+        ? (
+          <Map
+            obsLatitude={observation.latitude}
+            obsLongitude={observation.longitude}
+            mapHeight={230}
+          />
```
kueda, line 89: Doesn't need to happen here or block, but this map will need to handle obscured coordinates, e.g. showing when coordinates are obscured, showing the true coordinates if you can see them.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+          <Menu.Item
+            onPress={async ( ) => {
+
+            }}
+            title={t( "Copy-coordinates" )}
+          />
+        </KebabMenu>
+      </View>
+      {isOnline
```
kueda, line 83: Is this necessary? In the native Android app the maps SDK has cached tiles so the maps sometimes work while offline.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+      <View className={`mt-[11px] ${sectionClass}`}>
+        <ObservationLocation observation={observation} details />
+      </View>
+      <Divider />
+
+      <View className={sectionClass}>
+        <Heading4 className={headingClass}>{t( "DATE" )}</Heading4>
+        <DateDisplay
+          classNameMargin="mb-[12px]"
+          label={t( "Observed" )}
```
kueda, line 109: This could use a more descriptive key, like `date_observed_header` or `date_observed_header_short` or something. Otherwise translators won't know that this refers to a date and that it's a header and not some piecemeal part of a sentence.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+          dateString={checkCamelAndSnakeCase( observation, "timeObservedAt" )}
+        />
+        <DateDisplay
+          label={t( "Uploaded" )}
```
kueda, line 113: Same here.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+      <View className={`${sectionClass} flex-col`}>
+        <Heading4 className={headingClass}>{t( "DATA-QUALITY" )}</Heading4>
+        <View className="space-y-[15px]">
+          <View className="flex-row justify-around">
+            {displayQualityGradeOption( "casual" )}
+            {displayQualityGradeOption( "needs_id" )}
+            {displayQualityGradeOption( "research" )}
+          </View>
+          <Body4>
+            {t( "This observation needs more identifications to reach research grade" )}
```
kueda, line 128: I don't think we can use full sentences with spaces as keys, also I don't see this in strings.ftl.

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+
+      <Divider />
+
+      <View className={`${sectionClass} space-y-[11px]`}>
+        <Heading4 className={headingClass}>{t( "OTHER-DATA" )}</Heading4>
+        <Attribution observation={observation} />
+        {application && (
+          <View className="flex-row">
+            <Body4>{t( "Uploaded-via" )}</Body4>
+            <Body4>{` ${application}`}</Body4>
```
kueda, line 148: These two things need to be one string, like `t( "Uploaded_via_application", { application } )` because translators need control over word ordering (e.g. for RTL languages).

src/components/ObsDetails/DetailsTab.js
```diff
@@ -0,0 +1,157 @@
...
+          </View>
+        )}
+        <Body4>{t( "View in Browser" )}</Body4>
```
kueda, line 151: Missing from strings.ftl.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -159,6 +165,25 @@ const ObsDetails = (): Node => {
...
   };
 
+  const createIdentificationMutation = useAuthenticatedMutation(
+    ( idParams, optsWithAuth ) => createIdentification( idParams, optsWithAuth ),
+    {
+      onSuccess: data => setIds( [...ids, data[0]] ),
+      onError: e => {
+        let error = null;
+        if ( e ) {
+          error = t( "Couldnt-create-identification", { error: e.message } );
```
kueda, line 175: This key doesn't accept a variable, so you either need to alter it or make a different key.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -159,6 +165,25 @@ const ObsDetails = (): Node => {
...
+        } else {
+          error = t( "Couldnt-create-identification", { error: t( "Unknown-error" ) } );
```
kueda, line 177: If possible, try to avoid using translation strings with variables as they are almost always harder to translate. Something like `t( "Couldnt-create-identification-unknown-error" )`. It will feel a bit repetitive, but it keeps translators happy.

src/components/ObsDetails/ObsDetails.js
```diff
@@ -293,28 +351,37 @@ const ObsDetails = (): Node => {
...
+            size={25}
+            onPress={() => faveOrUnfave()}
+            textColor={colors.white}
+            className="absolute bottom-3 right-3"
+            accessible
+            accessibilityRole="button"
+            accessibilityLabel={t( "favorite" )}
+          />
+          <View className="absolute bottom-3 left-3">
+            <PhotoCount count={photos?.length} />
```
kueda, line 372: The `count` prop has to be a number, so either give it a default of 0 here or change the component to accept `undefined`.

src/components/SharedComponents/ObservationLocation.js
```diff
@@ -8,40 +8,73 @@ import useTranslation from "sharedHooks/useTranslation";
...
+  if ( !displayLocation ) {
+    displayLocation = t( "No-Location" );
+  }
+  if ( ( observation?.latitude !== null && observation?.latitude !== undefined )
     && ( observation?.longitude != null && observation?.longitude !== undefined )
   ) {
-    displayLocation = `${observation.latitude}, ${observation.longitude}`;
-  } else if ( !displayLocation ) {
-    displayLocation = t( "Missing-Location" );
+    displayCoords = `Lat: ${observation.latitude}, Lon: ${observation.longitude}`;
```
kueda, line 28: Needs to be globalized.

kueda (changes requested):
Most of these are i18n related, but I guess there's a lot of text on this view.

Maybe unrelated, but I'm getting an infinite series of `[TypeError: Network request failed]` when I view ObsDetail while offline.

src/components/ObsDetails/ObsDetails.js
angielt: Would `t( "Couldnt-create-identification", { error: e.message } );` be replaced with just `t( "Couldnt-create-identification")`? Or do we want strings for each error message?

src/components/ObsDetails/DetailsTab.js
angielt: Thanks for the heads up. I'll keep that in mind when I work on the map.

src/components/ObsDetails/DetailsTab.js
angielt: you're right. I've added the strings to strings.ftl. Only the needs-id string was included in the design docs so I added some placeholder strings for research and casual grade for now.

kueda (approved):

PR #608, opened 2023-05-12
Media Viewer with landscape UI

albullington:
Overhauled the media viewer. This is mostly ready to go. There's one part that I know is still buggy: when rotating the device, the FlatList does not always render the items correctly at the new size. I'm not sure what the best way to handle this is, so punting for another issue. I think our options are:

- find a way for FlatList to detect device rotation and rerender the list items at the appropriate size
- remove FlatList and use a gesture handler directly on the CustomImageZoom to detect panning left and right (I suspect this will be less buggy than using FlatList)


src/components/MediaViewer/MediaViewer.js
```diff
@@ -0,0 +1,215 @@
...
+        {
+          "left-[19px]": isTablet,
+          "left-[5px]": !isTablet
+        }
+      )}
+      >
+        <PhotoSelector
+          photoUris={photoUris}
+          scrollToIndex={scrollToIndex}
+          isLargeScreen={isTablet}
```
kueda, line 193: This is resulting in tiny icons on all phones, even larger phones. The designs specify a breakpoint / screensize approach.

src/components/MediaViewer/MediaViewer.js
```diff
@@ -0,0 +1,215 @@
...
+          <IconButton
+            onPress={handleArrowPressLeft}
+            icon="chevron-left-circle"
+            iconColor={colors.white}
+            size={26}
+          />
+        </View>
+      )}
+      {!atLastPhoto && (
+        <View className="absolute top-1/2 -mt-20 right-0">
```
kueda, line 173: I think here and for the other button you want `-mt-[20px]` right? `-mt-20` is the equivalent of `-mt-[80px]` I think.

src/components/MediaViewer/MediaViewer.js
```diff
@@ -0,0 +1,215 @@
...
+          <IconButton
+            onPress={handleArrowPressRight}
+            icon="chevron-right-circle"
+            iconColor={colors.white}
+            size={26}
+          />
+        </View>
+      )}
+      <View className={classnames(
+        "absolute bottom-[18px]",
```
kueda, line 183: Why does this need to be absolutely positioned? Is there a case where the `MainPhotoDisplay` needs to flow under the `PhotoSelector`?

kueda (changes requested):
Most of these are pretty minor. FWIW, I'm not seeing the issue with device rotation you described.

Here's one that didn't fit as an inline comment:

- [x] When viewing an observation that wasn't created on device from ObsEdit, it shows the square URL, so it looks pixelated. It should show medium or large size

src/components/MediaViewer/MediaViewer.js
```diff
@@ -0,0 +1,215 @@
...
+import useDeviceOrientation from "sharedHooks/useDeviceOrientation";
+import useTranslation from "sharedHooks/useTranslation";
+import colors from "styles/tailwindColors";
+
+import MainPhotoDisplay from "./MainPhotoDisplay";
+import PhotoSelector from "./PhotoSelector";
+
+export const PORTRAIT = "portrait";
+export const LANDSCAPE_LEFT = "landscapeLeft";
+export const LANDSCAPE_RIGHT = "landscapeRight";
```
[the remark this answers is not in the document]
albullington: whoops, I moved these into `useDeviceOrientation` as a shared hook so we could reuse the hook in the camera if needed, but I forgot to remove these lines.

Issue #606, opened 2023-05-11
ERROR Warning: Cannot update a component (`MyObservationsContainer`) while rendering a different component (`ObsDetails`)


jtklein:
This came up during #574:
Happens after getting an update about an unviewed comment and tapping on that observation:

```
 ERROR  Warning: Cannot update a component (`MyObservationsContainer`) while rendering a different component (`ObsDetails`). To locate the bad setState() call inside `ObsDetails`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    in ObsDetails (created by SceneView)
```

When I got to ObsDetails, the new comment did not appear (seems inconsistent).

The problematic `setState` is happening in this call: `setObservationList( [...collection] );` in a listener here `localObservations.addListener`.

In ObsDetail we are updating the observation in realm which triggers this listener that is attached in MyObs Container.

Chrischuck:
Think I can give this a shot

PR #609, opened 2023-05-14
606 - Fix render blocking state change

Chrischuck:
#606 

https://github.com/inaturalist/iNaturalistReactNative/assets/13311268/418bb9d2-263d-418b-a43e-c5b5e4588d5c

My first pass at a fix was to make the `MyObservations` page unmount on blur, while it worked, it lost the scroll position.

Second pass around, I used a ref to prevent any downstream rerenders and would only rerender if `MyObservations` (or any consuming user of that hook) is focused. Let me know what you think of this solution as I don't think it's particularly straightforward so I left a comment about it in the code.

On another note, I think we should default to pages unmounting on blur unless there's a specific reason not to. I feel most of the pages are lightweight enough that initial render isn't that expensive and it should save on memory usage.

### DIFF
--- a/src/components/SharedComponents/UserText.js
+++ b/src/components/SharedComponents/UserText.js
@@ -130,4 +130,5 @@ const UserText = ( {
   );
 };
 
+// Memoize to prevent excessive re-renders when HTML component is in a list
 export default React.memo( UserText, ( oldProps, newProps ) => isEqual( oldProps, newProps ) );
--- a/src/components/SharedComponents/UserText.js
+++ b/src/components/SharedComponents/UserText.js
@@ -6,6 +6,7 @@ import { Platform, useWindowDimensions } from "react-native";
 import HTML, { defaultSystemFonts } from "react-native-render-html";
 import WebView from "react-native-webview";
 import sanitizeHtml from "sanitize-html";
+import { isEqual } from 'lodash';
 
 const ALLOWED_TAGS = ( `
   a
@@ -130,4 +131,6 @@ const UserText = ( {
   );
 };
 
-export default UserText;
+export default React.memo(UserText, (oldProps, newProps) => {
+  return isEqual(oldProps, newProps)
+});
--- a/src/components/SharedComponents/UserText.js
+++ b/src/components/SharedComponents/UserText.js
@@ -130,4 +130,4 @@ const UserText = ( {
   );
 };
 
-export default React.memo( UserText );
+export default UserText;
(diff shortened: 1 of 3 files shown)


src/components/SharedComponents/UserText.js
```diff
@@ -130,4 +130,4 @@ const UserText = ( {
   );
 };
 
-export default UserText;
+export default React.memo(UserText);
```
Chrischuck, line 133: I was getting:

```
You seem to update props of the "TRenderEngineProvider" component in short periods of time, causing costly tree rerenders (last update was 59.31ms ago). See https://stackoverflow.com/q/68966120/2779871
```

Adding a memo seems to fix it

src/sharedHooks/useLocalObservations.js
```diff
@@ -1,14 +1,22 @@
...
 } from "react";
 import Observation from "realmModels/Observation";
 
 const { useRealm } = RealmContext;
 
 const useLocalObservations = ( ): Object => {
+  const isFocused = useIsFocused( );
+  // use a ref as a temporary store because MyObservations page doesn't unmount on blue
+  // only updating state when focused will help prevent MyObservations from
+  // blocking other components from rendering
```
kueda, line 17: How about something like, "Use refs to maintain state without triggering re-renders of hook consumers when they have lost focus, which prevents other views from rendering when they have focus."

src/components/SharedComponents/UserText.js
```diff
@@ -130,4 +130,4 @@ const UserText = ( {
   );
 };
 
-export default UserText;
+export default React.memo( UserText, ( oldProps, newProps ) => isEqual( oldProps, newProps ) );
```
kueda, line 133: Can you add a comment in the code explaining why this is necessary?

kueda (approved):
I think this is fine, though as you say, a bit convoluted. My only recommendations are regarding explanatory comments, which would be nice to have but don't need to block.

Regarding this weird behavior of react-navigation not unmounting components by default, I don't really understand it myself. We made the [`<Mortal>`](https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/components/SharedComponents/Mortal.js) component to wrap navigation screens that really don't need to live forever.

Chrischuck:
@kueda I don't think all Screens are wrapped w/ `<Mortal />` 

I __believe__ stack screens will get destroyed on blue, but default behavior for tab screens is to preserve the view unless stated otherwise with the [unmountOnBlur](https://reactnavigation.org/docs/bottom-tab-navigator/#unmountonblur) prop

Also just reading the `freezeOnBlur` prop, which i will see if it fixes it. Because if it does then this could be a one line fix

PR #635, opened 2023-05-23
587 add offline upload alert

Chrischuck:
Screenshot 2023-05-23 at 9 49 47 AM


src/i18n/strings.ftl
```diff
@@ -1022,3 +1022,6 @@ JOURNAL-POSTS-WITHOUT-NUMBER = {$count ->
     [one] JOURNAL POST
     *[other] JOURNAL POSTS
 }
+
+Internet-Connection-Required = Internet Connection Required
```
kueda, line 1026: Please add a context comment here, e.g. "Title for dialog telling the user that an Internet connection is required"

src/i18n/strings.ftl
```diff
@@ -1022,3 +1022,6 @@ JOURNAL-POSTS-WITHOUT-NUMBER = {$count ->
...
+Please-try-again-when-you-are-connected-to-the-internet=Please try again when you are connected to the internet.
```
kueda, line 1027: Please capitalize "Internet" and add a newline at the end of the doc.

src/sharedHooks/useIsConnected.js
```diff
@@ -1,15 +1,29 @@
...
+    netInfo.isConnected === null
+      ? true
+      : netInfo.isConnected
+  );
+
+  useEffect( () => {
+    const unsubscribe = addEventListener( state => {
+      // Note on ios simulator, these notifications are delayed
+      // https://github.com/react-native-netinfo/react-native-netinfo#issues-with-the-ios-simulator
+      setIsConnected( state.isConnected );
```
kueda, line 20: Might be a bit out of scope, but this and the initial value for `useState` should use `isInternetReachable`. If I'm running a VPN on my phone but the phone is in airplane mode, `isConnected` is true even when `isInternetReachable` is not.

kueda (changes requested):
Looking good, just need to resolve conflicts and make a few minor changes.

kueda (approved):

PR #644, opened 2023-05-25
Loading/disabled state for ObsEdit buttons

### LINKED ISSUE #627: Save button in ObsEdit needs a loading state
**Describe the bug**
When you save an observation on obs edit, saving and the return to My Observations is not instantaneous, so that button needs to be disabled on press so the user doesn't trigger a save twice, and it needs some kind of loading state to tell the user something is happening.

**To Reproduce**
1. Start a new observation and get to ObsEdit
2. Tap the "Save" button. Usually it will take so much time that you can tap it multiple times.

**Expected behavior**
When you tap save the button should become a disabled loading button to a) prevent further taps, and b) tell the user saving is in progress.

 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.4.0 (44)

**Additional context**
This isn't in the designs but this is just general good practice. When you tap an interactive element, something should happen immediately. If it's not the thing you wanted, then it should be something to tell you that what you wanted is going to happen soon.

albullington:
Closes #627

### DIFF
--- a/src/components/ObsEdit/BottomButtons.js
+++ b/src/components/ObsEdit/BottomButtons.js
@@ -21,7 +21,6 @@ const BottomButtons = ( ): Node => {
     saveObservation,
     saveAndUploadObservation,
     setNextScreen,
-    setLoading,
     currentObservation,
     unsavedChanges,
     passesEvidenceTest,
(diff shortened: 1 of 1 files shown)


kueda (approved):
FWIW, I never end up seeing the ActivityIndicator, but there's still a bit of a lag before saving. Can't decide if it's worth *intentionally* stalling save so we can show the user something is happening. Probably not. Anyway, good to merge.

Issue #641, opened 2023-05-24
Update toolbar upload status when uploading a single observation


kueda:
When you tap the upload button on a single observation, the upload status in the toolbar should update to show that 1 observation is being uploaded. Currently nothing happens.

So when you start uploading a single observation

- [ ] Nothing should change about the upload status on the observation item
- [ ] Toolbar upload button icon should rotate
- [ ] Toolbar upload status should say "Uploading 1 of 1 observation"
- [ ] Toolbar progress bar should reflect the same progress as shown on the observation item

PR #663, opened 2023-06-06
Simplify progress bar event emitters

### LINKED ISSUE #630: Toolbar upload progress bar fills up before upload is complete
**Describe the bug**
The progress bar on the bottom edge of the MyObservations toolbar fills up before all the observations have uploaded, possibly because it isn't accounting for multiple photos per observation

**To Reproduce**
1. Make a few new observations and make sure some have multiple photos
2. Tap on the sync button in the toolbar to start uploading
3. Watching the linear progress on the bottom of the Toolbar; it will get all the way to the right edge before all the observations have completed uploading

**Expected behavior**
The linear progress bar should show the progress of the entire upload and not suggest the upload is done when it isn't.

 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.4.0 (44)

**Additional context**
I'm guessing the progress bar is just getting its total from the number of observation but the progress gets updated for every individual photo.

FWIW, I also noticed some problems with the total number of observations being uploaded changing during the upload, sometimes resulting in impossible text like "Uploading 10 of 5 observations" but I don't have a way to replicate that.

albullington:
Closes #630

I think the code is a lot cleaner, but the UI is still a bit jumpy -- open to suggestions on:

1. how the toolbar should act when a user taps the circular progress bar, and vice versa -- I'm not sure if progress should be displayed on both or only one at a time
2. when to emit progress so the progress bar/circle advances smoothly -- right now I'm emitting progress at the start of an upload, after an observation has been created/updated, and when a photo is attached to an observation. we probably need to display progress for a few smaller in-between steps here, like when the photo is created


kueda (changes requested):
Regarding whether to update the toolbar status when uploading a single observation, there's an issue for that: https://github.com/inaturalist/iNaturalistReactNative/issues/641

Re: when to emit progress... I'm not really convinced we need this emitter strategy (maybe we need it to manage progress bars on observations *and* the toolbar?), but assuming we want to keep it, I think you're on the right track with emitting events when an upload starts and when it finishes, but I don't think the emitter needs to know how much progress goes with each event. IMO, the emitters should just emit separate events when they start and when they stop, and the events should include the UUID like we're already doing, and an error if the upload did not succeed. Listeners to those events should decide how much progress that counts for, e.g. `useUploadObservations` could maintain an array of all the observation _and_ photo UUIDs that it needs to upload, and the length of that array is the progress total. It could listen for the start and stop emitted events, and increment progress by 0.5 for each. That way all the places where we emit the events don't need to know anything about how much progress an event counts for.

If you want to stay on the general cleanup track, the upload logic seems pretty spread around (models, `useUploadObservations`, `ObsEditProvider`, etc.). I think most of this stuff could be probably be consolidated in `useUploadObservations` (or `ObsEditProvider`). I also don't think the API wrappers need to emit events since that doesn't really have anything to do with the API. Emitting stuff should probably happen in the place where those API wrappers get called. That said, I think that's probably beyond the scope of this issue.


The test failures seem legit, and there's a bug: uploads never finish for me. Or rather the uploads succeed, but the toolbar progress never completes:

Screenshot 2023-06-09 at 5 41 43 PM

albullington:
Ready for re-review. Mostly a lot of upload code cleanup, with some event emitter cleanup / simplification.

kueda (approved):
IMO this is all progress, and I especially appreciate moving a lot of the upload logic out of the model. My changes in #669 simplify things a little further, I think, but I don't think they're necessary. Either way, nothing blocking.

FWIW, I was noticing that progress indicators on individual observations never seem to complete, or rather they get to ~75% and then they're suddenly done without getting to 100%. I assume that's because 1 obs + 1 photo = 4x start/stop events, so when the final event happens when the obs is uploaded, it skips straight to a completed state. Might be nice to see the indicator get to 100% before that, but again, not a blocker.

Issue #545, opened 2023-03-15
Deleting an observation crashes the app


jtklein:
1. On the observation list
2. Navigate to one observation
3. ... button, delete, confirm
4. App navigates back to observation list and after few moments crashes.

With

```
ERROR : JS Error: Fatal: Error: Exception in HostFunction: Accessing object of type Observation which has been invalidated or deleted
    at get__synced_at (native)
    at wasSynced (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:112413:20)
    at ObsUploadStatus (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:322240:42)
    at renderWithHooks (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:9806:33)
    at updateFunctionComponent
```

and

```
Possible Unhandled Promise Rejection (id: 0):
Error: Exception in HostFunction: Cannot access realm that has been closed.
Error: Exception in HostFunction: Cannot access realm that has been closed.
    at objects (native)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:112638:53)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:112643:57)
    at anonymous (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:112467:54)
    at filter (native)
    at upsertRemoteObservations (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:112466:48)
    at _callee20$ (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=org.inaturalist.iNatMobileBeta:371049:62)
    at call (native)
    at tryCatch 
```

Chrischuck:
I can take a look at this

kueda:
[...]

PR #666, opened 2023-06-20
Fixed obs deletion bug and allow edit when obs has no media

### LINKED ISSUE #662: Can't edit an observation with no media
**Describe the bug**
ObsEdit has no edit button when the observation has no media

**To Reproduce**
1. Make an observation with no media
2. Upload it (otherwise you can't view it in ObsEdit)
3. View it in ObsEdit; there is no edit button

**Expected behavior**
You should be able to edit an observation without media, i.e. there should be an edit button in the header just like there is for obs with media.

**Screenshots**
Simulator Screenshot - iPhone 14 Pro - 2023-06-06 at 11 59 40

Happens in iOS and Android, phone and tablet.

kueda:
Obs deletion bug (#545) mostly has to do with Realm objects hanging around in memory when they're no longer in the database. I spent a long time trying to write a test that could see the custom header button menu we add via react-navigation but never got it to work.

I'm not sure #545 is really "done" since I didn't figure out a way to test it. If others think that needs to be tested... I'm open to advice, but if not I think it's worth merging this and closing that issue if this work doesn't have any negative side effects.

Closes #662 (edit button on obs detail when no media)

### DIFF
--- a/src/components/MyObservations/MyObservations.js
+++ b/src/components/MyObservations/MyObservations.js
@@ -220,7 +220,7 @@ const MyObservations = ( {
             />
             <AnimatedFlashList
               contentContainerStyle={contentContainerStyle}
-              data={observations}
+              data={observations.filter( o => o.isValid() )}
               key={layout}
               estimatedItemSize={
                 layout === "grid"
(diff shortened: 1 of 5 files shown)


albullington:
This fixes the crash for me for observations which have been created locally.

For observations which were synced from iNat, deleting the observation does not remove it from the observation list. Instead, I'm seeing the observation still in the list, with the photo thumbnail removed.

Not sure if we should fix that as part of this PR or create a separate issue.

Screenshot 2023-06-20 at 2 38 52 PM

albullington (approved):

PR #667, opened 2023-06-21
fix: MyObs log in button width issue on iPad

### LINKED ISSUE #665: MyObs Log in button doesn't resize on rotation on iPad
**Describe the bug**
The "LOG IN TO INATURALIST" button on My Observations looks fine on load, but is either too wide or too narrow after rotation on an iPad.

**To Reproduce**
1. Open the app on an iPad, make sure you're signed out. The "LOG IN TO INATURALIST" should fill the screen width (with some margin)
2. Rotate the device
3. If you started in landscape and rotated to portrait, the button now flows off the right side of the screen; if you did the opposite, the button is too narrow

**Expected behavior**
The button should remain 100% width like the other buttons in that view.

**Screenshots**
IMG_0127
IMG_0128

 - Device: iPad Air 5th gen
 - OS: iPadOS 16.5
 - App version: 0.6.0 (47)

AGCrossman95:
Closes #665


albullington (changes requested):
Hi @AGCrossman95, thanks for submitting a PR for this issue! The fix looks great to me; however, I don't think this PR needs to edit any files outside of `src/components/MyObservations/Header.js`. Would you be willing to resubmit with changes to that file only?

AGCrossman95:
> Hi @AGCrossman95, thanks for submitting a PR for this issue! The fix looks great to me; however, I don't think this PR needs to edit any files outside of `src/components/MyObservations/Header.js`. Would you be willing to resubmit with changes to that file only?

Hi @albullington, will do! I agree, didn't mean to include those files. 👍🏼 Will resubmit.

PR #669, opened 2023-06-22
Remove increment from upload event emitters

kueda:
This attempts to further simplify the upload event emitters by removing the increment, which never changes, and which should probably be controlled by the listener, not the emitter, since the listener is responsible for actually updating the progress. Also tries to ensure that events that get emitted to signal that an upload finished always happen even if the upload errors out.


albullington (approved):
This looks cleaner and more readable to me, and uploads are working as expected.

kueda:
I tried and failed to get this merged with main, so I'm just going to abandon it. IMO we should probably isolate the upload logic from ObsEditProvider somehow, but I'm not sure if a different context or using a hook would be best.

PR #684, opened 2023-06-28
ARCamera MVP: Basic screen with full screen camera and prediction labels

jtklein:


kueda (changes requested):
FWIW, I got this to run with some model files I had on hand from Seek, but either as a staff member or as an external developer, I don't know where I *should* get the latest versions of these files, or even some demo files to make sure the ARCamera is working as expected, so please add instructions to the README.

src/sharedHelpers/cvModel.js
```diff
@@ -0,0 +1,81 @@
+// @flow
+import { Platform } from "react-native";
+import RNFS from "react-native-fs";
+
+const modelFiles = {
+  IOSMODEL: "optimized_model.mlmodelc",
+  IOSTAXONOMY: "taxonomy.json",
+  ANDROIDMODEL: "optimized_model.tflite",
+  ANDROIDTAXONOMY: "taxonomy.csv"
+};
```
kueda, line 10: IMO, these values should be configurable in .env. Assuming these files have unique names each time we change them, that will help us know what version of the model we're building with.

src/sharedHelpers/cvModel.js
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  RNFS.readDirAssets( "camera" ).then( results => {
+    const model = modelFiles.ANDROIDMODEL;
+    const taxonomy = modelFiles.ANDROIDTAXONOMY;
+
+    const hasModel = results.find( r => r.name === model );
+
+    // Android writes over existing files
+    if ( hasModel !== undefined ) {
```
kueda, line 43: This results in a silent failure if the model files aren't in the correct places. Instead, I think it should provide an informative an alarming error, or it should cause a fatal crash. These files are required for some core app functionality.

kueda (commented):
Sorry, I guess I had more feedback. I don't think either of these are blockers, more like nice-to-haves.

src/sharedHelpers/cvModel.js
jtklein: I have added an error alert showing when there is no model available here. This does require to create a fake model though during e2e tests to avoid the error alert. I have added a node script to create such fake model. Without explanation in the Readme though, as its only use case would be here I think.

jtklein:
I have added a script to download the example model into the necessary folders, and added an explanation to the Readme.

src/sharedHelpers/cvModel.js
jtklein: Have refactored everything so that the filenames are read from the .env file. Please, let me know if the information in the readme would be enough to get started without any prior knowledge of how to set up the repository.

kueda (approved):
Ok, looking good. I think any developer should be able to get the app running now. I can't quite test on the ancient laptop I have at the moment, but I'll assume it works!

FWIW, in the future we'll need to add some strong checks to make sure we don't make production builds with the wrong model files (e.g. the example ones).

PR #692, opened 2023-06-29
Location picker improvements

### LINKED ISSUE #628: LocationPicker updates
- [x] Current location button needs a loading state; IMO we should just replace the crosshairs with an ActivityIndicator while location is being fetched
- [x] Don't use StickyToolbar for the footer; nothing ever needs to flow under the footer and doing it this way just makes spacing of the bottom-aligned elements hard to control in different contexts (e.g. on my Pixel 3 those buttons are too close to the footer)
- [x] Bottom buttons need a drop shadow in Android (maybe need the `elevation` prop)
- [x] Double check the size of the bottom buttons on Android, they seem too small; should be 46x46px
- [x] LocationPicker needs to lose its state when closed; currently if you enter it, drag the map around, close it without saving, then enter it again, it isn't reset back to the position of the observation
- [x] Zoom in while fetching current location so positional_accuracy is always the best possible

Here's what the bottom of the LocationPicker is looking like on my Pixel:
Screenshot_20230522-162755

albullington:
Closes #628


jtklein:
On the Android device I tested this it never resolves with a location. The default timeout for `Geolocation.getCurrentPosition` is 10 minutes. I think it makes sense to use a smaller timeout when the call comes fom pressing the "current location" button.

jtklein:
Also, currently, the crosshair element is somewhere in the center of the map+header elements, according to the design specs it should be in the center of the entire screen. I don't know if that is something you would incorporate here.

jtklein (changes requested):
Code seems mostly fine to me. However, I did notice a few things. The ones here in this comment are regressions from main I think and should be adressed by this PR. The ones in the previous comments can also be punted to different issues as they are not related to the changes introduced here.

On iOS, I have noted the following: On ObsCreate it shows my location with an accuracy of 35, and when I move to the LocationPicker it shows a green circle and again an accuracy of 35. However, I think the zoom level is not correct, because when I only pan around from the initial position the accuracy jumps to 545 (and orange circle) on finger release.

Also, on iOS it is not possible to pan the map around when pressing on the "Zoom in so that the observation's accuracy is as low as possible." white rectangle. On Android that is possible.

Another thing that I have noticed is that comparing to main now it is no longer possible to zoom out beyond a certain point. 545 meters of accuracy on my iPhone and 1090m on Android (don't know if the actual number is relevant).

Issue #694, opened 2023-07-06
Observation reappears after deletion


jtklein:
Delete one observation from the main screen, and after a short while the observation reappears on the main screen.

Screen recording for the first observation in the list:

https://github.com/inaturalist/iNaturalistReactNative/assets/17345891/2b66b635-0b76-4d29-93d4-5649524c2955

jtklein:
This seems to be flaky though and does not happen every time I delete an observation.

kueda:
FWIW, I *think* this might be related to caching in React Query, the app fetches your observation w/ React Query, you delete an observation and return to My Observations, which tries to fetch your obs again but it hits the React Query cache that still contains your deleted observation.

kueda:
@jtklein why did you close this? Did you test and could not replicate, or do you think it was fixed by a specific commit?

kueda:
FWIW, this is still happening for me in [v0.18.1+66](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.18.1%2B66) (Pixel 8, Android 13).

albullington:
So far I'm not able to replicate this in [v0.19.0+67](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.19.0%2B67) (iPhone 15 Pro, iOS 17)

kueda:
I can still replicate it with these steps on a simulated iPhone 15 Pro (iOS 17.2)

1. Sign in
2. Make a new obs without evidence
3. Tap the upload button the toolbar to upload, allow upload to complete
4. Delete the observation you just created
5. Back on My Observations, tap the sync button
6. The obs you deleted re-appears

jtklein:
I could not replicate this using the steps laid out above on a simulator iPhone 15 Pro (iOS 17.2) in [0.21.1](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.21.1%2B70). So, I assign it back to the creator.

kueda:
FWIW, I can still replicate this on device and in a simulator, but you have to sync quickly after deletion. I have a fix ready in #1240

Issue #683, opened 2023-06-28
Add e2e test for observation complete CRUD cycle 


jtklein:
It would be good to have a more detailed e2e test that goes through the cycle of:
- create an observation,
- update said observation,
- delete the observation

jtklein:
Almost working version here:
https://github.com/inaturalist/iNaturalistReactNative/pull/696

Testing for the comment to appear is not working because of: https://github.com/inaturalist/iNaturalistReactNative/issues/695

Also, currently blocked by https://github.com/inaturalist/iNaturalistReactNative/issues/694 because deleted observation does not disappear.

Issue #695, opened 2023-07-06
Adding comment twice


jtklein:
When adding a comment in the bottom modal, two activity items appear in the ObsDetail screen.

Screen recording:

https://github.com/inaturalist/iNaturalistReactNative/assets/17345891/bf167379-b78f-4cbc-939e-b815b2304ed3

angielt:
Fix included in #701 merged to main

PR #696, opened 2023-07-06
Crud e2e

jtklein:
This introduces an e2e test that:
1. Signs in
2. Creates an observation without evidence
3. Uploads the observation
4. Adds a comment to the observation
5. Deletes the observation


jtklein:
It is currently not checking if the comment appears because of this:  https://github.com/inaturalist/iNaturalistReactNative/issues/695

jtklein:
Not always, but sometimes the test fails because of this behaviour: https://github.com/inaturalist/iNaturalistReactNative/issues/694

jtklein:
Closes #694 
Closes #683

Issue #629, opened 2023-05-22
Trying to take a new photo from ObsEdit gets you stuck


kueda:
**Describe the bug**
When you try to take a new photo while editing an existing observation, a bottom sheet partially opens and seems to make the buttons at the bottom untappable. Even the system back button in Android doesn't let you leave and you have to force-quit to escape.

**To Reproduce**
1. Edit an existing observation
2. Tap the "+" button under EVIDENCE and choose to take a new photo
3. StandardCamera, but a few pixels of a bottom sheet appear at the bottom of the screen and none of the StandardCamera buttons at the bottom are tappable. Furthermore, the Android system back button does nothing, so you are trapped.

**Expected behavior**
I expect the StandardCamera to work normally and not to be trapped.

**Screenshots**
![Screenshot_20230520-132214](https://github.com/inaturalist/iNaturalistReactNative/assets/22691/5b836fd4-f513-4203-9935-890df09dc372)

**Context (please complete the following information):**
 - Device: Pixel 3
 - OS: Android 12
 - App version: 0.4.0 (44)

**Additional context**
This might only be an Android issue, or maybe just a phone issue, b/c it doesn't happen on my iPad.

albullington:
I can't replicate this on Pixel 7, Android 13

kueda:
I can't replicate the exact conditions I originally reported, but running [d58415d](https://github.com/inaturalist/iNaturalistReactNative/commit/d58415dd2749e0e8ee609abce59d1f767c1df0a6) on my Pixel 3 (Android 12), the ADD EVIDENCE bottom sheet is still visible when the StandardCamera appears and tapping the X button on the sheet does nothing. I can tap off the sheet to dismiss it now and the camera works normally.

PR #698, opened 2023-07-10
#629 - wait for bottom sheet to dismiss before navigating away to next screen

budowski:


kueda (changes requested):
This is working for me, these requests are largely cosmetic. I have a fix for the test failure that I'm about to push.

kueda (approved):

Issue #579, opened 2023-04-21
My Observations punchlist


kueda:
Abhas and I came up with this list last week and I meant to make individual issues but ran out of time. If someone wants to pick these up, great, and if they don't make any sense I can rework them.

- [x] #591
  kern), but it looks like it's book only on android, not fully centered,
  should go a pixel up
- [x] ~photo count sometimes gets duplicated (overlapping numbers) usually only
  on iphone, seems to be related to scrolling, so possibly a recycling views~ (@kueda can't replicate this on 2023-08-29)
  problem, look into PhotoCount component with ForeignObject
- [ ] #760
- [x] when at the top of the screen, if you pull the list down and let it
  bounce back, the header collapses. it should stay the same. only on ios
- [x] #590
  camera views
- [x] #586
- [ ] on abhas's samsung galaxy s8, he has granted storage permission but when
  he tries to open the gallery he gets the PermissionGate permission denied
  view
- [x] #588
  seconds, often it's so fast it doesn't appear at all
- [x] toolbar while uploading 1 of 1, non-progress bar contents should never
  flow out of margins, and status text should wrap
- [ ] #694
- [x] #587
- [ ] ~while offline, sometimes the Toolbar progress bar gets stuck. if it
  fails, the progress should reset~
  * still happening on 2023-08-31, but I think it would be handled by #621
- [x] similarly, while offline, if you try to tap upload on obs edit, there's
  a loading wheel, but nothing happens. it should just save the obs or show
  an error
- [x] #589

Chrischuck:
I can take a look at the offline issues and probably the the ios issue where you pull down and the header collapses.

kueda:
Thanks, @Chrischuck. If you comment on those issues I'll assign you.

kueda:
I'm going to close this because I think all the remaining items are handled in other issues.

PR #703, opened 2023-07-17
#579 - font weights fixes for obs list

budowski:


albullington (commented):
This fix works for me. @budowski is there a specific reason why you think we need to introduce a new Typography component here? I think this change (using `INatText` instead of `INatTextLight`) should apply to every instance of Body1 throughout the app, right?

Issue #710, opened 2023-07-21
Android location doesn't resolve in location picker


albullington:
On the Android device I tested this it never resolves with a location when pressing the "My location" button. The default timeout for `Geolocation.getCurrentPosition` is 10 minutes. I think it makes sense to use a smaller timeout when the call comes fom pressing the "current location" button. But this is not part of the PR spec, so we can also make a new issue for this.

_Originally posted by @jtklein in https://github.com/inaturalist/iNaturalistReactNative/issues/692#issuecomment-1625574916_

kueda:
This may have been addressed in #727 but I think it was fixed earlier.

PR #727, opened 2023-07-31
#710 - current location fetching timeout; fix crash when reaching location picker before location was fetched in observation viewer

budowski:


kueda (changes requested):
I wasn't able to replicate #710 in main, so while I think we can close that, I don't think this PR fixed it.

This does fix the crash you noted, but it leaves the screen in this state:

Screenshot_20230731-114558

The loading spinner is weirdly in a square, overlapping some text, and never goes away. Instead, the loading spinner should appear *instead* of any underlying text, have no background, and it should go away whenever a location fetch in flight has completed or timed out.

kueda (approved):

PR #741, opened 2023-08-19
Nested navigators

### LINKED ISSUE #737: Use nested navigators
Currently, the root navigator of the app is a [Drawer Navigator](https://reactnavigation.org/docs/drawer-navigator) (corresponding to the drawer that slides out from the left side of the screen), which contains a single screen, the [Bottom Tabs Navigator](https://reactnavigation.org/docs/bottom-tab-navigator), which contains *every* screen in the app. We did this to address a back button problem in #541, but it has the effect of making all screen transitions instantaneous, and I think it confuses the IA and discourages modularity (e.g. ObsEdit *should* be modular enough to exist in multiple stack navigators without any problems).

So my request here is research at first, maybe leading to a full-scale change:

1. Read up on [nesting navigators](https://reactnavigation.org/docs/nesting-navigators)
1. Experiment with using nested [Native Stack Navigators](https://reactnavigation.org/docs/native-stack-navigator) on a single flow, e.g. MyObs -> ObsDetail -> ObsEdit
1. Report back on whether this seems better and worth doing everywhere

albullington:
Closes #737 

- Creates stack navigators outside and inside of the tab navigator.

Pay attention to:
- AddID/ObsEdit screens and whether they're working correctly now that we have them in two separate stacks, both within and outside of the tab navigator.


src/navigation/StackNavigators/CameraStackNavigator.js
```diff
@@ -0,0 +1,216 @@
...
+  <Mortal>
+    <PermissionGate
+      permission={PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION}
+    >
+      <ObsEdit />
+    </PermissionGate>
+  </Mortal>
+);
+
+const CameraStackNavigator = ( ): Node => (
```
kueda, line 134: Can we call this the `AddObsStackNavigator` instead? It seems to encompass more than just camera-related screens, but it doesn't seem to handle more than the add obs flow. Alternatively, you could think of each add obs option as an entry point to its own stack, since you can't really navigate between them (e.g. switch to the ARCamera while you're importing photos). That might be excessive, though.

src/navigation/StackNavigators/ObservationsStackNavigator.js
```diff
@@ -0,0 +1,74 @@
...
+  removeBottomBorder,
+  showHeaderLeft,
+  showLongHeader
+} from "navigation/navigationOptions";
+import type { Node } from "react";
+import React from "react";
+
+const Stack = createNativeStackNavigator( );
+
+const ObservationsStackNavigator = ( ): Node => (
```
kueda, line 23: Is the intent to reuse this navigator in Explore? If not, can we call it the `MyObservationsStackNavigator`? Consider what should happen when the user taps the Explore button in the header of MyObservations. I don't think that should actually open Explore within the same stack, i.e. pushing a new window in from the right and allow the user to pop back out to My Obs. Instead, I think it should be like popping out to the root and entering a new stack, albeit with some params.

kueda (commented):
This is looking pretty good to me. Transitions feel more native, and I think the challenges w/ ObsEdit and AddID can be overcome by placing them in multiple navigators.

Haven't gone through this line by line, but I left some inline comments, and there are the following minor questions / issues:

* Shouldn't the CameraContainer be a part of the BottomTabsNavigator (or maybe a part of a new AddObsStackNavigator)?
* Do we need `react-native-modal` if `react-navigation` handles modal presentation?
* Re: handling `goBack()` from ObsEdit, it works fine for me if I add ObsEdit to the ObservationsStackNavigator. react-navigation seems fine with screens that have the same name in different navigators... unless I'm missing something?
* Can we make tapping the buttons in the BottomTabBar pop the user out to the root of the relevant navigator? I feel like that would be useful when you're nested down really deeply... but also dangerous if you're on a screen like ObsEdit
* Back button in the ObsEdit header seems to have shifted over to the right too far, esp. in Android

src/navigation/StackNavigators/ObservationsStackNavigator.js
kueda: Ok, @abhasinat and I [discussed this](https://inaturalist.slack.com/archives/C02G5AATVGA/p1692740155221459) and he made a good case for a combined stack because there are so many ways to get to explore (from MyObs, from user profile, from taxon detail...), so `ObservationsStackNavigator` it is!

albullington:
Rad, will take another look at it with this feedback in mind. To respond to your questions:

_Shouldn't the CameraContainer be a part of the BottomTabsNavigator (or maybe a part of a new AddObsStackNavigator)?_
- No, because if we put the CameraContainer or any of the AddObsStack screens in the Tab Navigator, the bottom tab nav will automatically show on those screens. Tab Navigators don't have access to see what is in nested stack navigators, so there's no way to selectively hide the tab navigator on those screens. (Per [this documentation](https://reactnavigation.org/docs/hiding-tabbar-in-screens))

_Do we need react-native-modal if react-navigation handles modal presentation?_
- I suspect we want to punt this exploration to a different issue, but I've been intentionally avoiding using react-navigation modal screens. My thinking on this before was that we're typically using modals to pass callbacks from screen components (for example, FlagItemModal receives a callback that handles an item being flagged). React-navigation frowns on passing callbacks through route parameters, so that's not an option if we do need to pass callbacks into the modals. It would be ideal if we could use modals without needing to pass in callbacks, and I'm definitely open to suggestions or some exploration on how we can refactor to accomplish this.


_Re: handling goBack() from ObsEdit, it works fine for me if I add ObsEdit to the ObservationsStackNavigator. react-navigation seems fine with screens that have the same name in different navigators... unless I'm missing something?_
- Will take a second look. I suspect every time I've tried this and gotten an error it's been a complaint about the same name being in a parent and child navigator, not unrelated navigators.


_Can we make tapping the buttons in the BottomTabBar pop the user out to the root of the relevant navigator? I feel like that would be useful when you're nested down really deeply... but also dangerous if you're on a screen like ObsEdit_
- I thought I did this for MyObservations (popped a user back to the ObservationsList screen) and the root of the navigator doesn't apply to Explore anymore since that will be in the same ObservationsStack navigator -- we can have that bottom button go back to the main Explore screen rather than Explore filters, etc.

_Back button in the ObsEdit header seems to have shifted over to the right too far, esp. in Android_
- Will double check. I think we could use an overall audit of header styles, either in this PR or a separate issue, to make sure all of our custom header buttons line up with the react-navigation header buttons

albullington:
Tried a version of putting ObsEdit and AddID into the TabNavigator -> ObservationsStack as well as the AddObsStack... and realized the reason I didn't do that before is because it means the versions in the ObservationsStack are stuck with the CustomTabBar. This wouldn't be the worst except that it covers the buttons in ObsEdit so a user can't save/upload an observation. So, we'll need a different solution for this part, even if it's just something like getting those buttons to sit on top of the nav bar.

kueda (approved):
K, I think this is good. I made a few minor layout related changes and fixed a typo preventing saving an obs in the ObservationsStackNavigator

PR #743, opened 2023-08-23
Permission gate layouts

kueda:
Numerous changes to PermissionGate to get it closer to the designs:

* Separated business and display logic
* Updated `react-native-permissions` and used it exclusively instead of mixing in `PermissionsAndroid`
* Consolidate logic for determining platform-specific permissions into `PermissionGateContainer`
* Allow `PermissionGate` to request multiple permissions

Questions

1. `PermissionGate` should never really get plugged into a container component other than `PermissionGateContainer`, so would it feel better to name the container `PermissionGate` and the view `PermissionGateView` or something?
2. Does the gate need to be optional sometimes, e.g. location for ObsEdit?


src/navigation/BackButton.js
```diff
@@ -4,12 +4,12 @@ import type { Node } from "react";
 import React from "react";
 import colors from "styles/tailwindColors";
 
-const BackButton = (): Node => {
+const BackButton = ( { tintColor = colors.black } ): Node => {
```
albullington, line 7: A little out of scope, but I think this custom BackButton should live in `SharedComponents/Buttons` so we know to reuse it throughout the app. Right now we're using HeaderBackButton in ~5 places

src/components/SharedComponents/PermissionGateContainer.js
```diff
@@ -0,0 +1,160 @@
+// @flow
+
+import { useNavigation } from "@react-navigation/native";
+import Modal from "components/SharedComponents/Modal";
```
albullington, line 4: These PermissionGates might actually be good candidates for using react-navigation modals, since they're full screen and don't need background opacity, etc.

albullington (commented):
Looks pretty good to me. I tried to permission gates on iOS iPhone 14 Pro simulator and an iPad Air simulator, and here's what I'm noticing:

1. On Explore (and Projects) I get the regular permission for location, not a permission gate. Is this intentional or should every permission show a PermissionGate now? (I see the Location PermissionGate when moving through PhotoGallery -> GroupPhotos -> ObsEdit flow or going straight to ObsEdit)

2. The first time I tried to open StandardCamera (after going to Explore) I got a blank gray screen, not the PermissionGate & not the Camera screen, but I think this was a fluke because I can’t replicate. The second time I restarted the app the PermissionGate appeared

3. I’m seeing a weird flicker effect when using the close button to close out of the permission modals. Going from PhotoGallery or RecordSound back to ObsList, it looks like ObsList goes dark for half a second

4. When I go to ObsEdit and tap the "only while using app" option for location permission, I land on ObsEdit and immediately get a second native permission that asks, "Are you sure you only want to fetch location while using the app?"

kueda:
> On Explore (and Projects) I get the regular permission for location, not a permission gate. Is this intentional or should every permission show a PermissionGate now? (I see the Location PermissionGate when moving through PhotoGallery -> GroupPhotos -> ObsEdit flow or going straight to ObsEdit)

Should be fixed.

> The first time I tried to open StandardCamera (after going to Explore) I got a blank gray screen, not the PermissionGate & not the Camera screen, but I think this was a fluke because I can’t replicate. The second time I restarted the app the PermissionGate appeared

I haven't seen this so I'm assuming it was solved with a clean build.

> I’m seeing a weird flicker effect when using the close button to close out of the permission modals. Going from PhotoGallery or RecordSound back to ObsList, it looks like ObsList goes dark for half a second

Haven't figured this out yet. Currently the PermissionGate does not render anything behind the gate until permission is granted, but the gate itself is rendered in a modal. Behind the modal is just a blank screen, so when you close the modal, that blank screen is visible right before you navigate back. This might be fixed if we used react-navigation modals, but I haven't experimented with that yet.

> When I go to ObsEdit and tap the "only while using app" option for location permission, I land on ObsEdit and immediately get a second native permission that asks, "Are you sure you only want to fetch location while using the app?"

Looked into this a bit and I *think* it's because https://github.com/michalchudziak/react-native-geolocation says we have to require `NSLocationAlwaysAndWhenInUseUsageDescription`. So we could probably address it by asking for that permission... but why do we even need that? We really only need to access the user's location when the app is in use.

FWIW, not sure this is mergeable yet. I think it's an improvement, but I still need to figure out how to fetch location again after permission has been granted (on Explore and Projects).

albullington (changes requested):
A few funky things for me: 

1. I'm seeing this pop up on the ObsEdit screen immediately after I tap confirm on the location PermissionGate... not sure why there are double permissions happening here. Maybe the "use location when not using the app" is something we can avoid asking the user at all?
Screenshot 2023-09-28 at 12 26 16 PM

2. I'm not seeing the PermissionGate pop up on Explore or Projects when I navigate to those screens first on a fresh build. Instead, I'm getting this warning message that says we probably need to use a PermissionGate. So it looks like that PermissionGate is only working on ObsEdit
Screenshot 2023-09-28 at 12 25 35 PM
Screenshot 2023-09-28 at 12 25 41 PM

3. For gallery, if I deny permissions, I'm seeing first the PermissionGate, then when I tap "Choose photos" it shows me the "denied gallery permissions text." A similar thing happens in the camera. Shouldn't we be able to detect user permissions and show the user the denied permissions screen off the bat instead of first showing them the standard permission gate screen?

4. If I deny location permissions on ObsEdit, there's no way to ever get to the ObsEdit screen. Is this intentional, or should there be a way for a user to navigate forward without location permissions enabled? (Like, if they're importing a gallery photo?)

5. If I deny location permissions on ObsEdit, then tap the X, I'm seeing the Camera Access Permissions screen instead of just going back to whatever screen I was on before.

kueda:
> I'm seeing this pop up on the ObsEdit screen immediately after I tap confirm on the location PermissionGate... not sure why there are double permissions happening here. Maybe the "use location when not using the app" is something we can avoid asking the user at all?

I think I fixed that. Seems like if the app declares the "always" version of that permission we end up asking for it even when we explicitly only ask for the "only while in use" version.

> I'm not seeing the PermissionGate pop up on Explore or Projects when I navigate to those screens first on a fresh build. Instead, I'm getting this warning message that says we probably need to use a PermissionGate. So it looks like that PermissionGate is only working on ObsEdit

Fixed in Explore, debateable whether it's worth addressing in Projects. That warning appears when useUserLocation tries to get the user's location but doesn't have permission. The way projects is set up right now, it tries to fetch that location on first load, no matter what tab is visible, even though it's not required and the user isn't expecting that to happen. When you tap the Nearby tab, it should show the Permission Gate and reload the nearby projects if permission is granted. IMO, it's fine the way it is because that warning should not appear in a release build, but we could redo projects so each tab is a separate component and responsible for fetching its own data, and just wrap the Nearby tap in a Permission Gate. What do you think? Worth it, and if so, worth doing in this PR or carving it off into a separate issue?

albullington:
I'd say probably not worth it to redo Projects now, since like you said, a user isn't expecting to need a location prompt anyway.

kueda:
@albullington I think I've addressed most of the things in your review. IMO the most glaring oddity remaining is that location does not fetch *immediately* when you get to ObsEdit, but I think I'd prefer to punt that to future issues since this is already an absolute beast of a PR. At this point I'm mostly just hoping nothing is drastically broken and things that are weird can be worked out after merging.

albullington (approved):
Looks good to me -- not seeing anything broken on iOS or Android.

PR #778, opened 2023-09-11
Show taxon pins on Explore Map

### LINKED ISSUE #750: Explore: tap observations to go to ObsDetail
A user should be able to zoom into the map and tap on individual observations -- this is functionality we currently have on the iOS/Android apps.

Tapping on an observation should take a user to the ObsDetail screen for that observation.

This means that we need to plot individual observations on the map, instead of using the map tiles overlay.

albullington:
closes #750


kueda (approved):
Looks fine for the locally-rendered markers approach. If we get infinite scroll working in Explore is that potentially going to render infinite map markers, though?

PR #780, opened 2023-09-14
Group photos UI

### LINKED ISSUE #480: GroupPhotos UI
Depends on
- [x] #469

Integration issue for GroupPhotos UI

- [x] #481
- [x] Replace existing header with a `<ContextHeader>`
- [x] "Group photos into observations– make sure there is only one species per observation" should be Body2 and be a header to the FlatList (or FlashList) of photos
- [x] Replace FlatList with FlashList (one could potentially import a large number of photos; is there a cap?)
- [x] #482
- [x] Use GroupPhotoObservations or whatever comes out of #482
- [x] #484
- [x] Add a StickyToolbar with a single button that says "IMPORT X OBSERVATIONS" where X is the number of observations; should take the user to ObsEdit (equivalent of current "Next" button)
- [x] #485
- [x] Add a FloatingToolbar with group, split, and delete buttons
- [x] group button should group selected observations into a single observation
- [x] split button should ungroup all selected observations so all their photos are separate observations
- [x] delete button should remove the selected observations and delete selected photo records
- [x] FloatingToolbar should only appear when you've selected an observation, should animate in from behind the StickyToolbar
- [x] Breakpoints change the number of columns and the size of the grid items (size needs to be calculated based on screen size):
  - [x] `sm`: 1 col
  - [x] `md`: 2 col
  - [x] `lg`: 2 col
  - [x] `xl`: 4 col
  - [x] `2xl`: 6 col

jtklein:
Closes #480

This PR includes several smaller UI changes, including dynamic sizing and number of columns for the photo's list.

Made a separate [issue](https://github.com/inaturalist/iNaturalistReactNative/issues/781) for what happens when pressed on the plus button.

### DIFF
--- a/src/components/SharedComponents/Buttons/INatIconButton.js
+++ b/src/components/SharedComponents/Buttons/INatIconButton.js
@@ -77,9 +77,12 @@ const INatIconButton = ( {
       onPress={onPress}
       style={( { pressed } ) => [
         {
-          opacity: pressed
-            ? 0.95
-            : 1,
+          // eslint-disable-next-line no-nested-ternary
+          opacity: disabled
+            ? 0.5
+            : pressed
+              ? 0.95
+              : 1,
           width,
           height,
           justifyContent: "center",
--- a/src/components/SharedComponents/Buttons/INatIconButton.js
+++ b/src/components/SharedComponents/Buttons/INatIconButton.js
@@ -67,6 +67,16 @@ const INatIconButton = ( {
       "Button needs an accessibility label"
     );
   }
+  const opacity = pressed => {
+    if ( disabled ) {
+      return 0.5;
+    }
+    if ( pressed ) {
+      return 0.95;
+    }
+    return 1;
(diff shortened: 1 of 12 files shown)


src/components/SharedComponents/Buttons/INatIconButton.js
```diff
@@ -77,9 +77,12 @@ const INatIconButton = ( {
...
         {
-          opacity: pressed
-            ? 0.95
-            : 1,
+          // eslint-disable-next-line no-nested-ternary
+          opacity: disabled
+            ? 0.5
+            : pressed
+              ? 0.95
+              : 1,
```
kueda, line 85: Is there a reason to ignore the rule here? I guess I'm on the fence about nested ternaries. Sometimes I find them hard to read, sometimes not.

kueda (approved):
Looks good. I'm going to commit a few minor style changes.

PR #798, opened 2023-09-25
Update Fastlane file to check for extraneous assets

### LINKED ISSUE #793: Fastlane check to prevent extraneous files in android/app/src/main/assets/camera
I noticed today that the APKs I build are ~350 MB, while the APKs Johannes builds are ~90 MB, and I realized the reason is that I have a bunch of extraneous files lying around in android/app/src/main/assets/camera that are getting bundled into the APK even though they aren't referenced anywhere and don't do anything.

So I'd like a Fastlane check that runs with `fastlane release` that throws an error if there's anything in android/app/src/main/assets/camera that is *not* referenced in the .env file. It would be great to perform similar checks on everything in the assets directory if possible, maybe using `link-asset-manifest.json`.

albullington:
closes #793

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -299,6 +299,34 @@ end
 
 lane :release do
   desc "Make github release for the latest tag and make builds"
+  ANDROID_ASSETS_FOLDER = Dir.children( "../android/app/src/main/assets/camera" )
+
+  ANDROID_ASSETS_FOLDER.each do | file |
+    next unless !ENV.has_value?( file ) 
+
+    UI.abort_with_message! <<~EXTRA_ANDROID_ASSETS_ERROR
+      Android assets folder has extraneous file: #{file}, which is not listed
+      as an environment variable. Remove any unreferenced files from
+      android/app/src/main/assets/camera before running a release build
+      EXTRA_ANDROID_ASSETS_ERROR
+  end
+
+  ASSETS = File.open( "../android/link-assets-manifest.json" ) {| f | JSON.parse( f.read )["data"] }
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -299,6 +299,34 @@ end
 
 lane :release do
   desc "Make github release for the latest tag and make builds"
+  ANDROID_ASSETS_FOLDER = Dir.children( "../android/app/src/main/assets/camera" )
```
kueda, line 302: This doesn't need to be a constant since it's not used outside of this context. You can just do something like `Dir.children( "../android/app/src/main/assets/camera" ).each do | file |`.

fastlane/Fastfile
```diff
@@ -299,6 +299,34 @@ end
...
+      EXTRA_ANDROID_ASSETS_ERROR
+  end
+
+  ASSETS = File.open( "../android/link-assets-manifest.json" ) {| f | JSON.parse( f.read )["data"] }
+  linked_assets = Array.new
+  ASSETS.each do | asset |
+    linked_assets.push( asset["path"].split( "assets/fonts/" )[1] )
+  end
+
+  FONTS_FOLDER = Dir.children( "../assets/fonts" )
```
kueda, line 320: Same here, doesn't really need to be a variable.

kueda (commented):
Nothing blocking, just minor feedback. Way to tackle some Ruby! My main feedback here is that this should probably all be inside 

```ruby
platform :android do
  lane :build do
    # android build related stuff here
  end
end
```

since this is all unique to an Android build.

PR #819, opened 2023-10-06
Vision camera v3

jtklein:


kueda (approved):
I think your patch approach is fine, Johannes. Not thrilled about upping the minimum Android version, but as we discussed, Android 8 is pretty old.

FWIW, I haven't had a chance to run this, but it looks fine after a read through.

PR #828, opened 2023-10-23
569 native image picker

budowski:
Use our own forked version of react-antive-image-picker library when choosing existing photos for an observation.


kueda (changes requested):
My Main request is to refactor to use PermissionGate, because that already encapsulates a lot of the complicated permissions logic for reading from the gallery in multiple versions of multiple operating systems. https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/navigation/StackNavigators/AddObsStackNavigator.js#L57-L70 shows how we do it using our own gallery component, which might be an option if we want to have a kind of empty gallery component that just launches the native picker. https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/components/ObsEdit/EvidenceSection.js#L145-L151 shows how to do this without actually treating it as a "gate" component that wraps a component that can only be used after granting permission.

LMK if that doesn't make sense, but my main goals are to isolate permissions logic in a single place.

Also, some bugs:

* while I can import in iOS, I do not get coordinates out of the metadata when I do so
* sharing photos into the app doesn't seem to work in an iOS simulator
* on a physical iPad Air (5th gen), the native photo picker doesn't appear

kueda (changes requested):
- [x] Need to resolve conflicts w/ main
- [ ] When I grant permission in a simulator, I get a blank white screen (iPhone 15 Pro and iPhone 8 simulators). The same seems to happen on my iPad Air (5th gen). This does not happen in Android (Android 12, Pixel 3)

src/components/PhotoImporter/PhotoGallery.js
```diff
@@ -1,261 +1,150 @@
...
+      response = await ImagePicker.launchImageLibrary( {
+        selectionLimit: MAX_PHOTOS_ALLOWED,
+        mediaType: "photo",
+        includeBase64: false,
+        includeExtra: true,
+        forceOldAndroidPhotoPicker: true,
+        chooserTitle: t( "Import-Photos-From" ),
+        presentationStyle: "fullScreen"
+      } );
+    } catch ( exc ) {
```
kueda, line 51: This catches every single thing that might go wrong with `await ImagePicker.launchImageLibrary`. Please catch a specific error or check the error to make sure you're only catching errors related to permissions.

src/i18n/strings.ftl
```diff
@@ -256,8 +256,8 @@ DISCARD-CHANGES = DISCARD CHANGES
 DISCARD-OBSERVATION = DISCARD OBSERVATION
 
 DISCARD-X-OBSERVATIONS = {$count ->
-    [one] DISCARD OBSERVATION
-    *[other] DISCARD {$count} OBSERVATIONS
+[one] DISCARD OBSERVATION
+*[other] DISCARD {$count} OBSERVATIONS
```
kueda, line 260: Are these manual? I don't think we have linting set up for the ftl files, and all the examples at https://projectfluent.org/ use indentation so I don't think this is conventional. If this is something your editor was doing automatically could you turn it off?

src/sharedHelpers/permissions.js
```diff
@@ -0,0 +1,41 @@
...
+  for ( let i = 0; i < permissions.length; i += 1 ) {
+    // eslint-disable-next-line no-await-in-loop
+    const result = await requestPermission( permissions[i] );
+    if ( result !== PERMISSION_GRANTED ) {
+      return result;
+    }
+  }
+
+  return PERMISSION_GRANTED;
+};
```
kueda, line 41: Is this file still necessary? It seems to do stuff that PermissionGate does.

kueda (changes requested):
Sorry, here's another review based on a read-over. There are also a number of lingering console.log states you should remove.

kueda (approved):

PR #834, opened 2023-10-25
Update of react-native-vision-camera to version 3.4

jtklein:


jtklein:
The changes necessary for this update in the codebase are pretty minor for it to work.
However, I also had to update a lot of the e2e CI codebase in order to cope with some errors that came up during testing. Some of them might be happening only, e.g. getting the location permission on iOS simulator is currently not working using the latest applesimutils together with detox. Overall, the iOS e2e CI tests seem to be even flakier than before now.

jtklein:
Found another bug when testing, reverting this back to draft for now.
When taking a photo with the ARCamera the app keeps saving the same photo over and over again to the phone.

jtklein:
Fixed the bug with taking a photo

src/components/Camera/ARCamera/ARCamera.js
```diff
@@ -111,11 +113,15 @@ const ARCamera = ( {
...
+  useEffect( () => {
+    if ( hasFinishedHere ) {
+      return;
+    }
     if ( photoSaved ) {
+      setHasFinishedHere( true );
       navToObsEdit( localTaxon );
     }
-  }, [photoSaved, navToObsEdit, localTaxon] );
+  }, [hasFinishedHere, photoSaved, navToObsEdit, localTaxon] );
```
kueda, line 124: What is this for? This either needs a better variable name or a comment explaining why `hasFinishedHere` is necessary.

.github/workflows/e2e_ios.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   setup:
-    runs-on: macos-latest
+    runs-on: macos-13
```
kueda, line 14: Can you add a comment to this file explaining why `macos-13` is required? It looks like `macos-latest` *is* `macos-13` right now. Are you just trying to avoid unpleasant surprises when it becomes `macos-14`?

.github/workflows/e2e_ios.yml
```diff
@@ -36,11 +36,17 @@ jobs:
   test:
     needs: checksecret
     if: needs.checksecret.outputs.is_SECRETS_PRESENT_set == 'true'
-    runs-on: macos-latest
-    # Kill the task if not finished after 60 minutes
-    timeout-minutes: 60
+    runs-on: macos-13
+    # Kill the task if not finished after 150 minutes
+    timeout-minutes: 150
```
kueda, line 41: I have never seen any of our e2e tests take this long, and while the e2e-iOS tests for this branch seem about twice as slow as main, they don't seem to be exceeding 40 minutes, so why is this necessary?

kueda (changes requested):
I'm not seeing any changes in camera behavior, so I think this is fine on the whole, just some minor requests regarding readability and docs.

Regarding the react-native-vision-camera patch, can you please explain what that's for in the body of the PR, preferably with a link to the relevant issue in the vision-camera repo?

This change also requires devs to update to XCode 15, so please make a note of that in the PR body.

.github/workflows/e2e_ios.yml
jtklein: Hm, I was checking here: https://github.com/actions/runner-images/blob/main/README.md#available-images and from that it seems that macos-12 = macos-latest. macos-12 did not have Xcode 15 so we had to switch to macos-13 for this PR.

src/components/Camera/ARCamera/ARCamera.js
jtklein: Have renamed and added a comment.

.github/workflows/e2e_ios.yml
jtklein: I did see some runs where the e2e failed with error message  "timeout exceeded". So I kept ramping up until I could see for most failed runs an error message that had nothing to do with timeouts. For a successful run this number is indeed way to high. I don't know what is a good number here, or at what point we think a timeout means failure.

PR #867, opened 2023-11-07
Performance testing with Reassure

albullington:
Testing framework for measuring how long a component takes to render. Uses MyObservations as an example. Couldn't quite get continuous integration working (maybe reassure needs to be installed in the main branch to start), but the ultimate goal would be to connect this to CI to make sure we're not adding significant render time with new PRs.


kueda (changes requested):
IMO reassure looks interesting and is worth trying, but I don't think we'll know if it tells us anything useful until we merge it into main. After we do that we can think about how to work it into our CI.

package.json
```diff
@@ -133,6 +134,7 @@
     "babel-plugin-module-resolver": "^5.0.0",
     "babel-plugin-transform-inline-environment-variables": "^0.4.4",
     "babel-plugin-transform-remove-console": "^6.9.4",
+    "danger": "^11.3.0",
```
kueda, line 137: Still including danger.

kueda (changes requested):

kueda (approved):

PR #879, opened 2023-11-09
Performance: move upload reducer out of ObsEditProvider into MyObservationsContainer

albullington:
It's likely that using Provider/Context for global state is slowing down our overall app performance by encouraging the whole tree to rerender. This is a first step towards removing ObsEditProvider, or at least minimizing the usage of global state.

With this rewritten code, a user should still be able to:
- sync observations from the server via the toolbar
- upload via the toolbar
- upload via individual upload buttons
- upload directly from the button on ObsEdit with a single observation or multiple observation screens
- see correct progress on toolbar and individual uploads in MyObservations

[upload-move-branch-revised.json](https://github.com/inaturalist/iNaturalistReactNative/files/13313798/upload-move-branch-revised.json)
[upload-main.json](https://github.com/inaturalist/iNaturalistReactNative/files/13313799/upload-main.json)

Here are Flipper profiles for the main branch and new changes. I saved two observations for upload, then pressed recording on Flipper, tapped the sync button on the Toolbar, and waited for 2 observations to upload. It looks like individual renders take roughly the same amount of time, but the uploads completed in 22 renders in the new branch and 44 in the main branch.

A next step might be to try to test performance in reassure.


kueda (changes requested):
I could replicate your results in Flipper (after your help understanding Flipper!), so that's great. Quantifying the performance improvement feels good. I agree it would feel even better to automate that quantification with reassure, but I don't think that needs to block merging this PR.

My comments here are mostly for readability and don't need to block merge either.

The e2e test in iOS seems real, though, and that is blocking. I can replicate by running the iOS e2e test locally, and on my iPad (you need to tap "UPLOAD NOW" on ObsEdit.

Issue #359, opened 2023-01-13
Post logs to internal logging


kueda:
Depends on #357. react-native-logs supports custom transports, so we can and maybe should write a custom transport that posts log data to api.inaturalist.org/v1/log. That way we can track custom events as well as crashes using our own services, no third party. This would be a lot easier with https://github.com/inaturalist/inaturalistjs/issues/68, and I think it absolutely depends on https://github.com/inaturalist/iNaturalistAPI/issues/358. I can flesh this issue out a bit more if/when those things get done.

kueda:
Paired with Amanda on this today. Got logs to make GET requests to the logging endpoint in e408cd93fcfbca5914d910ff3878a933a775e2b4. Some more todos / things to consider:

* what happens when you try to log while offline? does it raise an exception? if it is do we want to log that anywhere so we know logging is broken? what if THAT log system fails?!?!?!
* might we consider writing to some kind of queue that gets processed when online?
* should we be using a POST request for all these requests? they are all creating data, even if they're not errors
* do we want to log while the user is logged out?
* once we have *some* auth figured, do we want to merge this just so we have some logging while we're developing?

PR #904, opened 2023-11-15
Performance monitoring: Shopify React Navigation

albullington:
Here's a first pass at performance monitoring with Shopify's React Navigation performance tool. This wraps the Drawer navigation component and measures render time when a user navigates to a new screen. It reports out these linked [ types of data](https://shopify.github.io/react-native-performance/docs/fundamentals/render-pass-report) on render times / touch events.

A few notes:
- This tool seems to produce errors anywhere we have a custom back button
- I tried to set this up to measure all the core screens in the MyObservations -> ObsEdit flow. If we decide we find this tool useful, we can continue setting this up across the full app
- If we like this tool, we should figure out how we can use results to measure changes from one branch to another. Right now, there's no easy way to compare render times when we introduce new code. Shopify recommends using a visual tool like Amplitude (which is not free)
- This would likely be an alternative to reassure, which is the other performance monitoring tool we've been testing


kueda (commented):
FWIW, this seems a bit too invasive for my tastes, so I'd prefer not to merge for now. Most of what I can see is just render times, which is useful, but as you say, hard to use to compare branches at this point.

That said, it might be useful if we want / need to do some internal analytics in the future. At a minimum we could post render times to our own logging service if/when we get https://github.com/inaturalist/iNaturalistReactNative/issues/359 done.

PR #908, opened 2023-11-16
Performance: simplify ObsEditProvider

albullington:
This is a refactor which moves a lot of functionality out of ObsEditProvider and into the screens where state is needed. This PR has some performance gains, but the main purpose was to simplify ObsEditProvider enough to give us the option to explore a global state management library like [zustand](https://github.com/pmndrs/zustand).

I ran through profiling in Flipper, starting on the MyObservations screen, then:
- Adding an observation with no evidence
- Tapping add evidence button and adding a photo from the camera
- Tapping add evidence button and adding a photo from the gallery
- Saving and returning to MyObservations

At first, the total number of renders were the same as in main, with the number of renders caused by ObsEditProvider halved (12 instead of 22).

Then, I refactored so ObsEditContext is only called at the top level of a screen (i.e. in ObsEdit instead of in BottomButtons) and the relevant context is passed down as props. When I ran through the same profiling steps described above, the total renders during that profiling session were reduced to 149 from 194 (~25% less renders).

### DIFF
--- a/src/components/ObsEdit/BottomButtons.js
+++ b/src/components/ObsEdit/BottomButtons.js
@@ -1,17 +1,24 @@
 // @flow
 
+import { useNavigation } from "@react-navigation/native";
 import classnames from "classnames";
 import {
   Button, StickyToolbar
 } from "components/SharedComponents";
 import { View } from "components/styledComponents";
-import { ObsEditContext } from "providers/contexts";
+import { ObsEditContext, RealmContext } from "providers/contexts";
 import type { Node } from "react";
 import React, { useContext, useEffect, useState } from "react";
+import Observation from "realmModels/Observation";
+import { writeExifToFile } from "sharedHelpers/parseExif";
+import uploadObservation from "sharedHelpers/uploadObservation";
 import useTranslation from "sharedHooks/useTranslation";
 
+import { log } from "../../../react-native-logs.config";
 import ImpreciseLocationSheet from "./Sheets/ImpreciseLocationSheet";
 import MissingEvidenceSheet from "./Sheets/MissingEvidenceSheet";
+
+const { useRealm } = RealmContext;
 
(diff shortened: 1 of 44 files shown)


src/components/ObsEdit/BottomButtons.js
```diff
@@ -1,40 +1,115 @@
...
+    }
+    // Update all photos taken via the app with the new fetched location.
+    cameraRollUris.forEach( uri => {
+      logger.info( "writeExifToCameraRollPhotos, writing exif for uri: ", uri );
+      writeExifToFile( uri, exif );
+    } );
+  };
+
+  const saveObservation = async observation => {
+    ensureRealm( );
```
kueda, line 75: What happens when you don't do this?

kueda (approved):
I'm not seeing quite the reduction in renders that you are, but I am seeing a reduction, so that's progress. Inline comments are all nice-to-haves.

src/components/Suggestions/Suggestions.js
```diff
@@ -25,28 +25,39 @@ type Props = {
...
-  setSelectedPhotoUri
+  setSelectedPhotoUri,
+  nearbySuggestions,
+  onTaxonChosen,
+  comment,
+  loadingSuggestions,
+  setComment,
+  currentObservation,
+  loading,
+  setLoading
```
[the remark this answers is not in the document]
albullington: I've been likewise trying to remember to alphabetize as much as I can, and have also been wishing for a solution that would do it automatically. It's definitely hard to read long lists without alphabetizing.

src/components/ObsEdit/BottomButtons.js
albullington: I think it's safe to remove. That was code we had in ObsEditProvider but doesn't seem necessary.

Issue #851, opened 2023-11-04
Adding an ID from Explore takes you to ObsEdit


kueda:
**Describe the bug**
Trying to add an identification to an observation from Explore takes you to ObsEdit. No one should ever see ObsEdit unless they are editing their own observation.

**To Reproduce**
1. Go to Explore
2. Tap on an observation to get to ObsDetail
3. Tap on the SUGGEST ID button to get to Suggestions
4. Suggestions is blank, which is not great, but not the problem here. Tap on SEARCH FOR A TAXON to get to TaxonSearch
5. Search for a taxon and tap the green checkmark button
6. You land on ObsEdit

**Expected behavior**
Tapping on the green checkmark button should add the ID and take you back to ObsDetail where you should see your ID load (or it should take you to ObsDetail and you should see a loading status while your ID gets added). You should never get to ObsEdit for an existing observation you didn't create.

**Context (please complete the following information):**
 - Device: iPad Air (5th gen)
 - OS: iPadOS 16.6.1
 - App version: [v0.15.0+61](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.15.0%2B61)

**Additional context**
I haven't tried actually saving the observation b/c I don't want to mess with anyone's obs, but you should test that that's impossible too by trying it out while connected to staging.

PR #921, opened 2023-11-18
#851 - taxon suggestions - support for adding identification for someone else's observation (not your own)

budowski:


kueda (changes requested):
This looks good, but there is a legit test failure having to do with mocking. If you merge from main I think that should go away.

kueda (approved):

PR #931, opened 2023-11-28
Increase pressable area of back button, Obs Edit

### LINKED ISSUE #907: ObsEdit back button is very hard to tap
**Describe the bug**
The back button in the header of ObsEdit is quite difficult to tap. You need to tap the center very precisely to get it to work, which probably means it is not the minimum acceptable size for an accessible button.

**To Reproduce**
1. Start a new obs without evidence
2. Smash the back button the header with prejudicial inaccuracy
3. Feel the anger rising

**Expected behavior**
The button should be easy to tap, i.e. it should have a tappable area of at least 44x44, if not bigger.

 - Device: iPad Air (5th gen)
 - OS: iPadOS 16.6.1
 - App version: [v0.16.0+62](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.16.0%2B62)

**Additional context**
This might be about the tappable area, or maybe there are other elements overlapping the back button and intercepting taps. Try making a test for the MyObs Header that uses `.toBeAccessible()` to see if that complains about a small tappable area.

albullington:
Closes #907

- Use custom header rather than the react navigation header with setOptions, so we can control the padding around the back button

### DIFF
--- a/src/components/ObsEdit/EvidenceSection.js
+++ b/src/components/ObsEdit/EvidenceSection.js
@@ -89,7 +89,7 @@ const EvidenceSection = ( {
   };
 
   return (
-    <View className="mx-6 mt-6">
+    <View className="mx-6">
       <AddEvidenceSheet
         disableAddingMoreEvidence={obsPhotos?.length >= MAX_PHOTOS_ALLOWED}
         hidden={!showAddEvidenceSheet}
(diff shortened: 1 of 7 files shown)


kueda (changes requested):
This creates an extra blank space at the top of the screen on an iPhone SE and iPad Air.

PR #948, opened 2023-12-01
Replace ObsEditProvider with zustand for global state management

albullington:
Big overhaul to use [zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) instead of provider/context for state management. This change came out of research on performance improvements -- this library is recommended by [callstack](https://www.callstack.com/), since it doesn't cause extra rerenders the way providers do.

I tested in Flipper and unfortunately didn't see any immediate performance improvements when looking at number of renders or render time on the Flamegraph. I profiled using the steps in this [performance PR](https://github.com/inaturalist/iNaturalistReactNative/pull/908).

That said, I think there are some improvements to consider unrelated to render time:
- It feels easier to test components, since we no longer need to wrap components in <ObsEditProvider /> or mock the provider. Instead, we can add any mocked data directly to the `useStore` mock. A good example of this is the test to check for photo deletion in the PhotoCarousel.test.js file -- we can mock arrays as well as functions in global state.
- There's less boilerplate code, which makes everything in `useStore` feel more readable
- I believe there will be less temptation to add extra functions/variables to `useStore` the way we did in the provider component
- It feels easier to read and fix bugs in the Observation loop since all code lives in the screen components, not in a provider and reducer


src/stores/useStore.js
```diff
@@ -0,0 +1,106 @@
...
+  } ) ),
+  updateComment: newComment => set( { comment: newComment } ),
+  updateObservations: updatedObservations => set( state => ( {
+    observations: updatedObservations,
+    currentObservation: updatedObservations[state.currentObservationIndex],
+    unsavedChanges: true
+  } ) )
+} ) );
+
+export default useStore;
```
kueda, line 106: Even at this early stage, this store is becoming big and complicated. Have you thought about using [slices](https://docs.pmnd.rs/zustand/guides/slices-pattern)?

kueda (commented):
Easier testing is definitely an improvement improvement, so maybe worth trying in our next debug iteration. Bummer that it doesn't seem to have any performance gains. Why do you think it's less of a temptation to overload the store?

I'm also curious how you think this should be used with useReducer. Does it replace that pattern?

albullington:
I'd say useReducer is a good tool for local state management, so we can keep using that for any self-contained screens that don't need access to global state (like MyObservations). useStore should be for state that absolutely needs to be shared across screens, which currently only includes the Observation creation loop.

If we do find a compelling reason to add to global state, we should definitely use slices to keep the Observation loop separate from other stores, and I added a note about that on top of the useStore code.

Issue #584, opened 2023-04-25
Field changing value rapidly without user input


xurizaemon:
**Describe the bug**

When entering observation data, the "Organism is wild" field blinked "Yes" and "No" rapidly.

**To Reproduce**

I'm not confident about reproducing this :) but I'll describe how it happened.

1. First open of this app v0.1.1+40
2. Logged in OK
3. Created an observation, took a few photos of a dog
4. Went to modify observation data, changed "Organism is wild" to "No"
5. Tried to click next text input field (label not fully legible behind keyboard but it's "Add optional notes")
6. Noticed that "Organism is wild" field was alternating between Yes and No rapidly - no user input at this time
7. I was not able to set the value after this - it kept bouncing between yes/no

**Expected behavior**

Retain the set value

**Screenshots**

- https://user-images.githubusercontent.com/105608/234164395-6fd4d962-e7fc-4878-bea4-9eccf36b610f.mp4

**Context (please complete the following information):**
 - Device: Android OnePlus 8T
 - OS: Android 13
 - App version: v0.1.1+40

**Additional context**
I realise it's an early release and am excited to see it. While checking the app out for a first run I noticed this and thought to capture it.

jtklein:
Thank you for your report. I believe this is not an issue anymore and fixed recently by #453.

kueda:
This is still happening in [v0.16.0+62](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.16.0%2B62), albeit less frequently.

Here's [screen cap](https://photos.google.com/share/AF1QipNWgp3Z2K6BEanMSj-hPDevcHAaftOTHVTmW3uizNNFmZSHsPTN9I0Rn_jFCSKFNA/photo/AF1QipOT3PTND08H5MDhO-g_2ZmaBDpFF7U9Mv2sWBL1?key=NnY4R1FOcmE5bW9vTHRBdnlSQktmNy1DSThjS2tB) from an iPhone 13 Pro, iOS 17.0.3.

This remains hard to replicate, but as the screen cap suggests, it helps to use a real device, and it helps to be in airplane mode so the app is forced to use slower GPS coords. IMO, we probably do not need to be writing anything other than local state every time we get those coordinates, and only write them when the observation gets saved.

kueda:
This *might* be fixed by https://github.com/inaturalist/iNaturalistReactNative/pull/949

kueda:
Haven't seen this happening recently so closing for now.

PR #949, opened 2023-12-02
#584 - don't directly call updateObservationKeys from callback function, after fetching location

budowski:


kueda:
@budowski sorry for not seeing this earlier. Could you please resolve your conflicts?

PR #960, opened 2023-12-12
Refactor BottomSheet to always use dynamic snap points

jtklein:
For all usages in the app of the BottomSheet component, or components derived from it, this PR removes the possibility to specify snapPoints as a prop, and replaces it with a hook to dynamically calculate them onLayout. In other words, all instances of bottom sheet will scale depending on the content to display.
This necessitated some changes in paddings at the bottom of existing bottom sheets.

### DIFF
--- a/src/components/Camera/StandardCamera/DiscardChangesSheet.js
+++ b/src/components/Camera/StandardCamera/DiscardChangesSheet.js
@@ -23,7 +23,6 @@ const DiscardChangesSheet = ( {
   return (
     <WarningSheet
       handleClose={( ) => setShowDiscardSheet( false )}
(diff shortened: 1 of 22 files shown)


albullington:
Definitely not necessary, but it might be helpful to throw an error if a user tries to pass in hardcoded values for snapPoints. I could see us forgetting to use best practices here and reverting back to passing props at some future date. I'm imagining an error that would work similarly to the ones in [iNatIconButton](https://github.com/inaturalist/iNaturalistReactNative/blob/42ab5eafc5d91e04ca76e8b3b27a1398b132cfc7/src/components/SharedComponents/Buttons/INatIconButton.js#L60).

PR #963, opened 2023-12-13
Checkbox shared component

albullington:
Using a new library, bouncy checkbox, because it is more customizable than either of the other solutions we're using in different parts of the app. This will be used in the ExploreFilters screen.

We should also replace other checkboxes throughout the app (Settings, FlagItemModel) with this shared component in a separate PR.


jtklein (changes requested):
Mostly seems fine.
I did see on Android and iOS a small space between the inner circle and the more squary outline. Is that something we can get rid of? Also, if the component is stable, we could add a snapshot test as well.

Screenshot 2023-12-14 at 11 42 53

PR #971, opened 2023-12-15
Update checkboxes & fix tests

albullington:
- Replace react native community checkbox with bouncy checkbox
- Fix tests; make sure checkbox is still working when checked/onPress props are passed in from external components


albullington:
@jtklein does this address the problem you were bumping into with the checkmark icon still appearing after pressing? This seems to be working for me in the Settings examples.

Issue #871, opened 2023-11-08
Taxon Detail screen; Photo section


jtklein:
In the photo section according to the [designs](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=2106-15859&mode=dev) add a kebab (meetballs) menu to the section.
The opened menu has two items:
1) View in browser
On press it should: Opens an external browser with this taxon detail page on web. Opening the external browser should not deeplink back into the app directly, and ideally not show the "Open this page in the app banner", however, if you feel this is out of scope here, feel free to move this part to a separate issue.
2) Share
On press it should: Open the native share modal with a link (same URL as in 1) to this taxon detail page

kueda:
Closed in 1bc128bfc

PR #980, opened 2023-12-18
#871 - show pop-up menu for taxon photo - share / view in browser

budowski:


src/components/TaxonDetails/TaxonDetails.js
```diff
@@ -122,6 +141,33 @@ const TaxonDetails = ( ): Node => {
...
+              onPress={( ) => {
+                openURLInBrowser( taxonUrl );
+                setKebabMenuVisible( false );
+              }}
+              title={t( "View-in-browser" )}
+            />
+            <Menu.Item
+              onPress={( ) => {
+                Share.share( {
+                  message: taxonUrl,
```
albullington, line 162: On iOS, having both message and url keys results in text that says "2 Links", as shown below. We solved this on ObsDetails using a [conditional statement](https://github.com/inaturalist/iNaturalistReactNative/blob/94ea2b6054dc06a0fea8e234e28a525ff08377ec/src/components/ObsDetails/HeaderKebabMenu.js#L26) for iOS vs. Android.

Screenshot 2023-12-18 at 2 44 28 PM

src/components/TaxonDetails/TaxonDetails.js
```diff
@@ -64,6 +68,7 @@ const TaxonDetails = ( ): Node => {
     logger.error( `Failed to retrieve taxon ${id}: ${error}` );
   }
   const taxon = remoteTaxon || localTaxon;
+  const taxonUrl = `${TAXON_URL}/${taxon.id}`;
```
albullington, line 71: This line is throwing an error for me on Android.

Screenshot 2023-12-18 at 2 46 18 PM

albullington (changes requested):
Blockers:
1. The 2 links issue on iOS and
2. The null error that's preventing me from opening TaxonDetails at all on Android

A nice-to-have would be adding a few tests, similar to what we have for the [kebab/share menu on ObsDetails](https://github.com/inaturalist/iNaturalistReactNative/blob/94ea2b6054dc06a0fea8e234e28a525ff08377ec/tests/unit/components/ObsDetails/HeaderKebabMenu.test.js#L25), to make sure the kebab menu opens the expected browser link and the expected share link on iOS/Android.

budowski:
Thanks - but very weird with Android, since I had it working for me - what steps do you take to recreate this? How do you reach the taxon page and what taxon are you viewing?

albullington:
> Thanks - but very weird with Android, since I had it working for me - what steps do you take to recreate this? How do you reach the taxon page and what taxon are you viewing?

I tried as a logged out user. Went to Explore -> ObsDetails for Yellow Admiral -> TaxonDetails which resulted in a crash.

albullington:
fwiw, I believe all you need is the optional chaining operator here (i.e. `taxon?.id` instead of taxon.id) to prevent a crash -- JavaScript is complaining about looking for the id key on a null taxon object.

budowski:
Thanks, but the problem is that I need to show some URL there, otherwise it will be a taxon URL without the actual taxon ID - I'll look into why that object is empty

Issue #975, opened 2023-12-17
Crash when viewing UserProfile of certain users


kueda:
**Describe the bug**
Viewing the profile of certain users causes a "white screen of death" crash.

**To Reproduce**
1. Find an observation by the user laurabaker, e.g. go to Explore and search for Nidula candida in Sonoma County, CA, USA to get to https://www.inaturalist.org/observations/193169066
2. Tap on "@laurabaker" immediately below the photos
3. The nav header shows briefly, and then the screen goes all white

**Expected behavior**
The profile should show without crashing.

**Context (please complete the following information):**
 - Device: iPad Air (5th gen)
 - OS: iPadOS 17.1.2
 - App version: 0.19.0 (67)

**Additional context**
I'm not seeing anything useful in the logs.

joergmlpts:
This crash appears to happen since **laurabaker** has an account without site affiliation:

![Screenshot_20231218-141620](https://github.com/inaturalist/iNaturalistReactNative/assets/66563749/b808b8e5-e2c0-4fb4-93d5-2b4c5377aff0)

This API call [https://api.inaturalist.org/v1/users/laurabaker](https://api.inaturalist.org/v1/users/laurabaker) also shows that this is an account with no site:

```
{
  "total_results": 1,
  "page": 1,
  "per_page": 1,
  "results": [
    {
      "id": 17402,
...
      "site_id": null,
...
    }
  ]
}
```

albullington:
@joergmlpts thanks for your PR contribution! This issue is closed in #982

PR #982, opened 2023-12-19
don't show site affiliation when user doesn't have one; fix for issue #975

joergmlpts:
This change fixes issue #975.


albullington (approved):
Fixes #975

Issue #896, opened 2023-11-13
Explore screen; Map layer button


jtklein:
Similar to #872 add a Maplayer button to the Explore screen, as per the current [designs](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=7063-35744&mode=dev). On pressing this button the map toggles between Map and Satellite on iOS and Map, Satellite, and Terrain layers on Android.

kueda:
@jtklein and @abhasinat did we really want different functionality on iOS (Map & Satellite) and Android (Map, Terrain, & Satellite)? That's no in the designs, FWIW. If that's what we want, I'll re-open this.

jtklein:
Yes, the decision was taken because iOS native Maps app has Map & Satellite only, whereas Google Maps being the standard on Android has Map, Terrain & Satellite.
However, since now Map to Satellite is implemented, maybe @abhasinat might decide to punt the addition of Terrain to post-MVP?

abhasinat:
that's fine with me if that makes implementation easier.

PR #989, opened 2023-12-31
#896 - switch map layer button

budowski:


package.json
```diff
@@ -121,6 +121,7 @@
     "realm": "^12.3.0",
     "reassure": "^0.10.1",
     "sanitize-html": "^2.11.0",
+    "tailwind-merge": "^2.2.0",
```
kueda, line 124: Please remove this and use the classnames package instead. I'm pretty sure it does exactly the same thing.

kueda (changes requested):
Works for me, but we don't need the tailwind-merge package. There's a bit of a layout bug on tablets, but I think that's a separate issue.

kueda (approved):

Issue #993, opened 2024-01-02
Endless spinner when adding a photo to an existing observation


kueda:
**Describe the bug**
You see an endless spinner when you try to add a photo to an existing observation (one downloaded from the web).

**To Reproduce**
1. Sign in to an account with existing observations; allow existing observations to download
2. Edit an existing observation with a photo to get to ObsEdit
3. Tap the "+" button next to the existing photo
4. Tap the gallery button in the bottom sheet
5. Choose a photo and tap "Add"  in the upper right
6. You see a blank white screen with a green loading spinner that spins forever

**Expected behavior**
You should be able to add photos to existing observations just like you can for new observations.

**Context (please complete the following information):**
 - Device: iPad Air (5th gen)
 - OS: iPadOS 16.6.1
 - App version: [v0.19.0+67](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.19.0%2B67)

**Additional context**
Happens on Android too. While trying to replicate I occasionally got a crash immediately after tapping the "+" button so that's worth investigating too. In the log it looks like `Error: Accessing object which has been invalidated or deleted` so another fun Realm error.

albullington:
Note for @budowski : there's a PR open that may address this issue: https://github.com/inaturalist/iNaturalistReactNative/pull/998

joergmlpts:
This issue does not occur anymore in v0.20.0+68. The PR for issue #918 appears to have fixed this as well.

kueda:
Closed in #998

Issue #918, opened 2023-11-17
Uploading new photo to existing observation does not complete progress


kueda:
**Describe the bug**
When you upload a new photo for an existing observation, the circular progress indicator on the observation never completes its progress.

**To Reproduce**
1. Add a new observation with a photo and upload it
2. Tap on that observation to get to ObsDetail
3. Tap on the pencil button in the upper right get to ObsEdit
4. Tap the "+" button in the EVIDENCE section and choose to import another photo
5. Tap the "SAVE " button to save the observation
6. Return to My Observations. The toolbar should show the sync button with the "Upload 1 observation" label and the observation list item should show its own upload button
7. Tap "Upload 1 observation" in the toolbar
8. The upload should complete and the toolbar UI should return to its default state, but the circular progress indicator on the observation list item remains in a state of incomplete progress
9. Tap on another observation to get to ObsDetail and then return to MyObs
10. The circular progress indicator has gone away and we're showing obs status instead

**Expected behavior**
The circular progress indicator should complete whether you are uploading an observation or just the photos.

**Context (please complete the following information):**
 - Device: iPad Air (5th gen)
 - OS: iPadOS 16.6.1
 - App version: [v0.19.0+67](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.19.0%2B67)

Originally reported from iPhone 13 Pro, iOS 17.0.3 with this report:

> Make an obs and sync it. Then, edit it and add a photo to it and start syncing again. The sync animation stops at this point (see below), but the spinner keeps going ad infinitum. Go to another screen then back to the my obs screen and you’ll see the obs has synced correctly. Maybe related to https://github.com/inaturalist/iNaturalistReactNative/issues/905

**Additional context**
I noticed a few other hiccups when while trying to reproduce this, particularly when tapping "UPLOAD NOW" instead of saving and uploading all obs from the toolbar.

kueda:
Currently blocked by [#993](https://github.com/inaturalist/iNaturalistReactNative/issues/993)

joergmlpts:
I looked into this and noticed the new issue #993 as well. In a debug build on Android #993 looks like this:

![Screenshot_20231230-144138](https://github.com/inaturalist/iNaturalistReactNative/assets/66563749/6b0410c7-64fb-4d7c-9950-a335dd9383d4)

That happens in `Observation.appendObsPhotos()` where in
 
```
   const updatedObs = currentObservation;
[...]

PR #998, opened 2024-01-03
Fix for issues #918 and #993.

joergmlpts:


kueda (changes requested):
This introduces a different bug:

1. Follow the repro steps in #993; no more endless spinner, which is great
2. Save the observation
3. On MyObservations, the observation row now shows no photo, even though it does show a photo count indicator of 2
4. Tap on the observation
5. You get this error
    ```
    TypeError: Cannot read property 'toJSON' of null

    This error is located at:
        in PhotoDisplayContainer (created by ObsDetails)
        in RCTScrollContentView (created by ScrollView)
        in RCTScrollView (created by ScrollView)
        in ScrollView (created by ScrollView)
        in ScrollView
        in NativeWind.ScrollView
        in NativeWind.NoName
        in Unknown (created by ObsDetails)
        in RCTSafeAreaView
        in Styled
        in NativeWind.NoName
        in Unknown (created by ObsDetails)
        in ObsDetails (created by ObsDetailsContainer)
        in ObsDetailsContainer (created by SceneView)
    ```

joergmlpts:
That appears to happen here:

```
        // TODO replace this hack. Without this you get errors about the
        // photo objects being invalidated down in PhotoScroll, but the
        // questions remains, why are these objects getting invalidated in
        // the first place? We are not deleting them, so what's happening
        // to them and why?
        op => (
          op.photo.toJSON
            ? op.photo.toJSON( )
            : op.photo
        )
      )
```

joergmlpts:
Sorry for wasting your time.

Another attempt at a fix is coming soon.

budowski:
@joergmlpts - hint at what's causing the crash - that's because `toJSON` is a Realm object function (it's probably expecting an `ObservationPhoto` instance or similar), and instead it's getting a regular JS object

joergmlpts:
@budowski The crash in the original code happens because `updatedObs.observationPhotos` cannot be modified outside of a write transaction when `updatedObs` is the same object as `currentObservation` and `currentObservation` is a Realm object.

My illfated fix correctly used the JSON of `currentObservation` as `updatedObs` and thus it could modify `updatedObs.observationPhotos` - the drash was gone. It did not work properly because `currentObservationPhotos` was still obtained from the Realm object and that containted `photo` as a Realm object which downstream turned into `null`.

The new fix also uses the JSON of `currentObservation` as `updatedObs` but it obtains `updatedObs.observationPhotos` from this JSON object. Hence the issues downstream are gone. The type check is optional, the change also works when toJSON() is called for json object. I wanted to avoid a function call that might be a deep copy of an observation when it is not necessary.

PR #999, opened 2024-01-04
Add Photo Permission Gate

### LINKED ISSUE #830: Use Permission gate for photo library access on iOS
**Describe the bug**
On the first time using the app and taking a photo with the standard camera the user is presented with permission gates for camera permission and location permission. However, there seems to be no gate for the photo library permission.

**To Reproduce**
1. Fresh install of the app
2. Go to + (Start observation)
3. Use Standard camera
4. (Notice no native permission alert) Press "Observe Organisms"
5. Grant camera permission
6. Take photo
7. Press the checkmark button
8. Notice the native permission alert for ""iNaturalistReactNative" Would like to Add to your Photos. This permission has no permission gate?

**Expected behavior**
There should be a permission gate for the alert of granting photo library permission.

**Screenshots**

 - Device: iPhone 7
 - OS: iOS 15.7.9
 - App version: 0.14.0 + newer commits

albullington:
Closes #830 

- Adds a permission gate for saving photos to library. Permission Gate appears when a user taps the checkmark on the Standard Camera or snaps a photo / taps the checkmark on the AR Camera
- Code cleanup: using shared hooks for zoom, rotation, etc. in camera library to make camera functionality easier to reason about

### DIFF
--- a/src/components/Camera/ARCamera/hooks/useModelLoaded.js
+++ b/src/components/Camera/ARCamera/hooks/useModelLoaded.js
@@ -1,0 +1,81 @@
+// @flow
+
+import {
+  useState
+} from "react";
+import {
+  Platform
+} from "react-native";
+
+const useModelLoaded = ( ): Object => {
+  const [result, setResult] = useState( null );
+  const [modelLoaded, setModelLoaded] = useState( false );
+
+  const handleTaxaDetected = cvResults => {
+    if ( cvResults && !modelLoaded ) {
+      setModelLoaded( true );
+    }
+    /*
+      Using FrameProcessorCamera results in this as cvResults atm on Android
+      [
+        {
+          "stateofmatter": [
+            {"ancestor_ids": [Array], "name": xx, "rank": xx, "score": xx, "taxon_id": xx}
+          ]
+        },
+        {
+          "order": [
+            {"ancestor_ids": [Array], "name": xx, "rank": xx, "score": xx, "taxon_id": xx}
+          ]
+        },
+        {
+          "species": [
+            {"ancestor_ids": [Array], "name": xx, "rank": xx, "score": xx, "taxon_id": xx}
+          ]
+        }
+      ]
+    */
+    /*
+      Using FrameProcessorCamera results in this as cvResults atm on iOS (= top prediction)
+      [
+        {"name": "Aves", "rank": 50, "score": 0.7627944946289062, "taxon_id": 3}
+      ]
+    */
+    const standardizePrediction = finestPrediction => ( {
+      taxon: {
+        rank_level: finestPrediction.rank,
+        id: Number( finestPrediction.taxon_id ),
+        name: finestPrediction.name
+      },
+      score: finestPrediction.score
+    } );
+    let prediction = null;
+    let predictions = [];
+    if ( Platform.OS === "ios" ) {
+      if ( cvResults.length > 0 ) {
+        const finestPrediction = cvResults[cvResults.length - 1];
+        prediction = standardizePrediction( finestPrediction );
+      }
+    } else {
+      predictions = cvResults
+        ?.map( r => {
+          const rank = Object.keys( r )[0];
+          return r[rank][0];
+        } )
+        .sort( ( a, b ) => a.rank - b.rank );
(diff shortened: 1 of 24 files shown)


src/components/Camera/hooks/useCreateEvidence.js
```diff
@@ -0,0 +1,110 @@
...
+
+import { CameraRoll } from "@react-native-camera-roll/camera-roll";
+import {
+  useCallback
+} from "react";
+import Observation from "realmModels/Observation";
+import ObservationPhoto from "realmModels/ObservationPhoto";
+import useStore from "stores/useStore";
+
+import { log } from "../../../../react-native-logs.config";
```
kueda, line 11: FWIW, I added a shortcut for this: `import { log } from "sharedHelpers/logger";`

src/components/Camera/hooks/useCreateEvidence.js
```diff
@@ -0,0 +1,110 @@
...
+        .appendObsPhotos( obsPhotos, currentObservation );
+      observations[currentObservationIndex] = updatedCurrentObservation;
+      updateObservations( observations );
+      logger.info(
+        "addCameraPhotosToCurrentObservation, calling savePhotosToCameraGallery with paths: ",
+        evidenceToAdd
+      );
+      return savePhotosToCameraGallery( evidenceToAdd );
+    }
+    return createObsWithCameraPhotos( cameraPreviewUris, localTaxon );
```
kueda, line 91: There is a lot going on here, and I'm not sure this hook makes it easier for me to track it. The hook is named `useCreateEvidence`, but it returns a function that takes a taxon as the only arg... which doesn't seem to have anything to do with the media evidence. It seems more like this hook does a couple different things:

1. Saves new photos to the gallery
2. Create a new obs with new photos (which also saves photos to the gallery)
3. Updates zustand state with all these new values where they can be used by ObsEdit

It might make more sense to me if this was called `usePrepareStateForObsEdit` and exported functions that just do what they're called, like `createObservation` and `createPhotos`. Not sure that's the best approach. As the saying goes, people are usually right when they think something's wrong but almost always wrong when they propose a solution.

src/components/Camera/ARCamera/hooks/useModelLoaded.js
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  return {
+    handleTaxaDetected,
+    modelLoaded,
+    result
+  };
+};
+
+export default useModelLoaded;
```
kueda, line 81: This is another confusingly-named hook. It seems to do a lot more than simply track whether the model is loaded. Can you think of a name that describes the everything this hook does, or perhaps the most important thing it does?

kueda (commented):
As far as I can tell this works, so I'm not sure there's anything really blocking, but I do find some of the hooks a bit mystifying (see inline comments). One more non-blocking thing you might want to address:

- [ ] Delete PhotoLibraryPermissionGate (empty file)

src/components/Camera/hooks/useCreateEvidence.js
albullington: Ok, for now I renamed this. I agree there's a lot going on in here. The main point of moving everything into hooks for me was to start to separate the logic that happens when a photo is taken from the state of the camera (rotated, zoomed, etc.) since it was tricky to see which parts of the code were necessary for updating the PermissionGate. And now we're no longer reasoning about things like the AR model loading while looking at code for the StandardCamera (as example). But agree that this hook could probably be further simplified and detangled.

src/components/Camera/hooks/usePrepareStateForObsEdit.js
```diff
@@ -0,0 +1,109 @@
...
+    currentObservationIndex,
+    evidenceToAdd,
+    numOfObsPhotos,
+    observations,
+    savePhotosToCameraGallery,
+    updateObservations
+  ] );
+
+  return {
+    createEvidenceForObsEdit: localTaxon => createEvidenceForObsEdit( localTaxon )
```
kueda, line 105: Not sure if you want more feedback here, but FWIW, renaming the file helps, but I think this method is still misleading, since it does a lot more than just creating evidence.

PR #1002, opened 2024-01-08
Adding photo to already uploaded observation; fix for issue #918

joergmlpts:
The crash in the original code happens because `updatedObs.observationPhotos` cannot be modified outside of a write transaction when `updatedObs` is the same object as `currentObservation` and `currentObservation` is a Realm object.

My ill-fated fix correctly used the JSON of `currentObservation` as `updatedObs` and thus it could modify `updatedObs.observationPhotos` - the crash was gone. It did not work properly because `currentObservationPhotos` was still obtained from the Realm object and that containted `photo` as a Realm object which downstream turned into `null`.

This new fix also uses the JSON of `currentObservation` as `updatedObs` but it obtains `currentObservationPhotos` from this JSON object. Hence the issues downstream are gone. The type check is optional, the change also works when toJSON() is called for the json object as well. I want to avoid a function call that might perform a deep copy of an observation in the >98% of the calls where that is not needed.


src/realmModels/Observation.js
```diff
@@ -270,11 +270,13 @@ class Observation extends Realm.Object {
   };
 
   static appendObsPhotos = ( obsPhotos, currentObservation ) => {
+    const updatedObs = (currentObservation instanceof Realm.Object
+                        ? currentObservation.toJSON( ) : currentObservation);
+
```
kueda, line 275: This doesn't conform to our eslint rules. Please make sure you are running a linter in your editor and/or have set up [our precommit hooks](https://github.com/inaturalist/iNaturalistReactNative#set-up-pre-commit-hooks), which will run eslint and fix this kind of issue automatically before commit.

kueda (changes requested):
Cool, this seems to fix #918 and I'm not seeing additional issues. Once you fix that minor style issue this should be good to merge.

kueda (approved):

Issue #560, opened 2023-03-30
Scrolling to the top of MyObservations doesn't always bring back Header


albullington:
A user should be able to scroll to the top and see both the Header and Toolbar. Currently, scrolling to the top is inconsistent -- sometimes it shows the Header & Toolbar, sometimes it only shows the Toolbar, and sometimes it shows half of the header (the username, but not the Welcome back line of text).

<img width="377" alt="Screenshot 2023-03-30 at 1 59 59 PM" src="https://user-images.githubusercontent.com/35536439/228963309-1c73599e-9aa0-42e4-8cd8-07e46229e728.png">

kueda:
This is actually working pretty well for me, @albullington. Would you try and replicate again or close?

albullington:
@kueda included a screen recording with repro conditions. For the original issue here, the main thing that I'm seeing is that when a user scrolls to the top, sometimes the whole header is displayed, but sometimes the "Welcome back" text is cut off.

This screen recording also highlights a separate (maybe related) issue, that when a user scrolls down and loads more observations, the top part of the header disappears suddenly so only the toolbar is displayed.

https://github.com/inaturalist/iNaturalistReactNative/assets/35536439/6de931cc-9c32-47a6-ac85-b11451dbb063

So things to fix here are:

- [ ] Welcome back text is occasionally cut off when scrolling up on MyObservations
- [ ] Top half of header disappears suddenly when user pulls down to load more observations

kueda:
What device are you using? I'm not seeing that on an iPad Air (5th gen) or an iPhone SE, so maybe it has to do with using a device with a camera in the screen area?

albullington:
This is on iPhone 15 Pro -- maybe has something to do with the SafeAreaView, if you're not seeing it on the other iOS devices

PR #1032, opened 2024-01-15
#560 - my observations - over-scroll header bug fix

budowski:


kueda (changes requested):
I'm not sure this is an improvement. I'm seeing significant flicker around the header under normal scrolling conditions:

https://github.com/inaturalist/iNaturalistReactNative/assets/22691/550d8cc8-adaa-4983-906d-60302b86c9a6

Is there anything we can do about that?

budowski:
@kueda - I think I can make it a conditional state update - meaning, trigger the state update using a debounce (not on every update) + only update in case of over-scroll (when scrolling past the top of the list)

kueda (approved):

PR #1062, opened 2024-01-20
Obs details sounds

### LINKED ISSUE #885: Obs Detail screen; observation with sound first
When a user lands on an observation with a sound as the first piece of media. In the top section of the Obs Detail screen, show the play button and duration controls, according to the recent [designs](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=8379-46076&mode=dev). The screen starts in "Paused" state.
When tapped, the sound starts to play, and the UI switches to the pause "Playing" state. The Playing state shows a Pause button instead of the Play button, which when tapped pauses the playing of the sound.
In the Playing state, the Duration text as well as the slider show the current time of the playing recording, and the slider can be dragged to set a time at which to continue playing the recording after release of the sliding motion.

When the observation has multiple media, swiping to the next photo/or sound should not loose the state of the current paused time if playing the sound has been paused.

kueda:
* Adds sound player to ObsDetails
* Allow offline viewing of media on ObsDetails (if default cache has the image)
* Sticky nav header on ObsDetails until you scroll past tabs
* Refactor of components toward the top of ObsDetails (media, fave button, header)

Note that this introduces a somewhat significant bug by inventing UUIDs for remote sounds, because they're not available in API responses (pending https://github.com/inaturalist/iNaturalistAPI/issues/422). It's not really a problem at present since the user can't delete individual sounds yet, but it will be when that needs to get done.

Closes #885


kueda:
Some things that need to be fixed up before this should be merged:

- [ ] Sound needs to stop playing when the user leaves ObsDetails
- [ ] Player needs to be destroyed or reset when leaving ObsDetails (sometimes tapping play on a different obs plays the sound from an older obs)

Some other TODOs that don't need to block merge

- [ ] Progress bar indicator needs to support drag interaction for seeking forward and backward
- [ ] Progress bar needs to support touch interaction to support seeking
- [ ] Play button gets cut off by the progress bar in Android

kueda:
Alrighty, I did fix up the problem with playback continuing after you navigate away from ObsDetails, but I did not fix the problem with the player often playing a sound from a previous observation because it only seems to happen in Android and I'm almost out of ideas. It's possible we need to use a class-based component like the author of the library [uses in their example app](https://github.com/hyochan/react-native-audio-recorder-player/blob/main/Example/App.tsx#L146), or maybe this is a legit bug in the Android implementation of that library. Either way, I'm going to make a separate post-MVP issue for that.

Issue #1000, opened 2024-01-04
Remove positional_accuracy requirement for imported and existing observations


kueda:
Currently we have a soft requirement that observations have a `positional_accuracy` of 100 meters and a hard requirement of 500,000 meters, but a lot (most?) photos imported into the app will not have any accuracy info associated with them, and most people will not be able to improve on missing or very large `positional_accuracy` values, creating a frustrating experience where the user feels they've done something wrong but can't fix it.

Instead, let's remove the accuracy requirements for imported observations, while keeping the coordinate requirement.

- [x] When you import a photo and land on ObsEdit, there should be no warning if `positional_accuracy` is blank
- [x] When you import a photo and land on ObsEdit, there should be no warning if `positional_accuracy` is over 100 meters
- [x] When you import a photo and land on ObsEdit, there should be no warning if `positional_accuracy` is over 500,000 meters
- [x] When you create a new observation without evidence or with one of the cameras, these requirements and warnings should still apply
- [x] When you edit an existing observation, there should be no warning about `positional_accuracy`

PR #1089, opened 2024-01-29
#1000 - obs edit - positional_accuracy check - don't perform the check for new observations with imported photos

budowski:


kueda (changes requested):
Ok, removal of warnings works well, but it went a bit too far:

1. Start a new obs w/o evidence
2. Edit the location
3. Zoom out until the acc is over 600,000
4. On ObsEdit there red triangle warning appears next to evidence (good)
5. Tap SAVE
6. The obs saves w/o showing the [Location Too Imprecise bottom sheet warning](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=718-17887&mode=design&t=8FdFj0MO163u5pV6-4)

It should still show that bottom sheet when you manually set the location like that.

Now, if you,

1. Edit that observation w/ the enormous acc
2. Tap the SAVE button
3. You see the MISSING EVIDENCE bottom sheet

It's good that we're showing a bottom sheet here, but it's showing the wrong content.

kueda (approved):

Issue #1028, opened 2024-01-12
Exif data error


albullington:
<img width="352" alt="Screenshot 2024-01-12 at 12 19 50 PM" src="https://github.com/inaturalist/iNaturalistReactNative/assets/35536439/12838725-a433-4f92-b1bd-c0a060f3e438">

I get warnings about this promise rejection in pretty random places, like while I'm on the Explore screen or sometimes when I first start the app. We should make sure we're only parsing EXIF data when we need to.

budowski:
@albullington - I did a tiny fix (info instead of warning when encountering an EXIF error), in a new PR - however, I double checked all the code, this flow could only happen when either: Sharing a photo into iNat or importing a photo.
Could it be that these are older warnings (not "fresh") that were left in the call stack while you were previously using the app?

albullington:
@budowski I guess that's possible! It's a warning I was seeing pretty regularly, but I didn't pay much attention to whether I was doing a fresh install of the app or not before seeing it.

kueda:
So you have no repro conditions here, @albullington? Not even related to importing photos that might not have coordinates?

albullington:
No, it doesn't pop up for me specifically around imports -- I'd say I've seen it the most when tapping Explore on the TabBar, when I'm not importing anything at all. It's also possible it's happening for me in the simulator more than on a real device -- I can keep an eye out for this.

albullington:
Closing this for now since I'm not seeing it in the latest build.

PR #1107, opened 2024-02-04
#1028 - info instead of warning in case of EXIF parsing error

budowski:


kueda (changes requested):
Two problems:

1. There's an extraneous file that doesn't seem to have anything to do with #1028
2. This just silences a potentially useful error instead of solving the problem, which seems to be that EXIF is getting parsed long after an observation was imported

kueda:
We need better repro conditions to address #1028

Issue #1051, opened 2024-01-17
Cancelling photo import often takes you to the wrong screen


kueda:
**Describe the bug**
Sometimes cancelling a photo import lands you on a weird screen, like ObsEdit or StandardCamera

**To Reproduce**
1. Sign in
2. Start a new obs without evidence
3. Let the location load, then back out; you should be back on MyObs
4. Start a photo import
5. Tap "Cancel" in the upper left of the photo chooser UI
6. You've landed on ObsEdit again for some reason

**Expected behavior**
When you cancel a photo import, you should return to wherever you were when you started, so in the above steps, you should return to MyObs, but if you started from ObsDetails, you should return there.

**Context (please complete the following information):**
 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.2.1
 - App version: [v0.20.0+68](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.20.0%2B68)

**Additional context**
Maybe related to #943. There are a number of ways to repro this, usually requiring you to go to one screen first, then cancelling an import.

PR #1109, opened 2024-02-04
#1051 - when canceling photo import - go back to correct screen

budowski:


kueda (changes requested):
This always returns you to MyObs, which addresses the exact repro steps in #1051, but not the following:

> When you cancel a photo import, you should return to wherever you were when you started, so in the above steps, you should return to MyObs, but if you started from ObsDetails, you should return there.

kueda (approved):

PR #1110, opened 2024-02-05
Image predictions uri fix

### LINKED ISSUE #1038: Offline suggestions don't work in Android
**Describe the bug**
Offline suggestions don't work in Android

**To Reproduce**
1. Put the device in airplane mode
2. Import a photo of something you know the offline model will do well on
3. Tap "ADD AN ID" when you land on ObsEdit
4. On Suggestions, you get the activity indicator, but the offline suggestions never load

**Expected behavior**
Offline suggestions should load while offline

 - Device: Pixel 8
 - OS: Android 14
 - App version: [v0.20.0+68](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.20.0%2B68)

**Additional context**
I'm fairly sure this is a bug in https://github.com/inaturalist/vision-camera-plugin-inatvision that I had to [work around](https://github.com/inaturalist/iNaturalistReactNative/pull/1025/files#diff-3eb02ad95f41c6eab40496101db1257c63de98e39479f2e0c9d17e9021131de1R32-R37) here because it was causing a crash, but I don't feel like I know enough about the problem to file an issue there. Clearly android.graphics.Bitmap.getWidth() is getting called on a null object, but I don't know how to replicate it only within the context of that package.

jtklein:
This PR includes a new version of the vision-camera-plugin-inatvision after a merged PR there should fix the problem with reading images from file given a uri as input. With this PR the error of trying to `getWidth` from a null should not happen anymore. https://github.com/inaturalist/vision-camera-plugin-inatvision/pull/11

This should fix #1038 but I was not able to reproduce the wrong = error state on my devices, so I would need a review if this fixes the error seen on @kueda 's device as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25692,7 +25692,7 @@
     },
     "node_modules/vision-camera-plugin-inatvision": {
       "version": "0.1.0",
-      "resolved": "git+ssh://git@github.com/inaturalist/vision-camera-plugin-inatvision.git#350639f718fe0b5e14c8a7c0bb53cb7ecf929a4b",
+      "resolved": "git+ssh://git@github.com/inaturalist/vision-camera-plugin-inatvision.git#8526519edd6983b59e3ecf826bb26220b0d85f51",
       "license": "MIT",
       "engines": {
         "node": ">= 16.0.0"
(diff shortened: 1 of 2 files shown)


kueda (approved):
Looks good to me, but I'm still weirded out that you can't replicate.

Issue #1093, opened 2024-01-30
Auto-populate metadata from subsequent photos, ObsEdit


albullington:
**Describe the bug**
An observation should try to auto-populate the date/time and location, even if the first photo in an observation is missing either a date/time or a location. For example, if an observation has two photos and the second photo has a location but the first does not, the observation should use the second photo's location.

**To Reproduce**
1. Create a new observation with multiple photos, where the first photo is missing either a datetime, location, or both
2. ObsEdit will display datetime/location as missing, even if the second photo has that metadata
3. If new media is added to the unuploaded observation, ObsEdit will still display datetime/location is missing, even if the new media has metadata

**Expected behavior**
ObsEdit should display the first datetime and location found in the evidence list. If the second photo has that data, ObsEdit should display the data from the second photo and apply it to the observation.

**Context (please complete the following information):**
 - App version: 0.21.1

**Additional context**
In the future, we might want to consider other use cases, like when a user adds new media to an observation that's already uploaded remotely. However, if we're looking for web parity, adding evidence with more metadata to a previously uploaded observation on the web does not appear to add that metadata to the observation. So I think we should restrict this issue to observations which have not yet been synced with web for now.

PR #1128, opened 2024-02-11
#1093 - when importing photos, try to fill out EXIF data (location+da…

budowski:
…te) from all photos, by order


kueda (changes requested):
This addresses the first (1) repro case at https://github.com/inaturalist/iNaturalistReactNative/issues/1093, but not number 3, adding a photo to an existing unuploaded observation on ObsEdit.

kueda (approved):

Issue #940, opened 2023-11-30
Login flickers when typing username


kueda:
**Describe the bug**
When typing your username on the Login view the screen flickers with each key press.

**To Reproduce**
1. Sign out if signed in
2. Tap "Log in to iNaturalist" in the drawer menu to get to the Login screen
3. Start typing in the USERNAME OR EMAIL field. The screen above the keyboard flickers with each key press

**Expected behavior**
The screen should not flicker while signing in.

**Context (please complete the following information):**
 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.1.1
 - App version: [v0.17.1+64](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.17.1%2B64)

**Additional context**
Replicated on an iPhone 15 Pro and iPhone 12 mini too. Flickering does not happen when entering your password, and doesn't seem to happen on iPad or on Android.

kueda:
Angie could not replicate on an iPhone X 16.4

PR #1133, opened 2024-02-12
Allow password managers to fill in login fields

kueda:
Main required change here is `textContentType` prop in iOS.

### DIFF
--- a/src/components/LoginSignUp/LoginForm.js
+++ b/src/components/LoginSignUp/LoginForm.js
@@ -80,21 +80,24 @@ const LoginForm = ( {
         accessibilityLabel={t( "USERNAME-OR-EMAIL" )}
         autoComplete="email"
         headerText={t( "USERNAME-OR-EMAIL" )}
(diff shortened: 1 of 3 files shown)


albullington:
doesn't this reintroduce the screen flickering on iOS?

kueda:
I wasn't seeing that when I was testing on my iPhone SE today, @albullington, but I'll test again (and with the iPad) tomorrow). If you're seeing flicker in this branch on another iPhone, please holler.

kueda:
Alrighty, this does indeed re-introduce some flicker while typing on the login form, which is a [known bug](https://github.com/facebook/react-native/issues/39411) that Facebook doesn't think is their problem, and that Amanda fixed in #940 by changing `textContentType` to `oneTimeCode`. We [discussed](https://inaturalist.slack.com/archives/C02G5AATVGA/p1707867090475759) in Slack and the rough consensus is that access to a password manager is worth some flicker, at least for now.

PR #1155, opened 2024-02-15
Vision plugin add possibility to set confidence threshold for predictions from file

jtklein:


kueda (approved):
Looks ok. I'm ok w/ trying a 0.2 threshold for now. What I really wanted to do was provide a control so we can experiment with what threshold seems ok, but we can do that later.

I also really wish our references to this lib were versioned, even if it's just a hard reference to a tag, but that's not a blocker.

PR #1161, opened 2024-02-21
Some changes in Highest and Lowest Rank filter modals

### LINKED ISSUE #1139: hrank and lrank Explore filters
They're only partially there at present

- [x] Lists should be scrollable
- [x] I don't think we need radios, a single tap should select and then we don't need an additional button
- [x] Needs a header, e.g. LOWEST RANK and HIGHEST RANK
- [x] Should not flow under the status bar
- [x] Should be dismissible without making a choice, i.e. "x" button should work
- [x] Should be an "Any" or "None" option (I prefer the latter)

jtklein:
I could not figure out how to show a list of scrollable items in this modal. So, I changed it to use a Picker UI instead of a scrollable list. I hope this is enough for the MVP, and we can create a separate issue for further changes required after the release as for now it is at least functional.

So, this will close #1139 for now.


kueda (changes requested):
> Should be dismissible without making a choice, i.e. "x" button should work

This is only true in the iPhone implementation. In Android you have to make a choice. We should fix this eventually, but I don't think this needs to block merge.

The translation key issue is blocking, though, though admittedly kind of pedantic.

Issue #927, opened 2023-11-27
Settings UI


kueda:
Settings should be mostly functional, but it has not been brought in line with designs... which do not exist yet. This is a placeholder for that work.

jtklein:
As a prerequisite, the link to the settings page in the app drawer should be made to be only visible when a user is logged into the app. In other words if the current user is logged out do not show the link to the settings page in the drawer. Designs for the link text and icon are [here](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=3835-22631&mode=design&t=bsXQl5S2iInXhufe-4).

The settings page itself should be implemented according to the recent designs [here](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=11087-28248&mode=design&t=bsXQl5S2iInXhufe-4): 

- [ ] Style the header and back button according to the designs
- [ ] (Taxon names display section): Add the UI according to the designs
- [ ] (Taxon names display section): Changing this setting here changes it globally for this user, i.e. in the app as well as on the website. So, I guess what I am saying is check the website code on how to update this setting via the API.
- [ ] (iNaturalist account settings section): Add the UI according to the designs
- [ ] (Taxon names display section): Pressing on the button in this section should open an in-app browser to the logged-in user's account settings page on the website, e.g. https://www.inaturalist.org/people/1/edit

PR #1170, opened 2024-02-24
#927 - account settings

budowski:


src/components/FullPageWebView/FullPageWebView.js
```diff
@@ -0,0 +1,58 @@
+import { useFocusEffect, useNavigation, useRoute } from "@react-navigation/native";
+import { getAPIToken } from "components/LoginSignUp/AuthenticationService";
+import { ViewWrapper } from "components/SharedComponents";
+import React, { useState } from "react";
+import WebView from "react-native-webview";
+
+const FullPageWebView = () => {
```
jtklein, line 7: The designs for how the full page in-app browser should look like are here:
https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=4292-26271&mode=design&t=P0wnKV1p3oYLhoTF-4

src/components/Settings/Settings.js
```diff
@@ -1,165 +1,30 @@
...
-  Text,
-  View
+  StatusBar, View
 } from "react-native";
+import { useTranslation } from "sharedHooks";
 import useAuthenticatedMutation from "sharedHooks/useAuthenticatedMutation";
 import useUserMe from "sharedHooks/useUserMe";
-import { textStyles, viewStyles } from "styles/settings/settings";
-
-import SettingsAccount from "./SettingsAccount";
```
jtklein, line 20: Should we also remove all those files if they are not used anymore?

kueda (changes requested):
I merged from main to resolve conflicts and made a few minor changes, but updating react-native-webview introduces [this problem](https://github.com/react-native-webview/react-native-webview/issues/2959) with our testing setup. Do we need to be on v13? If so, please fix those tests. You probably need to write some mocks for both react-native-webview and webshell.

budowski:
@kueda - seems from the RN webview's release history, we can use v12.4.0 - I'll change the version and update the PR

budowski:
Shoot - misunderstood - it required downgrading to v11.23.1 - will fix

src/components/Settings/Settings.js
kueda: We probably don't need the styles, but I'm going to leave the bulk of the existing Settings components in case we change our minds and want to ditch the webview approach.

kueda (approved):

Issue #1046, opened 2024-01-16
Change display name of app to "iNat Beta"


kueda:
If there are long and short options, long should be "iNaturalist Beta" and short should be "iNat Beta." Note this should only affect the name displayed under the icon, we don't need to rename the repo or change the package name or anything.

PR #1174, opened 2024-02-25
#1046 - change app name to iNat beta

budowski:


kueda (approved):
Made a change so the name fits w/o getting ellipsized, otherwise looking good.

PR #1218, opened 2024-02-27
Taxon detail map and taxon map screen

### LINKED ISSUE #872: Taxon Detail map and Taxon Map screen
On the Taxon Detail screen, add a map preview component with section header following the [designs](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=2255-16637&mode=dev). The map view is a static representation, no interaction apart from pressing is possible, so no pan or zoom. The map preview shows only a region that "makes sense" based on the observations shown. So, the map extent depends on the range of the taxon (plus some padding to look better).
Pressing the map area opens a full screen modal screen on top of the Taxon Detail screen. This screen component needs to be added according to the most recent [designs](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=2414-22084&mode=dev). This full screen map is very similar to the Explore map and maybe reused with the same component.

- [x] The styling of the header section follow the Taxon Detail screen, therefore requires these issues (#871 and #873).
- [x] Add a Close button that on press closes this modal screen and shows Taxon Detail again.
- [x] Add a Map component to the screen. The map should pan and zoom. Even though the design specs show different shapes and polygons we only need to show all observations of this taxon worldwide. The design of how to show those observations is the same as on the Explore screen map.
- [x] Add a Maplayer button. On pressing this button the map toggles between Map and Satellite on iOS and Map, Satellite, and Terrain layers on Android.
- [x] Add a CurrentLocation button. On press is the same as on Explore screen map.

### LINKED ISSUE #1233: Disable map interaction on ObsDetails and TaxonDetails
Neither of these maps should support any pan or zoom interactions. They should only support a tap that opens up a detail map.

angielt:
-Taxon Detail map and map screen
- Moved DetailsTab to SharedComponents
- Map Preview may take a while to load and the change between map locations can be seen when you switch to a different taxon detail page (TODO: add a loading indicator for map preview or change how it loads)

Closes #872
Closes #1233


kueda (changes requested):
The map on TaxonDetails is not showing observation tiles for that taxon. It looks like it's showing tiles for all observations. It's also not zooming in on the extent of the taxon's observations.

kueda (approved):
FYI, I fixed a few things. Maybe more than a few. The bit that's probably the most worth your time to review is calculating the region given the observation bounds in https://github.com/inaturalist/iNaturalistReactNative/pull/1218/commits/521307f44d700353e9a1b83e947f788702f3f743#diff-f6d6a4b58e8ce0a2bb53983d3cc7c751763ff752af8b926ebba9d3f54bae3fc7

Issue #1211, opened 2024-02-27
The tab bar should not be visible on obs edit, even when editing an observation


abhasinat:
**Describe the bug**
When viewing obs edit as reached through the ObsCreate flow, you can't see the tab bar, and Obs Edit/Obs Create sits as a modal on top of all other screens. But if you edit an observation from home screen/my observations (whether it's saved or uploaded), you can see the tab bar, which you shouldn't be able to.

**To Reproduce**
1. Go to Home screen/My observations
2. Click any observation you've uploaded
3. Click the edit icon in the top right to open obs edit

OR
1. Go to Home screen/My observations
2. Click a saved observation to open obs edit

**Expected behavior**
Just like on obsedit through obscreate, obs edit when editing an existing or saved observation should not show a tab bar, and it should sit as a modal on top of everything else.

PR #1246, opened 2024-03-03
#1211 - don't show tab bar when editing an observation

budowski:


kueda (changes requested):
When I get to ObsEdit from ObsDetails and tap the back button, the header on ObsDetails is messed up: only the edit button appears and its in the upper left corner for some reason: 
Simulator Screenshot - iPhone 15 Pro - 2024-03-07 at 10 37 01

kueda (changes requested):
This solves the header problem, but now the StatusBar disappears from ObsDetails when you return from ObsEdit. The only situation where the StatusBar should not appear is in the cameras.

PR #1254, opened 2024-03-05
Taxon result updates

### LINKED ISSUE #1195: TaxonResult in ARCamera rarely shows remote data
**Describe the bug**
The ARCamera loads local taxon data, but it shows TaxonResult which attempts to load remote data (w/ common names and photos), but it rarely succeeds, sometimes only after leaving the ARCamera and coming back.

**To Reproduce**
1. Install a fresh copy of the app
2. Point the ARCamera at something you know it will get (e.g. a picture of a robin or something)
3. Chances are only the scientific name will appear, and no photo
4. Leave the ARCamera and try again
5. The common name and photo might appear, but might not

**Expected behavior**
The common name and photo should appear if you have reasonable network connectivity.

 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.3.1
 - App version: 0.23.0 (73)

**Additional context**
I looked into this a bit and commited b88de5809c8bd69f20814eb4dc6947cd2e86485a which will handle a lot of these problems, but here's some remaining stuff to look into:

* `Taxon.saveRemoteTaxon` returns a promise that we are not awaiting or catching, so if there's a problem in there, we're not seeing it
* We should continue saving to Realm, but we could also transform the API response to a Realm-like object literal and return that immediately in `useTaxon()`
* useTaxon() might not need to make a network request at all if there's a Realm Taxon; refreshing the local copy is good

Could also benefit from some UI improvements
- [ ] Show an ActivityIndicator instead of the photo if there is no photo URL and you're online (there will be some cases where the taxon actually has no photo, so maybe only do this when we're sure we haven't loaded remote data)

kueda:
Several minor changes to TaxonResult, mostly focused on improving ARCamera

* Stop requesting remote taxon if local exists and was updated in the last week
* Should show a localized version of a remote taxon without needing another render to load a newly-created local taxon
* Show ActivityIndicator while attempting to load a remote taxon
* Show iconic taxon icon if loading a remote taxon fails or there is no taxon photo
* Show iconic taxon as backdrop for ObsImage whenever possible, e.g. while loading the photo or when photo can't load

Closes #1195

### DIFF
--- a/tests/integration/sharedHooks/useTaxon.test.js
+++ b/tests/integration/sharedHooks/useTaxon.test.js
@@ -1,8 +1,32 @@
-import { renderHook } from "@testing-library/react-native";
+import { waitFor } from "@testing-library/react-native";
+import inatjs from "inaturalistjs";
 import safeRealmWrite from "sharedHelpers/safeRealmWrite";
 import { useTaxon } from "sharedHooks";
-import factory from "tests/factory";
+import factory, { makeResponse } from "tests/factory";
 import faker from "tests/helpers/faker";
+import { renderHook } from "tests/helpers/render";
+import setupUniqueRealm from "tests/helpers/uniqueRealm";
+
+// UNIQUE REALM SETUP
+const mockRealmIdentifier = __filename;
+const { mockRealmModelsIndex, uniqueRealmBeforeAll, uniqueRealmAfterAll } = setupUniqueRealm(
+  mockRealmIdentifier
+);
+jest.mock( "realmModels/index", ( ) => mockRealmModelsIndex );
+jest.mock( "providers/contexts", ( ) => {
+  const originalModule = jest.requireActual( "providers/contexts" );
+  return {
+    __esModule: true,
+    ...originalModule,
+    RealmContext: {
+      ...originalModule.RealmContext,
+      useRealm: ( ) => global.mockRealms[mockRealmIdentifier]
+    }
+  };
+} );
+beforeAll( uniqueRealmBeforeAll );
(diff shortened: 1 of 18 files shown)


tests/integration/sharedHooks/useTaxon.test.js
```diff
@@ -1,34 +1,66 @@
...
   }
 } );
 
-jest.mock( "sharedHooks/useAuthenticatedQuery", () => ( {
-  __esModule: true,
-  default: ( ) => ( {
-    data: mockRemoteTaxon
-  } )
-} ) );
-
```
kueda, line 19: @albullington I don't think you need to review this whole PR (unless you want to), but you might want to check out the changes to this test file. Here specifically, I chose not to mock `useAuthenticatedQuery` because this is an integration test, so we *want* to see how `useTaxon` integrates with the code in this repo, like `useAuthenticatedQuery`. Instead, for most of these tests, I've mocked `inatjs` responses, because `inatjs` is outside of this repo and thus not something we need to test in an integration test.

To achieve that, I needed to make our own version of `renderHook()` that includes our `QueryClient` so React Query will actually make those requests.

I also expanded upon the tests to make sure that the API gets called when it should. The main place I struggled was testing a request *failure*, and there I had to mock React Query itself, which is not great since we're not testing our own error processing code, but I couldn't figure out another way to do it.

albullington: Thanks for flagging. It's helpful to review this and also have it in mind as a model for future tests.

PR #1255, opened 2024-03-06
Cache user icon to resolve flicker

### LINKED ISSUE #1217: Bottom tabs user icon flickers when loading
**Describe the bug**
The profile pic / user icon in the bottom tabs flickers while loading under a poor Internet connection.

**To Reproduce**

This is hard to reproduce b/c you need a less than great Internet connection

1. Sign in, make sure the profile pic in the bottom tabs loads
1. Force quit the app
1. Go somewhere with an internet connection that isn't great... or use a network conditioner to simulate that.
3. Open the app
4. Watch the profile pic in the MyObs bottom tabs item; it will often flicker between your profile pic and the icon, and while it does so the rest of the bottom tabs seem unresponsive

**Expected behavior**
Either the profile pic should load or it should not, but it shouldn't flicker and/or disable the bottom tabs

 - Device: Pixel 8
 - OS: Android 14
 - App version: 0.23.0 (72)

I'm not convinced this is only an Android problem. I just see it there b/c that's the phone with the SIM card.

**Additional context**
Not sure what we're doing to show that pic now, but it feels like it's making too many network calls and loading the image over the Internet. IMO, a better approach would be to look for a cached file in the app file storage, and if it doesn't exist, asynchronously fetch users/me and download the profile pic there. Do that on sign in and sign out, and on app boot. Maybe do it every 10 minutes, or after you save your user settings.

albullington:
Closes #1217 

This felt kind of convoluted, so requesting a review to make sure this fixes the actual problem. A few new things here:

- Fetch user icon_url on the login screen before navigating to MyObservations, so icon can load immediately
- Use `FastImage` for caching the UserIcon
- Bust the cache for user's icon once on each app open, so if a user updates their icon on the website, it will be reflected in the cache
- Return cached images when available for current user in `User.uri`

For the end user, this should resolve flickering, allow a user to see their updated profile picture each time they load the app, and display the user profile icon immediately upon login.


kueda (commented):
I think removing `useIsConnected` will probably fix #1217 by itself, and fetching the user after sign in and using FastImage both seem like nice improvements.

I don't understand what useBustUserIconCache and user.cached_icon_url are doing, though. The `icon_url` in the users/me response should change whenever you change your profile pic, so there should not be a need to add a cache-busting param. Is there a specific bug you were trying to address with that part?

kueda (approved):

PR #1267, opened 2024-03-09
Improve algorithm for detecting nearby points when tapping on Explore map

budowski:


kueda:
Thanks for this, @budowski. This makes the map way more usable.

Issue #848, opened 2023-11-03
Importing a photo generated in the app does not import the date


kueda:
**Describe the bug**
Importing a photo that was generate by this app does not populate the date.

**To Reproduce**
1. Take a photo using the StandardCamera and save the obs
2. Confirm that the photo was saved out in a photos app (e.g. Apple Photos or Google Photos) and confirm that it has a date
3. Add a new obs by importing a photo
4. Import the photo you took in the app
5. The date field does not get filled in

**Expected behavior**
Photos that this app generates should write enough metadata that they can be imported again as additional observations. This definitely applies to the date, but also the coordinates.

**Context (please complete the following information):**
 - Device: iPad Air (5th gen)
 - OS: iPadOS 16.6.1
 - App version: [v0.15.0+61](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.15.0%2B61)

**Additional context**
I'm not sure if this is an export or an import problem. If it's an import problem it *might* be solved in https://github.com/inaturalist/iNaturalistReactNative/pull/828

jtklein:
I can not replicate this on my iPad with version 0.16.0 which includes #828.
Was this only on iPad or also on Android? Can you still reproduce this on your specific device?

budowski:
Cannot recreate this with latest version (tried both Android and iPhone physical devices) - @kueda / @jtklein - still not working for you?

kueda:
Yes, I can still replicate on my iPad Air, and also on an iPhone SE (3rd Gen).

kueda:
TODO, replicate and check the logs, see if there's anything there.

kueda:
Ok, I can replicate this on an iPhone SE. I've spent the last two days investigating what's going on and my conclusion is that at least on an iPhone SE, when you take a photo in the StandardCamera and save the observation, the `DateTimeOriginal` EXIF tag never gets written, which is what react-native-exif-reader uses to populate the `date` attribute it returns from its `readExif()` method.

IMO, the quickest fix would be to write the `DateTimeOriginal` tag like we write the GPS tags, but I am blocked in that by a [bug that creates duplicate photos when you write EXIF](https://github.com/inaturalist/react-native-exif-reader/issues/5).

Alternatively, we could update that library to return the GPS timestamp if DateTimeOriginal isn't there, but that's just a hack around the real problem that our camera isn't writing the EXIF tags that it should, certainly not the minimum set that the Apple Camera app does.

[...]

PR #1268, opened 2024-03-10
Use latest react-native-exif-reader package version; use writeLocation instead of writeExif

budowski:
See comment here:
https://github.com/inaturalist/react-native-exif-reader/issues/5#issuecomment-1987270604


src/sharedHelpers/parseExif.js
```diff
@@ -52,7 +52,7 @@ interface ExifToWrite {
 export const writeExifToFile = async ( photoUri: ?string, exif: ExifToWrite ): Promise<Object> => {
```
kueda, line 52: Since this no longer writes arbitrary EXIF, it should be renamed `writeLocation` and accept only the arguments it is capable of writing, e.g. `latitude` and `longitude` (can we write horizontal accuracy to a PHAsset?).

kueda (changes requested):
I think this should be merge-able after renaming that method, but it doesn't unblock me for #848, which specifically requires writing `DateTimeOriginal`, which I guess we can't do for an existing PHAsset. Instead, I'd like to write the EXIF to file *before* we add it to the gallery, but I'm not sure if that's possible in iOS, hence my question at https://github.com/inaturalist/react-native-exif-reader/issues/5#issuecomment-1998705470

src/sharedHelpers/parseExif.js
budowski: I think we should consider switching to [react-native-exify](https://github.com/lodev09/react-native-exify/) library - even though it's new, it does everything we need, including support for both `PHAsset` and regular URIs - for example here:
https://github.com/lodev09/react-native-exify/blob/main/ios/ExifyUtils.swift

What do you think?

kueda: It doesn't seem to write `GPSHPositioningError` in iOS, nor does it read it in iOS when we can access it (i.e. when we're not going through PHAssets), so I'd be up for using it if you want to fork it and add that functionality. Otherwise I'd prefer to work on our own library. There's a high probability that if we start using `react-native-exify` we're going to have to end up supporting a fork of it when it gets abandoned anyway.

budowski: Mmmm, true, didn't notice that issue - OK, I'll do the changes on our end

ios/iNaturalistReactNative.xcodeproj/project.pbxproj
```diff
@@ -436,6 +436,7 @@
 			);
 			name = "[CP] Check Pods Manifest.lock";
 			outputFileListPaths = (
+				"${PODS_ROOT}/Target Support Files/Pods-iNaturalistReactNative/Pods-iNaturalistReactNative-resources-${CONFIGURATION}-output-files.xcfilelist",
```
kueda, line 439: This is causing the following build errors for me in iOS:

```
❌  error: Multiple commands produce '/Users/kueda/Library/Developer/Xcode/DerivedData/iNaturalistReactNative-bodcddsbdubblxajopponrkjzdoy/Build/Products/Debug-iphoneos/iNaturalistReactNative.app/RNCAsyncStorage_resources.bundle'
❌  error: Multiple commands produce '/Users/kueda/Library/Developer/Xcode/DerivedData/iNaturalistReactNative-bodcddsbdubblxajopponrkjzdoy/Build/Products/Debug-iphoneos/iNaturalistReactNative.app/AntDesign.ttf'
❌  error: Multiple commands produce '/Users/kueda/Library/Developer/Xcode/DerivedData/iNaturalistReactNative-bodcddsbdubblxajopponrkjzdoy/Build/Products/Debug-iphoneos/iNaturalistReactNative.app/Entypo.ttf'
```

and that list goes on. Does this need to exist? Without it I think this is working as expected.

kueda (changes requested):

kueda (approved):

PR #1283, opened 2024-03-13
Try to catch and log errors during upload

kueda:
* Wraps some upload async methods with try/catch to attempt to log some potential bugs
* Adjusts upload param mappings
* Try to get upload status to more accurately reflect the upload state


src/components/MyObservations/ToolbarContainer.js
```diff
@@ -112,7 +112,7 @@ const ToolbarContainer = ( {
     if ( uploadInProgress ) {
       const translationParams = {
         total: numToUpload,
-        currentUploadCount: numFinishedUploads + 1
+        currentUploadCount: Math.min( numFinishedUploads + 1, numToUpload )
```
kueda, line 115: At least in tests I was seeing some cases of "Uploading 3 of 2 observations" so this should prevent that... I hope.

src/components/MyObservations/ToolbarContainer.js
```diff
@@ -122,13 +122,15 @@ const ToolbarContainer = ( {
       return t( "Uploading-x-of-y-observations", translationParams );
     }
 
+    if ( numUnuploadedObs && numUnuploadedObs !== 0 ) {
+      return t( "Upload-x-observations", { count: numUnuploadedObs } );
+    }
+
```
kueda, line 128: This might be a bit off spec, but if you upload one of several obs, I think it should say there are still obs to upload when it's done.

src/realmModels/ObservationPhoto.js
```diff
@@ -32,7 +32,9 @@ class ObservationPhoto extends Realm.Object {
 
   static mapPhotoForUpload( observtionID, observationPhoto ) {
     return {
-      "photo[uuid]": observationPhoto.uuid,
+      photo: {
+        id: observationPhoto.photo.id
+      },
```
kueda, line 37: Sending the ObservationPhoto UUID instead of the Photo UUID is dangerous here, because they will be different in server responses. The server doesn't seem to respond with the photo UUID at all, which is a separate problem.

tests/integration/MyObservations.test.js
```diff
@@ -125,6 +127,35 @@ describe( "MyObservations", ( ) => {
...
+        const mockPhotos = mockObsPhotos.map( op => op.photo );
+        const mockPhoto = mockPhotos.find( p => p.id === params.photo.id );
+        // Pretend this takes a bit
+        // await sleep( 200 );
+        return Promise.resolve( makeResponse( [{
+          id: faker.number.int( ),
+          uuid: mockPhoto.uuid
+        }] ) );
+      } );
+
```
kueda, line 158: I'm not sure why this *was* working, but I'm pretty sure it was not actually integrating the API wrappers and related networking code. This mocks inatjs responses to make sure that when those parts of the app get exercised they get reasonable responses.

kueda:
@jtklein I haven't had a chance to test this thoroughly under field conditions, but I got a bit into the upload code which I know you worked on recently, so I'd appreciate it if you could take a glance and see if anything seems glaringly wrong.

src/components/MyObservations/ToolbarContainer.js
jtklein: Okay, I did not see this particular problem in testing but I can imagine having a race condition where `numFinishedUpload` ticks one higher before `uploadInProgress` is false.

jtklein (changes requested):
Have started to review but when I upload an observation I am hitting this every time, and the upload does not finish:
```
LOG  2024-03-13T12:03:18.495Z | INatApiError | ERROR : Error requesting https://stagingapi.inaturalist.org/v2/observation_photos (status: 422):
    {"status":"422","errors":[{"errorCode":"422","message":{"errors":"No photo specified"},"from":"externalAPI"}]}
```
This would need to be fixed before I can continue.

src/components/MyObservations/ToolbarContainer.js
jtklein: When this branched off the "Upload success" message was sticking around forever, current main is it will disappear after 5s and then the "Upload x observations" state will reappear if there are `numUnuploadedObs !== 0 `. The way I handled it in my PR is to reset the upload state and therefore set `uploadsComplete` to false after 5s. With your changes, it will switch back to "Upload x observations" immediately. I am fine with merging this in if it seems more appropriate UX.

src/realmModels/ObservationPhoto.js
```diff
@@ -30,29 +30,31 @@ class ObservationPhoto extends Realm.Object {
...
-        name: `${observationPhoto.uuid}.jpeg`,
+        uri: photo.localFilePath,
+        name: photo.localFilePath,
         type: "image/jpeg"
       } )
     };
   }
 
-  static mapPhotoForAttachingToObs( id, observationPhoto ) {
+  static mapPhotoForAttachingToObs( observtionID, observationPhoto ) {
```
jtklein, line 43: The typo bothers me a bit.

jtklein (approved):
Overwhelmingly good to go I would say. Could not see anything glaringly off.

tests/integration/MyObservations.test.js
jtklein: Does it make sense to have those mocks in the inaturalistjs package itself? Would other clients benefit if we have a shared mock?

kueda: Not sure. Here the mock implementation is very specific to the test, like we want it to return very specific values. If inatjs were to have built-in mocks, they'd probably need to return generic mock data, so in this case we'd still need to make our own implementation. There might be other cases where that would be useful, though.

src/realmModels/ObservationPhoto.js
kueda: Agreed. I fixed it in a subsequent commit.

src/components/MyObservations/ToolbarContainer.js
kueda: I like your approach better, and I think that's what Abhas wants. I merged from main and tried to get the single-upload case to have that 5s delay as well.

PR #1291, opened 2024-03-20
Community Taxon improvement DQA

### LINKED ISSUE #1027: Data Quality Assessment needs ID vote
The DQA has a section where people can vote whether an observation should remain in the Needs ID quality grade, but we haven't implemented it yet. This is a vote (like a fave) and not a quality metric, so it needs different handling.

jtklein:
Closes #1027

### DIFF
--- a/src/components/ObsDetails/DQAContainer.js
+++ b/src/components/ObsDetails/DQAContainer.js
@@ -40,14 +40,10 @@ const DQAContainer = ( ): React.Node => {
   const {
     remoteObservation,
     refetchRemoteObservation,
-    isRefetching,
-    fetchRemoteObservationError
+    isRefetching
   } = useRemoteObservation( observationUUID, fetchRemoteObservationEnabled );
   const observation
     = localObservation || Observation.mapApiToRealm( remoteObservation );
-
-  // TODO: show some UI on remote observation fetch error
-  console.log( "fetchRemoteObservationError", fetchRemoteObservationError );
 
   const fetchMetricsParams = {
     id: observationUUID,
--- a/src/components/ObsDetails/DQAContainer.js
+++ b/src/components/ObsDetails/DQAContainer.js
@@ -209,31 +209,42 @@ const DQAContainer = ( ): React.Node => {
   };
 
   const checkTest = metric => {
-    if ( observation ) {
+    const obsDataToCheck = {
+      date: observation.observed_on,
+      location: [observation.latitude, observation.longitude],
+      evidence: compact( [
+        observation.observationPhotos || observation.observation_photos,
+        observation.observationSounds || observation.sounds
+      ] ),
+      taxonId: observation.taxon?.id,
+      rankLevel: observation.taxon?.rank_level,
+      identifications: observation.identifications
+    };
+    if ( obsDataToCheck ) {
       if ( metric === "date" ) {
-        return observation[metric] !== null;
+        return obsDataToCheck[metric] !== null;
       }
       if ( metric === "location" ) {
-        const removedNull = observation[metric]
-          .filter( value => ( value !== null ) );
+        const removedNull = obsDataToCheck[metric].filter(
+          value => value !== null
+        );
         return removedNull.length !== 0;
       }
       if ( metric === "evidence" ) {
-        const removedEmpty = observation[metric]
-          .filter( value => ( Object.keys( value ).length !== 0 ) );
+        const removedEmpty = obsDataToCheck[metric].filter(
+          value => Object.keys( value ).length !== 0
+        );
         return removedEmpty.length !== 0;
       }
-      if ( observation.taxon ) {
-        if ( metric === "id_supported" ) {
-          const taxonId = observation.taxon.id;
-          const supportedIDs = observation.identifications.filter(
-            identification => ( identification.taxon.id === taxonId )
-          ).length;
-          return supportedIDs >= 2;
-        }
-        if ( metric === "rank" && observation.taxon.rank_level <= 10 ) {
-          return true;
-        }
+      if ( metric === "id_supported" ) {
+        const { taxonId } = obsDataToCheck;
+        const supportedIDs = obsDataToCheck.identifications.filter(
+          identification => identification.taxon.id === taxonId
+        ).length;
+        return supportedIDs >= 2;
+      }
+      if ( metric === "rank" && obsDataToCheck.rankLevel <= 10 ) {
+        return true;
       }
     }
     return false;
--- a/src/components/ObsDetails/DQAContainer.js
+++ b/src/components/ObsDetails/DQAContainer.js
@@ -265,9 +265,9 @@ const DQAContainer = ( ): React.Node => {
   return (
     <>
       <DataQualityAssessment
-        loadingAgree={loadingAgree}
-        loadingDisagree={loadingDisagree}
         qualityMetrics={combinedQualityMetrics}
+        loadingAgree={loadingAgree || isRefetching}
+        loadingDisagree={loadingDisagree || isRefetching}
         loadingMetric={loadingMetric}
         qualityGrade={observation?.quality_grade}
         setMetricVote={setMetricVote}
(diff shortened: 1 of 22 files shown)


src/realmModels/Observation.js
```diff
@@ -397,6 +398,7 @@ class Observation extends Realm.Object {
         mapTo: "viewerTrustedByObserver",
         optional: true
       },
+      votes: "Vote[]",
```
jtklein, line 401: I just copied how the `faves` key is used here. I assume that syncing this key from the server to local is happening automatically. Because when searching for faves, I did not find any special handling.

src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -141,34 +140,17 @@ const ObsDetailsContainer = ( ): Node => {
 
   const localObservation = useLocalObservation( uuid );
 
-  const fetchRemoteObservationQueryKey = useMemo(
-    ( ) => ( ["fetchRemoteObservation", uuid] ),
-    [uuid]
-  );
   const fetchRemoteObservationEnabled = (
     !remoteObsWasDeleted
-    && !!isOnline
```
jtklein, line 150: `isOnline` is now checked in the `useRemoteObservation` hook.

src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -141,34 +140,17 @@ const ObsDetailsContainer = ( ): Node => {
...
-    && localObservation?.wasSynced( )
+    && ( !localObservation || localObservation?.wasSynced( ) )
```
jtklein, line 145: When opening ObsDetails of someone else's observation it did not fetch from the server. This is weird because I think I only refactored this code into the useRemoteObservation hook without changing the desired behavior. If I make a check here for not having a local observation it does fetch from the server and the observation loads in the UI. Please, let me know if that might break other things.

tests/unit/components/ObsDetails/DataQualityAssessment.test.js
jtklein, line 1: From this file I took out unit tests for the DQAVoteButtons into a separate file, and tests that check if the API was called and if container and button work together into integration tests.

src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -141,34 +140,17 @@ const ObsDetailsContainer = ( ): Node => {
...
-        fields: Observation.FIELDS
-      },
-      optsWithAuth
-    ),
-    {
-      keepPreviousData: false,
-      enabled: fetchRemoteObservationEnabled
-    }
-  );
+    fetchRemoteObservationError
```
jtklein, line 152: The two things from here that I refactored into a new hook are 1) fetching the remote observation by uuid, and 2) updating the local copy of a user's own observation, so that I can reuse it on the DQA page to load the observation from remote.
I am not sure if anything else that is done to the `remoteObservation` on this page is necessary whenever a remote observation is fetched, and therefore I should also do it on the DQA page.

src/components/ObsDetails/DetailsTab/DQAVoteButtons.js
```diff
@@ -8,46 +8,60 @@ import {
...
   removeVote: Function
 }
 
-const getUserVote = ( metric, qualityMetrics ) => {
-  if ( qualityMetrics ) {
-    const match = qualityMetrics.find( element => (
-      element.metric === metric && element.user_id ) );
+const getUserVote = ( currentUser, metric, votes ) => {
+  if ( votes && votes.length > 0 ) {
+    const match = votes.find( element => ( element.user_id === currentUser?.id ) );
```
jtklein, line 25: @angielt when I checked the schema of the `Vote`s that are returned for the dq metric "needs_id" from the server and used here from what I can tell all have a user_id which is just the id of the user who created the Vote. So, I changed the check here to find the element that is from the current user. so that the function checks wether the current user has voted up or down. The schema for the `QualityMetric` has the same structure with user_id, so I think it applies also to the other quality metrics. Let me know if you see anything off with this change.

src/components/ObsDetails/DataQualityAssessment.js
```diff
@@ -169,7 +171,7 @@ const DataQualityAssessment = ( {
           </View>
           <DQAVoteButtons
             metric="date"
-            qualityMetrics={qualityMetrics}
```
jtklein, line 172: @angielt I am only passing down the array of votes that are relevant for the particular metric. That makes the internal logic of each button more performant because it can return early if no data, or doesn't have to use Array.find/filter on a lot of elements. I have changed the qualityMetrics object so that each metric is a key that has the votes for that metric, so that data slicing should be pretty fast for this.

jtklein:
@kueda I refactored some of the logic that fetches a remote observation in ObsDetail into a hook and reused it in the DQA screen. I am not sure if I have taken everything I need for the remote observation to sync up properly. From my testing now it seem to work.
The main objective of this PR is to enable the "Community Taxon can be improved" quality metric and this is working from what I have tested. However, since it involves a new API endpoint I haven't worked with yet and I am not fully aware why this particular metric is a "fave" vote I would like to ask for a review if I missed something.
@angielt I also refactored some of the logic in the DQA screen and vote buttons,  so please could you check that it makes sense still with how you set up this screen.

src/components/ObsDetails/ObsDetailsContainer.js
kueda: Yeah, the logic in `main` is wrong, and it is only working by coincidence because when useQuery received an `enabled` value of `undefined`, it treats that as `true` and runs the query. This is apparently [desired behavior](https://github.com/TanStack/query/issues/2269), even though it's somewhat insane.

As far as I can tell, your changes both fix what was probably my logical problem in main, as well as prevent similar problems by ensuring the `enabled` value is always boolean.

src/components/ObsDetails/ObsDetailsContainer.js
kueda: IMO this looks good, I don't think you omitted anything.

src/components/ObsDetails/DQAContainer.js
```diff
@@ -81,26 +149,37 @@ const DQAContainer = ( ): React.Node => {
...
+      faveMutation.mutate( faveParams );
+      return;
+    }
+
+    const qualityMetricParams = {
+      id: observationUUID,
+      metric,
+      agree: vote,
+      ttyl: -1
+    };
```
kueda, line 177: I don't object to the way you're doing this, but I think it might further confuse developers in this repo as to how this particular part of the DQA is different from all of the others. Maybe a better way to put it is that it doesn't help developers understand that the `needs_id` part of the DQA is different from all the other parts.

src/realmModels/Observation.js
kueda: Yup, I think this works fine, though it does kind of duplicate the fave data here and in the `faves` property, so if you wanted to take it a step further you could remove `faves` and use `votes` to maybe replace it with a method like `faves()` that checks the `votes` property for fave votes. We'd be storing less data in Realm and fetching less data in requests. Definitely ok to merge as is, IMO, though.

kueda (approved):
Reminder to please name your branches starting with the issue number, e.g. `1027-dqa-needs-id-vote`

Otherwise, my main misgiving is continuing to obscure the difference between the `needs_id` `Vote` and the `QualityMetrics`, but as I commented, I don't think that needs to be a blocker. Otherwise this is looking good to me.

src/components/ObsDetails/DataQualityAssessment.js
angielt: Makes sense. Definitely a lot better than having each button do that find its respective metric.

src/components/ObsDetails/DetailsTab/DQAVoteButtons.js
angielt: Good catch, I don't see anything off with this change. I didn't know about the groupBy function in lodash until now. It seems very handy for the DQA.

angielt (approved):

PR #1302, opened 2024-03-24
Check for Internet connectivity in useOnlineSuggestions

### LINKED ISSUE #1144: useOnlineSuggestions should check for connectivity
The timeout in useOnlineSuggestions is useful when the connection is spotty, but a waste of time if the connection is definitely not there, so we should try to perform a connectivity check beforehand. Assuming that takes less than a second it's probably worth it.

- [x] Only enable the query in useOnlineSuggestions if there is an Internet connection
- [ ] Test it to ensure that this check takes less than a second under good, middling, and no Internet conditions (try the Network Link Conditioner)

budowski:
Closes #1144

### DIFF
--- a/src/sharedHooks/useIsConnected.ts
+++ b/src/sharedHooks/useIsConnected.ts
@@ -2,7 +2,10 @@ import { useNetInfo } from "@react-native-community/netinfo";
 
 // Note that a return value of null means the state is unknown
 const useIsConnected = ( ): boolean | null => {
-  const { isInternetReachable } = useNetInfo( );
+  const { isInternetReachable } = useNetInfo({
+    reachabilityShortTimeout: 1000,
+    reachabilityRequestTimeout: 1000
+  } );
   return isInternetReachable;
 };
 
--- a/src/sharedHooks/useIsConnected.ts
+++ b/src/sharedHooks/useIsConnected.ts
@@ -2,10 +2,7 @@ import { useNetInfo } from "@react-native-community/netinfo";
 
 // Note that a return value of null means the state is unknown
 const useIsConnected = ( ): boolean | null => {
-  const { isInternetReachable } = useNetInfo({
-    reachabilityShortTimeout: 1000,
-    reachabilityRequestTimeout: 1000
-  } );
+  const { isInternetReachable } = useNetInfo( );
   return isInternetReachable;
 };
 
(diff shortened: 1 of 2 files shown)


src/sharedHooks/useIsConnected.ts
```diff
@@ -2,7 +2,10 @@ import { useNetInfo } from "@react-native-community/netinfo";
 
 // Note that a return value of null means the state is unknown
 const useIsConnected = ( ): boolean | null => {
-  const { isInternetReachable } = useNetInfo( );
+  const { isInternetReachable } = useNetInfo({
+    reachabilityShortTimeout: 1000,
+    reachabilityRequestTimeout: 1000
+  } );
```
kueda, line 8: Why are you forcing the reachability check to time out after 1s? In the issue I asked to determine if the reachability check typically takes 1s, not to force it to do so. Forcing it to do so means a user in middling network conditions will never see the online suggestions, even if they wait.

If you're finding that the reachbility check *does* tend to take more than 1s under middling to poor network conditions, I don't even think this issue is worth it, or we need a different approach. I don't want the connectivity check and the suggestions request to take more than 5s total, but if the connectivity check takes 200ms then I want to show the offline suggestions in less than 5s. Does that make sense?

kueda (changes requested):

src/sharedHooks/useIsConnected.ts
budowski: Sorry, I misunderstood the issue ("Test it to ensure that this check takes less than a second"), but I agree, we can increase this. And the general 5-seconds-for-suggestions timeout is already implemented previously

kueda: In main there is a 5s timeout for the suggestions request, but the change I'm requesting here is that the connectivity check *AND* the suggestions request take no more than 5s.

kueda: Interestingly, this also seems to be the cause of the e2e failure in iOS.

budowski: I mean technically what's going on at the moment is this:
* If suggestions API call takes more than 5 seconds -> error
* If Internet connectivity fails -> error
* And Internet connectivity check happens simultaneously right now (not synchronously, before suggestions API call)

kueda: Ahh, I see what you're saying. In that case I think if we just get rid of the change to `useIsConnected` we should be good. Since the checks are in parallel, if we determine the user isn't connected before the 5s are up, we bail.

PR #1319, opened 2024-03-27
GitGuardian pre-commit hook

### LINKED ISSUE #25: GitGuardian: pre-commit hooks
We should implement GitGuardian [pre-commit hooks](https://docs.gitguardian.com/internal-repositories-monitoring/integrations/git_hooks/pre_commit). This will allow us to prevent secrets from being committed to GitHub (for example, in case a file is not yet listed in `.gitignore`).

We'll also need to add instructions for setting up the hook to `README.md`.

kueda:
Adds a pre-commit hook to run GitGuardian's `ggshield` command to check new code for secrets, passwords, API keys, etc.

I tested this by

1. Hard-coding an OAuth client secret
2. Hard-coding the Android Google Maps API key
3. Hard-coding the anonymous JWT secret

It catches 1 and 2, but 3 fails because GitGuardian's [generic high entropy secret detector](https://docs.gitguardian.com/secrets-detection/secrets-detection-engine/detectors/generics/generic_high_entropy_secret) relies on assignment, and statements like `jwt.encode( claims, "bsaruceobkoraebisroaecbu89", "HS512" );` won't trigger it because there's no assignment happening (that's one of their example secrets). Since GitGuardian doesn't seem to provide a mechanism for custom detectors, the only way I can think of to prevent that kind of secret leakage would be a custom eslint rule. I spent an hour trying to write one but I couldn't figure out how to detect a function call and check if its second argument is a string literal or not in that time span.

Another concern is that I didn't want to require the GitGuardian check for volunteer contributors who might not have it installed, since setup is already a pain and volunteers shouldn't have access to or be using critical secrets. Instead, I'm running the check if `ggshield` is installed OR if the git user has an inaturalist.org email address. That means team members who don't use inaturalist.org email addresses w/ git will need to install GitGuardian.

Questions for review:

1. Do we want to live with the vulnerability of secrets leaking in ways GitGuardian can't detect? We're living with that now.
2. Is it ok to only require the `ggshield` check if it's installed OR if the git user has an inaturalist.org email address?

Closes #25

### DIFF
--- a/.gitguardian.yml
+++ b/.gitguardian.yml
@@ -1,0 +1,8 @@
+version: 2
+secret:
+  # Exclude files and paths by globbing
+  ignored-paths:
+    - '**/README.md'
+    - 'node_modules/*'
+    - 'package-lock.json'
+    - '**/package-lock.json'
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,6 @@ artifacts/
 
 # Reassure performance testing #
 .reassure
+
+# GitGuardian cache file
+.cache_ggshield
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,15 +1,6 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-# Run lint and flow
-npm run lint
-
-# Generate translations from strings.ftl
-npm run translate
-# ...and add them to the commit; this assumes src/i18n/l10n/ *only* contains
-#    files you want to add
-git status --porcelain --untracked-files src/i18n/l10n/ | cut -c 4- | xargs git add
-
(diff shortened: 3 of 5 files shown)


albullington:
My two cents on the questions:

1. I think this is fine for now. If I recall, the main use case we were trying to prevent against was accidentally checking secrets in `.env` files into Github (for example, if a developer created a new env file like `.env.staging` but then switched branches, the file could be carried into the new branch without being properly added to .gitignore). I think preventing some developer error is better, even if we aren't protecting ourselves against every possibility.
2. This seems fine, since only iNaturalist team members are likely to be adding and committing secrets.

albullington (approved):

PR #1384, opened 2024-04-11
1364 UI setting

### LINKED ISSUE #1364: Observation button should default to AI Camera view & add a setting to switch between AI camera and obs wheel
Based on MVP changes, there are adjustments to the [designs for settings](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=16738-48917&mode=design&t=B2ikgku6nMA9WBXb-4).

- [x] Make the settings section visible in the hamburger menu for all users
- [x] Create a new section inside of settings that is visible to both logged in and logged out users titled "Observation Button" - for logged out users, this is the only section they will see (The option a user chooses here is preserved throughout the app, whether they log in or out or close the app)
- [x] For all users on first launch of the iNat app, the observation button (green button) should default to showing the AI Camera only. They can adjust this in settings.
- [x] When a user changes this setting, the AI Camera button in the tab bar should change to the observation wheel as it is now.

jtklein:
Closes #1364

This also adds react-native-mmkv as a replacement for AsyncStorage because it is more performant.


package.json
```diff
@@ -97,6 +97,7 @@
     "react-native-logs": "^5.1.0",
     "react-native-mail": "github:chirag04/react-native-mail",
     "react-native-maps": "^1.13.0",
+    "react-native-mmkv": "^2.12.2",
```
albullington, line 100: A couple questions on this:

1. Do we need a key/value storage solution for Settings, or can we store this information in Realm along with all our other local data? We already have a [LocalPreferences](https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/realmModels/LocalPreferences.js) model in Realm which could include this data without the complexity of introducing a new library.
2. If you're committed to using `react-native-mmkv` here, can you make a separate issue to replace AsyncStorage in the rest of the app? Right now, we're using 4 different libraries for local storage (`mmkv, async-storage, realm, and react-native-sensitive-info`) which creates developer confusion on which library to use and what data we're storing. I think it's better for us to make a clear decision on which storage solution(s) we really need and remove the rest. This way we have more of a single source of truth for local data.

src/sharedHooks/useStorage.ts
```diff
@@ -0,0 +1,12 @@
+import { useMMKVBoolean } from "react-native-mmkv";
```
albullington, line 1: I noticed you've been saving TypeScript files more often -- could you add this to the next mobile meeting and tell us a little about when and why you're using it? I suspect we all want to be moving in the direction of TypeScript, rather than Flow, and I think it would be a good thing for us to collaborate on. I added TypeScript extensions to Tailwind, so afaik we should be able to use TypeScript in all of our UI components now.

package.json
jtklein: 1.) We do not need a key/value storage for this setting. What we do need is to persist it between app launches. We can store it in Realm. I did not see the LocalPreferences before, thanks for mentioning it. I do think though that the overhead for using Realm is quite large for this specific boolean. So in terms of complexity it seems more complex to add it to Realm than a key/value storage. It also feels weird to use Realm which I think of more as being for collections for one item only.
2.) I am not commited to this dependency, so I will revert to another solution.
My main concern here was speed of retrieval of the value, as it is a value we need at app start, and everything that makes the app taking longer to start makes for a worse UX and higher churn.
But then my question becomes, why do we have three different storages already, plus `zustand` for app state?

src/sharedHooks/useStorage.ts
jtklein: Can do. But there is not much to say I guess. Whenever I create a new file that has no dependency on any of our own Flow-typed JS files I use Typescript.

package.json
albullington: Sounds good. To be clear, I'm not being prescriptive one way or another -- mmkv might make the most sense here if it's performant. I'm just feeling confused about why we have so many different sources of truth in the app & would like to reduce developer confusion if possible. Afaik the only reason we have three different storages is because we each have different developer preferences on which libraries we prefer to use -- it would make sense to me to pick one if possible.

`zustand` serves a different purpose since it preserves app state while a user is in the app, while the storage solutions preserve data between app launches.

src/sharedHooks/useStorage.ts
albullington: Thanks. I think it would be helpful to understand why you're using TS, not Flow, even if that reason is just "this is the direction the industry is going."

Issue #1371, opened 2024-04-08
App freezes on ObsDetails when navigating from Notifications


albullington:
**To Reproduce**
1. Go to Notifications screen for a logged in user
2. Find a notification that you haven't already viewed
3. Tap on that notification to go to ObsDetails
4. The app will freeze and no buttons will be tappable on the screen or bottom tab bar

**Expected behavior**
App should not freeze when navigating from Notifications -> ObsDetails

**Context (please complete the following information):**
 - OS: iOS
 - App version: [v0.29.2+83](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.29.2%2B83)

**Additional context**
Not sure what's causing this, but it might be something like `useEffect` being called an infinite number of times when a user lands on ObsDetails. I noticed this happening with the markViewedMutuation code and tried to fix [here](https://github.com/inaturalist/iNaturalistReactNative/commit/408f746701aa1b9d02c9560a5ce5e0a230886ec5#diff-d934a11d86e60ca6be2e38e42f1a159f3cc02ec8969771b52cebc5e1a5fa164a), but maybe that's still being called too many times in certain scenarios.

budowski:
@albullington - does not occur anymore with latest main branch, tested on iOS and Android (probably fixed by #1386 ?) - does it still happen to you

kueda:
Fixed by #1386

PR #1386, opened 2024-04-11
Mark viewed infinite loop

kueda:
For me this seems to solve a `ERROR  Warning: Maximum update depth exceeded` error I was seeing after viewing ObsDetails 2-3 times and was making it impossible for me to test anything else on ObsDetails. I *think* the point of `isMarkingViewed` is to prevent this kind of loop, so if that's the case I think we want to bail out of this effect whenever it's true, not only when it's true and there's no remote observation.

This might also address #1371


albullington (approved):
Good catch -- definitely was trying to avoid the Maximum Depth error with this.

PR #1387, opened 2024-04-11
Refactor navigation

albullington:
Our navigators are confusing to use. It's ambiguous whether we should save screens in the Drawer navigator, one of the many Stack navigators, or the Tab navigator. This PR is an attempt to:

1. Consolidate our navigators, making sure we're not duplicating any navigation screens unintentionally (like `FullPageWebView`)
2. Use Stack Groups to set screen options for screens with similar header treatment, instead of specifying options on every individual screen
3. Rename navigators so developers know whether to expect a screen to show bottom tab navigation or not
4. Create a navigation structure that's easier to reason about. There's a Drawer with a Tab Navigator and two Stacks. Inside the Tab Navigator is one Stack. There's no additional nesting

Open to feedback since I'm still not positive if this is the best way to do this -- I'm hoping it's at least an improvement over our current navigation system.


kueda (changes requested):
I like the overall direction. `NoBottomTabStackNavigator` is a mouthful, but the only alternative I can think of is `ModalStackNavigator` which might be confusing with other kinds of modals in play, so I'd just go with what you have unless you want to get a third opinion.

I'm struggling to test this thoroughly because of the infinite loop bug so you might want to merge in from main. Here's one blocking problem I found, though:

- [ ] If I am signed in at tap LOG OUT in the drawer, I land on the camera permission gate instead of the log out screen (you probably need a fresh install to replicate). If I then tap the X in the upper right of the permission gate and repeat the steps, I instead land on a black screen with no UI and I'm trapped.

albullington:
Merged from main and also fixed the blocking issue with the LoginStackNavigator.

PR #1396, opened 2024-04-15
1070 sound progress

### LINKED ISSUE #1070: SoundContainer player progress bar interactions
The audio player in the SoundSlide component *shows* the current progress of the audio playback, but that progress bar doesn't let you seek ahead or back.

- [x] Tapping the bar should seek to that position
- [x] Dragging the current position indicator should seek to that position
- [x] Neither interaction should change the playing status, so if you tap the bar while playing, it should continue playing at that position, and if you tap the bar while not playing, it should not start playing

@abhasinat holler if you have a different opinion re: MVP status. IMO it feels a bit buggy not to support this basic audio player functionality.

jtklein:
Closes #1070

### DIFF
--- a/src/components/ObsDetails/SoundContainer.js
+++ b/src/components/ObsDetails/SoundContainer.js
@@ -2,6 +2,7 @@ import {
   refresh as refreshNetInfo,
   useNetInfo
 } from "@react-native-community/netinfo";
+import Slider from "@react-native-community/slider";
 import { useFocusEffect } from "@react-navigation/native";
 import { Body1, INatIconButton, OfflineNotice } from "components/SharedComponents";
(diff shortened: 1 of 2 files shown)


src/components/ObsDetails/SoundContainer.js
```diff
@@ -28,6 +50,8 @@ const SoundContainer = ( { sizeClass, isVisible, sound } ) => {
   // Track whether the user swiped away since last playing
   const [swipedAway, setSwipedAway] = useState( false );
 
+  console.log( "sound", sound );
```
jtklein, line 53: This slipped through, needs to be deleted.

PR #1413, opened 2024-04-18
Support Typescript files in eslint

### LINKED ISSUE #1394: Eslint: Support TypeScript
Now that we're starting to use more TypeScript files as a replacement for Flow, we should make sure eslint works with TypeScript files (.ts and .tsx).

The [typescript-eslint](https://typescript-eslint.io/getting-started/typed-linting) library might be a great starting point. I believe this would replace the `@babel/eslint-parser` we're currently using. It recommends using a flat config instead of the legacy eslint config we have in the app at the moment.

`ExploreContext.tsx` is a good file to check to make sure our custom eslint rules are applied, since there are currently a few lines over 100 characters in that file that *should* be linted.

albullington:
Closes #1394

This PR:
- Allows us to use eslint in Typescript files
- Fixes all the linting errors we were previously missing in Typescript files
- Changes the props we weren't fully typing (Object, Function) into new types that can be accepted by both Flow and Typescript. I might add a new issue to encourage us to use real Typescript types in these files post-launch

Note: I think we should eventually remove Flow and support only Typescript, since we weren't totally sure if Flow was helping us from the start. Here's a [compelling list of reasons](https://www.skovhus.dev/blog/flow-to-typescript-migration) we should make this switch. But for now, I believe we need both to make sure we're type-checking in both .js/.jsx files and .ts/.tsx files.


src/components/AddObsModal.js
```diff
@@ -13,8 +13,8 @@ import { useTranslation } from "sharedHooks";
 import useStore from "stores/useStore";
 
 type Props = {
-  closeModal: ( ) => void,
-  navAndCloseModal: ( string, ?{ camera: string } ) => void
+  closeModal: any,
+  navAndCloseModal: any
```
kueda, line 17: This seems like a real loss of specificity. `( ) => void` doesn't trigger any alerts as a type for me, and IMO,

```
// eslint-disable-next-line @typescript-eslint/ban-types
navAndCloseModal: Function
```

at least signals to devs that this prop is supposed to be a function and not a boolean or some non-function primitive.

jest.config.ts
```diff
@@ -17,13 +17,14 @@ const config: Config = {
     "<rootDir>/tests/initI18next.setup.js"
   ],
   transformIgnorePatterns: [
+    // eslint-disable-next-line max-len
     "node_modules/(?!(jest-)?@react-native|react-native|react-clone-referenced-element|@react-native-community|expo(nent)?|@expo(nent)?/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base||(?!react-native-redash))|jest-runner"
```
kueda, line 21: This is one where I think it would be better to change the code to confirm to the rule instead of ignoring the rule. The rule is there to make the code more readable, and you can do so here by breaking this long regex into multiple lines.

.eslintrc.js
```diff
@@ -123,13 +125,19 @@ module.exports = {
     "react-native-a11y/has-valid-accessibility-descriptors": 1,
     "react-native-a11y/has-valid-accessibility-ignores-invert-colors": 1,
     "react-native-a11y/has-valid-accessibility-live-region": 1,
-    "react-native-a11y/has-valid-important-for-accessibility": 1
+    "react-native-a11y/has-valid-important-for-accessibility": 1,
+
+    "@typescript-eslint/no-unused-vars": 0,
+    "@typescript-eslint/no-var-requires": 0,
+    "@typescript-eslint/no-empty-function": 0,
```
kueda, line 132: I think this is fine, but you have one-off exceptions to this rule elsewhere in this PR. I'd leave this rule here and remove the one-offs.

kueda (commented):
Not sure this needs to block, but I don't love the more generic typing, especially when it comes to functions. I understand the reason for the rule, but that reasoning presumes a commitment to typing that we're not making outside of typescript files. I'm less concerned about distinguishing an Observation from an Error than I am distinguishing a Function from a boolean.

PR #1421, opened 2024-04-19
Always use latest added photo on suggestions

### LINKED ISSUE #1309: Fix crashes and stale data when a user backs out from ObsEdit -> Suggestions -> Camera
Following on #1281, we need to make some fixes to the user flow when a user goes from the Camera -> Suggestions -> ObsEdit and then backs out to the camera.

AR Camera
- [x] should no longer display the previous taxon result (addressed by [#1305](https://github.com/inaturalist/iNaturalistReactNative/issues/1305))
- [x] should replace the existing observation photo with any new photo taken, since the AR Camera is only designed to take one photo at a time (might want to check with Abhas on whether this is the expected user flow)
- [x] should show new top suggestion for new photo when progressing to Suggestions screen

Standard Camera
- [x] should treat any new photos taken like adding evidence, so a user can see all observation photos on Suggestions screen (might want to check with Abhas on this user flow)
- [x] should populate Suggestions screen with the latest photo

budowski:
Also  replaces existing photos on Suggestions when reached from AI camera, even if the user backs out of Suggestions to AICamera and then back to Suggestions again.

Closes #1309


kueda (changes requested):
Does not fulfill the second requirement in the issue: "should replace the existing observation photo with any new photo taken, since the AR Camera is only designed to take one photo at a time (might want to check with Abhas on whether this is the expected user flow)"

To repro:

1. Fresh install of the app (not signed in)
2. Tap the AI Camera button, either in the middle or the bottom tabs
3. Point the camera at something it will get a prediction for an tap the shutter
4. On Suggestions, the suggestions are for the thing you just photographed
5. Tapped the back button to return to AICamera
6. Point the camera at a different subject that generattes a prediction
7. Tap the shutter
8. On Suggestions, you are seeing the preview for the first photo you took and predictions for the first photo you took, not the second photo that you just took.

You should see the preview for the second photo and predictions for the second photo.

kueda (approved):

PR #1423, opened 2024-04-19
Change language from Settings while offline

budowski:


kueda (commented):
Technically it sort of works. It seems to load the locales, and they are available offline, and while online I can change to Spanish and see the "Bienvenido de nuevo" on MyObs. However, if I go offline and force quit the app, when I open it again I see "Welcome bacK." I should see "Bienvenido de nuevo".

Furthermore, I'm not entirely sure this issue should have been reopened and I don't remember why I did that. @abhasinat, is this still something you want done, allowing the user to choose a language from Settings while offline? If so, can you come up with a design? Yaron has implemented it as a bottom sheet that fills the entire screen because there are a lot of languages to choose from:

IMG_0525

kueda:
Ok, @abhasinat does want to keep this and is designing something.

abhasinat:
Here's a [design](https://www.figma.com/file/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?type=design&node-id=16738-48917&mode=design&t=sDHHn4dUWs1XsZaJ-4) for the settings screen for a language picker.

Clicking on the dropdown should open a picker that looks just like the rank picker in Explore Filters with the title "Language/Locale":
IMG_5591

PR #1465, opened 2024-04-25
Focus types

jtklein:
I added some of the recent updates to the Camera views into Seek and adapted some of the TypeScript code from those changes.
I am presenting the updates here as well.
Let me know what you think of it @albullington.

### DIFF
--- a/src/components/Camera/CameraView.tsx
+++ b/src/components/Camera/CameraView.tsx
@@ -2,7 +2,6 @@ import { useAppState } from "@react-native-community/hooks";
 import { useIsFocused } from "@react-navigation/native";
 import useFocusTap from "components/Camera/hooks/useFocusTap.ts";
 import VeryBadIpadRotator from "components/SharedComponents/VeryBadIpadRotator";
-import type { Node } from "react";
 import React, {
   useCallback
 } from "react";
@@ -38,7 +37,7 @@ interface Props {
 
 // A container for the Camera component
 // that has logic that applies to both use cases in StandardCamera and AICamera
-const CameraView = ( {
+const CameraView: React.FC<Props> = ( {
   animatedProps,
   cameraRef,
   device,
@@ -49,7 +48,7 @@ const CameraView = ( {
   onDeviceNotSupported,
   pinchToZoom,
   resizeMode
-}: Props ): Node => {
+}: Props ) => {
   const {
     animatedStyle,
     tapToFocus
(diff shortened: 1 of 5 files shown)


src/components/Camera/CameraView.tsx
```diff
@@ -25,8 +24,8 @@ Reanimated.addWhitelistedNativeProps( {
 
 interface Props {
   animatedProps: CameraProps,
-  cameraRef: Object,
```
jtklein, line 28: Those are more precise types available by the used libraries.

src/components/Camera/FocusSquare.tsx
```diff
@@ -1,14 +1,13 @@
...
 import _ from "lodash";
-import type { Node } from "react";
 import React from "react";
 import { Animated } from "react-native";
 
 interface Props {
   animatedStyle: Object
 }
 
-const FocusSquare = ( { animatedStyle }: Props ): Node => {
```
jtklein, line 11: Functional React components work well with inferred types: https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components

src/components/Camera/hooks/useFocusTap.ts
```diff
@@ -20,7 +29,8 @@ const useFocusTap = ( cameraRef: Object, supportsFocus: boolean ): Node => {
     } );
   }, [tappedCoordinates, focusOpacity] );
 
-  const onFocus = useCallback( async ( { x, y }: GestureResponderEvent ) => {
+  type TapEvent = GestureStateChangeEvent<TapGestureHandlerEventPayload>;
```
jtklein, line 32: Gesture-handler also have their own types. I am not sure if this is necessary but in any case `GestureResponderEvent` is not exported from gesture-handler.

src/components/Camera/FocusSquare.tsx
albullington: It sounds like we can remove this Node type from all TS files going forward, yeah?

albullington (approved):
Looks good -- it's helpful to see what types we should be using for refs & that there are newer types available for gesture handler (I was looking at some pretty old examples in the react-native-vision-camera library).

src/components/Camera/FocusSquare.tsx
jtklein: Yes

PR #1496, opened 2024-05-05
Display taxon name according to user settings

### LINKED ISSUE #1444: Names Display not showing on all screens
When I set my names display to Scientific Name (Common Name), the only place I see it displayed correctly is on home screen/my observations. On all other screens, it still appears as Common Name (Scientific name)

Please make sure that names display also applies to:
- [x] Obs Detail
- [x] Taxon Detail
- [x] Explore

budowski:
Closes #1444


kueda (changes requested):
Still a few places this preference isn't getting used:

- [x] Taxonomy section of TaxonDetails:
    IMG_0568
- [ ] Default options on TaxonSearch
    IMG_0569
- [ ] Search results in TaxonSearch
    IMG_0570
- [ ] AICamera suggestion
    IMG_0571

kueda (approved):

Issue #1481, opened 2024-04-29
Bulk upload text and progress bar doesn't match uploads in progress


albullington:
**Describe the bug**
The bulk upload text and progress bar on MyObservations still doesn't reflect the actual status of uploads. 

**To Reproduce**
1. Queue up a large number of observations (Carrie tried 72)
2. Start uploading all of them from MyObservations
3. Observe progress bar. It will say "Uploading 1 of 72" and won't make progress
4. Text may flash and occasionally say "19 observations uploading" but will change states instead of conveying a consistent, accurate message to the user

**Expected behavior**
A user should be able to see the real-time progress of their uploads, along with the correct message letting them know how many uploads are in progress.

**Additional context**
App version: [v0.32.0+88](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v0.32.0%2B88)

I suspect this is related to our lack of [sequential uploads](https://github.com/inaturalist/iNaturalistReactNative/issues/1427), which is an issue we've decided not to prioritize. The way our code is currently written, all 72 uploads will start at once, instead of moving through sequential progress. That means uploads will complete in a random order, instead of completing one at a time. All 72 uploads will create an observation, then all will try to append any photos or sounds to the observations via a different API call. I think we'll need to address this underlying code if we want to display reliable messaging to the user.

kueda:
Possibly related problem in v0.32.3 (91): progress bar can exceed the text progress:

![IMG_0841](https://github.com/inaturalist/iNaturalistReactNative/assets/22691/808d4715-89fc-4d6e-b216-6a7564bd72fb)

In this particular case there were a few observations with multiple photos, but I was able to replicate this with a couple single-photo observations as well:

![IMG_0856](https://github.com/inaturalist/iNaturalistReactNative/assets/22691/ab97d2bc-8c69-4c76-b5f2-2d61bdb3c404)

albullington:
Closed by PR [#1606](https://github.com/inaturalist/iNaturalistReactNative/pull/1606)

PR #1606, opened 2024-05-24
Refactor upload functionality into hook/zustand

albullington:
This is a draft PR because tests need updating, and there are likely bugs.

Working on consolidating our upload code into a few single sources of truth, because right now, the logic for uploading is spread all over the place and hard to reason about.

Kind of an initial step before I can really tackle [#1481](https://github.com/inaturalist/iNaturalistReactNative/issues/1481), since this helps us calculate the progress number for the toolbar and individual uploads in a single array of objects, rather than in individual components.

Feel free to let me know if you see anything glaringly obvious that we should move into zustand or that is broken in this draft PR.


src/components/MyObservations/hooks/useUploadObservations.ts
```diff
@@ -0,0 +1,237 @@
...
+import {
+  useTranslation
+} from "sharedHooks";
+import useStore from "stores/useStore";
+
+const { useRealm } = RealmContext;
+
+export default useUploadObservations = (
+  toggleLoginSheet,
+  showInternetErrorAlert,
```
kueda, line 22: FWIW, I don't think this hook needs to have any awareness of UI logic like this. If there's no internet connection, it should either throw an error or include an error in its return value, and the component using this hook can decide what do do with that in the UI. Same goes for not having a logged in user.

src/stores/createUploadObservationsSlice.ts
```diff
@@ -0,0 +1,153 @@
...
+  totalProgress: number
+}
+
+interface UploadObservationsSlice {
+  uploadError: string | null,
+  multiError: string | null,
+  errorsByUuid: Object,
+  uploaded: Array<string>,
+  singleUpload: boolean,
+  uploads: Array<Object>,
```
kueda, line 32: It would really help a reader if this was an array of typed objects and not just `Object`. What do we expect to find in here?

src/stores/createUploadObservationsSlice.ts
```diff
@@ -0,0 +1,153 @@
...
+  numToUpload: number,
+  numFinishedUploads: number,
+  totalToolbarProgress: number,
+  totalUploadProgress: Array<TotalUploadProgress>,
+  uploadStatus: string | null,
```
kueda, line 37: Here too. `uploadStatus` has a finite set of possible values, so they should be strongly typed.  If someone tries to set `uploadStatus` to `"whatever"` we should get an error

kueda (commented):
Not as much of a refactor as I thought it would be so hopefully not too buggy. I personally prefer a reducer over Zustand because if I'm centralizing state management I don't want anyone to be able to plug into that state except through the centralized source I've made, but I don't think that needs to block.

kueda (changes requested):
Still reviewing, but one blocking issue is that canceling uploads doesn't reset the status text correctly. If you have one obs to upload and you start uploading, then press the "x" button, the toolbar looks like there are no observations to upload, but it should say "Upload 1 observation" again.

Ok, more comprehensive review

- [x] Cancelling upload seems to reset the entire upload state and the only way to get it back is to force quit
- [ ] When upload is complete, the linear progress bar at the bottom of the footer often gets stuck in a partial state, i.e. it doesn't go all the way back to zero: IMG_0936
- [ ] The rotating button enters turns red and has an exclamation mark when 1 of several uploads errors out, resulting in a red, rotating exclamation mark. It should only enter this error state when all uploads have completed or errored out (the status text should update immediately to say "x uploads failed" which is working correctly): IMG_0949
- [ ] Hard to verify without instrumenting the code, but it feels like progress increments for the *start* of uploads aren't getting registered, i.e. the progress indicators only seem to update when an upload *finishes*

PR #1632, opened 2024-06-03
Fix reposition evidence through drag and drop

albullington:


kueda:
What was the problem here, @albullington? Changes to photo positions not getting uploaded?

albullington:
no, there was a crash when dragging and dropping which I introduced the week before by passing only uris instead of full observation photo objects into the drag and drop flatlist

PR #1648, opened 2024-06-04
Typescrpt path aliases; tried to resolve ts errors in one file

kueda:
Would like some feedback on the approach in src/api/log.ts. Is this how we're supposed to handle sometimes-present attributes on objects?

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,7 +23,7 @@ import { GestureHandlerRootView } from "react-native-gesture-handler";
 import { startNetworkLogging } from "react-native-network-logger";
 import { SafeAreaProvider } from "react-native-safe-area-context";
 import { reactQueryRetry } from "sharedHelpers/logging";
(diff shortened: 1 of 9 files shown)


jtklein (commented):
So, I am not sure I understand the comment above the new interface. From the code I think you want to accept two different types of params, either something that is an instance of Error or an object that can have a message and backtrace attribute. If you want to say that ErrorWithContext might not have either of those attributes you can type them as `string | undefined` or with the `?` shorthand as I have added in the line comment.

Or are you rather asking how to combine the fact that some error objects have the message attribute and some don't into one type?

kueda:
@jtklein I guess it just seems awkward to declare a one-off interface just to handle this situation where an Error might have some optional attributes. Taking your advice about the optional attributes makes it a bit cleaner, thanks.

PR #1656, opened 2024-06-06
Don't put error in Alert; globalize error message

kueda:
Addresses crash reported when trying to follow a user that you already have a Friendship record for (Friendships can exist without a `following` attribute). Actually that problem was addressed by adding upsert functionality to `POST /relationships`, but this addresses a related problem in which adding an Error object as the body of an Alert resulted in a native crash. This also globalizes the error message when this happens and logs the error to logstash.


kueda:
@angielt, you might want to take a look at this b/c I think you wrote some of this stuff. Some main takeaways are

1. We should not put Errors directly into Alerts
2. All user-facing text should be globalized
3. Try to make sure variable names describe what they are; `id` did not actually contain an ID but a params object

PR #1670, opened 2024-06-10
Changes to syncing process

### LINKED ISSUE #1630: Tapping sync button in toolbar syncs forever
**Describe the bug**
Tapping the sync button in the toolbar when you have no observations to upload syncs forever

**To Reproduce**
1. Sign in to an account with observations
2. Tap the sync button in the toobar
3. Toolbar icon starts to rotate and status says "Syncing..."... and stays that way until you force quit the app or sign out

**Expected behavior**
Syncing status should stop when syncing stops, and syncing should not last forever/

 - Device: iPhone SE (3rd gen)
 - OS: 17.4.1
 - App version: 0.32.6 (94)

**Additional context**
Abhas and I are both experiencing this under very good network conditions, so this isn't because the request is timing out. That case *might* be worth addressing but this is something else.

albullington:
Closes #1630

In this issue, we spelled out the sequential way syncing should happen. In this PR:

When a user lands on MyObservations, syncing automatically starts. Syncing involves:
1. Fetching deletions from the server (logged in users)
2. Handling local deletions (logged in/out users)
3. Fetching remote observations from the server (logged in users)

During this process, the toolbar UI does not alert the user that syncing is in progress. Right now, the sync button is temporarily disabled while the automatic syncing happens. I don't think we'll want this long term, but for now, I wanted to make sure the sync button process and the automatic process aren't interfering with each other.

The one UI exception here is when a user navigates to ObsEdit, deletes an observation, and returns to MyObservations. They will then see "1 observation deleted" in the toolbar. This message should time out after a few seconds.

When a user taps the sync button, a new syncing process starts. This involves:
1. Fetching deletions from the server (logged in users)
2. Handling local deletions (logged in/out users)
3. Fetching remote observations from the server (logged in users)
4. Uploading all unuploaded observations (logged in users)

During this process, the toolbar UI says "Syncing..." until uploads start, then the toolbar displays the uploading UI.

I still need to get tests passing, but I think it would be helpful to get a review on this in the meantime since there are a lot of moving parts.


kueda (changes requested):
In terms of functionality, there seems to be a blocking issue:

- [ ] If you record a new observation and save without uploading, the upload button in the toolbar is permanently disabled and you can't upload until you leave MyObs and come back

I personally don't like disabling the sync button while we're doing automatic sync. IMO it should look just like it does during manual sync so the user knows what's happening, but that's a decision for @abhasinat.

In terms of engineering and maintainability, I don't have any blocking problems if this works, but I don't like the idea of multiple hooks that are basically `useEffect` callbacks waiting for some change in state to run. That seems like a recipe for race conditions to me. You've laid out a clear sequence for how sync should operate in your PR description, so if that's how it should work I expect to see that in code, e.g. something like

```typescript
async function syncAutomatically() {
  if ( loggedIn ) await fetchRemoteDeletions();
  await deleteLocalObservations();
  await fetchRemoteObservations();
}

async function syncManually() {
  if ( loggedIn ) await fetchRemoteDeletions();
  await deleteLocalObservations();
  await fetchRemoteObservations();
  if ( loggedIn ) await uploadObservations();
}
```

Not only is that easier to read, but it's easier to test. Again, not blocking.

albullington:
Alright, tried to clean up my convoluted code and rewrite in a way that's easier to understand (though I'm still using a `useSyncObservations` hook). Added some `useSyncObservations` tests to make sure the right functions are getting called when a user is logged in / out and when there's a manual vs. automatic sync. I removed the code related to disabling the button and the UI is currently the same whether a user is manually or automatically syncing -- I think we can tweak this part later if it's not what @abhasinat wants, but for now, it seems easier to test and to show the user what's happening if we use the same UI.

kueda (approved):
Looking better. My only caveat is that `useSyncObservations.test.js` should use a unique realm instance to avoid flakiness when multiple parallel tests might be altering realm at the same time, but I don't think that has to block.

abhasinat:
sounds good, happy to keep an eye on this in testing next round to see how it feels.

Issue #1594, opened 2024-05-21
Changes to permissions


abhasinat:
After user testing, we found this was overwhelming for a user and the additional gallery permission led to negative feedback

**Current permissions flow for a first time user**
Home screen -> AI Camera button -> _camera permissions_ -> AI Camera -> Shutter button -> _Location permissions_ -> _Saving to gallery permissions_ -> _Additional gallery permissions_ -> Suggestions

**Desired permissions flow for a first time user**
Home screen -> AI Camera button -> _camera permissions_ -> AI Camera -> Shutter button -> _read AND write to gallery permission_ -> Suggestions -> "Improve these suggestions by using your location" button -> _location permissions_

### Key changes to make for permissions
- [x] whenever we ask to read OR write to the gallery, ask for a single modal to read AND write to the gallery - this should be the case whether the first permission a user sees is to save to the gallery or upload from the gallery.
- [x] prevent location permissions from automatically triggering on suggestions
- [x] Instead, location permissions should trigger opt-in on suggestions and explore and projects, and any time the user opens the location picker on obs edit
    - [x] On suggestions, the user should see a persistent button that says ["Improve these suggestions by using your location"](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=17975-46329&t=oRoFSIcq98XobLV2-11) if they have not allowed location access. Clicking this should open the location permission gate or location permissions denied screen. (this should be covered by #1592)
    - [x] On explore, when a user is viewing species nearby (the default state), they should see an [opt-in message to allow location access](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=17969-41209&t=oRoFSIcq98XobLV2-11). Clicking this should open the location permission gate or location permissions denied screen.
    - [x] On obs edit, the user will by default see an error state for location that says "add location." [Opening the location picker](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=707-8217&t=oRoFSIcq98XobLV2-11) on obs edit when a user has not allowed location should open the location permission gate or location permissions denied screen.
    - [x] On any map screen (including explore map, taxon detail map, obs detail map) whenever a user clicks the "current location" button, the location permission gate or location permissions denied screen.
[...]

PR #1680, opened 2024-06-17
Close #1612 - permission handling in relation to denying gallery access

### LINKED ISSUE #1612: Gallery permissions can't be granted after initial denial
**Describe the bug**
If a user denies gallery access by selecting "keep add only," they won't be able to use the gallery at all later.

**To Reproduce**
1. Install a new build
2. Press the AI camera button
3. Allow camera access
4. Take a photo
5. Allow location access while using app
6. Allow to save photos to gallery
7. When prompted about photo library access, select “Keep add only”
8. Turn on all observation options
9. Tap gallery
10. Tap “open settings” from “please allow gallery access” screen
11. Tap “Photos” in iNat Next settings
12. See that the only options are “None” or “Add Photos Only”

**Expected behavior**
A user should be able to edit their permissions later in iOS settings.

**Screenshots**
Screenshot 2024-05-29 at 1 19 20 PM

 - Device: iPhone 13 pro
 - OS: 17.4.1

**Additional context**
This probably requires some investigation of what should happen after a user says "keep add only" and maybe more digging into [react-native-permissions](https://github.com/zoontek/react-native-permissions)

budowski:


jtklein (changes requested):
This does not fix the problem. The change on my device is that it asks for the two permissions, read and write, immediately one after the other. If I press "Keep Add Only" on the second permission I still have the same situation when navigating to the Settings app that I can not change the Setting into allowing access to all photos.

jtklein:
#1594 is now completed. So, you can check if this applies at all anymore.

budowski:
Does not happen anymore after refactor work by @jtklein

PR #1717, opened 2024-06-27
#1104 - when searching for taxon, use the general /search endpoint

### LINKED ISSUE #1104: Multiple common names appear differently on TaxonSearch and ObsEdit
**Describe the bug**
When you have multiple common names enabled, they often appear differently on TaxonSearch vs ObsEdit

**To Reproduce**
1. On the web, set your Common Name Lexicon Display Order to match locale first (United States), then Korean: Screenshot 2024-02-01 at 2 06 25 PM
2. In the app, start a new obs w/o evidence
3. Tap the ADD AN ID button
4. Tap the SEARCH FOR A TAXON button
5. Search for "bats"; I see this Screenshot 2024-02-01 at 2 08 15 PM, even though the taxon Chiroptera does have a Korean common name
6. Search for "vascular plants"; I see this: Screenshot 2024-02-01 at 2 09 32 PM, which is what I'm expecting
7. Tap the green check button to choose Tracheophyta
8. On ObsDetail, the taxon is displayed with a single common name: Screenshot 2024-02-01 at 2 10 27 PM

**Expected behavior**
TaxonSearch and ObsEdit should both show multiple common names if that's the user's preference, and the name should look the same on both screens.

 - Device: Simulated iPhone 15 Pro
 - OS: iOS 17.2
 - App version: v0.21.1+70

**Additional context**
We don't do any handling of multiple common names in the app, so I suspect this is a matter of not requesting the correct data at the right time, or it's a problem with not authenticating API requests, or maybe it's a problem with the API or cached content from the API. If you've ruled out problems with the app, reach out to the web team to figure out if it's an API issue.

budowski:
Close #1104


kueda:
@budowski you've got a legit test failure here. You probably need a mock the response for the different endpoint.

PR #1749, opened 2024-07-02
1593 potential disagreement bottom sheet and disagreeing ids

angielt:


src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -121,6 +129,7 @@ const ObsDetailsContainer = ( ): Node => {
   const { params } = useRoute();
   const {
     suggestedTaxonId,
+    taxon,
```
albullington, line 132: One issue I'm seeing with passing this full taxon object through params is the following warning. If we're going to use params, I'd say we should only use the parts of the taxon object we need to show the user or send to the API (i.e. taxon id, photo, name, common name). 

```
 WARN  Non-serializable values were found in the navigation state. Check:

TabNavigator > TabStackNavigator > ObsDetails > params.taxon.default_photo._synced_at (Tue Jul 02 2024 08:57:56 GMT-0700)

This can break usage such as persisting and restoring state. This might happen if you passed non-serializable values such as function, class instances etc. in params. If you need to use components with callbacks in your options, you can use 'navigation.setOptions' instead. See https://reactnavigation.org/docs/troubleshooting#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state for more details.
```

src/components/ObsDetails/Sheets/PotentialDisagreementSheet.js
```diff
@@ -0,0 +1,88 @@
...
+      key={radioRow}
+      value={radioValues[radioRow].value}
+      checked={checked === radioValues[radioRow].value}
+      onPress={() => setChecked( radioValues[radioRow].value )}
+      label={radioValues[radioRow].label}
+    />
+  );
+
+  return (
+    <BottomSheet
```
albullington, line 62: Could we use `SharedComponents > Sheets > RadioButtonSheet` here instead of a custom implementation? It's generally better to reuse our existing components instead of rewriting code. You would need to modify that bottom sheet to accept the additional elements here (maybe it could take in props like `optionalText` and `showsTaxon`.

src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -416,6 +458,44 @@ const ObsDetailsContainer = ( ): Node => {
     createIdentificationMutation.mutate( { identification: idParams } );
   };
 
+  const onSuggestId = useCallback( ( ) => {
+    // check for disagreement
+    let observationTaxon = observation.taxon;
```
albullington, line 463: For code like this, it can be helpful to leave few comment lines above with a link to the existing iNaturalist web code. That way other devs know where it's coming from and it's easy to double check that it's actually at parity with the web code.

```
// based on disagreement code in iNat web
// https://github.com/inaturalist/inaturalist/blob/30a27d0eb79dd17af38292785b0137e6024bbdb7/app/webpack/observations/show/ducks/observation.js#L827-L838
```

src/components/ObsDetails/Sheets/PotentialDisagreementSheet.js
albullington: A few UI issues here:
- there's an extra space in the "Is the evidence..." test

I believe these three would be addressed by using `RadioBottomSheet`:
- text is running off the right side of the screen
- the top line of radio text is supposed to align with the button, rather than text being centered next to the buttons
- the top radio button should be selected to start

Screenshot 2024-07-02 at 9 17 33 AM

src/i18n/strings.ftl
```diff
@@ -288,6 +288,7 @@ Deleting-x-of-y-observations =
 DETAILS = DETAILS
 # Button that disables the camera's flash
 Disable-flash = Disable flash
+Disagreement = @{ $username } disagrees this is { $commonName } ({ $scientificName })
```
albullington, line 291: this line of text should start with this symbol *

albullington (changes requested):
looking good overall -- the bottom sheet and disagreement text are both working as expected for me. there are a few small UI and text fixes to make sure we're in line with what's in Figma. other than that, I'd simplify what we're sending through the navigation parameters, make sure we're clearing state when we need to in ObsDetails, and use our reusable sheet component. nothing really blocking here except for the UI changes and the need to reset taxon state between agreement/disagreement.

PR #1765, opened 2024-07-03
Show all obs photos in StandardCamera preview

kueda:
After discussing with Abhas, we want the StandardCamera to show all photos associated with an observation, even after backing out to the StandardCamera as an entry point.

This also tries to make some parts of the Zustand state a little more consistent and descriptive, and tries to clean things out of state when they're no longer needed.

### DIFF
--- a/src/components/Camera/Buttons/GreenCheckmark.js
+++ b/src/components/Camera/Buttons/GreenCheckmark.js
@@ -9,10 +9,12 @@ import { useTranslation } from "sharedHooks";
 import colors from "styles/tailwindColors";
 
 type Props = {
+  disabled?: boolean,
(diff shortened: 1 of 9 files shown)


albullington (approved):
Much more readable. I wonder if it's worth leaving a comment somewhere noting that these are the uris we've already rotated, not the exact uris coming from the camera library. My original thinking here was that we're doing a lot of photo manipulation (rotating, resizing, adding exif) in different places, so it's nice for developers to know exactly what stage the camera uris are at once they're in the store. On the other hand, hopefully this code is starting to be stable enough that we won't need to remember these details.

kueda:
Thanks, I added a comment to that effect.

Issue #1755, opened 2024-07-02
Online suggestions location permissions


albullington:
More changes to the online suggestions screen.

This screen is the default state for all users when they first use the app - a user should not see any location permissions modals in the app without opting in (see https://github.com/inaturalist/iNaturalistReactNative/issues/1594)

- [x]  just like in regular online suggestions, the results will effectively only use the Vision score. Since no location is included in the API request, the Geo score is empty, and the Combined score is the same as the Vision score.
- [x]  add a primary green button that always appears first beneath the photos that says "Improve these suggestions by using your location" - clicking this button should open the location permission gate or the location permissions denied screen.
- [ ]  If a user gives location permissions after prompting the location permissions modal, the suggestions screen should reload and then show the online suggestions state with location enabled.
- [x] per [this comment](https://github.com/inaturalist/iNaturalistReactNative/issues/1592#issuecomment-2183421638), make sure the Improve With Location button does not appear unless a user is coming from the AICamera or Camera screen. It should not appear when a user is coming from the PhotoGallery, GroupPhotos, PhotoSharing, ObsEdit, ObsDetail, etc.

This is split off from the bigger enhancement issue with changes to suggestions (https://github.com/inaturalist/iNaturalistReactNative/issues/1592) because it's dependent on the changes to permissions issue (https://github.com/inaturalist/iNaturalistReactNative/issues/1594). Once that is finished, we'll be able to prompt a user for location permissions on Suggestions when coming from the AICamera or StandardCamera.

jtklein:
As exemplified here [const { hasPermissions, renderPermissionsGate, requestPermissions } = useLocationPermission( );](https://github.com/inaturalist/iNaturalistReactNative/blob/6fa6db58111365eff19478903b25c0b69b0c9324/src/components/Suggestions/SuggestionsContainer.tsx#L53) I have added a hook to check for location permission now.
You can use the `hasPermissions` variable for conditional rendering of components that should only show with or without location permission.

[...]

Issue #1744, opened 2024-07-01
Refactor location fetch using watchPosition


kueda:
Current useUserLocation and useCurrentObservationLocation try to fetch location until the accuracy is good enough, but they do so using the Geolocation library's getCurrentPosition() method. Given my experience in the field, I'm not convinced this actually works very well. I often see the fetch giving up when the accuracy is still 50m, especially while offline, whereas switching the native iOS app quickly gets the accuracy down below 10m.

Instead, I think we should attempt a refactor using the watchPosition() method, which would start that process up and receive location updates until the accuracy is good enough, and then stop watching. I've experimented with this approach a bit, but I struggled to successfully stop watching, but I was probably messing something up.

So the task here is to start by refactoring useUserLocation to use watchPosition(). It should expose the same API (i.e. the hook should return the same values).

I would not spend more than a day or two on this without taking stock and discussing. This might also be the solution to #1688, which I'm stuck on and need to hand off.

albullington:
Closed by https://github.com/inaturalist/iNaturalistReactNative/pull/1788

PR #1788, opened 2024-07-10
Refactored location fetching for accurate locations

### LINKED ISSUE #1688: Location not getting refetched from camera after previous uses fell below accuracy threshold
**Describe the bug**
Hard to replicate, but as reported by Carrie, you can make a lot of observations and if one use of the camera within the session gets coordinates with accuracy that are below 10m, you may end up with all your subsequent observations having exactly the same coordinates. Restarting the app or otherwise unloading the camera component would reset it.

**To Reproduce**
(Theoretical) steps to reproduce the behavior:
1. Make sure you are using an iPhone with a SIM card so you are getting the finest location accuracy possible, or prepare to wait a long time for GPS positions to get really accurate
2. Open the StandardCamera
3. Take photos or wait for 10 seconds to make sure you're getting a good location read
4. On ObsEdit, confirm the coordinates are below 10m. If not, repeat previous steps until you get to that state
5. Walk for 5 minutes
6. Make a new observation with the StandardCamera
7. Walk for 5 minutes
8. Make a new observation with the StandardCamera
9. All of the observations you just created have exactly the same coordinates

**Expected behavior**
The coordinates should get refetched every time the camera is opened.

 - Device: iPhone 13 Pro
 - OS: iOS 17.5.1
 - App version: 0.33.1+96

**Additional context**
Carrie reported this, and at present her symptomatic observations are at https://www.inaturalist.org/calendar/carrieseltzer/2024/6/15. This is almost certainly related to my recent work on continuous location fetch. It's probably just holding on to the one user location when it falls below the accuracy threshold, and I didn't catch it b/c my iPhone SE w/o a SIM card generally never gets much bettern than 20-30 m.

albullington:
Closes #1688

- Location fetching uses watchPosition, as suggested in https://github.com/inaturalist/iNaturalistReactNative/issues/1744
- Exposes `userLocation` for scenarios where we want a location value that's not attached to an observation, like ProjectContainer (replacing the `useUserLocation` hook). Otherwise, updates observation keys with the new location
- Stops watch when an accurate enough location is found

Note: I resisted the urge to refactor `useWatchPosition` out of the Camera and into Suggestions, since that should be addressed in part by the changes to permissions issue: https://github.com/inaturalist/iNaturalistReactNative/issues/1594. Right now, I think having `useWatchPosition` in the Camera slows down performance a tiny bit, but we need permission changes and a permission gate (https://github.com/inaturalist/iNaturalistReactNative/issues/1755) for user location fetching on the Suggestions screen

A few things to think about when reviewing:

1. Does this impact performance, since this hook is called multiple times? (My sense is yes for the current Camera screen which we'll refactor into Suggestions soon, and no for other flows)
2. Are locations accurate when walking to different locations online/offline? Are there any moments when they're not updating?
3. Are gallery photos still using the correct evidence-location? Are there scenarios when this is being overwritten by the watch position hook? (It seemed like this was happening for me out in the field, but not at home...)
4. Do we need a timeout so the loading wheel doesn't show forever on ObsEdit?


src/sharedHooks/useWatchPosition.ts
```diff
@@ -0,0 +1,125 @@
...
+  distanceFilter: 0,
+  enableHighAccuracy: true,
+  maximumAge: 0,
+  timeout: 2000
+};
+
+const useWatchPosition = ( options: {
+  retry: boolean
+} ): WatchPosition => {
+  const updateObservationKeys = useStore( state => state.updateObservationKeys );
```
kueda, line 40: I strongly advise removing this and any observation-specific logic from this hook. It's called watchPosition, so it should just watch the position. If consumers of this module need to do other stuff, they should pass in callbacks as arguments.

In general, you want code to be modular (i.e. reusable) and encapsulated (with as few external dependencies as possible).  There are all kinds of reasons we might want to watch the user's location, so we don't want to tie this functionality to one specific use case.

src/sharedHooks/useWatchPosition.ts
```diff
@@ -0,0 +1,125 @@
...
+  const [subscriptionId, setSubscriptionId] = useState<number | null>( null );
+  const [locationPermissionResult, setLocationPermissionResult] = useState<string | null>( null );
+  const [errorCode, setErrorCode] = useState<number | null>( null );
+  const hasLocation = currentObservation?.latitude && currentObservation?.longitude;
+  const [shouldFetchLocation, setShouldFetchLocation] = useState<boolean>( true );
+  const [userLocation, setUserLocation] = useState<UserLocation | null>( null );
+
+  const watchPosition = ( ) => {
+    const success = ( position: GeolocationResponse ) => {
+      setLocationPermissionResult( "granted" );
```
kueda, line 52: Elsewhere you configured `@react-native-community/geolocation` to `skipPermissionRequests` on the assumption that we will handle that... so there's no real reason to do this unless you really want to change `locationPermissionResult` to a new string object every single time we get a new a new location. I think we can just set the default value of `locationPermissionResult` to "granted" and only change it in the failure callback.

kueda (changes requested):
Doesn't work on my iPhone SE.

1. Fresh install
1. Change to advanced mode
1. Add an obs w/o evidence
1. Immediately get this infinite recursion leading to a freeze:
    ```
    ERROR  Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
       in PermissionGateContainer (created by LocationPermissionGate)
       in LocationPermissionGate (created by EvidenceSection)
       in RCTView (created by View)
       in View
       in NativeWind.View
       in NativeWind.NoName
       in Unknown (created by EvidenceSection)
       in EvidenceSection (created by EvidenceSectionContainer)
   ```

I get the same problem when starting with a camera and granting all permissions, except it locks up on Suggestions, even if I run from 3f823d0e33dbc5e0ad759ee2852872d1b212c2b0 (i.e. w/o the typescript changes).

jtklein:
Changes to permissions PR #1594 are now merged in.
I did update the way useCurrentObservation is using location permission. Instead of checking for permission every time in the useEffect I added the `hasPermissions` state in as a hook param. I was not sure if `useUserLocation` and `fetchUserLocation` should check for permission every time they run or if the state kept in the functional component that make use of them is enough, and felt changing it is out of scope for my PR.
So, what I can see in the `useWatchPosition` hook here, you can decide if it is enough to know wether the rendered component has location permission or if the hook should check every time.

PR #1791, opened 2024-07-11
Hide skip ID button on Suggestions sometimes

### LINKED ISSUE #1776: Remove "add an id later" button in suggestions when coming from obs detail
Right now, the "add an id later" button on suggestions is acting weird or crashing when you come from obs detail via the suggest id button on your own or others' observations. It also seems strange to have it when a user navigates to "edit an observation" that already has an ID associated with it.

- [ ] remove the "add an ID later" button inside suggestions when a user navigates to suggestions from an obs detail page by clicking "suggest ID," whether on their own observations or others' observations.
- [ ] remove the "add an ID later" button inside suggestions when a user clicks the pencil icon next to an already-added ID under the "identification" section of obs edit. this can be whether the obs has been uploaded already or not. If they have already selected an ID and are editing it, then the "add an id later" button is irrelevant.

Put another way:
"Add an ID later" should only appear on the suggestions screen when the previous screen was:
- AI camera
- standard camera
- 1 photo selected from gallery (#1779)
- 1 observation created in group photos screen

kueda:
We only want to show the text "Add an ID later" when one needs to move forward without adding an ID, so this hides it

* When reached from ObsDetails (you can go back)
* When reaced from ObsEdit and there is an existing identification (you've already added an ID so if you're editing it you need to choose a taxon or back out)

Closes #1776


kueda:
I guess we both worked on this by accident, @albullington, so you should decide which solution you like better.

albullington (approved):

PR #1796, opened 2024-07-13
Show unsynced obs before all other obs

### LINKED ISSUE #1577: Show unsynced observations before all others in MyObs
Currently we just sort MyObs by `_created_at`, which means that any unsynced observations will get buried under newly-created observations on the web or in other apps. Instead,

* show all unsynced observations above other observations
* continue sorting them by `_created_at`, just show them above all the others
* as they upload they can go back to their previous positions

This will probably involve some weird ordering code.

budowski:
Achieves unsynced-first sort with a new built-in needs_sync property

Closes #1577


albullington:
@budowski is this working for you? I tried logging into my regular iNat account on staging, then creating ~9 new observations. I uploaded a couple of observations in the middle of the list, expecting that once they were uploaded, they would be displayed below the unsynced observations. At first, nothing changed, even when I closed and reopened the app.

Eventually I did see a shift in the order of observations, but the most recently uploaded observations are above the unsynced observations in the sorted list, and all my observations downloaded from web are below the unsynced observations.

albullington (changes requested):
Sort doesn't seem to work in the way I expect. It leaves unsynced observations in the middle of the list, below observations which have been recently uploaded.

albullington (approved):

PR #1864, opened 2024-07-26
1843 offline taxon search

### LINKED ISSUE #1843: OfflineNotice for TaxonSearch
- [x] Add iconic taxa as the default results for TaxonSearch the same way it works for ExploreTaxonSearch (why are these different components, anyway?)
- [ ] Add an OfflineNotice to the search results area of ExploreTaxonSearch and TaxonSearch when the user enters a query but there are no results because they are offline.
- [ ] Tapping the OfflineNotice should try to redo the search

jtklein:
Closes #1843


src/components/SharedComponents/TaxaList/TaxaList.tsx
```diff
@@ -0,0 +1,75 @@
+import { refresh, useNetInfo } from "@react-native-community/netinfo";
+import { ActivityIndicator, OfflineNotice } from "components/SharedComponents";
+import { View } from "components/styledComponents";
+import React from "react";
+import { FlatList } from "react-native";
+import { useIconicTaxa } from "sharedHooks";
+
+interface Props {
+  isLoading: boolean;
+  taxa: {}[];
```
jtklein, line 10: Not empty object as type

Issue #1798, opened 2024-07-15
UI fixes for label headers on radio button sheets


albullington:
**Describe the bug**
Label headers in the radio bottom sheets are misaligned and using the wrong font.

**To Reproduce**
1. Go to Explore screen
2. Tap on leaf icon to bring up Explore View bottom sheet
3. Header labels are misaligned -- too far from radio button and too close to icons. It looks like the font might have changed from the Figma designs, too

or

1. Take an observation with no evidence
2. Open the Geoprivacy sheet
3. See header labels far away from radio buttons in wrong font

**Expected behavior**
Label headers should match the Figma designs: https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=14942-36008&t=9boOhQ8bJ8oQjW18-4

**Screenshots**
<img width="483" alt="Screenshot 2024-07-15 at 1 42 05 PM" src="https://github.com/user-attachments/assets/47129756-97cf-49f7-8fd5-733ba07ebb47">

**Context (please complete the following information):**
 - OS: iOS
 - App version: v0.36.0+101

jtklein:
From what I have poked around I think this was introduced with #1749 ,
maybe that helps with fixing it.
Also, see https://inaturalist.slack.com/archives/C02G5AATVGA/p1721117575103769

albullington:
@budowski not sure if you've started working on this, but I believe this will be fixed via https://github.com/inaturalist/iNaturalistReactNative/pull/1865

abhasinat:
The UI for all radio bottom sheets is still not matching the Figma designs. Here are some places to reference:
- [explore view chooser](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=8008-40019&t=Fpd9fOjOVmOwoZy7-11) bottom sheet
- [geoprivacy & wild status ](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=711-9042&t=Fpd9fOjOVmOwoZy7-11)bottom sheets

PR #1865, opened 2024-07-26
Use DisplayTaxonName component for disagreements

### LINKED ISSUE #1852: Disagreement notice does not work when disagreeing with a taxon that has no common name
**Describe the bug**
When you view ObsDetail and there is an identification that disagrees with a taxon that has no common name, you see a notice like "@kueda disagrees this is {%commonName} (_Something something_)"

**To Reproduce**
1. Go to https://www.inaturalist.org/observations/200550773
2. Note the erroneous text in the disagreement section

**Expected behavior**
* the text should work when there's no common name
* the text should respect the user's name display preferences (e.g. if they've chosen sciname first, it should show the sciname first

**Screenshots**
Screenshot 2024-07-24 at 4 19 53 PM

 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.5.1
 - App version: 0.36.1 (102)

**Additional context**
This really shouldn't just a single i18n string. We should probably use DisplayTaxonName here.

albullington:
Closes #1852

### DIFF
--- a/src/components/ObsDetails/ActivityTab/ActivityItem.js
+++ b/src/components/ObsDetails/ActivityTab/ActivityItem.js
@@ -48,8 +48,6 @@ const ActivityItem = ( {
     && isFirstDisplay
     && currentUserId;
 
-  console.log( currentUserId );
-
   const navToTaxonDetails = ( ) => navigation.navigate( "TaxonDetails", { id: taxon.id } );
 
   return (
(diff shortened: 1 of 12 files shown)


albullington:
@angielt thought you might want to take a look since this is code you worked on. The main changes here are:

- Getting `DisplayTaxonName` to display in a plain text component with formatting for each of the three user settings, common name first, scientific name first, and only sci name
- Passing `DisplayTaxonName` as a component into `<Trans />`, which prevents a user from seeing the actual variable names (like commonNames) displayed as part of the text string when the variable is null
- Moving logic specific to disagreement out of `RadioButtonSheet` so that sheet doesn't need to know details about a taxon, only whether or not to display additional components (I believe removing `showTaxon` code also fixed https://github.com/inaturalist/iNaturalistReactNative/issues/1798)

PR #1951, opened 2024-08-11
Show generic profile icon if user has no profile image

budowski:


kueda (changes requested):
- [x] UserProfile looks like this: Screenshot 2024-08-12 at 4 41 50 PM
- [x] Drawer icon gets cut off at the bottom: Screenshot 2024-08-12 at 4 42 29 PM

kueda (approved):

PR #1971, opened 2024-08-14
Fixes and error handling for API throttling

### LINKED ISSUE #1661: Alert user to API throttling and try to prevent it
Several of our test users have hit problems where their usage of this app triggers our own [API throttling limits](https://www.inaturalist.org/pages/api+recommended+practices). From the user's perspective this will manifest as something not working, e.g. taxon search results or following someone. The software will see this as 429 errors from our own API. Obviously the user should never encounter this, but to facilitate debugging and to shame us into fixing it, I think we should alert the user when this happens. And, of course, we should try to prevent it from happening.

- [x] Make a new error class called `INatApiTooManyRequestsError` that extends `INatApiError` and throw it when we encounter a 429 response from our own API. This error should not be caught, should be fatal, and should crash the app.
- [x] Audit our use of `logger` to make sure we're only using it to record information we're paying attention to and *not* as a stand-in for `console.log`. This was the most-hit endpoint from this app over the past week, more than 3x the next most-hit endpoint.
- [ ] Try to refactor SpeciesSeenCheckmark so that every single instance of this component doesn't trigger a separate request to /v2/observations to determine if the user has observed that taxon. This might require some changes to the API ([see slack convo](https://inaturalist.slack.com/archives/C04MF2V4ZAN/p1718044108328169))

albullington:
Closes #1661


albullington:
@kueda review would be helpful to know:
1. does this meet the requirements for fatal error handling?
2. are there any `logger` statements I removed that we do need to track? 
3. are there any `logger` statements I left that we don't really need? I'm assuming we're using `pickup` and `isAdvancedUser` for tracking purposes

src/components/Explore/SpeciesView.js
```diff
@@ -67,11 +73,42 @@ const SpeciesView = ( {
...
+  const taxonIdsSeenByCurrentUser = useMemo( ( ) => seenByCurrentUser?.results?.map(
+    r => r.taxon.id
+  ) || [], [seenByCurrentUser?.results] );
+
+  useEffect( ( ) => {
+    if ( taxonIdsSeenByCurrentUser.length > 0 ) {
+      const uniqueSeenList = _.uniq( taxonSeen.concat( taxonIdsSeenByCurrentUser ) );
+      if ( uniqueSeenList.length > taxonSeen.length ) {
+        setTaxonSeen( uniqueSeenList );
+      }
```
kueda, line 108: If you're maintaining a unique list of numbers, you probably want to use a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead of an array. That way you don't need to use `_.uniq()`. 

Also, `taxonSeen` sounds like it's either a taxon or a boolean. Since it is instead a set of IDs of taxa observed by the current user, a better name might be something like `observedTaxonIds`.

src/components/Explore/SpeciesView.js
```diff
@@ -67,11 +73,42 @@ const SpeciesView = ( {
...
+    } ),
+    {
+      enabled: !!( taxonIds.length > 0 && currentUser )
+    }
+  );
+
+  const taxonIdsSeenByCurrentUser = useMemo( ( ) => seenByCurrentUser?.results?.map(
+    r => r.taxon.id
+  ) || [], [seenByCurrentUser?.results] );
```
kueda, line 101: This variable really just captures a *page* of taxon IDs seen by the current user, so to avoid confusion with other variables that hold similar data, this might be called `pageObservedTaxonIds` or something similar.

src/components/Explore/SpeciesView.js
```diff
@@ -67,11 +73,42 @@ const SpeciesView = ( {
     {
       ...queryParams,
       fields: {
-        taxon: Taxon.TAXON_FIELDS
+        taxon: Taxon.SCORE_IMAGE_FIELDS
```
kueda, line 76: What does this have to do with "scoring"?

src/components/Explore/SpeciesView.js
```diff
@@ -67,11 +73,42 @@ const SpeciesView = ( {
...
       }
     }
   );
 
+  const taxonIds = data.map( r => r.taxon.id );
+
+  const { data: seenByCurrentUser } = useAuthenticatedQuery(
+    ["fetchSpeciesCounts", taxonIds],
+    ( ) => fetchSpeciesCounts( {
```
kueda, line 85: You need to use `optsWithAuth` to actually authenticate this query and get the correct preferred_common_name values in the response.

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -236,7 +231,7 @@ const getJWT = async ( allowAnonymousJWT = false ): Promise<string | null> => {
     try {
       response = await api.get<{api_token: string}>( "/users/api_token.json" );
     } catch ( getUsersApiTokenError ) {
-      logger.error( "Failed to fetch JWT: ", getUsersApiTokenError );
+      if ( !getUsersApiTokenError ) { return null; }
```
kueda, line 234: `getUsersApiTokenError` cannot be falsy. I don't see this error in the logs so I don't think it's a liability, so I would err on the side of logging it. If you disagree, I don't think we need this try/catch.

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -455,14 +450,10 @@ const authenticateUser = async (
...
 
-  logger.debug( "writing current user to realm: ", localUser );
   safeRealmWrite( realm, ( ) => {
     realm.create( "User", localUser, UpdateMode.Modified );
   }, "saving current user in AuthenticationService" );
-  const currentRealmUser = User.currentUser( realm );
-  logger.debug( "Signed in", currentRealmUser.login, currentRealmUser.id, currentRealmUser );
-  const realmPathExists = await RNFS.exists( realm.path );
-  logger.debug( `realm.path exists after sign in: ${realmPathExists}` );
+  await RNFS.exists( realm.path );
```
kueda, line 456: I don't think we need this if we're not going to log whether it exists.

src/components/ObsDetails/DQAContainer.js
```diff
@@ -131,8 +128,7 @@ const DQAContainer = ( ): React.Node => {
         await refetchRemoteObservation( );
         setNotLoading( );
       },
-      onError: error => {
-        logger.error( "createQualityMetricMutation failure", error );
+      onError: ( ) => {
```
kueda, line 131: This seems like a useful log event, no? Was it happening a lot?

src/components/ObsEdit/EvidenceList.js
```diff
@@ -193,11 +191,7 @@ const EvidenceList = ( {
     if ( obsSound?.id ) {
       deleteObservationSoundMutation.mutate( { uuid: obsSound.uuid }, {
         onSuccess: removeLocalSound,
-        onError: deleteRemoteObservationSoundError => {
-          logger.error(
-            "[EvidenceList.js] failed to delete remote observation sound: ",
-            deleteRemoteObservationSoundError
-          );
+        onError: ( ) => {
```
kueda, line 194: If this failed because there's no internet connection, then sure, we should show an alert and not log the error, but if it failed for any other reason, that's a bug and we should log it.

kueda (changes requested):
Needs a few changes but overall looks ok. I like the refactor of SpeciesSeenCheckmark. Some comments and questions in addition to inline ones:

* In general we should log errors. We seem to have a bad habit of assuming a errors that come from React Query can only mean there's no internet connection, but that isn't true. If we want to show the user something when a requests fails due to lack of internet and skip logging, we should check the error to make absolutely sure that's what the error was for. Otherwise we're going to miss valuable information about unexpected problems.
* I am seeing logs for requests for `/v2/observations/species_counts?return_bounds=true`. Is that new? Why is that happening? `species_counts` never needs the bounds

albullington:
@kueda `/v2/observations/species_counts?return_bounds=true` isn't new, and we're passing that as query parameters for all of the infinite scroll queries in Explore views (Observations, Species, Identifiers, Observers). It's currently the default state value in `ExploreContext`. I'm not sure if we need it for loading the correct map in ObservationsView or if we can remove it.

src/components/Explore/SpeciesView.js
albullington: I didn't think we needed `optsWithAuth` since we're only returning `taxonIds` from this query, and taxon ids don't need translation or any current user preferences...

kueda: True, but in that case this doesn't need to be authenticated, right?

PR #1990, opened 2024-08-22
Preserve MyObs scroll position

### LINKED ISSUE #1957: Navigation issue: MyObs -> ObsEdit -> back scrolls you back to the top of the screen
**Describe the bug**
Tapping from MyObs to ObsEdit via an unsynced observation scrolls you back to the top of the MyObs screen.

**To Reproduce**
1. Make ~20 unsynced observations via the photo importer. Save all of them
2. Scroll down on MyObs
3. Tap into an unsynced obs to go to ObsEdit
4. Tap back button
5. Land back at the top of MyObs, rather than seeing your scroll state preserved

**Expected behavior**
A user should not lose their scroll position when navigating away from MyObs

**Additional context**
This is probably a problem of jumping from one stack navigator to another and back, so it might be tricky to solve.

kueda:
Symptom: when you scroll MyObs, tap an unsynced obs to enter ObsEdit, then go back, MyObs is always scrolled back to the top.

Diagnosis: entering ObsEdit requires opening a parallel stack navigator to the one that contains MyObs, which means the MyObs screen gets destroyed and it loses its scroll position.

Treatment: the ideal treatment would be to not use parallel stack navigators, but since we don't have a way to do that *and* ditch the bottom tabs for certain screens, my approach here was to store the scroll offset in Zustand when we navigate to ObsEdit, and when we return to MyObs we scroll the list to that position. We do *not* want this behavior after creating an observation, b/c we want the newly-created obs to be visible at the top of the list.

Questions for review:

* Feels real awkward. Is there a better way?
* Is there a way to test this? I want to write a test that scrolls MyObs, expects the first obs to be off-screen, taps an unsynced obs, then goes back, then expects the first obs to still be off-screen, but I don't see a way to test if view is off-screen.

Closes #1957

### DIFF
--- a/src/components/MyObservations/MyObservations.js
+++ b/src/components/MyObservations/MyObservations.js
@@ -15,12 +15,15 @@ type Props = {
   currentUser: Object,
   handleIndividualUploadPress: Function,
   handleSyncButtonPress: Function,
-  isFetchingNextPage: boolean,
   isConnected: boolean,
+  isFetchingNextPage: boolean,
   layout: "list" | "grid",
+  listRef?: Object,
   numUnuploadedObservations: number,
   observations: Array<Object>,
   onEndReached: Function,
+  onListLayout?: Function,
+  onScroll?: Function,
   setShowLoginSheet: Function,
   showLoginSheet: boolean,
   status: string,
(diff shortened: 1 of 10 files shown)


kueda:
Sorry Amanda, didn't check the calendar to see if you were out. Happy to discuss and adjust when you get back if you like, though.

PR #1999, opened 2024-08-23
DRY Explore containers

kueda:
RootExploreContainer and ExploreContainer were basically duplicates with minor tweaks, which was frustrating when you needed to change something in both, and as a result they were starting to drift. This PR unifies all of the logic they share into ExploreContainer and separates the ways in which they differ into RootExploreContainer and StackedExploreContainer.


kueda:
@albullington weird, nothing should have changed with signing. Did you try running from xcode and not using the commandline react native tools? Either way, we can revisit next week after soft launch. Nothing here is critical for functionality.

Issue #1570, opened 2024-05-20
Attribution should not appear on MediaViewer from Suggestions for new photo


kueda:
**Describe the bug**
When you are making a new obs with the camera and you land on Suggestions, you can tap the photo to view it in the MediaViewer. There we show an attribution button that probably shows the wrong license choice

**To Reproduce**
1. Open one of the cameras and take a photo
2. When you land on Suggestions, tap the photo you just took to enter the MediaViewer
3. The attribution button shows the copyright icon no matter what your default license choice is
4. Tap the attribution
5. The alert says "Copyright. all rights reserved" no matter what your default license choice is

**Expected behavior**
Te information should be accurate or there should be no attribution

**Screenshots**

**Context (please complete the following information):**
 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.4.1
 - App version: 0.32.3 (91)

**Additional context**
Talked w/ Abhas and we agreed to just remove the attribution from the MediaViewer when reached from Suggestions for a *new* photo (it should still appear when you're viewing suggestions for one of your synced observations or someone else's observation.

israr002:
@jtklein i have worked on this and raised the pr

PR #2040, opened 2024-09-04
Remove attribution for new photo in MediaViewer

israr002:
closes https://github.com/inaturalist/iNaturalistReactNative/issues/1570

### DIFF
--- a/src/components/MediaViewer/AttributionButton.tsx
+++ b/src/components/MediaViewer/AttributionButton.tsx
@@ -1,25 +1,22 @@
-// @flow
-
 import {
   TransparentCircleButton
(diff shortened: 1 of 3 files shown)


jtklein:
Thank you @israr002 for your contribution.
I have:
- Changed the test to query and find the button by accessibility label instead of test ID, as this is the preferred approach by react-native-testing library
- Migrated the AttributionButton to Typescript. This is something we decided as a team, to slowly migrate whenever we touch a small leaf component or a component that solely imports TS children.

Have tested your branch in the app and I think it works as it should, in order to fix the issue.

Again, thank you.

Issue #1885, opened 2024-07-29
Explore Species view shows global results for Nearby


kueda:
**Describe the bug**
When you go to to Explore with location permission, the Species view shows global results even though the header says it's showing Nearby results

**To Reproduce**
1. Make sure you've granted location permission
2. Restart the app
3. Go to Explore. You should be in Species view. Let it load
4. The header says you're viewing Nearby results, but no matter where you are you see Mallard, Western Honey Bee, Asian Lady Beetle, House Sparrow, Monarch, American Robin, etc, i.e. the top *global* results (if you're the SF Bay Area it should be more like Western Fence Lizard, CA Poppy, Anna's Hummingbird, etc).

**Expected behavior**
Explore should show Nearby results by default, not just say it's showing nearby results

**Context (please complete the following information):**
 - Device: iPhone SE (3rd gen)
 - OS: 17.5.1
 - App version: 0.37.0 (103)

**Additional context**
Tapping into observations view seems to actually show Nearby results. If you have not granted location permission and you do so from Explore, we also seem to show Nearby results.

This might also explain why Explore feels slow to load for some people.

PR #2045, opened 2024-09-05
Stop Explore from defaulting to worldwide in Android

### LINKED ISSUE #2032: Explore defaults to worldwide
**Describe the bug**
On Android, going to Explore defaults to worldwide, even when you have already granted location permissions.

**To Reproduce**
1. Make sure you have granted location permissions
2. Go to Explore
3. Note that the header says "Nearby" briefly then switch to "Worldwide"
4. Woldwide results load

**Expected behavior**
Nearby results should load be default.

 - Device: Pixel 8
 - OS: Android 14
 - App version: 0.44.0

israr002:
Closes #2032


kueda (changes requested):
If you disable high accuracy in *all* contexts (iOS, obs creation, etc), people are going to end up with unacceptably imprecise coordinates for their observations. The problem in #2032 is only in Explore, so please try to refactor in a way that addresses that problem without impacting other uses of user location, e.g. you might add an option to the fetchuserLocation hook like `fetchUserLocation( {enableHighAccuracy true } )` and make sure you use it whenever we're creating an observation or showing the user's location on a map.

kueda (changes requested):
That makes it specific to Android, but it's still misleadingly named, since it's going to request a fine location on iOS and a coarse one on Android. My original qualm about affecting observations isn't really valid since we're not using this hook for that, but I'm still concerned about future confusion regarding this platform-specific alteration, so how about this:

- [x] Rename this hook to `fetchCoarseUserLocation` so it's super clear that this is only to be used for purposes that don't require high accuracy
- [ ] Remove the platform-specific check for `enableHighAccuracy` since we're declaring that this is only for low-accuracy situations
- [ ] Leave the platform-specific check on `maximumAge`;  if possible, we want people who just got off a plane to get an up-to-date location
- [ ] Add a comment linking to https://github.com/michalchudziak/react-native-geolocation/issues/272 so it's clear to future maintainers (like me) that this is a bug in the underlying library

israr002:
@kueda I have updated as per your suggestion

kueda:
@israr002 you've still got one more thing in the checklist to do.

israr002:
No i have added a comment https://github.com/michalchudziak/react-native-geolocation/issues/272#issuecomment-2349543220

kueda:
@israr002, I'm asking for a comment in *this* repo linking out so that in the future when we're wondering why there is this weird OS-specific conditional, we know where to look to check if the underlying bug has been fixed.

israr002:
@kueda Sry my bad i misunderstood
i have added in the repo

kueda (changes requested):
While I think the work so far improves the situation, it doesn't fix the bug in #2032. When I go to Explore on my Pixel 8, the header says "Nearby" now, but the content is all for a worldwide search, e.g. the top species are the worldwide species, not the ones for where I am in California, and the map is showing tiles for the entire world.

israr002:
@kueda 
The issue you're facing was related to a different problem, which is already being fixed in #1885 . I've synced the code with the latest from main.

Regarding the map tiles issue, it's happening because we're currently passing a hardcoded parameter worldwide: true in the navigation flow. The relevant code is in the TaxonDetails screen:

navigation.navigate("TabNavigator", {
  screen: "TabStackNavigator",
  params: {
    screen: "Explore",
    params: {
      taxon,
      worldwide: true, // Hardcoded value
    },
  },
});
Before making any changes, I wanted to confirm if there's a specific reason why we're forcing worldwide: true. If not, we can  dynamically set the worldwide parameter based on state.placeMode. Let me know your thoughts!

kueda:
After your merge, @israr002, it seems to be working, or at least I can't reproduce with the steps in #2032, so I'm going to merge.

Regarding the worldwide: true param on TaxonDetails, I don't think that's related, but the reason that's there is that when you navigate to Explore from TaxonDetails, we want you to see *all* the observations of that taxon, not just the ones near you.

kueda (approved):

PR #2057, opened 2024-09-10
Username mentions navigate to User profile screen in-app

### LINKED ISSUE #1975: mentions in comments and remarks should open user profile
Right now, if you click a mention tag in a comment or remark, it opens the user's profile in an external browser. If a URL leads to https://www.inaturalist.org/people/[username], it should open in the iNat Next app.

jtklein:
Closes #1975


jtklein:
When a non-existing user is pressed the user profile screen is opening but no data is loaded. This needs to be changed before merging, I guess.

angielt (approved):

PR #2073, opened 2024-09-16
2024 faster explore load

### LINKED ISSUE #2024: Explore worldwide species load as quickly as the web
**Describe the bug**
Worldwide species in Explore is slow, even though it's showing exactly the same data you see at https://www.inaturalist.org/observations, which is very fast, so we're probably appending some unnecessary extra params that are busting the cache. I'm seeing the following that look unnecessary b/c they're defaults:

- `order_by=created_at`
- `order=desc`
- `quality_grade=research,needs_id`

And the following that we could probably change to mimic the web

- `per_page=20`

If removing / changing these doesn't get you a fast response, consult with the web team to choose the params that will utilize the cache.

jtklein:
Closes #2024

### DIFF
--- a/src/providers/ExploreContext.tsx
+++ b/src/providers/ExploreContext.tsx
@@ -52,10 +52,6 @@ export enum SORT_BY {
   MOST_FAVED = "MOST_FAVED",
 }
 
-// TODO: this should be imported from a central point, e.g. Taxon realm model
-// TODO: this is probably against conventioins to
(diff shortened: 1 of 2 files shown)


jtklein:
The issue was fixed by enabling server-side caching for queries that have the `fields` param set for the API v2 with https://github.com/inaturalist/iNaturalistAPI/compare/dd1471211c25...7f169fdae905

PR #2090, opened 2024-09-17
fix: do not allow rotation in PermissionGate on Android

### LINKED ISSUE #1205: Camera permission gate allows rotation on Android
**Describe the bug**
The camera permission gate layout rotates with the device. It should not.

**To Reproduce**
1. Remove camera permission
2. Enter the StandardCamera
3. The camera permission gate should appear
4. Rotate the device 90 degrees into landscape orientation
5. The layout rotates into landscape

**Expected behavior**
The permission gate should not rotate on phones under any circumstances. It should on tablet.

**Screenshots**
Screenshot_20240225-133122

 - Device: Pixel 8
 - OS: Android 14
 - App version: 0.23.0 (72)

**Additional context**
Note that the camera does provide some rotation sensitive elements like the preview and camera controls, so those should continue to work as is.

israr002:
Closes #1205

### DIFF
--- a/src/components/SharedComponents/PermissionGateContainer.tsx
+++ b/src/components/SharedComponents/PermissionGateContainer.tsx
@@ -8,8 +8,6 @@ import React, {
   useState
 } from "react";
 import { Platform } from "react-native";
-import DeviceInfo from "react-native-device-info";
-import Orientation from "react-native-orientation-locker";
 import {
   AndroidPermission,
   checkMultiple,
@@ -21,8 +19,6 @@ import {
 } from "react-native-permissions";
 
 import PermissionGate from "./PermissionGate";
-
-const isTablet = DeviceInfo.isTablet( );
 
 const usesAndroid10Permissions = Platform.OS === "android" && Platform.Version <= 29;
 const usesAndroid13Permissions = Platform.OS === "android" && Platform.Version >= 33;
@@ -104,9 +100,6 @@ interface MultiResult {
   [permission: string]: PermissionStatus;
 }
 export function permissionResultFromMultiple( multiResults: MultiResult ) {
-  if ( !isTablet ) {
-    Orientation.lockToPortrait( );
-  }
   if ( typeof ( multiResults ) !== "object" ) {
     throw new Error(
       "permissionResultFromMultiple received something other than an object. "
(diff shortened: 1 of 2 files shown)


src/components/SharedComponents/PermissionGateContainer.tsx
```diff
@@ -100,6 +104,9 @@ interface MultiResult {
   [permission: string]: PermissionStatus;
 }
 export function permissionResultFromMultiple( multiResults: MultiResult ) {
+  if ( !isTablet ) {
+    Orientation.lockToPortrait( );
+  }
```
kueda, line 109: This is an odd place to put this. I can see why this addresses the issue (orientation gets locked before PermissionGate renders), but it also does this locking unnecessarily *every* time we process permission results.

kueda (changes requested):
This does address the symptom, but I don't think it treats the disease. It also results in unnecessary calls to `Orientation.lockToPortrait( )`. The `useLockOrientation()` hook *should* be locking the entire app to portrait orientation by default on non-tablets, so why isn't that working on Android?

src/components/SharedComponents/PermissionGateContainer.tsx
israr002: I understand that this might seem like an odd place to lock the orientation. I initially took reference from src/components/Camera/CameraWithDevice.tsx, where a similar approach was used to lock the orientation. However, I will dive deeper into the root cause of the issue and explore why its not working in the PermissionGateContainer

kueda:
@israr002 are you still working on this? If not I'll close.

israr002:
@kueda  I haven't been able to get the root cause yet, but I'll continue working on it and aim to have it resolved by tomorrow. I'll keep you updated on the progress.

israr002:
@kueda The issue was caused by the orientation: "all" parameter in CAMERA_SCREEN_OPTIONS, which was added to allow the camera screen to rotate. However, it unintentionally removed the portrait lock.
this parameter is unnecessary because the camera library automatically handles screen orientation without affecting the portrait lock. Therefore, I've removed the orientation setting to restore the intended behavior.

kueda (approved):

kueda:
Thanks again, @israr002. That's a much simpler fix.

PR #2092, opened 2024-09-17
fix: hide Wikipedia button if server provides no Wikipedia URL

### LINKED ISSUE #1740: Hide wikipedia link in TaxonDetails if there's no URL
**Describe the bug**
There are some edge cases like [this one](https://www.inaturalist.org/taxa/231783-Strangalia-acuminata) where a taxon can have a wikipedia description but not a url. So we're showing a wikipedia link that the user can't open.

**To Reproduce**
1. Go to Strangalia acuminata Taxon Details screen via Explore
2. Tap on Read more on Wikipedia link
3. Nothing happens

**Expected behavior**
Right now, we're checking to see if a user can open the link and doing nothing if the wikipedia url doesn't exist. Instead, we should hide the "Read more on Wikipedia" link (or defer to @abhasinat on what the preferred UX is here).

OS: iOS
App version: v0.34.0+97

budowski:
Closes #1740

### DIFF
--- a/src/components/TaxonDetails/Wikipedia.js
+++ b/src/components/TaxonDetails/Wikipedia.js
@@ -30,11 +30,6 @@ const Wikipedia = ( { taxon }: Props ): React.Node => {
   const { t, i18n } = useTranslation();
   const { language } = i18n;
 
-  const openWikipedia = ( ) => {
-    if ( taxon?.wikipedia_url ) {
-      openExternalWebBrowser( taxon.wikipedia_url );
-    }
-  };
 
   let wikipediaUrl = taxon.wikipedia_url;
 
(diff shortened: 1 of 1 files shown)


kueda (changes requested):
The example of _Strangalia acuminata_ in the issue is really a server-side problem. IMO, we should do what the issue requested and hide the link if the server didn't provide `wikipedia_url`, because in theory it's the server's responsibility to provide a valid Wikipedia URL, and if it doesn't, the client should just assume there's a good reason for that.

So,

- [x] Hide the Wikipedia button if the server didn't provide `wikipedia_url`, or if it did and it was blank
- [x] Remove that vestigial code constructing the Wikipedia URL from the taxon name

kueda (approved):

PR #2093, opened 2024-09-17
Force refresh of home screen when drawer is opened/closed

### LINKED ISSUE #1886: Signing out of an account with 0 observations doesn't always work properly
**Describe the bug**
When you click signout of a logged in account that has zero observations (in mine and @tiwane's cases, we made brand new accounts) - the home screen still seems to display the username of that account that just logged out. in tony's case, the old username remained while being populated with his new account's data.

**To Reproduce**
1. Create a new account
2. make 0 observations
3. Try to sign out
4. View the home screen still displaying the username of the user that just logged out
5. log in with normal inat account and see if the old account name is still displayed

**Expected behavior**
This isn't happening when I log out of my normal account - the home screen immediately refreshes to be the logged out home screen, which it should always do when a user signs out. When I log into my regular account after I log out of my test account, the old name should definitely not show.

**Screenshots**
Tony's [screen recording](https://inaturalist.slack.com/archives/C02G5AATVGA/p1721918059349559) in slack.

Device: iPhone 13 Pro
OS: iOS 17.5.1
App version: 0.73.0 (103)

budowski:
Closes #1886

### DIFF
--- a/src/components/MyObservations/MyObservationsContainer.js
+++ b/src/components/MyObservations/MyObservationsContainer.js
@@ -3,11 +3,12 @@
 import {
   useNetInfo
 } from "@react-native-community/netinfo";
+import { useDrawerStatus } from "@react-navigation/drawer";
 import { useFocusEffect, useIsFocused } from "@react-navigation/native";
 import { RealmContext } from "providers/contexts.ts";
 import type { Node } from "react";
 import React, {
-  useCallback,
+  useCallback, useEffect,
   useRef,
   useState
 } from "react";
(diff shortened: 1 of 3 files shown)


kueda (changes requested):
I think we need a different solution here. Add `useDrawerStatus()` to MyObsContainer causes MyObs to rerender *every* time the user opens the drawer, adding a lag to the appearance of the drawer.

I think you need to figure out *why* MyObs is rerendering when a user w/ observations signs out but *not* when a user without observations signs out and work from there. I would also expect `useCurrentUser( )` to return a different value and trigger a re-render when the user signs out, so that might be another way to come at this.

kueda (changes requested):
Drawer is much more responsive now and the signout bug is not happening, but have you have some failing tests.

kueda:
App restart on sign out added in #2294 makes this unnecessary, I think.

PR #2097, opened 2024-09-18
add condition to show count for more than one obs

### LINKED ISSUE #2081: Hide multiple photos indicator on obs detail for obs with a single image
On Obs Detail, let's hide the multiple photos indicator when the observation only has a single image, since everywhere else in the app, we only show that indicator on observations with more than 1 photo.

image

israr002:
closes #2081

### DIFF
--- a/src/components/ObsDetails/ObsMediaDisplay.js
+++ b/src/components/ObsDetails/ObsMediaDisplay.js
@@ -40,7 +40,7 @@ const ObsMediaDisplay = ( {
           sounds={sounds}
           tablet={tablet}
         />
-        {!loading && (
+        {!loading && items.length > 1 && (
           <View className="absolute bottom-5 left-5">
             <PhotoCount count={items.length} />
           </View>


albullington:
@israr002 thanks for working on this. This approach is fine, but I think it would actually be better to add this conditional into the `PhotoCount` component (and update `ObsImagePreview` to remove the conditional there) so we can make sure we're showing the count consistently across the app. Does that make sense?

albullington (changes requested):
See comment above-- prefer an approach where we update conditional in the PhotoCount component for consistency.

israr002:
@albullington i have doubt here in the photoCount component there is a condition to add noevidence icon if there is no obs
and  we are not rendering photoCount if observation is less than or equal to 1 in obs media so if remove condition here then for 0 evidence it will show the noEvidence icon
WhatsApp Image 2024-09-20 at 10 53 30 PM (1)
But in the current app and web we are not showing that icon for obs without photos so shall i remove that piece of code ??

PR #2146, opened 2024-09-21
fix: pass PermissionGate after granting permission in Settings and returning to app

### LINKED ISSUE #1206: Granting camera permission from Settings requires an app restart in Android
**Describe the bug**
If you deny camera permissions, the app prompts you to open the Settings app to grant the permission. If you do so and return to the app, it still asks you to do the same thing.

**To Reproduce**
1. Fresh install of the app (or deny camera permission in Settings)
2. Enter the StandardCamera
3. You should see the camera permission gate
4. Tap OBSERVE ORGANISMS
5. Tap "Don't allow"
6. Tap OPEN SETTINGS
7. In the Settings app, grant camera permission (tap Permissions, tap Camera, tap "Allow only while using the app"
8. Return to iNat (I usually use the app switcher)
9. The permission gate still displays and is prompting you to open settings... even though you just did that
10. Tap the x button in the upper right to close
11. Enter the StandardCamera again
12. It's still asking you to open settings

**Expected behavior**
Once you've granted camera permission in Settings, iNat should realize that, e.g. when the permission gate has focus, it should check for the permission again.

 - Device: Pixel 8
 - OS: Android 14
 - App version: 0.23.0 (72)

**Additional context**
Works on iOS, so be careful that any changes don't break this there.

israr002:
closes #1206 

https://github.com/user-attachments/assets/3e5bddc1-4477-4906-8a0d-2e52e53ac40c

### DIFF
--- a/src/components/SharedComponents/PermissionGateContainer.tsx
+++ b/src/components/SharedComponents/PermissionGateContainer.tsx
@@ -5,9 +5,10 @@ import React, {
   PropsWithChildren,
   useCallback,
   useEffect,
+  useRef,
   useState
 } from "react";
-import { Platform } from "react-native";
+import { AppState, AppStateStatus, Platform } from "react-native";
 import {
   AndroidPermission,
   checkMultiple,
@@ -149,6 +150,7 @@ const PermissionGateContainer = ( {
 }: Props ) => {
   const [result, setResult] = useState<PermissionStatus | null>( null );
   const [modalShown, setModalShown] = useState( false );
+  const prevAppState = useRef<AppStateStatus>( AppState.currentState );
 
   const navigation = useNavigation();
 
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/PermissionGateContainer.tsx
```diff
@@ -204,6 +206,32 @@ const PermissionGateContainer = ( {
     }
   }, [result, children, setModalShown] );
 
+  useEffect( () => {
+  // We need to handle permission changes manually on Android
+  // Permissions modified from the settings are not captured automatically
+  // So we check permissions when the app is transitioned from background to foreground
+  // This check is performed only when the platform is Android and the permission result is BLOCKED
+    if ( Platform.OS === "android"
+       && ( result === RESULTS.BLOCKED ) ) {
```
kueda, line 215: Unnecessary parens here.

src/components/SharedComponents/PermissionGateContainer.tsx
```diff
@@ -204,6 +206,32 @@ const PermissionGateContainer = ( {
...
+
+      const subscription = AppState.addEventListener( "change", onAppStateChange );
+
+      return () => {
+        subscription?.remove();
+      };
+    }
+
+    return () => undefined;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
kueda, line 232: Why are you ignoring the exhaustive deps warning?

kueda (changes requested):
This works well, but I'm not sure why you're disabling the exhaustive deps warning. In general if you disable an linting rule, please explain why in a comment.

src/components/SharedComponents/PermissionGateContainer.tsx
israr002: removed

src/components/SharedComponents/PermissionGateContainer.tsx
israr002: added checkPermission dependency

kueda (approved):

PR #2157, opened 2024-09-23
feat: phone storage full alert

### LINKED ISSUE #1399: Phone Storage Full alert
When a user's device is full and they take a photo, iNat may not be able to save their observation, photos, or may even crash. When iNat detects a users' device's storage is full, I would like a native modal to appear that says "Device storage full. iNaturalist may not be able to save your photos or may crash." just so a user understands what is happening. I imagine this is most likely to happen in the multicapture and AI cameras when a user takes a photo, but there may be other places where it could be shown as well– @kueda may have suggestions.

Todo:

- [x] Check to see if device storage is full when
  - [x] a user opens the app or
  - [x] takes a photo from the multicapture camera or ai camera, c
- [x] If it is, show the following native modal:
    Title: Device Storage Full
    Body: iNaturalist may not be able to save your photos or may crash.
    1 Button: OK

This was inspired by some of Carrie's experiences using the app with full storage and not understanding what was happening.

angielt:
Closes #1399

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ import {
 import App from "components/App";
 import ErrorBoundary from "components/ErrorBoundary";
 import initI18next from "i18n/initI18next";
+import { t } from "i18next";
 import inatjs from "inaturalistjs";
 import INatPaperProvider from "providers/INatPaperProvider";
 import RealmProvider from "providers/RealmProvider";
@@ -55,6 +56,14 @@ if (
 
 // I'm not convinced this ever catches anything... ~~~kueda 20240110
 const jsErrorHandler = ( e, isFatal ) => {
+  if ( e.match( /No space left on device/ ) ) {
+    Alert.alert(
+      "Device-storage-full",
+      "iNaturalist may not be able to save your photos or may crash.",
+      [{ text: t( "OK" ) }]
+    );
+  }
+
   // not 100% sure why jsErrorHandler logs e.name and e.message as undefined sometimes,
   // but I believe it relates to this issue, which reports an unnecessary console.error
   // under the hood: https://github.com/a7ul/react-native-exception-handler/issues/143
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -7,7 +7,7 @@ import useDeviceStorageFull from "components/Camera/hooks/useDeviceStorageFull";
 import RootDrawerNavigator from "navigation/rootDrawerNavigator";
 import { RealmContext } from "providers/contexts.ts";
 import type { Node } from "react";
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { LogBox } from "react-native";
 import Realm from "realm";
 import { addARCameraFiles } from "sharedHelpers/cvModel.ts";
@@ -59,6 +59,7 @@ const App = ( { children }: Props ): Node => {
   const realm = useRealm( );
   const currentUser = useCurrentUser( );
   const { deviceStorageFull, showStorageFullAlert } = useDeviceStorageFull();
+  const [deviceStorageFullShown, setDeviceStorageFullShown] = useState( false );
 
   // this ensures that React Query has the most accurate depiction of whether the
   // app is online or offline. since queries only run when the app is online, this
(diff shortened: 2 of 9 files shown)


index.js
```diff
@@ -55,6 +56,14 @@ if (
 
 // I'm not convinced this ever catches anything... ~~~kueda 20240110
 const jsErrorHandler = ( e, isFatal ) => {
+  if ( e.match( /No space left on device/ ) ) {
+    Alert.alert(
+      "Device-storage-full",
+      "iNaturalist may not be able to save your photos or may crash.",
```
kueda, line 62: Both of these strings need to be internationalized.

src/components/App.js
```diff
@@ -27,6 +29,9 @@ import useTaxonCommonNames from "./hooks/useTaxonCommonNames";
 
 const { useRealm } = RealmContext;
 
+// 200MB - number in bytes
+const MIN_DEVICE_STORAGE = 200000000;
```
kueda, line 33: IMO, long numbers are a lot easier to read with separators, e.g. `200_000_000`. It could also use a name that better describes its purpose. Is this referring to total size of the disk? The smallest amount of *available* disk we can use?

src/components/Camera/StandardCamera/StandardCamera.js
```diff
@@ -100,6 +105,19 @@ const StandardCamera = ( {
...
+    [{ text: t( "OK" ) }]
+  );
+
+  DeviceInfo.getFreeDiskStorage().then( freeDiskStorage => {
+    // considered full when 200MB left
+    if ( freeDiskStorage <= MIN_DEVICE_STORAGE ) {
+      setDeviceStorageFull( true );
+    }
+  } );
+
```
kueda, line 120: Don't repeat yourself. This is the 4th time you've had almost identical code showing this modal, and the third time you're doing the same check with `DeviceInfo.getFreeDiskStorage()`. If you do the same thing twice, it's time to pull that logic out into a module (seems like a hook would do the trick here). If there are small tweaks to how it has to behave, like making the alert have a cancel button or not, those can be options passed into whatever module you create.

src/components/App.js
```diff
@@ -78,6 +83,25 @@ const App = ( { children }: Props ): Node => {
...
   useTaxonCommonNames( );
 
+  const showStorageFullAlert = () => Alert.alert(
+    "Device Storage Full",
+    "iNaturalist may not be able to save your photos or may crash",
+    [{ text: t( "OK" ) }],
+    {
+      cancelable: true
+    }
+  );
```
kueda, line 93: This gets redefined on every render. Either stabilize it with useCallback or define it inside the useEffect callback so it only gets defined with the effect runs.

src/components/Camera/hooks/usePrepareStoreAndNavigate.js
```diff
@@ -65,6 +65,11 @@ export async function savePhotosToCameraGallery(
         // the EXIF metadata of these photos, once we retrieve a location.
         onEachSuccess( savedPhotoUri );
       } catch ( cameraRollSaveError ) {
+        if ( cameraRollSaveError.message.match( /No space left on device/ )
+          || cameraRollSaveError.message.match( /The operation couldn’t be completed/ ) ) {
+          console.log( "cameraRollSaveError: ", cameraRollSaveError );
+          return;
+        }
```
kueda, line 72: Remove the `console.log` statement since that's only for dev purposes. Also, you're silently suppressing a real error here. Don't you want to alert the user instead? If you really don't, please leave a comment explaining why this error is ok to ignore.

kueda (changes requested):
I'm still working on filling up my phone to test this under real conditions, but I think there are several fixes that need to be made anyway, most importantly refactoring the repeated code.

src/components/Camera/hooks/useDeviceStorageFull.js
```diff
@@ -0,0 +1,36 @@
+import { t } from "i18next";
```
kueda, line 1: When you import `t` like this instead of using the hook, the translations won't change until the app restarts, so if the user changes their language preference, the text in this hook won't change to the new language until restart. Try using the hook instead.

src/components/Camera/hooks/usePrepareStoreAndNavigate.js
kueda: Better, but you should never show the user the actually error object. It will look like a bunch of incomprehensible code nonsense. The `message` attribute is better, but probably still too jargon-y and definitely not localized. Instead, tell the user what actually went wrong with globalized text that a layperson would understand. You're catching two separate kinds of errors here, which probably requires two separate messages to the user.

kueda (changes requested):
Much better, just the two remaining todos.

PR #2159, opened 2024-09-24
fix: replaying and scrubbing audio repeatedly eventually stops playback

### LINKED ISSUE #1540: Replaying and scrubbing audio repeatedly eventually stops playback
**Describe the bug**
Replaying and scrubbing audio repeatedly eventually stops the playback indicator from moving as expected.

**To Reproduce**
1. Make a sound observation
2. Replay and scrub audio via the playback UI
3. Eventually the playback indicator will stop moving / the UI will not match the playback

**Expected behavior**
A user should be able to consistently playback sounds and see the correct scrubbing UI without errors.

**Additional context**
There seem to be two related promise rejections:
`Possible unhandled promise rejection: Error: Player has already stopped` - this happens when starting and stopping playback over and over on the Sound Recorder screen
`Possible unhandled promise rejection: Error: Player is null` - this happens when starting and stopping scrubbing over and over on the scrubbing screen

angielt:
Closes #1540

### DIFF
--- a/src/components/ObsDetails/SoundContainer.js
+++ b/src/components/ObsDetails/SoundContainer.js
@@ -26,6 +26,8 @@ const SoundSlider = ( { playBackState, onSlidingComplete } ) => {
       style={sliderStyle}
       minimumValue={0}
       maximumValue={playBackState.duration}
+      lowerLimit={0}
+      upperLimit={playBackState.duration}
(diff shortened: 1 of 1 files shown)


kueda (changes requested):
This is working well for me, just need the one tiny change for clarity.

PR #2198, opened 2024-09-30
Fix #1613 - when marking notification as read, use that in useInfinit…

### LINKED ISSUE #1613: Mark notifications as read on press
On Notifications, you press on an unread observation notification to get to ObsDetails, and on ObsDetails we set observationMarkedAsViewedAt in Zustand to signal to other parts of the app that an observation was viewed and we should refetch the updates from the server. However it doesn't seem like we're using that at all in useInfiniteNotificationsScroll, and furthermore, it might just be a better experience to update the viewed state in the local copy of the observations/updates response so the NotificationsList UI updates immediately. Otherwise users have to wait for Notifications to refresh and may tap the same notification twice and wonder why they're just doing the same thing over and over (the app should manage state, not the user).

- [x] ~Make the useInfiniteNotificationsScroll query key use `observationMarkedAsViewedAt` so it definitely re-fetches when a user returns to Notifications from ObsDetails~
- [ ] ~Make sure that doesn't mean that query runs *every* time a user views ObsDetails, only when they view Notifications~
- [ ] Try updating the local state of the updates response immediately after the user presses a notification, so if for some reason they come back and the list does not refresh, the notification still looks read.

budowski:
…eNotificationsScroll

### DIFF
--- a/src/sharedHooks/useInfiniteNotificationsScroll.js
+++ b/src/sharedHooks/useInfiniteNotificationsScroll.js
@@ -6,6 +6,7 @@ import { RealmContext } from "providers/contexts.ts";
 import { useCallback } from "react";
 import Observation from "realmModels/Observation";
 import { useAuthenticatedInfiniteQuery, useCurrentUser } from "sharedHooks";
+import useStore from "stores/useStore";
 
 const { useRealm } = RealmContext;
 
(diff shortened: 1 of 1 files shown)


albullington:
@angielt assigning you to review this one since you worked on notifications

angielt (approved):

PR #2216, opened 2024-10-02
fix: add debounce to navigate function

### LINKED ISSUE #2204: Multiple taps on an observation can open multiple instances of obs detail
**Describe the bug**
Right now, if you tap an observation from my obs or explore quickly multiple times, it can open multiple instances of that same obs detail screen on top of each other, and then you have to click back multiple times through the same screen to get back to the previous screen.

**To Reproduce**
1. Go to Home screen/My obs in list view
2. Quickly tap on a single observation several times super fast
3. Notice how it opens multiple instances of the obs detail page for that obs on top of each other
4. Click back and see the same obs detail that you need to back out of multiple times

Note: this happens on explore as well.
May take a couple tries (you have to click it really fast)

**Expected behavior**
We should disable multiple navigation taps on an observation on explore or my observations once one is in progress. No UI changes needed.

 - Device: iPhone 13 Pro
 - OS: 18.0
 - App version: 0.49.0 (118)

israr002:
closes #2204


src/components/ObservationsFlashList/MyObservationsPressable.js
```diff
@@ -23,15 +24,15 @@ const MyObservationsPressable = ( { observation, testID, children }: Props ): No
...
+  const navigateToObservation = debounce( ( ) => {
     const { uuid } = observation;
     if ( unsynced ) {
       prepareObsEdit( observation );
       navigateToObsEdit( navigation, setMyObsOffsetToRestore );
     } else {
       navigation.push( "ObsDetails", { uuid } );
     }
-  };
+  }, 200 );
```
kueda, line 35: This reduces the problem but does not remove it. I can still tap quickly and get two instances of ObsDetails pushed onto the stack.

tests/integration/PhotoDeletion.test.js
```diff
@@ -164,10 +165,11 @@ describe( "Photo Deletion", ( ) => {
     await confirmPhotosAndAddTopId();
     await saveAndEditObs();
     // Enter camera to add new photo
-    const addEvidenceButton = await screen.findByLabelText( "Add evidence" );
-    await actor.press( addEvidenceButton );
-    const addEvidenceSheet = await screen.findByTestId( "AddEvidenceSheet" );
-    const cameraButton = await within( addEvidenceSheet ).findByLabelText( "Camera" );
+    const addEvidenceButton = await waitFor( () => screen.findByLabelText( "Add evidence" ) );
```
kueda, line 168: I think all of the findBy* helpers basically do `waitFor` internally, so this should not be necessary, especially if you're not manually specifying a timeout.

kueda (changes requested):
IMO this just makes things marginally more confusing to read without actually fixing the problem. Instead, check out https://github.com/react-navigation/react-navigation/issues/271#issuecomment-362726430. That's the solution recommended by the maintainers of react-naviation. Our situation is a little more complicated because we really do want to use `push()` and that doesn't seem to support extra stuff like the `key` param, but I think there's a way to do it with `navigate({ screen: "foo", key: "bar" })` if we construct our keys so that that they are always unique within the context of a single list of observations.

What we want is for something like this to open all new screens, no navigating back:

1. Go to MyObs
2. Tap on an observation
3. Tap on the observer's user icon to get to UserProfile
4. Tap on the observations stat to get to Explore
5. Tap on the same observation from step 2
6. A new instance of ObsDetails should be pushed onto the stack, you should *not* navigate back to the same ObsDetails instance from step 2

israr002:
@keuda I have done the required changes Please review.

kueda (changes requested):
I can still reproduce the bug described in #2204 with these changes. I end up with keys like this:

`Obs-ObsList-wzflxz7sg-6FR9VWZxy6z-61152031-becd-4702-b281-ace2f4e936b5`
`Obs-Obs-ObsList-wzflxz7sg-6FR9VWZxy6z-61152031-becd-4702-b281-ace2f4e936b5-61152031-becd-4702-b281-ace2f4e936b5`

because `getCurrentRoute()` isn't telling you the route at the time the tap occurred, it's telling you the current route when `navigateToObservation()` gets called, which might actually be when you've already navigated to ObsDetails, resulting in that nested key.

To address this, I think you need to pass a prop down from every unique observation list, i.e. from MyObs, from every unique instance of Explore, etc. LMK if that doesn't make sense and I can take this one b/c I think I know how I would address this.

PR #2222, opened 2024-10-03
fix: nearby button in Explore not working after granting permission

### LINKED ISSUE #2207: Nearby button in Explore Filters doesn't work first time permissions are given
**Describe the bug**
Nearby button in Explore Filters > Explore Location Search doesn't work the first time permissions are given

**To Reproduce**
1. Go to Explore, then Explore Filters
2. Make sure your location settings are set to Ask Next Time Or When I Share
3. Tap on Edit Location, then Nearby
4. You'll be prompted for location. Say yes, give access once or always
5. Nothing happens, and you land back on the Explore Filters screen with the "Apply Filters" button disabled. At this point, you need to either exit out of the screen or change another filter to be able to tap "Apple Filters"
6. Nearby does seem to applied if you change another filter and go back to Explore

**Expected behavior**
The first time Nearby is applied in Explore Filters after permissions are given, we should enable the "Apply Filters" button to allow a user to apply this change and return to Explore.

**Screenshots**
Screenshot 2024-09-30 at 2 39 04 PM

 - Device: iPhone
 - App version: 0.45.0

israr002:
closes #2207 
When we select nearby, lat and long keys are added to the state, we also need to check 
the length of snapshot and state, as we are currently only checking for value changes.


kueda (changes requested):
I'm not convinced all this complexity is necessary. Did you try something simpler like this?

```typescript
// Check if the current state is different from the snapshot
// const checkSnapshot = () => {
//   if ( !snapshot ) {
//     return false;
//   }
//   return Object.keys( snapshot ).some( key => !isEqual( snapshot[key], state[key] ) )
//   || Object.keys( snapshot ).length !== Object.keys( state ).length;
// };
const differsFromSnapshot: boolean = React.useMemo(
  () => !isEqual( snapshot, state ),
  [state, snapshot]
);
```

I haven't tested that thoroughly but I think it makes better use of `isEqual()`, and it fixes the bug. If it doesn't work, we should probably add a comment explaining why `some()` is necessary.

israr002:
@kueda  I thought of something similar as well, Since the some() logic was already there, I left it untouched, assuming it might be used for specific scenarios and just added one more condition to the existing logic. However, i have verified the isEqual() check will work without adding complexity.

kueda (approved):

PR #2235, opened 2024-10-06
fix: photocount text overflows icon

### LINKED ISSUE #2131: PhotoCount text overflows icon
Happening when system font is set to the largest size, probably everywhere PhotoCount gets used but particularly on obs grid items and ObsDetails. If possible we should try and scale up the icon along with the text size, but failing that we should probably just not allow that text to scale up.

IMG_675790649A5A-1 IMG_F6F3347FD7BD-1

israr002:
closes #2131

### DIFF
--- a/src/components/SharedComponents/PhotoCount.js
+++ b/src/components/SharedComponents/PhotoCount.js
@@ -5,14 +5,12 @@ import { Body3, INatIcon } from "components/SharedComponents";
 import { View } from "components/styledComponents";
 import type { Node } from "react";
 import React from "react";
-import { I18nManager } from "react-native";
+import { I18nManager, PixelRatio } from "react-native";
 import { useTheme } from "react-native-paper";
 import Svg, { Path } from "react-native-svg";
 import { useTranslation } from "sharedHooks";
 import { dropShadow } from "styles/global";
 import colors from "styles/tailwindColors";
-
-const HEIGHT = 24;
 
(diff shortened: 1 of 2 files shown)


kueda (changes requested):
You've got a test snapshot you need to update, but more importantly this still isn't looking very good at the largest sizes:

Screenshot 2024-10-16 at 6 06 45 PM

The scaling of the photo count icon probably needs to be limited, and we need to try and make sure the text is always inside the white square part of the icon.

israr002:
@kueda I have tested the photo count on the observation screen, and it's working fine on my device. Could you please share the steps you followed to generate the scenario you encountered?

https://github.com/user-attachments/assets/3497b558-c4ae-4d8f-b41a-80dedf0fb85f

kueda:
@israr002 you need to run the app in iOS and set the font size to the maximum allowed size. You might be able to achieve something similar in Android at the max size, but since we're not yet supporting Android, it's most important that this work in iOS at the moment.

angielt:
Might have to check with @abhasinat again if this issue is still needed. While I was working on #2257, he mentioned to set the maxFontSizeMultiplier for the PhotoCount number to 1.

kueda:
@israr002 are you still planning to work on this?

israr002:
Yes I am working on this

kueda (approved):

PR #2286, opened 2024-10-21
fix: active tab doesn't change when swiping back from notifications

### LINKED ISSUE #2034: Active bottom tab doesn't change when swiping back from Notifications
**Describe the bug**
The active bottom tab doesn't change when go back via swipe.

**To Reproduce**
1. Open the app to land on MyObs. Note that that tab in the bottom tabs has a green outline to indicate it's the current tab
2. Go to Notifications. The notifications icon is now green to indicate you're on that tab
3. Swipe from the left side of the screen to go back
4. You're back on MyObs, but the Notifications icon is still green

**Expected behavior**
When you swipe back to MyObs, the notification icon should go back to being dark gray and the MyObs icon should have the green outline.

**Screenshots**
IMG_1956

 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.6.1
 - App version: 0.44.0

israr002:
closes #2034


src/navigation/BottomTabNavigator/CustomTabBarContainer.js
```diff
@@ -21,6 +22,23 @@ const CustomTabBarContainer = ( { navigation }: Props ): Node => {
   const currentUser = useCurrentUser( );
   const [activeTab, setActiveTab] = useState( OBS_LIST_SCREEN_ID );
   const isDrawerOpen = useDrawerStatus() === "open";
+  const route = getCurrentRoute();
+  const currentRoute = route?.name || "";
+  useEffect( () => {
```
jtklein, line 27: Sorry for the late review.
This is not a good use of an useEffect call. We have not been too strict with that ourselves but will try to be more strict about that in future. useEffects are primarily for handling changes that need to happen because something outside React has changed, like on the server for example.

Please, check out his documentation here and refactor this code accordingly:
https://react.dev/learn/you-might-not-need-an-effect
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

jtklein (changes requested):
Please refactor the useEffect call.

src/navigation/BottomTabNavigator/CustomTabBarContainer.js
israr002: @jtklein i have made the changes

jtklein:
Thanks for this! Have tested and:
1) It does fix the bug as described.
2) All tests and e2e tests are passing locally for me.

I have updated from main, and refactored a bit getting the screen names constants directly from the navigation setup as central point. Am merging this now.

PR #2305, opened 2024-10-26
Update some packages to fix compilation

### LINKED ISSUE #2302: Can't compile app in development: react-native-reanimated
**Describe the bug**
The command `npm start`

> Execution failed for task ':react-native-reanimated:configureCMakeDebug[arm64-v8a]'.
> [CXX1400] Gradle project cmake.path is /home/<USER>/iNaturalistReactNative/node_modules/react-native-reanimated/android/CMakeLists.txt but that file doesn't exist

**To Reproduce**
Not sure. Might be a problem with my versions of JAVA, Node etc..

**Expected behavior**
App should compile.

 - Device: PC
 - OS: Ubuntu 24.10
 - App version: current main branch
 - Node: v20.16.0

**Additional context**
After running `npm i react-native-reanimated@^3.16` the compilation continued and the emulator started.

tobias47n9e:
Closes #2302


tobias47n9e:
Not sure why I needed to update those packages, but they fix the compilation for me of the Android app on Linux.

jtklein (commented):
Many thanks for your contributions to our app!
Regarding your changes here, personally, I don't think this is the way forward for your troubles in compiling the app. Some reasoning:
1) None of us has currently problems compiling the app with those packages.
2) Reanimated and gesture-handler can be quite tricky to update in combination with react-native-vision-camera, so I would require extensive evidence of e2e tests passing as well as manual testing before merging.
3) Bumping a major version as in bottom-sheet usually signifies breaking changes, I can not see if those have been addressed.

I think if this is the only way you get it to compile, the way forward would be to keep a fork with those changes and make PRs without those changes if possible.

PR #2308, opened 2024-10-26
Fix resetting character count in feedback form

### LINKED ISSUE #2307: Clearing feedback form does not reset character count
**Describe the bug**
The character count in the feedback form does not reset when pressing "Clear".

**To Reproduce**
1. Go to Feedback
2. Type a text
3. Press "Clear"
4. The character count will stay the same and not switch to zero

**Expected behavior**
Count should reset when pressing "Clear".

**Screenshots**

 - Device: Android Emulator
 - OS: Android
 - App version: current main branch

**Additional context**
-

tobias47n9e:
Closes #2307

### DIFF
--- a/src/components/SharedComponents/Sheets/TextInputSheet.js
+++ b/src/components/SharedComponents/Sheets/TextInputSheet.js
@@ -9,7 +9,7 @@ import {
 import { Pressable, View } from "components/styledComponents";
 import type { Node } from "react";
 import React, { useMemo, useRef, useState } from "react";
-import { Keyboard } from "react-native";
+import { Keyboard, Platform } from "react-native";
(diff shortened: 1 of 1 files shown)


kueda (changes requested):
Thanks, this works, we just need two changes

1. Since this isn't necessary in iOS and the code looks redundant, please either add a comment explaining this is necessary in Android, or only do this extra step for Android
2. Please make sure you have eslint running and are abiding by our eslint rules. This currently violates `space-in-parens`. If you [set up the pre-commit hooks](https://github.com/inaturalist/iNaturalistReactNative?tab=readme-ov-file#set-up-pre-commit-hooks), this should have prevented you from committing.

PR #2309, opened 2024-10-26
fix : flagged comments still appears flagged after being resolved 

### LINKED ISSUE #2262: Flagged comment still appears flagged after being resolved
**Describe the bug**
Flagged comments still appear flagged after being resolved.

**To Reproduce**
1. View https://www.inaturalist.org/observations/245155834 in the app
2. The comment by alr1243 shows an orange flag and the label "Flagged"
3. View https://www.inaturalist.org/observations/245155834 on the web to confirm there are no open flags

**Expected behavior**
If a comment (or ID) has flags but they are all resolved, it should not appear flagged.

**Screenshots**
IMG_2231

 - Device: iPhone SE (3rd gen)
 - OS: 17.6.1
 - App version: 0.50.0

**Additional context**
The flagged state should only appear if there are any unresolved flags on the comment or ID.

israr002:
closes #2262

### DIFF
--- a/src/components/ObsDetails/ActivityTab/ActivityHeaderContainer.js
+++ b/src/components/ObsDetails/ActivityTab/ActivityHeaderContainer.js
@@ -31,11 +31,10 @@ const ActivityHeaderContainer = ( {
   belongsToCurrentUser
 }:Props ): Node => {
   const [currentUser, setCurrentUser] = useState( false );
-  const [flagged, setFlagged] = useState( false );
   const [loading, setLoading] = useState( false );
   const { user } = item;
 
-  const numFlags = item.flags?.length || 0;
+  const flagged = item.flags?.some( flag => !flag.resolved );
 
   useEffect( ( ) => {
     const isActiveUserTheCurrentUser = async ( ) => {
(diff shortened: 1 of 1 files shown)


src/components/ObsDetails/ActivityTab/ActivityHeaderContainer.js
```diff
@@ -35,19 +35,21 @@ const ActivityHeaderContainer = ( {
...
   useEffect( ( ) => {
     const isActiveUserTheCurrentUser = async ( ) => {
       const current = await isCurrentUser( user?.login );
       setCurrentUser( current );
     };
     isActiveUserTheCurrentUser( );
+  }, [user] );
 
-    if ( numFlags > 0 ) {
+  useEffect( ( ) => {
```
albullington, line 48: We really don't want to use `useEffects` in code unless they're absolutely necessary, since they cause extra rerenders. In this case, you've already made `isFlagged` into a boolean, so there's no need for this line `const [flagged, setFlagged] = useState( false )` and there's no need for this effect. I'd remove both and rename `isFlagged` to `flagged`.

albullington (changes requested):

albullington (approved):

PR #2324, opened 2024-10-29
refactor: restore ObsGridItem and ObsListItem UI demos

kueda:
Pulls store interaction up a few levels in the hierarchy to make demos and tests of upload-related components simpler.

### DIFF
--- a/src/components/Developer/UiLibrary/ObsGridItemDemo.js
+++ b/src/components/Developer/UiLibrary/ObsGridItemDemo.js
@@ -65,31 +65,30 @@ const ObsGridItemDemo = ( ) => (
       <Heading2 className="my-2">Synced</Heading2>
       <ObsGridItem
         observation={{ uuid: "the-uuid", _synced_at: new Date( ) }}
-        uploadState={false}
       />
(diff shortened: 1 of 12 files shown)


albullington (approved):
Lgtm. The only thing I'm wondering is if the naming for `uploadProgress` would benefit from more specificity, like `individualUploadProgress` or even `circularUploadProgress`. I think it's easy to get confused on a quick glance at this code on whether `uploadProgress` refers to the individual upload or the total upload progress.

kueda:
> The only thing I'm wondering is if the naming for uploadProgress would benefit from more specificity, like individualUploadProgress or even circularUploadProgress. I think it's easy to get confused on a quick glance at this code on whether uploadProgress refers to the individual upload or the total upload progress.

Definitely a matter of opinion. To me, each of the components here that receives an `uploadProgress` or `progress` prop only represents the state of a single observation, so the fact that its value reflects the upload progress of a single observation is implied. Personally I would not assume that an individual obs component would ever need to know about the batch upload state, so if that were being passed in as a prop, that would justify a more specific name. IMO, `circularUploadProgress` would only be justified if the component had multiple progress bars and one of them was circular.

PR #2329, opened 2024-10-31
Restore language chooser and queue offline locale changes

### LINKED ISSUE #2242: Restore language chooser in the app
Currently it's visible in debug mode but is itself kind of buggy.

- [x] Language choose should be visible without debug mode
- [x] Should work offline, i.e. we should know the list of available locales based on translation files in the app, and the user should be able to change the app language while offline. A job to update the user's preference on the server should either be queued
- [x] Language should change immediately after the user makes their choice. If that requires restarting the app, notify the user in a confirmation dialog

albullington:
Closes #2242

- Restores the language chooser which was in debug mode
- On a fresh install of iNat Next, checks for system language and updates the app locale accordingly
- When a user first signs in, updates the app locale with the user's server locale preferences
- If a user changes locale in LanguageSettings, queue this request in Realm to be handled whenever a user is online
- When queued request updates the server locale successfully, remove locale change from queue


albullington:
@kueda requested a review to 1) make sure locale changing is working the way you're expecting it to work and 2) double check that this custom-built queue in realm is generalizable enough for feedback and/or other use cases

src/components/hooks/useChangeServerLocale.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const useChangeServerLocale = ( ) => {
+  const realm = useRealm( );
+  const [isMutating, setIsMutating] = useState( false );
+
+  const queue = realm.objects( "Queue" ).sorted( "id" );
+  const firstQueueItem = queue?.[0];
+  const mutationParams = firstQueueItem
+    ? JSON.parse( firstQueueItem?.payload )
+    : null;
```
kueda, line 19: Figuring out which item is first in the queue and rehydrating its JSON string into an object belongs in the `dequeue()` method, because *every* consumer that wants to get something out of the queue will have to do that, and the FIFO processing order is the very definition of the queue so that needs to be internal. If a consumer can get the *last* item in the queue, then it's not a queue.

Since this is a work queue and you probably only want one worker working at a time, you might want to take a different approach where the worker claims the item at the head of the queue so no one else can work on it *but* doesn't remove it, when other potential workers peek at the head of the queue, they know not to work on it. When the worker is done working on the queue item, then it deletes it from the queue.

src/components/Settings/LanguageSetting.tsx
```diff
@@ -19,32 +21,41 @@ type Props = {
...
+  const [webLocales, setWebLocales] = useState<LocalesResponse>( [] );
+  const webLocalesOptions = Object.fromEntries(
+    webLocales?.map( locale => [locale.locale, {
       label: locale.language_in_locale,
       value: locale.locale
     }] )
   );
   const [localeSheetOpen, setLocaleSheetOpen] = useState( false );
 
+  const offlineLocalesOptions = Object.fromEntries(
```
kueda, line 33: Now that I'm trying this, I'm not even sure this is even necessary. The only situation in which a user might need these locales is if they sign in, then go offline, then go to Settings... which is so rare I don't think we need to handle it. In all other cases they're going to see fresh or cached web locales. So I think we can remove this logic and just hide the language chooser if the user does takes that very unusual path to Settings.

src/components/Settings/Settings.js
```diff
@@ -67,12 +67,15 @@ const Settings = ( ) => {
 
   const queryClient = useQueryClient();
 
+  const invalidateQuery = useCallback( ( ) => {
+    queryClient.invalidateQueries( { queryKey: ["fetchUserMe"] } );
+  }, [queryClient] );
+
```
kueda, line 73: Does this really force a re-fetch of users/me? If so can you please explain why in a comment in the code?

src/components/hooks/useFreshInstall.js
```diff
@@ -15,6 +19,18 @@ const installStatus = new MMKV( {
...
+  // only use system locale if this is a fresh install of the app;
+  // otherwise, locale should be set by LanguageSettings with a
+  // fallback to the web locale
+  const changeToSystemLocale = useCallback( ( ) => {
+    const systemLocale = getInatLocaleFromSystemLocale();
+    i18n.changeLanguage( systemLocale );
+  }, [
+    i18n
+  ] );
+
```
kueda, line 33: The other time we want to use the system locale is when the user signs out, and this is not handling that.

kueda (changes requested):
Re: queuing work to be done offline, moving in the right direction. What you're building here is not just a queue, but job processing queue, so you should probably name it to reflect that. Similarly, the work is getting done in a hook that claims to change the server locale, but really it works on anything in the queue, so it needs a new name too. Currently you just have one kind of job that can be worked on, but if the intent is to make this generic, the job should be generic and sub-typable, e.g. when the worker starts working on a job in the queue, it should know what kind of job it is so it knows how to work on it (or what other method to call to work on it).

The server seemed to get updated when I got back online after changing my locale while offline... but I'm not sure what cause App to re-render. Did you consider making it poll at an interval?

Some bugs

- [ ] When you sign out, the app locale does not revert to the system locale
- [ ] When you change the system locale, the app adopts the system locale even if you're signed in, e.g. sign in as a user with locale set to English, then change the system locale to French, and when you return to the app, it's in French

src/components/Settings/LanguageSetting.tsx
albullington: I was wondering about that. Will remove

src/components/Settings/Settings.js
albullington: My understanding is that anytime we invalidate queries, as long as we're making a fetch via useQuery, the fetch will automatically be refetched in the background. That's per this line in the [query invalidation docs](useUserMe) in React Query:

```
When a query is invalidated with invalidateQueries, two things happen:

It is marked as stale. This stale state overrides any staleTime configurations being used in useQuery or related hooks
If the query is currently being rendered via useQuery or related hooks, it will also be refetched in the background
```

This has been working well for me on other screens, but I'll verify that it's working here and add a comment.

kueda: Is invalidating an existing query really what we want here? What if there is no existing query? The other issue here is that complicated logic surrounding how the `useUserMe` hook works should really be encapsulated in that hook, so I like the fact that it exported a `refetchUserMe` method. If invalidating an existing query is preferable to calling the `refetch` method returned by `useQuery`, I think we should change the implementation of that method rather than remove it and rely on other code to know how to invalidate queries.

albullington: The reason I updated this code was because we're invalidating the query and then refetching, which I'm assuming means that React Query is running one refetch and then we're immediately repeating our own manual refetch. But I'm not attached to this, nor is it super relevant to this issue, so restoring for now.

albullington:
Restore language chooser with system locale (signed out users), web locale (signed in users), or app locale (as selected by user)

PR #2336, opened 2024-11-01
Add large font size scaling in webviews

### LINKED ISSUE #2137: Control font scaling in web views
For our own text components we can control font scaling with the `maxFontSizeMultiplier`, but web views don't seem to respect larger system font sizes. Look into our options for controlling this in our app (is there a prop? maybe we can set a base CSS font size for the web view?), and if there aren't any, ask the web team if there are options for our own website, at least.

angielt:
Closes #2137

### DIFF
--- a/src/components/FullPageWebView/FullPageWebView.tsx
+++ b/src/components/FullPageWebView/FullPageWebView.tsx
@@ -198,16 +198,17 @@ const FullPageWebView = ( ) => {
     }, [navigation, params.loggedIn, params.title] )
   );
 
-  const fontScalingJS
-  = `
+  const fontScalingJS = `
     if( ${isLargeFontScale} ){
-      document.body.style.fontSize = '40px';
+      document.getElementById("wrapper").style.fontSize = '2rem';
       let paragraphs = document.querySelectorAll('p');
       paragraphs.forEach(p => {
-        p.style.fontSize = '50px';
+        p.style.fontSize = '3rem';
       })
     }
-    true; // note: this is required, or you'll sometimes get silent failures
+    true; 
+    // note: string should evaluate to vaild type, can throw exception otherwise
+    // https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#injectedjavascript
   `;
 
   return (
--- a/src/components/FullPageWebView/FullPageWebView.tsx
+++ b/src/components/FullPageWebView/FullPageWebView.tsx
@@ -14,6 +14,7 @@ import { EventRegister } from "react-native-event-listeners";
 import WebView from "react-native-webview";
 import { log } from "sharedHelpers/logger";
 import { composeEmail } from "sharedHelpers/mail.ts";
+import { useFontScale } from "sharedHooks";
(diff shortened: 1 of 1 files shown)


src/components/FullPageWebView/FullPageWebView.tsx
```diff
@@ -196,11 +198,26 @@ const FullPageWebView = ( ) => {
     }, [navigation, params.loggedIn, params.title] )
   );
 
+  const fontScalingJS
+  = `
```
kueda, line 202: Please don't start a line with an equals. IDK why eslint allows this.

src/components/FullPageWebView/FullPageWebView.tsx
```diff
@@ -196,11 +198,26 @@ const FullPageWebView = ( ) => {
...
+    if( ${isLargeFontScale} ){
+      document.body.style.fontSize = '40px';
```
kueda, line 204: Do you really want to hard-code the size like this? Did you try a percentage?

src/components/FullPageWebView/FullPageWebView.tsx
```diff
@@ -196,11 +198,26 @@ const FullPageWebView = ( ) => {
...
+      let paragraphs = document.querySelectorAll('p');
+      paragraphs.forEach(p => {
+        p.style.fontSize = '50px';
+      })
```
kueda, line 208: Why is modifying the P elements necessary if you're setting a base size for the BODY?

src/components/FullPageWebView/FullPageWebView.tsx
```diff
@@ -196,11 +198,26 @@ const FullPageWebView = ( ) => {
...
+    }
+    true; // note: this is required, or you'll sometimes get silent failures
```
kueda, line 210: What kind of errors?

src/components/FullPageWebView/FullPageWebView.tsx
angielt: I'll try percentages

src/components/FullPageWebView/FullPageWebView.tsx
angielt: Not 100% sure why. Modifying the BODY only seems to affect the headings. My guess is that there might be some css that defines the font size specifically for P tags that overrides the BODY.

src/components/FullPageWebView/FullPageWebView.tsx
angielt: Haven't experienced any errors personally but it was listed in the docs to make sure the string evaluates to a valid type

src/components/FullPageWebView/FullPageWebView.tsx
kueda: That sounds like a web page hard-coding a font size... which is kind of out of our control, unless it's our own web page. Where were you seeing this?

kueda (approved):
I don't think this looks quite right in the Settings webview, but I do think it's an improvement, so let's merge.

PR #2419, opened 2024-11-12
Validate date formats in FTL files

### LINKED ISSUE #2394: Add date format validation to FTL validation
We already [validate the format of FTL](https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/i18n/i18ncli.js#L181-L223) translation files we get from Crowdin, but we don't validate the content, mostly because Fluent is pretty permissive, BUT we should validate date formats, as pointed out in #2387 and as made painfully obvious by some recent Android releases.

- [ ] for every string beginning with `date-format-`, [format](https://date-fns.org/v4.1.0/docs/format) it with date-fns, catch any formatting errors, and add them to the list of errors to report
- [ ] ensure that when these problems are present in the relevant source strings, `npm translate` fails with an error code

In theory that will be enough to cause our translation integration pipeline to fail when these formatting problems occur.

tobias47n9e:
Closes #2394

### DIFF
--- a/src/i18n/i18ncli.js
+++ b/src/i18n/i18ncli.js
@@ -18,6 +18,8 @@ const {
   sortBy,
   uniq
 } = require( "lodash" );
+const { format, parseISO } = require( "date-fns" );
+const { enUS } = require( "date-fns/locale" );
 
 // Exceptions to the rule that all locales should be specified as two-letter
 // language code only
@@ -178,6 +180,18 @@ const writeLoadTranslations = async ( ) => {
   out.write( "\n];\n" );
 };
 
+function isTranslationContainingDateFormat( item ) {
+  if ( item.type !== "Message" ) {
+    return false;
+  }
+
+  if ( item.id.name.startsWith( "date-format" ) ) {
+    return true;
+  }
+
+  return item.id.name.startsWith( "datetime-format" );
+}
+
 async function validateFtlFile( ftlPath, options = {} ) {
   const ftlTxt = await fsp.readFile( ftlPath );
   const ftl = parseFtl( ftlTxt.toString( ) );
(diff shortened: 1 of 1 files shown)


tobias47n9e:
The command currently reports 1 error in the checked-in files for me:

```shell
node src/i18n/i18ncli.js validate
```

```
✅ /.../iNaturalistReactNative/src/i18n/strings.ftl validated
❌ 1 errors found in /.../iNaturalistReactNative/src/i18n/l10n/pt-BR.ftl:
[Error] Invalid date/datetime format: MM AA
```

src/i18n/i18ncli.js
```diff
@@ -205,6 +207,17 @@ async function validateFtlFile( ftlPath, options = {} ) {
...
       }
     }
+    if ( item.type === "Message" && item.id.name.startsWith( "date-format" ) ) {
+      item.value.elements.forEach( textElement => {
+        try {
+          format( parseISO( "1970-01-01T00:00:00Z" ), textElement.value, {
+            locale: enUS
+          } );
+        } catch ( error ) {
+          errors.push( `Invalid date/datetime format: ${textElement.value}` );
```
kueda, line 217: Would you make this something more like this?

```javascript
errors.push( `${item.id.name} = ${textElement.value}: ${error.message}` );
```

When this *does* fail, it helps to know exactly what key failed and why.

kueda (changes requested):
Looks good, I think it should be good to merge after the one change. I would love to have a way to prevent translations like `date-format-short = M/d/aa` but I'm not sure we can since `aa` is valid and will be interpolated... just not as the year.

src/i18n/i18ncli.js
tobias47n9e: I changed the line. The error does now look more useful:

```
❌ 1 errors found in /.../iNaturalistReactNative/src/i18n/l10n/pt-BR.ftl:
[Error] date-format-month-year = MM AA: Format string contains an unescaped latin alphabet character `A`
```

About the `date-format-short = M/d/aa`: I also think it would be difficult to catch those errors. Hopefully those strings will stabilize within a few iterations :smile:

tobias47n9e: Oh I just noticed I still need to check for items that start with `datetime-format`

tobias47n9e: Now also pushed the checks for `datetime-format`

kueda (approved):

PR #2424, opened 2024-11-14
Keep saved but not uploaded audio recordings when clearing cache

angielt:


src/sharedHelpers/clearCaches.ts
```diff
@@ -53,13 +53,13 @@ const clearSyncedMediaForUpload = async realm => {
...
 
-  // Clean out photos
+  // Clean out sounds
   const unsyncedObservationsWithSounds: RealmObservation[] = realm
     .objects( "Observation" )
     .filtered( "observationSounds._synced_at == nil" );
   const unsyncedSoundFileNames = unsyncedObservationsWithSounds
     .map( observation => observation.observationSounds.map(
-      os => os.sound.fileUrl?.split( "soundUploads/" )?.at( 1 )
+      os => os.sound.file_url?.split( "soundUploads/" )?.at( 1 )
```
kvangork, line 62: Would it be a good idea to also modify line 19 in this file? The interface there no longer matches this property name.

angielt: good catch

kvangork: Thanks @angielt! Looks good to me now.

PR #2439, opened 2024-11-18
#1433 - when viewing explore map from project details - zoom on project's place

### LINKED ISSUE #1433: Project Explore screen opens world map, not project place map
**Describe the bug**
On iNat web, when a user opens the Explore map from a project, the map shows the project place. In iNat RN, opening Explore from a Project opens the worldview, so a user has to zoom and pan the map to find the project observations.

**To Reproduce**
1. Go to Projects
2. Tap a project you have joined
3. Go to "Observations" which opens the Explore screen
4. The map is set to worldwide, not the project place

A user should be able to immediately see project observations in the map without needing to zoom/pan.

budowski:
Fix #1433


albullington (changes requested):
This is working for me, but the UX is a little jarring -- when you open the map from project details, you can see the map from the previous Explore screen you were on. There's a loading wheel in the gray header bar, but there's otherwise no indication to the user that the map is about to reload. Then, after a delay, the map reloads to the correct project location.

Can we show the user the project map first, instead of displaying the previous Explore map? I think it would be better to show a loading wheel like we do when a user first lands on `RootExplore` from the tab bar than a stale map.

budowski:
Added a loading indicator - basically we have to wait for the explore API call to return the bounds of the map (which is always dependent on the observations of the project, not just the place ID)

albullington (approved):

PR #2467, opened 2024-11-20
fix: update user language on server after changing in app

kueda:
Previously these update requests were getting sent and returning with a success status... but they weren't actually changing the language on the server. This updates inaturalistjs to fix a bug and submits the JSON body of the request in a way the server will understand.

It also updates QueuedItem to prevent it from processing tasks indefinitely if they fail for some reason.


albullington (approved):
Looks good, and is updating the web language for me with both online and offline functionality.

PR #2473, opened 2024-11-21
fix: prevent overlay of tiles from multiple queries in Explore on Android

### LINKED ISSUE #2026: Map tiles don't change after you redo search in map area
**Describe the bug**
In Android, the map tiles don't change after you redo search in map area

**To Reproduce**
1. Go to Explore
2. Go to observations view, map layout, and let whatever the defaults are load so you can see some map tiles
3. Zoom into an area and tap REDO SEARCH IN MAP AREA
4. The obs count updates, but the map tiles don't. You may need to zoom out to prove this to yourself.

**Expected behavior**
Map tiles should update whenever the query parameters do.

 - Device: Pixel 8
 - OS: Android 
 - App version: 0.44.0

**Additional context**
Not sure why this would only happen on Android.

### LINKED ISSUE #2432: Explore rapidly alternates between different maps
**Describe the bug**
In Android, Explore rapidly alternates between different maps

**To Reproduce**
1. Go to Explore. Explore starts with a species view and all organisms. Set the location to Greater San Francisco Bay Area:
Screenshot_20241114-193943

2. Go to Observations and select the map view. Wait for the map to display.
Screenshot_20241114-194030

4. Tap on 'All Organisms' and select taxon *Scoliopus bigelovii*.
5. Two or three maps are displayed over and over in rapid succession. They differ in their center coordinates and zoom. These are two of them:
Screenshot_20241114-194124 Screenshot_20241114-194136

**Expected behavior**
The map should not change quickly. Only a single map should be displayed or the oscillation should at least come to an end.

The map should be overlayed with tiles for the selected taxon. The maps are still shown for all organisms. The map should look like this:
Screenshot_20241114-194540

 - Device: Pixel 7 Pro, also Android emulator w/ image Pixel 6a
 - OS: Android 15 on Pixel 7 Pro, Android 11 in Android emulator
 - App version: 0.53.1 (123)

**Additional context**
A workaround is to tap My Observations and then tap Explore again. This causes the oscillation to stop and the correct tiles for the selected taxon to be displayed.

### LINKED ISSUE #2522: Explore's map view opens ObsDetails beyond observation tiles
**Describe the bug**
Explore's map view accepts taps outside the map region and navigates to ObsDetails for them.

**To Reproduce**
1. Go to Explore, select Observations, and enter Map View.

Screenshot_1733419469

2. Tap on the map outside the region of observation tiles. In this example, I aimed for Jepson Prairie and tapped near the top of the map. The app navigates to ObsDetails:

Screenshot_1733419462 Screenshot_1733419528

Oddly enough, the observation that opened is not even from where I tapped. It is from much further south.

**Expected behavior**
Ignore taps on the map where tiles do not show observations. 

 - Device: Pixel 7 Pro
 - OS: Android 15
 - App version: 0.55.3+130

**Additional context**
This issue has been observed on Android. It is unknown if this is also an issue on iOS.

joergmlpts:
This is a fix for the second problem seen in issue #2432. It also fixes issue #2026.

React-native-mapview on Android does not reliably reflect changes to the observation tile url template. This fix does not change the url template behind MapView's back anymore . Instead, it shuts down the observation tiles first and then restarts  observation tiles with the updated url template.

This change detects changes to the url template and it first renders \<MapView\> without \<TileUrl\> for one frame. It then renders \<MapView\> with \<TileUrl\> for the new url template.

- These changes test for Android and there should be absolutely no change to the app on iOS.
- The plan was to use a handler installed on `onMapReady`. But it turned out `onMapReady` is is not called at all on Android. Since the code uses `onRegionChangeComplete`, in order to get `onRegionChangeComplete` to fire it needs to slightly modify the region when \<MapView\> is rendered without \<TileUrl\>.
- This reverts a refactoring that @albullington committed in April. I am sorry for undoing someone's work but this code needs access to data that had been pulled out.

Closes #2432
Closes #2026
Closes #2522


src/components/SharedComponents/Map/Map.tsx
```diff
@@ -294,7 +314,25 @@ const Map = ( {
...
+    ? {
+      ...curRegion,
+      latitudeDelta: 1.00001 * curRegion.latitudeDelta,
+      longitudeDelta: 1.00001 * curRegion.longitudeDelta
+    }
+    : curRegion );
+  const mapRegion = fuzzRegion( setRegion( ) );
+  const mapInitialRegion = mapRegion === null
+    ? fuzzRegion( initialRegion )
+    : initialRegion;
```
kueda, line 335: I find this hard to read. If you have a function called `fuzzRegion()` I expect it to fuzz a region, not fuzz a region sometimes. I also don't understand why you're fuzzing a region for `mapRegion`, then potentially fuzzing it again if somehow it was null. If this has to do with different behavior in iOS and Andoid, please be more clear in your variable and method naming, or have some clear branching logic for iOS and Android. I understand from your comment *why* you're doing this, and it does work, but I think the code could express how it works more clearly.

`setRegion()` is also confusing, even though that's not your problem, per se. It doesn't seem to actually set a region, just get an initial region for... reasons. @albullington do you remember why you named it that way? It seems more like an accessor than a mutator, but it's accessing a "region" that is not the same as the `region` prop.

kueda (changes requested):
Apologies for the extreme delay in reviewing, we've had a lot going on in the last week or two. This looks good and fixes the problems raised in the two issues. I would like a small rewrite of the region fuzzing code for clarity, or I know I will have trouble maintaining this in the future. If my feedback is too vague there, let me know and I'll take a crack at it.

src/components/SharedComponents/Map/Map.tsx
albullington: No idea, I'm pretty sure like many things this was a refactor I did of something that was named long ago.

kueda: Thanks, @joergmlpts, I find this version easier to read. The only additional change I might make is to pull `fuzzRegion()` out of the component. It doesn't need to be redefined on every render.

kueda (approved):

PR #2492, opened 2024-11-27
Get altitude for user location; Add Geo model

jtklein:
If we merge it in as is now, the results from AICamera and the suggestions screen will differ. This is because on suggestions screen we are using the function to predict for a given image file url. This function does not yet use the geo model.
Nevertheless, I think this is a small enough PR to merge. If the UX is unsatisfactory we can hide the geo model behind a debug flag.

GitHub Copilot summary:
This pull request introduces several changes to support the addition of a new geo model file for both Android and iOS platforms. The changes include updates to environment configuration, build scripts, and various components to handle the new geo model file.

#### Environment Configuration:
* Updated `README.md` to include instructions for adding the geo model file and the corresponding environment variables (`ANDROID_GEO_MODEL_FILE_NAME`, `IOS_GEO_MODEL_FILE_NAME`).
* Modified `env.example` to include commented-out entries for the geo model file names for both Android and iOS.

#### Build Scripts:
* Updated `ios/link-inat-model-files.sh` to link the geo model file if it exists.
* Modified `scripts/add-example-model.js` to copy the geo model placeholder file.
* Updated `scripts/add-github-actions-test-model.js` with a placeholder for future geo model download logic.

#### Component Updates:
* Updated the AI Camera to send the user's location to the frame processing which then uses the geo model in combination with the cv model.


kueda (approved):
This looks good. I was able to follow the updated instructions and make debug and release builds, and the model still works on my iPhone SE. I made some minor edits to the README for clarity.

PR #2495, opened 2024-12-01
feat: use physical volume buttons to capture photo

### LINKED ISSUE #2212: Physical volume button triggers shutter in camera
Pressing the volume button on the side of an phone can be used as a shutter button to trigger the camera.

I'm not sure how to implement this in React Native, so we might need to use the underlying system methods.

- iOS:  https://developer.apple.com/documentation/avkit/avcaptureeventinteraction
- Android: Handle KeyDown for KeyEvent.KEYCODE_VOLUME_DOWN and UP

budowski:
Closes #2212


albullington (changes requested):
This is working well for me in StandardCamera. Let's add this to the AICamera too so users have a consistent experience across our product -- I think users will be incredibly confused if they can use these buttons in one camera but not the other.

albullington (changes requested):
This is close, but currently tapping the AI Camera volume button navigates twice, not once, which means Suggestions is getting pushed onto the stack twice in a row. We need to make sure all the functionality related to taking a photo is only handled once, so we're not accidentally saving duplicate photos to the camera roll, making extra observations, or navigating multiple times.

PR #2496, opened 2024-12-01
feat: retain local media files for one month or until 5GB threshold

### LINKED ISSUE #2263: Hold onto local copies of media longer
Consider holding onto local copies of obs photos and maybe sounds for longer. When you try to look back at the photos the day after with little or no internet it kinda sucks to not be able to see the photos. 

OS: iOS, Android

budowski:
Closes #2263

### DIFF
--- a/src/sharedHelpers/removeSyncedFilesFromDirectory.ts
+++ b/src/sharedHelpers/removeSyncedFilesFromDirectory.ts
@@ -1,34 +1,114 @@
 import RNFS from "react-native-fs";
 import { unlink } from "sharedHelpers/util.ts";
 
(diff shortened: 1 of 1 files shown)


albullington (approved):
This is retaining local media files for me. I have some outstanding questions about whether we *need* to be keeping `galleryPhotos` -- they're often between 5 MB and 11 MB on iOS so they're the first media type that would push a user over the 5GB threshold, and I'm not certain why we're storing them at all -- do we perhaps need them to write exif data at some point after an observation is created? The only concrete use case I see is knowing whether or not a saved photo was originally from the gallery, which we could certainly do in a less size intensive way. But that's all best saved for another issue.

PR #2507, opened 2024-12-03
fix: load new Explore results when filters change

kueda:
URL params got removed from the query key as a part of pull-to-refresh, but that's how changes to filters were triggering a new request. This restores that behavior, adds a test to ensure changes to filters call the API again, and fixes the prior test of pull to refresh which was not testing the equivalent behavior when pulling to refresh (API on first load, so the mock needed to be cleared).

### DIFF
--- a/src/sharedHooks/useFontScale.ts
+++ b/src/sharedHooks/useFontScale.ts
@@ -1,15 +1,20 @@
 import {
+  useEffect,
   useState
 } from "react";
 import DeviceInfo from "react-native-device-info";
 
-const useFontScale = ( ):Object => {
+const useFontScale = ( ): {
+  isLargeFontScale: boolean
+} => {
   const [isLargeFontScale, setIsLargeFontScale] = useState<boolean>( false );
 
-  DeviceInfo.getFontScale().then( fontScale => {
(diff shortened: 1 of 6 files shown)


src/sharedHooks/useFontScale.ts
```diff
@@ -1,15 +1,20 @@
 import {
+  useEffect,
```
kvangork, line 2: The changes to this file (`useFontScale.ts`) look unrelated to the PR's primary purpose, but I don't think they're harmful.

albullington (approved):
Thanks for fixing -- this is a good example of where I don't use Explore Filters and therefore didn't test that functionality when building pull-to-refresh & didn't notice that we still needed to fetch with multiple query keys.

src/sharedHooks/useFontScale.ts
kueda: This was mostly to remove some effects that were running after tests were finishing, which triggers warnings. If I'm touching tests I generally try to make them pass as cleanly as possible. The only risk I could see was that this hook would run less frequently and the app would not respond immediately to changes in system font size, but as far as I can tell it still does.

PR #2524, opened 2024-12-05
feat: remove interaction from AI camera suggestion

### LINKED ISSUE #2456: Remove info and checkmark buttons from suggestion displayed at top of AI camera
The AI camera suggestion should not have visible buttons. This will help users avoid confusion about how to capture their observation. By taking a photo and giving us the chance to use the online vision api they're also likely to get a better set of identification suggestions.

kueda:
Closes #2456

### DIFF
--- a/src/api/types.d.ts
+++ b/src/api/types.d.ts
@@ -33,40 +33,6 @@ export interface ApiObservationsUpdatesParams extends ApiParams {
 
 // Model types, need to be ordered by reference
 
-export interface ApiTaxon {
-  id?: number;
-  name?: string;
-  preferred_common_name?: string;
-}
-
-export interface ApiUser {
-  id?: number;
-  login?: string;
-}
-
-export interface ApiComment {
-  body?: string;
-  user?: ApiUser;
-}
-
-export interface ApiIdentification {
-  body?: string;
-  taxon?: ApiTaxon;
-  user?: ApiUser;
-}
-
-export interface ApiNotification {
-  comment?: ApiComment;
-  comment_id?: number;
-  created_at: string;
-  id: number;
-  identification?: ApiIdentification;
-  identification_id?: number;
-  notifier_type: string;
-  resource_uuid: string;
-  viewed?: boolean;
-}
-
 interface ApiFlag {
   id?: number;
 }
@@ -111,6 +77,42 @@ export interface ApiObservationSound {
   sound?: ApiSound;
   position?: number;
   uuid?: string;
+}
+
+export interface ApiTaxon {
+  default_photo?: ApiPhoto;
+  iconic_taxon_name?: string;
+  id?: number;
+  name?: string;
+  preferred_common_name?: string;
+}
+
+export interface ApiUser {
+  id?: number;
+  login?: string;
+}
+
+export interface ApiComment {
+  body?: string;
+  user?: ApiUser;
+}
+
+export interface ApiIdentification {
+  body?: string;
+  taxon?: ApiTaxon;
+  user?: ApiUser;
+}
+
+export interface ApiNotification {
+  comment?: ApiComment;
+  comment_id?: number;
+  created_at: string;
+  id: number;
+  identification?: ApiIdentification;
+  identification_id?: number;
+  notifier_type: string;
+  resource_uuid: string;
+  viewed?: boolean;
 }
 
 export interface ApiObservation {
--- a/src/components/SharedComponents/TaxonResult.tsx
+++ b/src/components/SharedComponents/TaxonResult.tsx
@@ -1,4 +1,5 @@
 import { useNavigation } from "@react-navigation/native";
+import type { ApiTaxon } from "api/types";
 import classnames from "classnames";
 import ObsImagePreview from "components/ObservationsFlashList/ObsImagePreview";
 import {
@@ -6,7 +7,9 @@ import {
   INatIconButton
 } from "components/SharedComponents";
 import { Pressable, View } from "components/styledComponents";
-import React from "react";
+import React, { PropsWithChildren } from "react";
+import type { GestureResponderEvent } from "react-native";
+import type { RealmTaxon } from "realmModels/types";
 import { accessibleTaxonName } from "sharedHelpers/taxon";
 import { useCurrentUser, useTaxon, useTranslation } from "sharedHooks";
 import colors from "styles/tailwindColors";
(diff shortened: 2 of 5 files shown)


src/components/SharedComponents/TaxonResult.tsx
```diff
@@ -192,51 +230,52 @@ const TaxonResult = ( {
...
-            )}
-            accessibilityLabel={t( "More-info" )}
-            accessibilityHint={t( "Navigates-to-taxon-details" )}
-          />
-        )}
-        { showCheckmark
-          && renderCheckmark()}
-        { showEditButton
+      </TaxonResultMain>
+      { !unpressable && (
```
kvangork, line 234: An alternative approach to consider: using `unpressable` with boolean operations to modify the values of `hideInfoButton` and `showCheckmark` and `showEditButton` either inline below or when unpacking `TaxonResultProps` would avoid the nesting introduced here. This might help future readers to understand which specific elements are being rendered or hidden, and why.

src/components/SharedComponents/TaxonResult.tsx
```diff
@@ -192,51 +230,52 @@ const TaxonResult = ( {
...
+                  onPress={handleTaxonOrEditPress}
+                  accessibilityLabel={t( "Edit-identification" )}
+                  accessibilityHint={t( "Edits-this-observations-taxon" )}
+                />
+              )}
+          { showRemoveButton && handleRemovePress
             && (
               <INatIconButton
-                icon="edit"
+                icon="close"
```
kvangork, line 270: These lines highlight an avoidable difficulty caused by nesting.

src/components/SharedComponents/TaxonResult.tsx
```diff
@@ -141,20 +183,16 @@ const TaxonResult = ( {
...
         }
-        onPress={handleTaxonOrEditPress || navToTaxonDetails}
-        accessible
-        accessibilityRole="link"
-        accessibilityLabel={accessibleName}
-        accessibilityHint={t( "Navigates-to-taxon-details" )}
       >
         <View className="w-[62px] h-[62px] justify-center relative">
           <ObsImagePreview
+            // TODO fix when ObsImagePreview typed
```
kvangork, line 195: What needs to be fixed?
Will it be enough to grep the codebase for the class name whenever something is typed, or do we need another mechanism for finding a TODO of this kind?

src/components/SharedComponents/TaxonResult.tsx
kvangork, line 1: A question/proposal for (future!) refactoring: Is there something like a non-interactive TaxonLabel (TaxonDisplay? Naming is hard.) component implicit in the `TaxonResultMain` addition? I'm envisioning a component which could then be enhanced with usage-specific interactive elements.

kvangork (approved):
No changes _necessary_, but several comments intended to help future efforts.
In that vein: to help reviewers, please separate changes into multiple atomic commits, even if they end up combined into an omnibus pull request.

src/api/types.d.ts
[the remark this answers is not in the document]
kueda: I added `default_photo` to `ApiTaxon`, and since `ApiPhoto` extends `ApiMedia` and `ApiMedia` refers to `ApiFlag`, they all needed to be defined below `ApiFlag`.

None of this typescript conversion and typing was necessary for this PR, but if we're going to require PRs to be strictly on topic, I fear this kind of cleanup work will never get done.

src/components/SharedComponents/TaxonResult.tsx
```diff
@@ -75,23 +84,56 @@ const TaxonResult = ( {
...
+    navToTaxonDetails,
+    t,
+    unpressable
+  ] );
+
+  // useTaxon could return null, and it's at least remotely possible taxonProp is null
+  if ( !usableTaxon ) return null;
+
+  const taxonImage = {
+    uri: ( usableTaxon as ApiTaxon )?.default_photo?.url
```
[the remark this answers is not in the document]
kueda: Again, this was for TypeScript. The `taxon` prop could be from the API or Realm, and `useTaxon()` will return a Realm object, so to satisfy TypeScript we need these casts, and since there is no runtime type checking in TypeScript since it all compiles to JS, we have to do this duck type check.

src/components/SharedComponents/TaxonResult.tsx
kueda: Not in code, but in the designs this is what Abhas calls a "Taxon List" even though it's more like a "Taxon List Item" and it doesn't always occur in a list. That's basically what we call a TaxonResult, though.

src/components/SharedComponents/TaxonResult.tsx
kueda: ObsImagePreview is in Javascript but has flow types, and those flow types are not great. They claim that the `source` prop must have a `uri` attribute, even though that component explicitly works around a missing source URI and adds a placeholder image. If you follow and fix all these problems when doing piecemeal TypeScript conversion, you end up converting the whole repo, which would be way out of scope.

As to what to do about fixing these problems if/when the underlying problem is fixed, IDK. I just assume if someone sees a message like this and there's no associated linter warning, they delete the comment, or they fix the problem.

src/components/SharedComponents/TaxonResult.tsx
kueda: I hear you, but personally when I'm reading I like to see variables close to their effects. If `hideInfoButton` hides an info button, I'd like to see the boolean get checked close to where the info button is rendered so I'm not jumping up and down in the file. IMO that's worth a little extra indentation. If I'm misinterpreting, could you provide an example?

src/components/SharedComponents/TaxonResult.tsx
kueda: What is the difficulty, exactly? Do just mean you've got some junky lines in the diff due to changes in indentation?

kueda:
> to help reviewers, please separate changes into multiple atomic commits, even if they end up combined into an omnibus pull request.

Will try, though it can be a bit challenge with the partial TypeScript approach we have. There are always a million moles to whack.

src/api/types.d.ts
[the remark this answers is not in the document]
kvangork: Let's split the difference. Cleanup & refactoring while you've got a file loaded into the mental cache is efficient and good and will help us avoid neglecting that important work. Splitting it into a separate PR would be excessive and likely to add delays and bugs. Splitting into separate commits in the same PR? Helpful in the moment for review and later when we're looking at a file's history.

src/components/SharedComponents/TaxonResult.tsx
kvangork: Agree about variables being close to their effects. I also want to avoid adding complex boolean logic at lower levels (e.g. I would not support L236 changing from ` { !hideInfoButton && (` to ` { !hideInfoButton && !unpressable && (`)
What I was trying to suggest was that somewhere prior to this view being rendered, we'd have applied the effects of the override to the other boolean values like this: `hideInfoButton ||= unpressable; showCheckmark &&= !unpressable;`
That's marginally readable, but it's as far as I thought would be reasonable to suggest without asking for a complete refactor of the file.

src/components/SharedComponents/TaxonResult.tsx
kvangork: Yes. 
CleanShot 2024-12-11 at 17 11 53@2x

src/components/SharedComponents/TaxonResult.tsx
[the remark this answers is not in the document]
kvangork: Let me put it another way, since I didn't notice the more appropriate line when I made this comment: L40 says `taxon: RealmTaxon | ApiTaxon;` and I expected (perhaps from too many years in object-oriented code) that it would say `taxon: Taxon;` so that this component wouldn't have (or need!) knowledge of its data's source. I'm curious to learn if there's a different pattern here that I should learn.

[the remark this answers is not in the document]
kueda: Ah, yeah, it should be a single class that gets instantiated with data from the API or the database and presents a single interface, but it's not. Since we have only partially adopted TypeScript, a lot of what we're doing with it is just trying to work around some of the JS problems we have, one of which is throwing around POJOs from the API and Realm objects as if they were equivalent, but they don't even present consistent camelcase / snakecase members. Using TypeScript to at least describe the madness is an improvement, but you're right, the correct solution would be a single class.

PR #2551, opened 2024-12-14
Refactor: upload code

albullington:
Creative day project. I bailed on background uploading and settled for refactoring in an attempt to make our upload code make slightly more sense. A few things in here:

- Use a single `useEffect` in `useUploadObservations` to make it clear when and why uploads are starting
- Refactor toolbar progress and individual progress to be derived any time upload progress is incremented, and simplify calculations
- Use a standardized `startIndividualUpload` function that works across the multi upload flow and MyObs
- Break up all functionality in `uploadObservation` into bite-sized functions with clear naming
- Create `addObservationsToUploadQueue` function to set upload state (adding to queue, prepping the progress status) in a single function
- Remove eventListeners to listen for progress updates and instead pass this into `uploadObservation` via the `updateTotalUploadProgress` callback, which makes it easy to see exactly when and why we're incrementing progress in a single file
- Break the toolbar timeout into a separate hook from the rest of the upload code, `useToolbarTimeout`, since it has a single purpose to update state & doesn't have to do with the rest of the upload queue / progress

Would be good to get a review on this since I'm sure I broke something / missed something. Hopefully this makes things a little less complicated.


albullington:
@angielt would you mind reviewing this? doesn't have to be super thorough, just want a pulse check before merging to make sure nothing related to uploads/deletions is broken

angielt (changes requested):
Uploading and deleting works for the most part. I did experience some count inconsistencies when uploading batches of observations. Also when uploading multiple obs from group photos, there were times where the last observation wouldn't upload(and the dots didn't spin or stopped spinning). 
This might be a change/update I was unaware of, if it is please ignore, but I didn't see any individual upload progress ring in list view for each upload, only spinning dots. 
IMG_3184

https://github.com/user-attachments/assets/c2128ba2-a662-4eb3-9a56-6fe578d3c158

albullington:
Closing this for now because it's buggy and it was a creative day experiment. Might revisit at some point and try to implement in more incremental changes.

PR #2584, opened 2024-12-24
feat: update vision-camera-plugin-inatvision to a version that fully supports the geomodel

jtklein:
In this PR the vision-plugin was updated (changes see below).
In the app the `taxonomyRollupCutoff` is removed. If no value is passed into the frame processor it uses a dynamic calculation of this value of 0.0001 times the top combined score. Combined score is equal to vision + geo scores or vision scores only if no geo scores obtained.

Changes in the vision-plugin from 4.1.4 to 4.2.0 are:
1. A passed in location (lat, lng) is now used to look up a standardized location and elevation. For this we are using the h3, hex grid framework with the elevation around the world binned to resolution 4, from the 5 minute bioclim elevation tiff:
https://www.worldclim.org/data/worldclim21.html
The location used for getting geomodel scores then consists of the average elevation as well as centroid of the h3 cell the passed in location is in.
2. Fixes a bug with geomodel scores caching on iOS. The geomodel was run for each frame. Now only when the location changes.
3. Adds geomodel to Android.


src/components/Camera/AICamera/FrameProcessorCamera.js
```diff
@@ -134,6 +132,10 @@ const FrameProcessorCamera = ( {
 
   const patchedOrientationAndroid = orientationPatchFrameProcessor( deviceOrientation );
   const patchedRunAsync = usePatchedRunAsync( );
+  const hasUserLocation = !!userLocation && !!userLocation?.latitude && !!userLocation?.longitude;
+  const location = hasUserLocation
```
jtklein, line 136: TODO: Explain this

src/components/Camera/AICamera/FrameProcessorCamera.js
```diff
@@ -134,6 +132,10 @@ const FrameProcessorCamera = ( {
 
   const patchedOrientationAndroid = orientationPatchFrameProcessor( deviceOrientation );
   const patchedRunAsync = usePatchedRunAsync( );
+  const hasUserLocation = !!userLocation && !!userLocation?.latitude && !!userLocation?.longitude;
```
jtklein, line 135: TODO: Refactor this

src/components/Camera/AICamera/FrameProcessorCamera.js
```diff
@@ -162,16 +164,15 @@ const FrameProcessorCamera = ( {
             modelPath,
             taxonomyPath,
             confidenceThreshold,
-            taxonomyRollupCutoff,
             numStoredResults,
             cropRatio,
             patchedOrientationAndroid,
-            useGeomodel: !!userLocation,
+            useGeomodel: !!location,
```
jtklein, line 170: TODO: Use hasLocation here

src/components/Suggestions/hooks/useOfflineSuggestions.ts
```diff
@@ -14,21 +12,29 @@ const { useRealm } = RealmContext;
 
 const useOfflineSuggestions = (
   selectedPhotoUri: string,
-  options: Object
+  options: {
+    dispatch: () => void,
+    flattenedUploadParams: Object,
```
jtklein, line 17: See above

src/sharedHelpers/mlModel.ts
```diff
@@ -52,12 +52,18 @@ export const predictImage = ( uri: string ) => {
   if ( !url ) {
     throw new Error( `predictImage received invalid URI: ${uri}` );
   }
+  const hasLocation = !!location && !!location?.latitude && !!location?.longitude;
```
jtklein, line 55: TODO: refactor here

src/components/Camera/AICamera/FrameProcessorCamera.js
kueda: Why is this even necessary? You're only setting it when we already know the user location, so why would you ask InatVision for some other location when we already know the user's location?

kueda (commented):
As far as I can tell this is working fine (I can build, AICamera and Suggestions work as I would expect both online and offline). I'm not noticing an improvement in results, but I also don't have a good example of something the vision model performed poorly on without geomodel input.

My only caveat is my confusion over why we need to re-request location info from the geomodel before, which doesn't need to block.

PR #2594, opened 2024-12-29
Fixes for Map in Android

### LINKED ISSUE #2529: Zoom level resets on every zoom in ObsDetails map on Android
@kueda said:

> Android bug when trying to zoom the map from obsdetails

Map resets after almost every pinch-to-zoom action

[4357bdd3-1800-4fbd-8011-78bae1295c65](https://uploads.linear.app/4a6b5374-94c1-40ab-a1be-cd1d70b88f69/8b50d16f-29ee-4a4e-949a-eb0f1e44e5ca/4357bdd3-1800-4fbd-8011-78bae1295c65?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzRhNmI1Mzc0LTk0YzEtNDBhYi1hMWJlLWNkMWQ3MGI4OGY2OS84YjUwZDE2Zi0yOWVlLTRhNGUtOTQ5YS1lYjBmMWU0NGU1Y2EvNDM1N2JkZDMtMTgwMC00ZmJkLTgwMTEtNzhiYWUxMjk1YzY1IiwiaWF0IjoxNzMzNzY2NzA3LCJleHAiOjMzMzA0MzI2NzA3fQ.hvOpTclC3RwqeVCZenparfEFG79M31GNrrWzh62evyc)

v0.56.0 (131), Android 15, Pixel 8.

### LINKED ISSUE #2555: Current Location button does not work properly on Android
**Describe the bug**
The Current Location button in the map view of Explore does not center the map on the current location on Android.

**To Reproduce**
1. Go to Explore, select Observations and Map View

Screenshot_1734320520

2. Pan the map such that the current location is off the screen.

Screenshot_1734320530

3. Tap on Currrent Location button.

Screenshot_1734320544

4. An animation centers the map on the current location for a very short time and then returns to the previous map region, as shown above. Observation tiles are missing.

**Expected behavior**
The map animation should center on the current location.

 - Device: Pixel 7 Pro
 - OS: Android 15
 - App version: v0.57.0 (132)

### LINKED ISSUE #2593: Search in Location Picker does not update Map in Android
**Describe the bug**
In Android, a search in Location Picker updates the latitude and longitude displayed but it does not center the map on the new location.

**To Reproduce**
1. Go to My Observations, select an observation, edit observation
2. Tap on location

Screenshot_1735481070

4. Tap on search bar, enter a new location and click on one of the choices.

Screenshot_1735481116

6. The latitude, longitude, and name for the new location are displayed correctly, but the map has not been updated. Sometimes a spinner runs indefinitely:

Screenshot_1735481161

**Expected behavior**
The map should be centered on the new location.

 - Device: Pixel 7 Pro
 - OS: Android 15
 - App version: v0.57.0 (132)

**Additional context**
This issue can be assigned to me. I think I understand the root cause.

joergmlpts:
+ The map from ObsDetails does not reset after each zoom anymore. A state for the region needed to be introduced and updated after each gesture.
+ The current location button now centers the map on the current location. A callback needed to be called explicitly after `animateToRegion()`.
+ A search in LocationPicker now centers the map on the selected location.  A state for the region needed to be introduced and a callback needed to be called explicitly after `animateToRegion()`.

Closes #2529
Closes #2555
Closes #2593

### DIFF
--- a/src/components/SharedComponents/Map/Map.tsx
+++ b/src/components/SharedComponents/Map/Map.tsx
@@ -150,25 +150,27 @@ const Map = ( {
 
   // In Android, we maintain a state for defaultInitialRegion and initialRegion
   // that is updated on gestures like pan and zoom. This state is always null in iOS.
-  const [localRegion, setLocalRegion] = useState<Region|null>( Platform.OS === "android"
-    ? initialRegion || defaultInitialRegion
-    : null );
+  const [androidLocalRegion, setAndroidLocalRegion] = useState<Region|null>(
+    Platform.OS === "android"
+      ? initialRegion || defaultInitialRegion
+      : null
+  );
 
   // In Android, onMapReady does not fire when we pass parameter region instead
   // of parameter initialRegion. This state allows us to fire onMapReady and
   // fire it only once. This state is always false in iOS.
-  const [onMapReadyHasFired, setOnMapReadyHasFired] = useState( false );
+  const [onMapReadyHasFiredAndroid, setOnMapReadyHasFiredAndroid] = useState( false );
 
   // In Android, animateToRegion animates to the given region but the map then
   // immediately returns to the previous region. We fake a gesture to the
   // desired region to make it stick. This state stores the region for this
   // gesture. This state is always null in iOS.
-  const [animateRegion, setAnimateRegion] = useState<Region|null>( null );
+  const [androidAnimateRegion, setAndroidAnimateRegion] = useState<Region|null>( null );
 
   const animateToRegion = ( newRegion: Region ) => {
     mapViewRef.current?.animateToRegion( newRegion );
     if ( Platform.OS === "android" ) {
-      setAnimateRegion( newRegion );
+      setAndroidAnimateRegion( newRegion );
     }
   };
 
@@ -216,7 +218,7 @@ const Map = ( {
     onPermissionGranted
   } );
 
-  // In Android, we always return a state, either region or localRegion.
+  // In Android, we always return a state, either region or androidLocalRegion.
   const setRegion = ( ) => {
     if ( Platform.OS !== "android" && initialRegion ) {
       return null;
@@ -225,7 +227,7 @@ const Map = ( {
       return region;
     }
     return Platform.OS === "android"
-      ? localRegion
+      ? androidLocalRegion
       : defaultInitialRegion;
   };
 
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/Map/Map.tsx
```diff
@@ -148,12 +148,36 @@ const Map = ( {
...
+  // In Android, we maintain a state for defaultInitialRegion and initialRegion
+  // that is updated on gestures like pan and zoom. This state is always null in iOS.
+  const [localRegion, setLocalRegion] = useState<Region|null>( Platform.OS === "android"
+    ? initialRegion || defaultInitialRegion
+    : null );
+
+  // In Android, onMapReady does not fire when we pass parameter region instead
+  // of parameter initialRegion. This state allows us to fire onMapReady and
+  // fire it only once. This state is always false in iOS.
+  const [onMapReadyHasFired, setOnMapReadyHasFired] = useState( false );
```
kueda, line 160: Same here. This will be very confusing if for some reason we need to know if `onMapReady` has fired in iOS. The reader's assumption should be that it has if `onMapReadyHasFired` is true given the name, but it will be false. As above, I'd prefer this OS-specific functionality to be in a separate file, but failing that I think `onMapReadyHasFiredAndroid` or something would be more explicit.

src/components/SharedComponents/Map/Map.tsx
```diff
@@ -192,14 +216,17 @@ const Map = ( {
     onPermissionGranted
   } );
 
+  // In Android, we always return a state, either region or localRegion.
```
kueda, line 219: Does it cause a problem to always return a state in iOS? Why do we need OS-specific logic here?

kueda (changes requested):
This fixes all three bugs, which is great, but there are a lot of Android-specific carve-outs in this code that I suspect are going to cause confusion in the future. I want to be sure these changes cause problems in iOS and thus merit separate behavior, and if so, that we use names to reflect the OS-specific usage of these new variables.

src/components/SharedComponents/Map/Map.tsx
joergmlpts: I cannot build and test for iOS. So I kept the iOS functionality exactly the same.

Whether these states work for iOS is not clear. For example, on Android, `onMapReady` only fires when `initialRegion` is passed. A new state ensures that it will now fire in Android where we pass `region` instead of `initialRegion`. In iOS, if `onMapReady` also only fires when `initialRegion` is passed, iOS will not work with this code: The functionality that emulates `onMapReady` in Android is in a callback that is called when no gesture is detected. However, iOS only calls the callback when a gesture is detected.

PR #2596, opened 2025-01-07
2210 new buttons for user filters in explore filters

### LINKED ISSUE #2210: New Buttons for User Filters in Explore Filters
Replace the "My Observations" button in ExploreUserSearch with two buttons "By Me" and "Not By Me".
The buttons should have the following functionality:
- filter by a user: 1 user you choose
- by me: my observations
- not by me: all users excluding me

primarily so that on species view in explore, you can filter by species you have not observed yet, which is a highly requested feature

angielt:
Closes #2210 

Edit (kvangork):
Linking to Linear issue, because doing this in a comment did not seem to work to create the link and update status there.
Closes MOB-188


kvangork:
Linking to Linear issue.
Closes MOB-188

src/components/SharedComponents/ButtonBar.tsx
```diff
@@ -51,7 +51,7 @@ const ButtonBar = ( {
         const {
           title, onPress, isPrimary, ...props
         } = button;
-        console.log( "props", { ...props } );
+
         return (
           <Button
```
albullington, line 56: We should add a `key` to this component since we're mapping over an array and currently getting a warning about that. (Not a new problem, but let's fix it).

src/realmModels/Observation.js
```diff
@@ -110,7 +110,12 @@ class Observation extends Realm.Object {
       rank_level: true
     },
     taxon_geoprivacy: true,
-    time_observed_at: true
+    time_observed_at: true,
+    user: {
```
albullington, line 114: What's the reason we're storing a user object here? Afaik, we should *only* be storing observations that belong to the logged in user, so this feels redundant -- this would always store the same user for every observation.

src/components/Explore/hooks/useInfiniteExploreScroll.js
```diff
@@ -96,8 +98,16 @@ const useInfiniteExploreScroll = ( { params: newInputParams, enabled }: Object )
     await refetch( );
   };
 
-  const observations = flatten( data?.pages?.map( r => r.results ) ) || [];
+  let observations = flatten( data?.pages?.map( r => r.results ) ) || [];
   let totalResults = data?.pages?.[0].total_results;
+
+  // filter out obs from excluded user and adjust count
+  if ( excludedUser && observations ) {
```
albullington, line 105: I think this code is causing a bunch of rerenders in certain circumstances. If I filter by excluding the current user, I'm seeing the screen rerender over and over again with a different number of species and different results. If I then search by a different taxon, the screen rerenders 4-5 times, showing a blank screen in between each time.

src/components/Explore/SearchScreens/ExploreUserSearch.js
```diff
@@ -74,6 +74,30 @@ const ExploreUserSearch = ( { closeModal, updateUser }: Props ): Node => {
...
+          onUserSelected( currentUser );
+        }
+      },
+      isPrimary: false,
+      className: "w-1/2 mx-2"
+    },
+    {
+      title: t( "NOT-BY-ME" ),
+      onPress: () => {
+        console.log( "not by me pressed" );
```
albullington, line 91: let's try to remove extra log statements

albullington (changes requested):
A few things blocking here: many rerenders when filtering by excluded user and the `ButtonBar` styling that impacts other screens.

One other thing I'm noticing is that filtering by the excluded user while all organisms / worldwide is selected results in an endless scroll wheel and results never load. Is this what we're expecting to happen here? If this API call is too slow to load, we should show a timeout message instead of an infinite loading wheel; otherwise, we should make sure this API call is actually getting called and results are shown.

src/realmModels/Observation.js
angielt: Good point. I originally included it there to query for users in useInfiniteExploreScroll. I removed it from the realm file and only included the user object in the query fields in useInfiniteExploreScroll.

albullington (changes requested):
ButtonBar looks good. I'm still seeing a questionable number of total results when using the NOT BY ME filter. Here's what I'm trying:

- filter by plants in Alameda County. go to Explore -> ObservationsGridView. see that there are 173824 total observations of plants
- add the NOT BY ME filter. see that there are now only 20 observations of plants. wonder how i managed to make hundreds of thousands of observations of plants.
- tap into an ObsDetail screen then back to Explore. now, it says there are 40 observations of plants not by me.

BY ME appears to be working fine -- i have 170 observations of plants in Alameda County which is about what i would expect.

The actual infinite scroll for NOT BY ME appears to be working, so I think it's only the total results number that is impacted. This is also true in SpeciesView. (in SpeciesView I'm trying to infinite scroll with Protozoans in Alameda County because I expect this to be a smaller number than plants, and every time I pull down to infinite scroll I'm seeing another 10 species being added to the total count until I finally hit 72 species)

This might be a good place to add an integration test, just to make sure NOT BY ME is returning the expected number of total results in SpeciesView and ObservationsView.

albullington (approved):
Looks ready to merge.

PR #2599, opened 2025-01-08
feat: add simplified MyObservations for logged out users with at least 1 obs

kueda:
Reviewer notes:

* Only works in debug mode
* Adding obs is in debug mode is current broken pending completion of the new confirmation screen, so to test this PR, you need to add at least one obs while signed out in the default mode, *then* enter debug mode
* I tried to avoid polluting MyObservations with experimental code, so there's some degree of awkward duplication
* MyObservationsSimple is intentionally monolithic; IMO we can clean and modularize stuff like calculating leaf taxa from the Realm db when we're closer to releasing this; happy to field critiques of style and organization, but I'm mostly interested in whether this meets the product requirements

Closes #MOB-317


src/components/MyObservations/MyObservationsContainer.js
```diff
@@ -188,6 +191,30 @@ const MyObservationsContainer = ( ): Node => {
   // this component again after returning from ObsEdit
   const onScroll = scrollEvent => setMyObsOffset( scrollEvent.nativeEvent.contentOffset.y );
 
+  if ( isDebugMode() && !currentUser && observations.length > 0 ) {
+    return (
+      <MyObservationsSimple
```
albullington, line 196: @kueda curious if you considered making `MyObservationsSimple` into its own navigation route? I liked the simplicity of letting the navigator [figure out conditionally](https://github.com/inaturalist/iNaturalistReactNative/blob/f84d09648d6b01f727436689c9a02ced0758892d/src/navigation/StackNavigators/TabStackNavigator.js#L162) which screen to show for ObsDetails.

kueda: I did not consider it, but now that I am considering it, I wouldn't do it that way. That approach in the navigator assumes that we *always* show a different version of ObsDetails in debug mode, but the requirement here was to only show a different version of MyObs if the user is signed out and has more than one obs. To do that, I actually want almost all of the functionality in MyObservationsContainer intact, i.e. I want identical controller logic but different view logic.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -0,0 +1,245 @@
...
+        <Body1 className="mb-[4px]">{ t( "Intl-number", { val: stat } ) }</Body1>
+        <Heading5>{ label }</Heading5>
+      </View>
+    );
+  }, [t, numTotalObservations, numTotalTaxa] );
+
+  return (
+    <>
+      <ViewWrapper isDebug>
+        { numUnuploadedObservations > 0 && (
```
kvangork, line 142: Is this for the green bar at the top of the screen? I saw it appear with 2 observations, but the issue spec says it should appear with 10+

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -0,0 +1,245 @@
...
+    ).filter( id => Number( id ) !== Number( o.taxon?.id ) );
+    return taxonAncestorIds;
+  } ).flat( Infinity );
+  const leafTaxonIds = taxonIds.filter( taxonId => !ancestorIds.includes( taxonId ) );
+  const numTotalTaxa = leafTaxonIds.length;
+
+  // Get leaf taxa if we're viewing the species tab
+  let leafTaxa: Realm.Results | Array<RealmTaxon> = [];
+  if ( activeTab === TAXA_TAB ) {
+    // IDK how to placate TypeScript here. ~~~kueda 20240108
```
kvangork, line 111: This may be my worst nitpick ever: s/2024/2025/
Serious suggestion: git history (blame) should always show the author, date, and—even more helpfully—the context for a comment.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -0,0 +1,245 @@
...
+    paddingTop: 10
+  } ), [flashListStyle] );
+
+  // Calculate obs and leaf taxa counts from local observations
+  const realm = useRealm();
+  const numTotalObservations = realm.objects( "Observation" ).length;
+  const distinctTaxonObs = realm.objects( "Observation" )
+    .filtered( "taxon != null DISTINCT(taxon.id)" );
+  const taxonIds = distinctTaxonObs.map( o => ( o.taxon as RealmTaxon ).id );
+  const ancestorIds = distinctTaxonObs.map( o => {
```
kvangork, line 96: Without knowing enough about the design patterns in our codebase to request a specific change, I'll say that it surprised me to see this much data logic so close to the view. Does the subtlety of filtering happening here warrant moving it into a utility function or helper class? (Particularly if it's reusable or shared with some other part of the app (e.g. Explore))

src/components/ObservationsFlashList/ObsListItem.js
```diff
@@ -76,24 +78,28 @@ const ObsListItem = ( {
...
-            observation.time_observed_at
-            || observation.observed_on_string
-            || observation.observed_on
-          }
-          classNameMargin="mt-1"
-          geoprivacy={geoprivacy}
-          taxonGeoprivacy={taxonGeoprivacy}
-          belongsToCurrentUser={belongsToCurrentUser}
-        />
+        {!hideMetadata && (
```
kvangork, line 83: I low-key love the property naming here. In my world, hiding metadata should always require an active choice to deviate from the norm.

src/components/SharedComponents/Buttons/RotatingINatIconButton.js
```diff
@@ -28,7 +28,7 @@ type Props = {
   icon: string,
   mode?: "contained",
   onPress: Function,
-  rotating: ?boolean,
+  rotating?: boolean,
```
kvangork, line 31: Nice catch. This seems like something the linter should have raised before now. 🤨

src/realmModels/Taxon.js
```diff
@@ -124,12 +125,22 @@ class Taxon extends Realm.Object {
   }
 
   static forUpdate( taxon, extra = {} ) {
-    return {
+    const taxonForUpdate = {
```
kvangork, line 128: I have no idea what's going on here. Is it safe to ship? Do we need additional testing? I did _not_ test this branch beyond the specific feature changes being implemented.

kvangork (approved):

src/components/MyObservations/MyObservationsSimple.tsx
kueda: Good catch, will fix.

src/components/MyObservations/MyObservationsSimple.tsx
kueda: This absolutely does not belong here and if I thought this code was going to last I would probably create a separate MyObservationsSimpleContainer for business logic and move this particular chunk into a reusable hook like `useRealmObsStats()`. However, as I mentioned in the PR description, I suspect this will change significantly given the divergence between the requirements in the issue (only when signed out, only w/ one obs) and the designs (numerous designs showing signed in users with this layout, which will require a completely different way of calculating these stats).

So for now, it is hacky. LMK if you want me to refine this now rather than later.

src/components/SharedComponents/Buttons/RotatingINatIconButton.js
kueda: Flow makes no sense and is mostly useless, but it looks enough like TypeScript that my TS linter harangues me about things like this.

src/realmModels/Taxon.js
kueda: What's going on: in order to determine a leaf count of taxa represented by *local* observations, we absolutely need every taxon in the local database to have the `ancestor_ids` attribute populated. This change throws an error if any part of the app tries to save a taxon without that attribute.

Is it safe: I didn't notice any problems but this should noticeably crash if I overlooked any situations where we're not setting `ancestor_ids`, so I would expect to catch those bugs during internal or beta testing.

Even more: to my knowledge, Realm does not provide a way to write lifecycle callback functions like `beforeSave()`, so `Taxon.forUpdate()` kind of does that if people remember to use it before saving a taxon.

PR #2607, opened 2025-01-09
fix: save or discard the observation from the match screen

kueda:
* Add save/discard button functionality to match screen (closes MOB-353)
* Updates to TaxonDetails via MatchScreen (closes MOB-358)
* Saving from MatchScreen -> TaxonDetails -> Search (closes MOB-359)

These all ended up being kind of interrelated so apologies for the big PR. Probably the biggest / most invasive part of this is supporting search from TaxonDetails via the match screen.

I honestly did try to resist refactoring but couldn't help myself in some places. I don't love the approach in TaxonDetails here but untangling that component was more than I wanted to get into.


albullington (approved):
Lgtm. Save and discard from the Match Screen and TaxonDetails screen are working the way I'd expect them to, and so is taxon search. I'm going to make a related issue to hide the Explore button on TaxonDetails when coming from the MatchScreen, since that's one thing we need that wasn't specified in the issue.

PR #2609, opened 2025-01-11
 Add suggestions to match screen

albullington:
Closes MOB-352

- Creates a `useSuggestions` hook that isolates the online/offline suggestions logic from SuggestionsContainer
- Calls `useSuggestions` in `AdditionalSuggestionsScrollContainer` in the new Match screen to show a horizontal list of suggestions


albullington:
@kueda open to suggestions on code -- I pulled `useSuggestions` out of the `SuggestionsContainer` to try to have less duplication overall between Match/Suggestions, but I could take another pass next week and remove even more duplication. I didn't build out the functionality for what happens when you press a suggestion here and will write up a separate issue for that. Also not totally sure which suggestions we're showing here in terms of the nitty gritty (like, do we show AICamera suggestion or top online/offline suggestion at the top of the screen? if so, that means we should filter it out in the bottom list) so I think we can iron out those details when Abhas is back.

kueda (commented):
I think the code looks good, and I think pulling suggestions out into a hook is the right way to re-use that functionality. I left some comments that are mostly about form and readability, but I don't think any of those are blocking.

There are some functionality issues that feel like blockers to me, though:

- [ ] If you photograph something with a single suggestion and the match screen shows that species, that species appears again under "It might also be". The design specifies that these should not include the taxon on the match screen. I would just hide this entire section if there's only one suggestion, but regardless we should be sure to filter out the taxon that's on the match screen.
- [ ] I am seeing a lot of situations where there are suggestions with "0% confidence". These are taxa that show up in the regular suggestions screen with one dot. Not sure if this is a calculation error or if we're literally showing suggestions with a score of 0 on the regular suggestions, but it looks especially bad to present the user with a suggestion we have absolutely no confidence in. Can we just filter those out for this screen?
- [ ] I think the x padding in the suggestions is a lot bigger than in the designs

PR #2616, opened 2025-01-16
feat: show simplified MyObservations to logged in users

kueda:
Review notes

* You need to be in debug mode to view these changes
* In theory nothing should have changed outside of debug mode
* I don't love that the local taxa count gets calculated on every render in `MyObservationsSimplContainer` while logged out. In theory it only needs to be recalculated when an observation gets edited. Open to ideas on how to achieve that.
* Incomplete typing in useInfiniteScroll is causing TS errors in `MyObservationsSimpleContainer`; believe it or not I resisted the urge to refactor, but that would be the way to deal with those

Closes MOB-318


albullington (changes requested):
Seeing one thing here that's different from the Figma designs: in the new designs, we no longer want to show `ObsStatus` in list view (though we'll still need `ObsUploadStatus`.

Also, there are two things from the current version of `MyObservations` that have been removed here -- tapping the username no longer directs a user to their own profile, and the sync button no longer spins or changes colors to let a user know what's happening when tapped. I don't think we should remove this functionality.

Feels reasonable to create separate issues for any of these three things if you prefer that to making changes in this already large PR.

kueda:
The issue and the designs linked in the issue didn't mention anything about ObsStatus so I didn't do anything about it. Same goes for tapping the user icon or any upload state. @kvangork can you confirm that any or all of those are blockers here? If they are, I can do the profile pic and obs status ones, but I'll need more guidance on what upload state is supposed to look like. If they're not, someone else should make issues b/c I don't know how this is supposed to work.

albullington:
This is the [design](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=22778-76740&t=BwE4cRiZUn7vw8vJ-0) I'm looking at for upload state, and there's a nearby version that shows these without ObsStatus. But again, feels reasonable to leave for a separate issue.

kvangork:
> Feels reasonable to create separate issues for any of these three things if you prefer that to making changes in this already large PR.

This is the right solution. We should prefer to keep PRs small, and creating more issues to track next steps, regardless of whether they were in the original issue's spec or not, will help us keep moving.

kvangork:
> This is the [design](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=22778-76740&t=BwE4cRiZUn7vw8vJ-0) I'm looking at for upload state, and there's a nearby version that shows these without ObsStatus. But again, feels reasonable to leave for a separate issue.

Taking a look at the Figma links related to this, I'm confused too. Referring to a "Nearby version" is a red flag that we need to try something different from Figma links when communicating specifications. Maybe screenshots?

kueda:
Ok, I made https://linear.app/inaturalist/issue/MOB-387/remove-obs-status-except-for-upload-button-in-default-mode and https://linear.app/inaturalist/issue/MOB-387/remove-obs-status-except-for-upload-button-in-default-mode. I don't know enough to write an issue for how this screen should behave while uploading. Does that meet your review requirements, @albullington?

albullington:
Yep, thanks for making issues for those.

PR #2617, opened 2025-01-16
feat: select other suggestions in Match screen

albullington:
Closes MOB-368

- Reorders the other suggestion list, moves the top suggestion into the match header, and updates taxon photos with the new top suggestion

### DIFF
--- a/src/components/Match/Match.js
+++ b/src/components/Match/Match.js
@@ -4,6 +4,7 @@ import MapSection
   from "components/ObsDetailsDefaultMode/MapSection/MapSection";
 import { Button, Divider, ScrollViewWrapper } from "components/SharedComponents";
 import { View } from "components/styledComponents";
+import _ from "lodash";
 import React from "react";
 import { useTranslation } from "sharedHooks";
 
@@ -20,8 +21,9 @@ type Props = {
   handleSaveOrDiscardPress: ( ) => void,
   navToTaxonDetails: ( ) => void,
   handleLocationPickerPressed: ( ) => void,
-  suggestions: Object,
-  onTaxonChosen: ( ) => void
+  topSuggestion: Object,
+  otherSuggestions: Array<Object>,
+  onSuggestionChosen: ( ) => void
 }
 
 const Match = ( {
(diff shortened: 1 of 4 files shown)


src/components/Match/Match.js
```diff
@@ -20,8 +21,9 @@ type Props = {
   handleSaveOrDiscardPress: ( ) => void,
   navToTaxonDetails: ( ) => void,
   handleLocationPickerPressed: ( ) => void,
-  suggestions: Object,
-  onTaxonChosen: ( ) => void
+  topSuggestion: Object,
+  otherSuggestions: Array<Object>,
```
kueda, line 25: This file is not using flow and is not typescript, so these types probably shouldn't be here. If you do decide to convert to TypeScript, using `Object` types doesn't achieve the goal of allowing IDEs to provide developers with hints about how to use the component (which I think is why we decided to write new code in TS in the first place, right?), nor does it provide much type safety. IMO it's totally fine to use JS for speed, but I don't think incomplete typing like this helps.

src/components/Match/MatchContainer.js
```diff
@@ -148,12 +149,37 @@ const MatchContainer = ( ) => {
...
+      suggestionsList,
+      suggestion => suggestion.combined_score || suggestion.score,
+      ["desc"]
+    );
+
+    // order the chosen suggestion at the beginning of the list, so it
+    // can become the new top suggestion
+    const chosenIndex = _.findIndex(
+      sortedList,
+      s => s === selection
```
kueda, line 168: This use of object equivalence strikes me as a little risky. For example, if you wanted to use truly immutable data here and replaced

```javascript
const suggestionsList = [...orderedSuggestions];
```

with

```javascript
const suggestionsList = _.cloneDeep( orderedSuggestions );
```

it would not work. It might be a little more defensive to do something like `suggestion => suggestion.taxon.id === selection.taxon.id`.

kueda (approved):
Performs the way it's spec'd in the issue so I think it's good to go. My inline comments are nice-to-haves, not blockers.

src/components/Match/Match.js
albullington: Yeah, I see that. Total habit after 6 years of using... mostly Flow in this way. I think I'm going to leave for now because this warrants a bigger audit of all the files I've written in the past few weeks. I'm also sure we need to have a team convo soon about TS vs. Flow vs. no typing.

PR #2627, opened 2025-01-21
Fix #2573 - don't show location indicator if observation has no coord…

### LINKED ISSUE #2573: react-native-maps exception after selecting taxon for first observation in Android
**Describe the bug**

When you're a new user and you aren't signed in and haven't granted location permission, you get an error after choosing a taxon when creating an obs for the first time.

**To Reproduce**

1. Fresh install
2. Change to advanced mode
3. Take a photo of something human w/ StandardCamera
4. Choose human on suggestions
5. On TaxonDetails, you get an exception screen w/ `Error while updating property 'center' of a view managed by AIRMapCircle`
6. Tapping on SELECT THIS TAXON does nothing

**Expected behavior**

You should not see an exception and tapping the button should do what it does in iOS, which is to show a bottom sheet with options to upload now or save later

**Screenshots**

* App: iNat Next
* Device: Pixel 8
* OS: Android 15
* App version, or Browser: 0.57.0 (132)

**Additional context**

budowski:
…inates

### DIFF
--- a/src/components/SharedComponents/Map/Map.tsx
+++ b/src/components/SharedComponents/Map/Map.tsx
@@ -444,6 +444,10 @@ const Map = ( {
     !observation.obscured || observation.privateLatitude
   );
 
+  const latitude = observation?.privateLatitude || observation?.latitude;
+  const longitude = observation?.privateLongitude || observation?.longitude;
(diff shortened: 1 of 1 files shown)


angielt (approved):
Seems good. I don't see the error anymore

PR #2631, opened 2025-01-22
fix: prevent iconic taxon common names from disappearing in the local db

kueda:
Ensure useAuthenticatedQuery only executes the query when it knows if the user is signed in or not, and appends that state to the query key so signed in and signed out results are distinct.

This bug had nothing to do with Maori. When the user started the app while offline, `useIconicTaxa` would run and, for reasons I still don't understand, it was returning cached results from an unauthenticated request to`/v2/taxa?iconic=true`... even though the app had been restarted and in theory React Query should not have had any cached data.

This fix is more sweeping and a bit more risky, though. useAuthenticatedQuery really should return distinct results depending on whether the user is logged in, so that state should always be a part of its query key. There's no way for us to determine that state synchronously, though (well, we could use Realm, but that seems even riskier), so the approach here is to for `useAuthenticatedQuery` not to perform its request at all until we know whether the user is logged in or not. That might cause a slight delay and some additional renders before RNSI is fully functional.

Closes MOB-383


src/sharedHooks/useAuthenticatedQuery.js
```diff
@@ -1,29 +1,60 @@
...
-    };
-    return queryFunction( options );
-  },
-  retry: ( failureCount, error ) => reactQueryRetry( failureCount, error, {
-    queryKey
-  } ),
-  ...queryOptions
-} );
+) => {
+  const [userLoggedIn, setUserLoggedIn] = useState( LOGGED_IN_UNKNOWN );
```
albullington, line 18: We check for login state in so many places throughout iNatNext that I'm often wondering if we should be leveraging react navigation or persisting the login state in store somehow. I guess the main question here -- is there a reason to write a custom `useEffect` instead of using the `useCurrentUser` hook to check for login that we use everywhere else?

I don't *quite* know what solution we could implement with react navigation, but here's their documentation on authentication flows that makes me curious about how else we could do this: https://reactnavigation.org/docs/auth-flow/#what-we-need

src/sharedHooks/useAuthenticatedQuery.js
```diff
@@ -1,29 +1,60 @@
...
+
+  // The results will probably be different depending on whether the user is
+  // signed in or we wouldn't be using useAuthenticatedQuery in the first
+  // place, so we need to redo this request if the auth state changed
+  const authQueryKey = [...queryKey, queryOptions.allowAnonymousJWT, userLoggedIn];
+  if ( authQueryKey.includes( "useIconicTaxa" ) ) {
+    localLogger.debug( "authQueryKey: ", authQueryKey );
+  }
+
+  return useQuery( {
```
albullington, line 38: Is this all necessary? I thought react-query was already handling this when you add more items into the query key, i.e. if the query key in `useIconicTaxa` had three+ items in the array: `const queryKey = ["useIconicTaxa", reload, queryOptions.allowAnonymousJWT, userLoggedIn]` then it would fetch, cache, and return different query results for a query key with `userLoggedIn = true` vs. `userLoggedIn = false` and with or without an anonymous jwt.

src/sharedHooks/useTaxonSearch.ts
```diff
@@ -24,9 +24,10 @@ function saveTaxaToRealm( taxa: Taxon[], realm: Realm ) {
   }, "saving remote taxon from useTaxonSearch" );
 }
 
-const useTaxonSearch = ( taxonQuery = "" ) => {
+const useTaxonSearch = ( taxonQueryArg = "" ) => {
   const realm = useRealm( );
   const iconicTaxa = useIconicTaxa( { reload: false } );
+  const taxonQuery = taxonQueryArg.trim();
```
albullington, line 30: comment might be helpful, since I have no idea why we need to trim the query here.

albullington (approved):
This fixes the bug stated in the issue for me. I added some comments about two places where I'm surprised we need custom code -- feel free to ignore and merge if it's necessary to do things in a custom way in both of these.

src/sharedHooks/useAuthenticatedQuery.js
kueda: `useCurrentUser()` was my first attempt here, but it only returns null or a Realm user object. Here, we need to not execute the query if we don't know whether the user is logged in or not, and `useCurrentUser` doesn't tell us that. It conflates the unknown state with the logged out state. I could have altered `useCurrentUser` to have a yes|no|dunno response, but that seemed more invasive than this work merited. Altering `useAuthenticatedQuery` seemed unvasive enough.

src/sharedHooks/useAuthenticatedQuery.js
kueda: It's not necessary to fix this bug, but I am anticipating other, similar bugs where we accidentally use results from an unauthenticated query when we thought we were using results from an authenticated one. So yes, I could have simply altered the key we pass from `useIconicTaxa` to `useAuthenticatedQuery` and that would have fixed MOB-383, but this solution guards against similar bugs in the future.

src/sharedHooks/useTaxonSearch.ts
kueda: Agreed. Added.

PR #2634, opened 2025-01-23
fix: allow Explore searches in places again

kueda:
* remove assumed and undocumented meaning of page being zero from our use of useInfiniteScroll in Explore; uses explicit type checks instead
* document byzantine pagination logic in useInfiniteExploreScroll
* fix broken infinite scroll when sorting by faves (was only showing first page)

The underlying problem from 96c316a was setting initialPageParam to 1 in useAuthenticatedInfiniteQuery, which seems rational, except useInfiniteExploreScroll was using a false-ish value of page to detect when it was requesting its first page. This does the same by being more explicit about page's type without injecting undocumented meaning into the params we hand to useInfiniteScroll.

Closes MOB-400

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,7 @@ fastlane prod
             1. Delete the observation you just made
             1. Go to Explore and view a taxon
             1. Change to observations view and view an observation
+            1. Search for observations in a place
             1. Go to the observers profile
             1. Go to one project this user joined (if not available try another user)
             1. Sign in
--- a/src/api/types.d.ts
+++ b/src/api/types.d.ts
@@ -119,9 +119,50 @@ export interface ApiNotification {
   viewed?: boolean;
 }
 
-export interface ApiObservation {
+export interface ApiRecord {
+  created_at?: string;
+  id?: number;
+  updated_at?: string;
+}
+
+export interface ApiObservation extends ApiRecord {
   observation_photos?: ApiObservationPhoto[];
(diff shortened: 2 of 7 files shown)


kueda:
@kvangork I don't need a review here, but I would like your opinion on the need for a test. Writing an integration test that would have caught this bug would be quite hard: you have to mock species count results, place search results, and then finally the important part, the observation search, and then you can't really test whether the API is going to return stuff, just that we requested the right pagination params... which would be something, but a complicated and brittle test. IMO, if e2e wasn't its own slowdown / liability, I would say this would be a good e2e test. Do you think a test is worth it here?

kvangork:
Caveat: It's likely I'm misunderstanding some of the subtleties and that this advice is invalid.

This seems like a potential problem anywhere we have an "infinite" scroll. The documentation updates in the PR are a great way to mitigate that risk.
Looking at the code and reading your description of the difficulty in writing an integration test: rather than writing that test, I'd start by refactoring the way we generate pageparams, so that initial and next page params pass through the same code path which encapsulates the subtleties around first page detection, and would be more amenable to unit testing.

An expensive and brittle integration test with so many mocks seems like a poor investment of time.

kueda:
FWIW, the initial page param right now is just the default: `undefined`. I refactored the pagination stuff into a file that can be unit tested and I added unit tests. So not quite what you requested but at least there's some coverage for this logic.

PR #2636, opened 2025-01-25
feat: display time zones and times in time zones

kueda:
Some notable changes

* we're now storing observed_time_zone in realm; this only gets set by the server, so if we don't know it, we show the literal time
* date-fns-tz needed to be patched to work around a bug in Hermes that would only print "GMT" for some time zone / locale combos
* the latest version of date-fns has its own time zone support, but it doesn't print time zone names, just offsets, so I didn't use it; presumably it would be limited by the same Hermes bug
* I don't like the prop name `literalTime` but was struggling to think of a better, more adjectival name; it tells DateDisplay to interpret the time in the date string literally, e.g. if the string says "04:35", that should get printed

Closes MOB-149
Closes MOB-200


src/components/ObsDetails/DetailsTab/DetailsTab.js
```diff
@@ -138,7 +138,12 @@ const DetailsTab = ( { currentUser, observation }: Props ): Node => {
         <DateDisplay
           classNameMargin="mb-[12px]"
           label={t( "Date-observed-header-short" )}
-          dateString={checkCamelAndSnakeCase( observation, "timeObservedAt" )}
+          dateString={
+            checkCamelAndSnakeCase( observation, "timeObservedAt" )
+            || checkCamelAndSnakeCase( observation, "observed_on_string" )
```
jtklein, line 143: checkCamelAndSnakeCase checks if a snake_case version of a camelCase key is present but not vice versa. Maybe it would be a bit less confusing to access the key directly here instead.

src/components/SharedComponents/DateDisplay.js
```diff
@@ -37,40 +52,37 @@ const DateDisplay = ( {
...
     || taxonGeoprivacy === "private";
 
-  const formatDate = useMemo( () => {
-    if ( !belongsToCurrentUser && obscuredDate ) {
+  const formattedDate = useMemo( () => {
+    if ( !belongsToCurrentUser && dateObscured ) {
       return formatMonthYearDate( dateString, i18n );
     }
-    return formatApiDatetime( dateString, i18n );
+    if ( asDifference ) {
```
jtklein, line 64: First I thought this needs to be the first check because if we want to force to show a difference instead of a date what does it matter if the item is obscured. But apparently on the website we also don't show the precise date of an identification or comment on an obscured observation, so all good.

jtklein (changes requested):
Overall, only a few minor changes and questions.

Only one major thing I am unsure about is why did we need a different key for `observed_time_zone` in an Observation. Is this information not already present in the `observed_on` information, and then for things like `timeZone` and `literalTime` it could be inferred if a time zone is present or not directly in the format functions and DateDisplay component?

src/realmModels/types.d.ts
```diff
@@ -45,28 +55,43 @@ export interface RealmTaxon extends RealmObject {
...
+  taxon_geoprivacy?: "open" | "private" | "obscured" | null;
+  time_observed_at?: string;
+  timeObservedAt?: string;
   uuid: string;
 }
 
+export interface RealmObservation extends RealmObservationPojo {
+  missingBasics: ( ) => boolean;
+  needsSync: ( ) => boolean;
+  observationPhotos: Array<RealmObservationPhotoPojo>;
```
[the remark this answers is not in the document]
kueda: Yup, this was a mistake. Will fix. Do you think the POJO naming is confusing? I could also call it `RealmObservationJson` if that seems clearer.

kueda:
> Only one major thing I am unsure about is why did we need a different key for observed_time_zone in an Observation. Is this information not already present in the observed_on information,

`observed_on` contains a date, not a datetime, and thus has no time zone information. `time_observed_at` is in ISO 8601 format, which means it has a UTC offset, *not* a time zone. What we need is an IANA (tz database) time zone identifier so we can display the abbreviated name of the timezone, e.g. "CET" and not "GMT+1", which date-fns-tz does at least some of the time for some locales.

> and then for things like timeZone and literalTime it could be inferred if a time zone is present or not directly in the format functions and DateDisplay component?

Not sure I understand you here. Activity item header is a good example where we cannot infer the values of `timeZone` and `literalTime`. We don't have an IANA time zone for `comment.created_at` or `identification.created_at` (no `timeZone` value), but we do not want a literal interpretation of those strings because we want to display those datetimes in the viewer's time zone.

PR #2652, opened 2025-02-06
MOB-399 - make default mode opt-in for users

budowski:


jtklein (commented):
I was testing this just now.
From looking at the code changes I think this should satisfy the ticket, however, I can not test if the changes actually work, because default mode is only available from debug mode and debug mode overrides the advanced mode.

So, I think this PR should move ahead after default mode is out of debug mode.

Edit after a few days: I am not confident in my understanding of the underlying issue so maybe better for someone else to review this.
There are some implications for users that have the app already installed with the published version prior to this PR and how the app should behave with the next update after this is merged that I don't understand.

albullington (approved):

PR #2665, opened 2025-02-14
During signup check first if an email is taken or not

jtklein:
Closes MOB-321


jtklein:
E2e test passing locally so I will merge this in now.

PR #2666, opened 2025-02-15
Initial app state is in advanced mode (for now)

jtklein:
Closes MOB-453


jtklein:
E2e test fails in CI unrelated to changes, locally they pass.

PR #2678, opened 2025-02-20
User info to logged-in my obs empty

jtklein:
Closes MOB-462

Adds a user icon and name to the screen when a user is logged in but has no observations yet.


jtklein:
E2e tests are passing locally.

PR #2695, opened 2025-02-24
Setting to control if screen after camera/library is suggestions

jtklein:
Closes MOB-468


src/components/Camera/hooks/usePrepareStoreAndNavigate.ts
```diff
@@ -25,6 +25,7 @@ const usePrepareStoreAndNavigate = ( ): Function => {
   const setSavingPhoto = useStore( state => state.setSavingPhoto );
   const setCameraState = useStore( state => state.setCameraState );
   const setSentinelFileName = useStore( state => state.setSentinelFileName );
+  const isSuggestionsFlowMode = useStore( state => state.layout.isSuggestionsFlowMode );
```
albullington, line 28: could we add "advanced" to the naming here? maybe something like `isAdvancedSuggestionsMode`? i'd love to remember that this is a power user setting when i look at this code, and right now i'd be kind of confused by the naming about why we need this mode at all

src/components/PhotoImporter/PhotoLibrary.js
```diff
@@ -62,16 +63,21 @@ const PhotoLibrary = ( ): Node => {
   const advanceToMatchScreen = lastScreen === "Camera"
     && isDefaultMode;
 
-  const navToMatchOrSuggestions = useCallback( async ( ) => {
+  const navToNextScreen = useCallback( async ( ) => {
```
albullington, line 66: i could also see naming like `navBasedOnUserSettings` being helpful for looking at this in the future and trying to understand why we have so many different navigation paths and where they're coming from

albullington (approved):
looks good and is working as expected for me. i tried sharing into the app and it goes to ObsEdit no matter what the user setting is -- did you try changing that navigation based on the ObsEdit/Suggestions setting too? (seems reasonable to break that into a separate issue, but curious if you tried it already). as mentioned below, i think it could be helpful to name all these variables to reflect the fact that this is an advanced user setting option that's not the default, but i'm not tied to what those names actually are

src/components/PhotoImporter/PhotoLibrary.js
jtklein: Renamed

src/components/Camera/hooks/usePrepareStoreAndNavigate.ts
jtklein: Renamed

PR #2703, opened 2025-02-28
Mob 485 after a user grants location permissions on the match screen

angielt:
[MOB-434](https://linear.app/inaturalist/issue/MOB-434/match-screen-add-location-buttons-should-prompt-for-gps-permissions)
[MOB-485](https://linear.app/inaturalist/issue/MOB-485/after-a-user-grants-location-permissions-on-the-match-screen-return)


src/components/Match/MatchContainer.js
```diff
@@ -153,6 +144,36 @@ const MatchContainer = ( ) => {
     onlineSuggestionsAttempted
   } );
 
+  const getCurrentUserLocation = async ( ) => {
+    const currentUserLocation = await fetchUserLocation( );
```
jtklein, line 148: The result of this can be null. For example if the request for location errors out we will get null back here. In SuggestionsContainer there is also no reaction like for example an early return when this is null, so I am assuming it is fine. updateObservationKey is done there with null but updating the image params is done with optional chaining so best to copy from there: https://github.com/inaturalist/iNaturalistReactNative/blob/b43350730828b1b43a9a868a06d7008f791c20d1/src/components/Suggestions/SuggestionsContainer.tsx#L336-L337
 
Unclear to me what a good UI for an error state should be here, i.e. when null was obtained.

src/components/Match/MatchContainer.js
```diff
@@ -153,6 +144,36 @@ const MatchContainer = ( ) => {
...
+      longitude: currentUserLocation.longitude,
+      place_guess: placeName
+    } );
+    const newScoreImageParams = {
+      ...scoreImageParams,
+      lat: currentUserLocation.latitude,
+      lng: currentUserLocation.longitude
+    };
+    dispatch( {
+      type: "TOGGLE_LOCATION",
```
jtklein, line 162: Not really a toggle here just a one time update as there is no way to toggle it back off.

src/components/Match/MatchContainer.js
```diff
@@ -153,6 +144,36 @@ const MatchContainer = ( ) => {
     onlineSuggestionsAttempted
   } );
 
+  const getCurrentUserLocation = async ( ) => {
+    const currentUserLocation = await fetchUserLocation( );
+    const placeName
+     = await fetchPlaceName( currentUserLocation.latitude, currentUserLocation.longitude );
```
jtklein, line 150: There were some user bugs reported with updating this immediately as done here. If you compare the way it is done in ObsEdit: https://github.com/inaturalist/iNaturalistReactNative/blob/b43350730828b1b43a9a868a06d7008f791c20d1/src/components/ObsEdit/EvidenceSectionContainer.js#L148

Not sure if we need the same safeguards here. But maybe better to follow exactly what we are doing there.

jtklein (changes requested):
Few inline comments.
One thing I found not to spec is:
When I go to match screen, press Add location, and then deny permission the permission gate does not disappear but change to the location deny state instead. According to MOB-485 the permission gate should disappear and return to match screen without changes to UI there.

src/sharedHooks/useSuggestions/useSuggestions.js
```diff
@@ -39,6 +40,12 @@ export const useSuggestions = ( photoUri, options ) => {
     resetTimeout
   };
 
+  const refetchSuggestions = () => {
+    if ( shouldFetchOnlineSuggestions ) {
+      refetchOnlineSuggestions();
+    }
```
kvangork, line 46: This needs to refetch offline suggestions too

src/components/Match/MatchContainer.js
kvangork: Probably an alert would be sufficient error handling, but let's first wait to see if it's an actual problem.

PR #2705, opened 2025-03-01
Guard against combined_score being undefined

jtklein:
Closes MOB-513


kvangork:
I'd prefer to return null instead of zero. That causes a visible error when the score fields aren't present, and will make it easier to debug. Updated my [PR](https://github.com/inaturalist/iNaturalistReactNative/pull/2704) accordingly, and updated a comment there based on thoroughly reading all the online and offline paths for getting predictions.

PR #2707, opened 2025-03-03
Offline suggestions should have parity with online

kvangork:
Closes MOB-373

Changes to allow for matching usage of common ancestor from offline results. See atomic commit messages for intent details.

### DIFF
--- a/__mocks__/vision-camera-plugin-inatvision.ts
+++ b/__mocks__/vision-camera-plugin-inatvision.ts
@@ -1,4 +1,7 @@
-export const getPredictionsForImage = jest.fn( () => Promise.resolve( { predictions: [] } ) );
+export const getPredictionsForImage = jest.fn( () => Promise.resolve( {
+  predictions: [],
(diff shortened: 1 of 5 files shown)


albullington (approved):
It might be nice for us to leave a comment with some examples of what organisms we might expect to roll up to a common ancestor -- I tried quite a few organisms but didn't have a great way to visually test if this is working. Code-wise, this lgtm.

kvangork:
> It might be nice for us to leave a comment with some examples of what organisms we might expect to roll up to a common ancestor -- I tried quite a few organisms but didn't have a great way to visually test if this is working. Code-wise, this lgtm.

Good idea. I took it one step further: created an issue for the CV/ML team [asking for a set of test images](https://linear.app/inaturalist/issue/ML-236/create-a-set-of-test-images-which-should-always-get-a-common-ancestor) which are expected to produce common ancestors, to help us reduce uncertainty when testing our interface code.

PR #2709, opened 2025-03-03
Refactor match screen hooks / loading wheel

albullington:
Closes Mob-493

This is not perfect code but it's an upgrade from what we have, which hopefully makes things easier to read and debug. I focused on making sure:

- the loading screen UI is added on Match (not AICamera, as discussed with Kirk)
- users can skip online results on the loading screen
- offline/online are parallel processes
- store is in zustand so we can share it with Suggestions if/when needed instead of keeping logic in useReducers with the rest of the UI, which is hard to read
- a user can give permissions and receive new results after their location is added to an observation
- online suggestions still time out after 5 seconds / offline suggestions is a fallback when online suggestions aren't what we want to display to a user because of connectivity or user opt out


albullington:
@jtklein don't expect a super thorough review on this (it's a lot), but would appreciate if you take a look on a device to see if there's anything glaringly broken. and codewise, would be nice to double check that I incorporated all the changes from our team's recent vision / common ancestor / add location PRs.

jtklein (changes requested):
First, I don't know if it was intended but I like the fact that the loading state also comes up when turning location permission on.

I only got half-way through this today, my recommendation at this point would be to split it up into several PRs if possible. The new loading state could probably work also on the match screen as is on main if I am not mistaken.

I left two relevant screenshots/recordings in Slack.

I also encountered several crashes when pressing the shutter button, but I could not reproduce consistently. There are many confounding variables introduced now to track down where the crash results from I think, another reason to split into different PRs.

I am seeing this in the debug portion of the match screen directly coming from the camera but I doubt it's relevant to anything else than debugging.
Fetch status:null

Already left some inline comments in the code.

PR #2713, opened 2025-03-04
Remove initial block function from location permission gate

jtklein:
This was introduced because of a misunderstanding of requirements in a previous ticket. Now the permission gate behaves again the same on all screens.


jtklein:
Not ready. Merge after #2709 to avoid merge conflicts there.

kvangork:
What was the effect? What does this restore? Does it merit an issue, and is it severe enough to block the build?

PR #2714, opened 2025-03-05
Implement simple un-uploaded logged-out mode for ObsDetail

kvangork:
Closes MOB-512

Simple ObsDetail mode is activated only for logged-out users who are not using advanced settings (aka default mode), on observations which have not been uploaded.

### For Simple ObsDetail mode
- Removes the header, community discussion, details, and more sections.
- Swaps out the match-screen-style taxon display and replaces it with one showing the taxon photo, but with larger fonts per Figma.

### For all ObsDetail modes:
- Updates map height to match Figma.
- Fixes a crash when pressing on the taxon display when "Unknown" was shown.

### Todo before merge:
- [x] Update tests (they all passed, no updates needed)
- ~~[ ] Add new tests~~ We might delete this in the morning. I'm not going to spend any more time attempting to write tests for it until we do some user testing.

### Someone else's observation (reached by opening a link from outside the app or in Explore): 
IMG_8040

### DIFF
--- a/src/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeContainer.js
+++ b/src/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeContainer.js
@@ -11,6 +11,7 @@ import type { Node } from "react";
 import React, {
   useCallback,
   useEffect,
+  useMemo,
   useReducer,
   useState
 } from "react";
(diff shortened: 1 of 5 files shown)


src/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeContainer.js
```diff
@@ -180,6 +181,17 @@ const ObsDetailsDefaultModeContainer = ( ): Node => {
     || ( !observation?.user && !observation?.id )
   );
 
+  const isSimpleMode = useMemo( () => (
+    // Simple mode applies only when:
+    // 1. It's the current user's observation (or an observation being created)
+    // 2. AND the observation hasn't been synced yet
+    // 3. AND the user isn't logged in
+    ( belongsToCurrentUser || !observation?.user )
```
jtklein, line 189: I haven't looked into this condition any further (because the new screen mode works as expected) but having `belongsToCurrentUser` and `!currentUser` seems a weird combination just based on names.
With `!currentUser` I would expect `belongsToCurrentUser` to always be false just based on the name. Anyway, does not need to concern for this PR.

jtklein (approved):
Everything seems to work as expected. I did not encounter any crashes, tried going to ObsEdit from the new mode, to taxon details, media viewer, map, all still working. Taxon being "Unknown" also did look okay.

I made one change myself that I would otherwise have requested, so now I can approve instead.

src/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeContainer.js
kvangork: Agreed, it's terrible confusing naming, but it does seem to work correctly. Hopefully the comments help us when we revisit this while merging the two obsdetail implementations.

PR #2716, opened 2025-03-06
Convert advanced mode setting control from radio buttons to a boolean switch

kvangork:
Closes MOB-519

- Creates a new SwitchRow setting component appropriate for booleans.
- Uses SwitchRow to control Advanced Mode

## Screenshots
IMG_8066
IMG_8065


e2e/sharedFlows/switchPowerMode.js
```diff
@@ -10,10 +10,10 @@ export default async function switchPowerMode() {
   const settingsDrawerMenuItem = element( by.id( "settings" ) );
   await waitFor( settingsDrawerMenuItem ).toBeVisible().withTimeout( 10000 );
   await settingsDrawerMenuItem.tap();
-  // Tap the settings radio button for advanced interface mode
```
jtklein, line 13: Have you tested this? Pretty sure this will break e2e. Because a press on an already set advanced mode setting as is now changes nothing, bit a press on a radio does toggle every time.

src/i18n/strings.ftl
```diff
@@ -68,7 +68,7 @@ Add-optional-notes = Add optional notes
 Adds-your-vote-of-agreement = Adds your vote of agreement
 # Hint for a button that adds a vote of disagreement
 Adds-your-vote-of-disagreement = Adds your vote of disagreement
-Advanced--interface-mode-with-explainer = Advanced (Upload multiple photos and sounds)
+Advanced--interface-mode-with-explainer = Show Advanced Settings (Upload multiple photos and sounds)
```
jtklein, line 71: This doesn't seem right. The toggle not only reveals additional settings but also changes the UI of the app. In the absence of a design spec I would say to be more conservative with changes.

src/components/SharedComponents/SwitchRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <Pressable
+      className={classNames}
+      testID={testID}
+      accessibilityRole="button"
+      accessibilityState={{ checked: value, disabled }}
+      onPress={handlePress}
+      disabled={disabled}
+    >
+      <View className="flex-row items-center">
+        <Switch
```
jtklein, line 58: This works in my tests but in the absence of a design spec I would prefer to reuse already existing styled and unit tested components here. Probably best: src/components/SharedComponents/Checkbox.tsx

jtklein (changes requested):
The failing integration, unit, and e2e tests where still build for the assumption of advanced mode being turned off by default. They were built with a press on the advanced radio button as first step (which actually had no effect because advanced mode was already on). With these changes here this turned into a press on the switch which actually turned advanced mode off then. I updated all tests to pass now.

Two changes I, personally, would prefer but in the absence of designs for this I guess they might not be blockers after all.

src/i18n/strings.ftl
albullington: We should also never be editing existing translation keys. Anytime there is new text, it needs a new key, so this should read something like Show-Advanced-Settings--interface mode. If you change a key like this, translators will never be alerted that it changed so it will only ever be changed in English. And that's something that deeply aggravates our volunteer translator community.

src/components/SharedComponents/SwitchRow.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const Label = smallLabel
+    ? Body2
+    : Body1;
+
+  return (
+    <Pressable
+      className={classNames}
+      testID={testID}
+      accessibilityRole="button"
```
albullington, line 52: Afaik we're still trying to be diligent about adding accessibility hints and labels to buttons -- otherwise we can assume the app will be totally unusable for some percentage of users.

src/components/Settings/Settings.js
```diff
@@ -257,24 +258,13 @@ const Settings = ( ) => {
...
-          <View className="mt-4">
-            <RadioButtonRow
-              testID="advanced-interface-option"
-              smallLabel
-              checked={!isDefaultMode}
-              onPress={( ) => {
-                setIsDefaultMode( false );
-                setIsAllAddObsOptionsMode( true );
+              value={!isDefaultMode}
+              onValueChange={newValue => {
```
albullington, line 265: We're not diligent about doing this everywhere in our codebase, but in general in React it's preferred not to add functions in the render function, since this function will get recreated every time the component re-renders. My two cents is that these mild performance implications add up. One thing we've gotten feedback on in Seek and earlier versions of iNat Next is that it "doesn't feel native" because of performance, and I think that's something that will become more of a headache down the road if we don't think about it while we're coding now.

src/i18n/strings.ftl
kvangork: Thank you both. The label text needs improvement before this can ship, and I'll make a new key for the new text!

src/components/SharedComponents/SwitchRow.tsx
kvangork: Tested and confirmed this currently reads out the label text along with the state. It's imperfect, but usable.

src/components/SharedComponents/SwitchRow.tsx
kvangork: The checkbox isn't in conformance with either platform's interface guidelines for settings. Our SwitchRow component used in Settings could also benefit from unit tests, so I'll make an issue for them both.

PR #2722, opened 2025-03-06
Remove Announcements from default mode

jtklein:
Closes MOB-524


kvangork (approved):
I was simply going to delete line 208. Commenting it out with an explanation as you did is a better choice. Thanks @jtklein !

PR #2742, opened 2025-03-11
Retain user mode (advanced/default) after login

albullington:
Closes Mob-542


jtklein:
E2e were passing locally, so I merged it in.

PR #2764, opened 2025-03-14
fix location accuracy and using evidence location in match

angielt:
[MOB-523](https://linear.app/inaturalist/issue/MOB-523/fix-location-accuracy-and-using-evidence-location-in-match-screen)


jtklein (changes requested):
On a fresh app install I am seeing a crash when taking a photo and navigating to the Match screen:
from line 75 in MatchContainer.js `queryKey: setQueryKey( action.scoreImageParams.image.uri, action.shouldUseEvidenceLocation )`

 ERROR  TypeError: Cannot read property 'uri' of undefined

This error is located at:
    in MatchContainer (created by FadeInMatchContainer)
    
Is it one of the new useEffects firing before the imageParams are set?

PR #2776, opened 2025-03-16
MOB-514 - locally update total observation count when deleting/adding…

budowski:
… observations; update obs count from remote only once when app loads

### DIFF
--- a/src/components/MyObservations/MyObservationsContainer.js
+++ b/src/components/MyObservations/MyObservationsContainer.js
@@ -6,7 +6,7 @@ import { useFocusEffect, useRoute } from "@react-navigation/native";
 import { RealmContext } from "providers/contexts.ts";
 import type { Node } from "react";
 import React, {
-  useCallback,
+  useCallback, useEffect,
   useRef,
   useState
 } from "react";
@@ -28,7 +28,7 @@ import { isDebugMode } from "sharedHooks/useDebugMode";
 import {
   UPLOAD_PENDING
 } from "stores/createUploadObservationsSlice.ts";
-import useStore from "stores/useStore";
+import useStore, { zustandStorage } from "stores/useStore";
 
 import FullScreenActivityIndicator from "./FullScreenActivityIndicator";
 import useSyncObservations from "./hooks/useSyncObservations";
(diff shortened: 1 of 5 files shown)


src/components/MyObservations/MyObservationsContainer.js
```diff
@@ -101,6 +101,16 @@ const MyObservationsContainer = ( ): Node => {
       user_id: currentUserId
     }
   } );
+  const myObsLoaded = useStore( state => state.myObsLoaded );
+  const setMyObsLoaded = useStore( state => state.setMyObsLoaded );
+
+  useEffect( () => {
```
albullington, line 107: I'd love for us to avoid using `useEffect` as much as we can and be more declarative.

albullington (changes requested):
This isn't working the way I would hope. A few things:

1. When I upload an observation, I expect to see the observations number instantly increment. That's not happening, and I have to close the app and reopen it to see my new upload number.
2. When I delete an observation, I expect to see the observation number decrement. That did happen, but my species data didn't change, so I see two observations but three species with the species from my old observation still visible. This isn't a great user experience.

How can we write this so a user has an ideal experience and sees their data changing responsively?

budowski:
@albullington -
1. What flow are you using exactly to add an observation? Because the observation count does increase when I take a photo via the app.
2. Regarding the species count - I wrote this in Linear (MOB-514):
I think updating species counts this way should be in a separate ticket, since it's much more complicated to implement:
Since in most cases, for a user with a few hundred observations or more, this means we won't have all observations downloaded locally - so in order to accurately calculate if a newly added or removed observation changes the total taxon count, we'll need to somehow go over all observations (which are not locally downloaded), and then count the species there.

albullington:
Maybe @kvangork can weigh in, but I don't think we should add one of these bug fixes but not the other when we're this close to a launch date. I think the current user experience where this is buggy consistently across Observations and Species tab is less noticeable for users than a partial fix.

albullington:
I'm going to close this for now since there doesn't seem to be a great way to resolve the observation/species count being out of sync, and we haven't heard about this as a user concern since launch.

PR #2785, opened 2025-03-18
MOB-562 Common names are not translated until signup

sylvain-morin:


albullington (changes requested):
Nice contribution Sylvain! (And you even convinced me to put my phone into French this morning...)

This works well for me as a logged out user. But before merging, let's apply the same treatment to our [`useTaxon` hook](https://github.com/inaturalist/iNaturalistReactNative/blob/3aa40e1b27f31daa5f16c36e4f9fc46de2d5e632/src/sharedHooks/useTaxon.js#L20)? This is the code that shows the common name in the AICamera, so without this I'm seeing AICamera (English common name) followed by Suggestions (French common names) which is a bit of a strange user experience.

sylvain-morin:
Thanks @albullington for the remark and for testing!
I will have a look at this part.

sylvain-morin:
@albullington  could you please confirm:

`useTaxon` change will solve this :

image

'useOnlineSuggestions' change will solve this :

image

I suppose it will also solve this list of suggestions :

image

But what about this one ? when we click on "Learn More" button
This taxon page quotes wikipedia in english.

image

albullington:
Yes, `useTaxon` will solve the AICamera suggestion showing in device language while logged out... most of the time. The one edge case I'm seeing is if I switch device languages, we might end up with some common names stored locally in French and some in English (for example) for any of our code using `useTaxon`.

Did a bit more of an audit and also added this locale key to the API calls for Explore > SpeciesView, ObservationsView, ObsDetails, and TaxonDetails.

sylvainm0674:
Changing the device language is probably a rare edge case, to address later.

That's so great If new users can see common names in their own language! Thank you for helping and adding these other changes!

albullington (approved):

albullington:
Yw! And let us know if you end up finding other places in the app where we should internationalize common name that I missed...

PR #2786, opened 2025-03-18
Only show no  suggestions match screen UI without additional suggestions

jtklein:
So, in a case of having no top suggestion but low confidence additional suggestions we show all UI available for the match screen, so that a user can potentially also improve them by giving location permission.

Tested with the first photo here: 
https://linear.app/inaturalist/issue/MOB-600/match-screen-special-handling-of-potential-dual-species-situations


albullington (approved):
Yeah, that works for me. Maybe worth a comment? I'm not sure we're going to remember why we're doing this special conditioning in the future.

PR #2788, opened 2025-03-19
Make MyObservationsSimple the standard UI across default/advanced mode

albullington:
Closes Mob-587

Deletes outdated MyObservations code while preserving the differences between these components. Things that are different:

In default mode:
- Pivot cards are shown
- Missing evidence note is shown
- No obs status is shown in list view
- Observations without evidence cannot be uploaded
- Upload progress bar is hidden with less than 2 observations


src/realmModels/Observation.js
```diff
@@ -460,10 +460,32 @@ class Observation extends Realm.Object {
...
+      } );
+    }
+  }
+
   static schema = {
     name: "Observation",
     primaryKey: "uuid",
     properties: {
+      _pending_deletion: "bool?",
+      _deletion_attempted_at: "date?",
```
jtklein, line 488: This key does not seem to be used anywhere `_deletion_attempted_at`.

src/sharedHooks/useLocalObservations.js
```diff
@@ -31,7 +31,9 @@ const useLocalObservations = ( ): Object => {
         [["needs_sync", true], ["_created_at", true]]
       );
 
-      const obsNotFlaggedForDeletion = sortedCollection.filtered( "_deleted_at == nil" );
+      // eslint-disable-next-line max-len
+      const deletionFilters = "_deleted_at == nil OR _pending_deletion == false OR _pending_deletion == null";
```
jtklein, line 35: The "null" needs to be "nil" here, right?
Also, does the order matter for the realm ORs, should == nil be before == false, or it doesn't matter?

src/components/MyObservations/SimpleUploadBannerContainer.js
```diff
@@ -176,7 +180,7 @@ const SimpleUploadBannerContainer = ( {
 
   // hide when there are less than two observations to upload
   // but make sure completed status is displayed when uploads are finished
-  if ( numOfUserObservations < 2 && progress === 0 ) {
+  if ( isDefaultMode && numOfUserObservations < 2 && progress === 0 ) {
```
jtklein, line 183: From what I can see in the designs, I think this check here should not be for default mode but for currentUser.
 In these screens here in the column for all users the number shown is 1:
 https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=23655-68746&t=pMuk9C20LY8bFrMb-4
 https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=23655-73876&t=pMuk9C20LY8bFrMb-4
 Only for logged out users the number 1 is not shown:
 https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=23655-66337&t=pMuk9C20LY8bFrMb-4

src/components/MyObservations/SimpleErrorHeader.tsx
```diff
@@ -17,7 +17,12 @@ export interface Props {
 const SimpleErrorHeader = ( {
   isConnected
 }: Props ) => {
+  const { isDefaultMode } = useLayoutPrefs( );
   const { t } = useTranslation( );
+
+  if ( !isDefaultMode ) {
```
jtklein, line 23: From the designs, I think this is also supposed to be shown in advanced mode:
https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=23663-77654&t=pMuk9C20LY8bFrMb-4

jtklein (changes requested):
Mostly looking good.

I wanted to get this review over the finish line today, so I have not looked at:

Changes in tests.
I did not check the deleted files.
I did not check if there are some things in the codebase that were only referenced in one of the deleted files but are not deleted and now possibly obsolete.

src/sharedHooks/useLocalObservations.js
albullington: good call on nil. i believe filtering in Realm with ORs is purely filtering and will remain in the same order as the code above:
```
const sortedCollection = collection.sorted(
        [["needs_sync", true], ["_created_at", true]]
      );
```

PR #2790, opened 2025-03-21
Add a UI for when there is no top suggestion but other suggestions with low confidence on match screen

jtklein:
Closes MOB-611

All new design items are being added. The new UI is only to be shown when there is no top suggestion but other suggestions are present.

I am seeing some flickering in the additional suggestions list, but that is already present on current main.

### DIFF
--- a/src/components/Match/AdditionalSuggestions/AdditionalSuggestionsScroll.js
+++ b/src/components/Match/AdditionalSuggestions/AdditionalSuggestionsScroll.js
@@ -9,6 +9,7 @@ import { useTranslation } from "sharedHooks";
 import SuggestionsResult from "./SuggestionsResult";
 
 const AdditionalSuggestionsScroll = ( {
+  noTopSuggestion,
   otherSuggestions,
   suggestionsLoading,
   onSuggestionChosen
@@ -116,7 +117,11 @@ const AdditionalSuggestionsScroll = ( {
 
   return (
     <View className="mt-4 mb-7">
-      <Heading3 className="mx-5 mb-3">{t( "It-might-also-be" )}</Heading3>
+      <Heading3 className="mx-5 mb-3">
+        {noTopSuggestion
+          ? t( "It-might-be-one-of-these" )
+          : t( "It-might-also-be" )}
+      </Heading3>
 
       {!suggestionsLoading
         ? (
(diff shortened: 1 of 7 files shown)


src/components/Match/Match.js
```diff
@@ -78,20 +82,69 @@ const Match = ( {
...
+      <>
+        <ScrollViewWrapper scrollRef={scrollRef}>
+          <View className={cardClassTop}>
+            {
+              suggestionsLoading
+                ? (
+                  <ActivityIndicator size={33} />
+                )
+                : (
+                  <Body2>
```
albullington, line 118: It looks like the styling of this is off just a touch from Figma. In Figma the padding above and below this text is 18px, and here it appears to be 16px.

src/components/Match/AdditionalSuggestions/AdditionalSuggestionsScroll.js
```diff
@@ -116,7 +117,11 @@ const AdditionalSuggestionsScroll = ( {
 
   return (
     <View className="mt-4 mb-7">
-      <Heading3 className="mx-5 mb-3">{t( "It-might-also-be" )}</Heading3>
+      <Heading3 className="mx-5 mb-3">
```
albullington, line 120: Also looks like the styling is off a touch here. In Figma there are 20px above this text and 15px below, while here it's 12px below and I believe 24px above (mt-4 is 16px and there's all the pt-2 of 8px on the parent View of this component).

src/components/Match/Match.js
```diff
@@ -78,20 +82,69 @@ const Match = ( {
...
+              noTopSuggestion
+              onSuggestionChosen={onSuggestionChosen}
+              otherSuggestions={otherSuggestions}
+              suggestionsLoading={suggestionsLoading}
+            />
+            {!latitude && (
+              <Button
+                className="mx-5 mb-7"
+                level="neutral"
+                text={t( "ADD-LOCATION-FOR-BETTER-IDS" )}
```
albullington, line 141: I tried tapping ADD LOCATION on the Ivy photo that was shared in the [dual-species issue](https://linear.app/inaturalist/issue/MOB-600/match-screen-special-handling-of-potential-dual-species-situations), but it didn't appear to do anything. Tapping the button shows me the same not confident screen and the ADD LOCATION button is still visible. It does pop up the location permission gate but I don't think it's fetching new suggestions after that.

albullington (changes requested):

PR #2793, opened 2025-03-21
Add long press to AICamera gradient button in default/advanced

albullington:
Closes Mob-614

### DIFF
--- a/src/components/AddObsModal/AddObsButton.js
+++ b/src/components/AddObsModal/AddObsButton.js
@@ -77,6 +77,7 @@ const AddObsButton = (): React.Node => {
       />
       <GradientButton
         sizeClassName="w-[69px] h-[69px] mb-[5px]"
+        onLongPress={!isAllAddObsOptionsMode && openModal}
         onPress={isAllAddObsOptionsMode
           ? openModal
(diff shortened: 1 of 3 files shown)


src/components/AddObsModal/AddObsButton.js
```diff
@@ -77,6 +77,7 @@ const AddObsButton = (): React.Node => {
       />
       <GradientButton
         sizeClassName="w-[69px] h-[69px] mb-[5px]"
+        onLongPress={!isAllAddObsOptionsMode && openModal}
```
jtklein, line 80: Have you considered always using `openModal` here? Then the prop wouldn't need to be optional.

jtklein (approved):
Looking good.

I am wondering how this will be read out with Voice Over, but that can be it's separate issue down the line.

src/components/AddObsModal/AddObsButton.js
albullington: I considered this, but the way the issue is written we don't need the long press for the case where there's a plus button since the short press opens the modal. I imagine we'll end up tweaking this and maybe making the long and short press do the same thing for the plus button eventually.

PR #2795, opened 2025-03-24
Add pivot card for the tenth observation

jtklein:
Closes MOB-615


src/components/Developer/UiLibrary/PivotCardsDemo.js
```diff
@@ -3,7 +3,8 @@ import {
   FiftyObservationCard,
   FirstObservationCard,
   NotificationOnboarding,
-  SecondObservationCard
+  SecondObservationCard,
+  TenObservationCard
```
albullington, line 7: same naming feedback -- either `Ten` or `Tenth` is good as long as we're consistent with the other pivot cards.

src/i18n/l10n/en.ftl
```diff
@@ -731,6 +731,7 @@ MONTHS = MONTHS
 # Label for section of ObsDetails with information about observation projects, DQA, etc.
 More = More
 More-info = More info
+More-ways-to-observe = More ways to observe
```
albullington, line 734: This could use a comment for translators, something like `# Header for pivot card explaining long press functionality to users`

albullington (changes requested):
Needs the context comment for translators (like Ken-ichi requested [here](https://inaturalist.slack.com/archives/C02G5AATVGA/p1742242878123579)) and potential renaming -- otherwise works for me and good to go.

PR #2796, opened 2025-03-24
Change pivot card timing from on second to fifth observation

jtklein:
Closes MOB-616

### DIFF
--- a/src/i18n/strings.ftl
+++ b/src/i18n/strings.ftl
@@ -1542,7 +1542,7 @@ Youre-always-in-control-of-the-location-privacy = You’re always in control of 
 # Text prompting the user to open Settings to grant permission after
 # permission has been denied
 Youve-denied-permission-prompt = You’ve denied permission. Please grant permission in the settings app.
-Youve-made-2-observations = You’ve made 2 observations!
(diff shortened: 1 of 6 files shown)


src/i18n/strings.ftl
```diff
@@ -1542,7 +1542,7 @@ Youre-always-in-control-of-the-location-privacy = You’re always in control of
 # Text prompting the user to open Settings to grant permission after
 # permission has been denied
 Youve-denied-permission-prompt = You’ve denied permission. Please grant permission in the settings app.
-Youve-made-2-observations = You’ve made 2 observations!
+Youve-made-5-observations = You’ve made 5 observations!
```
albullington, line 1545: Is there ever a reason we would need to use the `{ $count }` plural rules here? If we did, it would allow us to change our minds on when this pivot card gets triggered without requiring new translations in all languages.

albullington (approved):
lgtm and working as expected on device.

realizing naming-wise it would make sense to have all of these cards named for either ordinal numbers or regular numbers, i.e. `FirstObservation`, `FifthObservation` and `FiftiethObservation` or `OneObservation` `FiveObservation` `FiftyObservation`, but feel free to ignore that nitpick.

PR #2801, opened 2025-03-25
Show a simple unviewed-activity-indicator for default obs list

jtklein:
Closes MOB-633


src/components/ObservationsFlashList/ObsListItem.js
```diff
@@ -33,7 +33,8 @@ type Props = {
   uploadProgress?: number,
   unsynced: boolean,
   hideObsUploadStatus?: boolean,
-  hideObsStatus?: boolean
+  hideObsStatus?: boolean,
+  simpleObsStatus?: boolean
```
albullington, line 37: since `simpleObsStatus` is a boolean, can we use one of the naming conventions for booleans? it's generally best to name them in a way that conveys true/false, like `isSimpleObsStatus`, `showSimpleObsStatus`, `displaySimpleObsStatus`.

albullington (approved):

PR #2804, opened 2025-03-27
Mob 596 sharing one image at a time in default mode shows outdated

angielt:


albullington (changes requested):
@angielt fyi some merge conflicts here related to the Advanced Settings PR.

Question: is `useEffect` the only way to fix this? I was hoping using `resetObservationFlowSlice` in the PhotoSharing app would help with this problem originally -- it would be great to make sure Match/Suggestions are reset completely before even landing on those screens, rather than after.

If we really have to fix this on the screen, is there a way we could use a unique queryKey for react query rather than a `useEffect`? Imo effects are hard to read and debug, and they come with some performance cost.

I'm also seeing some `console.logs` that would be helpful to clean up.

angielt:
@albullington i'm fixing the merge conflicts now. With the changes to advanced settings it seems the `useEffect` solution for `SuggestionsContainer` doesn't work as intended anymore so I'm open to new ideas on how to tackle this. It seems like the selectedPhotoUri and query key are the issues. In the PhotoSharing `useEffect`, `resetObservationFlowSlice` is already called before navigating

albullington:
@angielt hmm if the goal is to fetch a new query instead of showing stale data, maybe try taking a look at [queryClient.invalidateQueries](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientinvalidatequeries)? we're using that a few places in the app (Announcements, LocationSearch, etc.) so you can see how it works if you haven't used it before -- but tldr is that it should clear the previously cached queries. not sure where the right place for it would be, but you could try calling it in PhotoSharing before navigating which might be helpful since it would work for both Suggestions/Match (if you try this, add a comment about why it's necessary because i'm sure we'll forget), or when `navigation.addListener( "focus" )` is called on Suggestions and Match -- that way it's handled immediately when the screen opens

angielt:
@albullington I decided to reset the navigation state. I noticed a bug where someone was in default mode and shared a photo in the app and then switched to advanced  mode to share a photo(or vice versa), and then pressed back, it would show the previous mode's screen instead.

albullington (approved):

PR #2809, opened 2025-03-28
MOB-589 - call iNat /signout endpoint when signing out (or before signing in)

budowski:


angielt (changes requested):
@budowski  I'm still seeing the same behavior as in the issue. When I create an account, sign out of the new account, and log back in to another account, I'm still signed into the account i created.

budowski:
@angielt - can you send logs and a screen recording of this? I just tried the same scenario (registering a new account via the app, signing out, then signing in to another account) - and it worked fine...

angielt:
@budowski I tested using a release build and I see it working now. Just to make sure could either @albullington or @jtklein test this as well?

albullington:
Tested in a release build (create an account, sign out of the new account, and log back in to another account) and this is working for me. Fwiw I signed in with Google, and I made an observation on the new account before logging out for maybe slightly different repro steps.

PR #2820, opened 2025-04-01
Fix: limit API calls in AICamera to avoid delay in advancing navigation

albullington:
Closes Mob-635

I wasn't able to reproduce this, so taking a multi-pronged approach to avoid lengthy network requests here:

- Make sure we're not accidentally calling the API for iconic taxon whenever there's a new prediction. We fetch that data to store in Realm on app startup, so we can get it directly from Realm instead of the API.
- Set no retries for each TaxonResult (prediction) in the AICamera, since this API is being fetched every second anyway, and it seems wise to make sure these API calls are not stacking up (3 retries each) in low connectivity conditions
- Set a 2.5 second timeout for the geocoded observation place name

### DIFF
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -277,6 +277,10 @@ const AICamera = ( {
                 taxon={result?.taxon}
                 testID={`AICamera.taxa.${result?.taxon?.id}`}
                 white
+                // my thinking here is that we're already making this API call over and over
+                // every second when a prediction comes back, and we likely don't want to
+                // 3x that in low network conditions if every request is failing
+                retryQuery={false}
               />
             )
             : (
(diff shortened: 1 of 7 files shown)


albullington:
FYI for reviewers -- I made a release build outside and tested on 1 bar and 2 bars of cell service with location permission off, then on. The AICamera delay seems a lot quicker, which makes the Match screen loading screen feel slower by comparison. I think this PR will illuminate some other bugs -- occasional times when the loading wheel stops while the camera preview is still frozen, and occasional times when the Match screen loading state seems to reload twice. But hopefully a net benefit and helpful to see what else needs to be fixed.

angielt (approved):
Looks good. I tried using it with one bar and using the AI camera feels a lot better with the changes. Suggestions does seem to take longer in comparison but it is more expected on that screen than the camera.

PR #2827, opened 2025-04-03
multicapture camera photos UI is not accounting for iphone

angielt:
Closes MOB-612


albullington (changes requested):
IMG_9070

here's what i'm seeing on iPhone 15 Pro. I think there are going to be different notch/island sizes on different devices that we'll need to account for. did you try the SafeAreaView wrapper here to see if we can do this with less pixel pushing?

albullington (approved):
This works for me and does not seem to distort the camera preview on an iPhone 15 Pro. I can't test what this looks like on a phone without a notch, but I'm going to merge with the following assumptions:

1. We want to prioritize UI for newer devices, which all have notches/islands.
2. The majority of our iPhone users have these newer devices, so we'll be making the experience better for the bulk of our users either way.

PR #2829, opened 2025-04-04
Fix: revert upload code

albullington:
Another attempt to close MOB-639.

After a lot of looking at our upload code, I realized I didn't take my own advice a few weeks ago to not touch the upload code. I'm now suspicious that [this PR](https://github.com/inaturalist/iNaturalistReactNative/pull/2783) to try to handle jwt errors is actually what recently made bulk uploads spotty. Before that, we hadn't touched that code since November, so I put this upload code back to the state it was in before.

With this reversion, I made a local release build. Unlike the current build from `main` (or in TestFlight) I can now background the app, come back, and see my uploads continue as expected with no photos dropped.

Would be great to review & merge this if it's working for others before the next public release.

### DIFF
--- a/src/realmModels/ObservationPhoto.js
+++ b/src/realmModels/ObservationPhoto.js
@@ -1,11 +1,8 @@
 import { Realm } from "@realm/react";
 import inatjs, { FileUpload } from "inaturalistjs";
-import { log } from "sharedHelpers/logger";
 import * as uuid from "uuid";
 
 import Photo from "./Photo";
-
-const logger = log.extend( "Realm ObservationPhoto" );
 
 class ObservationPhoto extends Realm.Object {
   static OBSERVATION_PHOTOS_FIELDS = {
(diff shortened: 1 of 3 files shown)


kvangork:
Looks fine based on a quick read-through, but I'm too sleepy to be confident. I'm going to come back for a deeper review in the morning, and my hope is that we get it merged and into a TestFlight build for testing over the weekend.

kvangork (approved):
I finished reading this again, and also the original PR. Thank you @Amanda for linking it in the description, that was very helpful context.

I think we should merge this as-is and test it internally over the weekend. I have a bunch of saved observations ready to use in testing bulk uploads. We should meet and review the results on Monday to evaluate which version we're most comfortable including in the RC build.

Edit: I also reviewed the test failures. They look like runner problems, rather than test failures caused by this revert. The updated state of this code was previously tested and I am comfortable merging it without trying to fix the test runners.

PR #2832, opened 2025-04-04
Populate taxon in AICamera -> ObsEdit with AICamera suggestion

albullington:
Closes MOB-654

### DIFF
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -171,11 +171,18 @@ const AICamera = ( {
   const handleTakePhoto = useCallback( async ( ) => {
     await logStage( sentinelFileName, "take_photo_start" );
     setHasTakenPhoto( true );
+    // this feels a little duplicative, but we're currently using aICameraSuggestion
+    // to show the loading screen in Suggestions *without* setting an observation.taxon,
+    // and we're using visionResult to populate ObsEdit *with* the taxon
+    // before aICameraSuggestion has finished being stored.
+    // would be nice to refactor and set this more uniformly once the UX is more stable
+    // and we're fully certain we don't want to populate observation.taxon on Suggestions -> ObsEdit
     setAICameraSuggestion( result );
     await takePhotoAndStoreUri( {
       replaceExisting: true,
       inactivateCallback: () => setInactive( true ),
-      navigateImmediately: true
+      navigateImmediately: true,
+      visionResult: result
     } );
     setHasTakenPhoto( false );
   }, [
--- a/src/components/Camera/CameraContainer.tsx
+++ b/src/components/Camera/CameraContainer.tsx
@@ -138,12 +138,13 @@ const CameraContainer = ( ) => {
   // passing newPhotoState because navigation to SuggestionsContainer for AICamera
   // happens before cameraUris state is ever set in useStore
   // and we want to make sure Suggestions has the correct observationPhotos
-  const handleNavigation = useCallback( async ( newPhotoState = {} ) => {
+  const handleNavigation = useCallback( async ( newPhotoState, visionResult ) => {
     await prepareStoreAndNavigate( {
       ...navigationOptions,
       newPhotoState,
       logStageIfAICamera,
-      deleteStageIfAICamera
+      deleteStageIfAICamera,
+      visionResult
     } );
   }, [
     prepareStoreAndNavigate,
@@ -152,9 +153,9 @@ const CameraContainer = ( ) => {
     deleteStageIfAICamera
   ] );
 
-  const handleCheckmarkPress = useCallback( async newPhotoState => {
+  const handleCheckmarkPress = useCallback( async ( newPhotoState, visionResult ) => {
     if ( !showPhotoPermissionsGate ) {
-      await handleNavigation( newPhotoState );
+      await handleNavigation( newPhotoState, visionResult );
     } else {
       await logStageIfAICamera( "request_save_photo_permission_start" );
       requestSavePhotoPermission( );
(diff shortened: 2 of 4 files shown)


src/components/Camera/hooks/usePrepareStoreAndNavigate.ts
```diff
@@ -93,14 +95,25 @@ const usePrepareStoreAndNavigate = ( ): Function => {
         position: 0,
         local: true
       } );
+    if ( !isDefaultMode
+        && screenAfterPhotoEvidence === SCREEN_AFTER_PHOTO_EVIDENCE.OBS_EDIT
+        && visionResult ) {
+      console.log( visionResult.taxon, "ai camera suggestion" );
```
jtklein, line 101: Remove log

src/components/Camera/AICamera/AICamera.js
```diff
@@ -171,14 +171,22 @@ const AICamera = ( {
...
     await logStage( sentinelFileName, "take_photo_start" );
     setHasTakenPhoto( true );
-    setAICameraSuggestion( result );
+    // this feels a little duplicative, but we're currently using aICameraSuggestion
+    // to show the loading screen in Suggestions *without* setting an observation.taxon,
+    // and we're using visionResult to populate ObsEdit *with* the taxon
+    // before aICameraSuggestion has finished being stored.
+    // would be nice to refactor and set this more uniformly once the UX is more stable
+    // and we're fully certain we don't want to populate observation.taxon on Suggestions -> ObsEdit
+    setAICameraSuggestion( showPrediction && result );
```
jtklein, line 180: I think we should not mix types here as long as useStore is returning the `any` type for all state. It makes it more predictable to use aiCameraSuggestion from state in knowing that it can only be of one type. So, eg if showPrediction is false we store a boolean and otherwise a result object (or null). So, here if I am not mistaken we could change to `showPrediction && setAICameraSuggestion( result );` instead.

src/components/Camera/CameraContainer.tsx
```diff
@@ -138,12 +138,13 @@ const CameraContainer = ( ) => {
   // passing newPhotoState because navigation to SuggestionsContainer for AICamera
   // happens before cameraUris state is ever set in useStore
   // and we want to make sure Suggestions has the correct observationPhotos
-  const handleNavigation = useCallback( async ( newPhotoState = {} ) => {
+  const handleNavigation = useCallback( async ( newPhotoState, visionResult ) => {
```
jtklein, line 141: Adding types here would help making it clear what is expected to be used here.

src/components/Camera/CameraContainer.tsx
```diff
@@ -152,9 +153,9 @@ const CameraContainer = ( ) => {
     deleteStageIfAICamera
   ] );
 
-  const handleCheckmarkPress = useCallback( async newPhotoState => {
+  const handleCheckmarkPress = useCallback( async ( newPhotoState, visionResult ) => {
```
jtklein, line 156: Adding types here would help make it clear what is expected to be used here.

src/components/Camera/CameraContainer.tsx
```diff
@@ -219,7 +220,7 @@ const CameraContainer = ( ) => {
     const newPhotoState = await updateTakePhotoStore( uri, options );
     setTakingPhoto( false );
     if ( options?.navigateImmediately ) {
-      await handleCheckmarkPress( newPhotoState );
+      await handleCheckmarkPress( newPhotoState, options?.visionResult );
```
jtklein, line 223: From how it is used in AICamera `options?.visionResult` can be `false` or an actual `StoredResult` I think it would be clearer if we only make it `StoredResult | null`. Null check in usePrepareStore is still covering the null case from what I can see. So, no other changes required I think then to change `showPrediction && result` to only send up result when showPrediction is true.

src/components/Camera/AICamera/AICamera.js
```diff
@@ -171,14 +171,22 @@ const AICamera = ( {
...
     await takePhotoAndStoreUri( {
       replaceExisting: true,
       inactivateCallback: () => setInactive( true ),
-      navigateImmediately: true
+      navigateImmediately: true,
+      visionResult: showPrediction && result
```
jtklein, line 185: Since `result` can be `StoredResult | null` I would change here to `showPrediction ? result : null` in that way we don't send up a boolean (= false).

jtklein (changes requested):
I would call it "Approve with minor changes".
The only change I would like to see is to not mix the types of Boolean and StoredResult. The way it is now the visionResult can be "false" but can also be a "StoredResult" object. This propagates through to the store.aiCameraSuggestion which is not typed yet. So, to avoid developer to developer miscommunication I think we should restrict ourselves to one type here. So, to remove the possibility of this state to become = false.

src/components/Suggestions/SuggestionsLoading.tsx
```diff
@@ -21,9 +21,6 @@ const SuggestionsLoading = ( {
   const { t } = useTranslation( );
   const aiCameraSuggestion = useStore( state => state.aICameraSuggestion );
 
-  const showPrediction = ( aiCameraSuggestion && aiCameraSuggestion?.taxon?.rank_level <= 40 )
```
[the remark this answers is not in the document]
albullington: This changed to make sure we're showing these predictions in a uniform way -- I thought it made more sense to move in this into a single location (the AICamera) now that we need to show the AI Camera suggestion on multiple screens / UX flows. So I think of it as less of a change and more of a standardization. We *could* add this logic in a different place for the AICamera -> ObsEdit flow and duplicate logic, but I prefer consistency and only needing to tackle this in a single place for all of our UX flows.

src/components/Camera/hooks/usePrepareStoreAndNavigate.ts
albullington: Once upon a time I had a precommit hook to make sure we didn't accidentally commit console.logs...

PR #2843, opened 2025-04-09
Add error context to crash reporting

albullington:
For better or worse we haven't gotten the 429 errors in Grafana for build 0.59.16 yet that would confirm that this is working in a production environment, but we can simulate that it's working in <Developer /> by pressing the relevant 429 error button. This is intended to help us understand what a user was doing in the app before they receive a 429 error.

### DIFF
--- a/src/api/error.js
+++ b/src/api/error.js
@@ -11,10 +11,14 @@ export class INatApiError extends Error {
   // HTTP status code of the server response
   status: number;
 
-  constructor( json: Object, status?: number ) {
+  // Additional context information
+  context: ?Object;
+
+  constructor( json: Object, status?: number, context?: Object ) {
     super( JSON.stringify( json ) );
     this.json = json;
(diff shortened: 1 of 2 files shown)


jtklein (approved):
I was not able to trigger the error in the StartupService.js file within a timebox, but I am also sure that the change there is not a regression.

I am seeing some context in the server logs when triggered from logging.js, e.g. 
`429 in reactQueryRetry: {"queryKey":"[\"fetchTaxon\",48623]","failureCount":2,"timestamp":"2025-04-10T17:41:27.779Z","errorType":"INatApiTooManyRequestsError","status":429}`.
I am not seeing context in the server logs when triggered from index.js
`{"error":"Too Many Requests","status":429,"context":null}`.

My current summary is that this needs more work to be helpful to find patterns of usage that trigger 429 errors.
But I don't see any harmfull regressions here so I Approve.

PR #2844, opened 2025-04-09
Prevent multiple taps on all Button components

albullington:
Adds a brief debounce to `<Button />` to make sure the user cannot press the button in rapid succession and thereby get lost in a navigational time warp. (Like needing to navigate back through multiple stacked versions of Suggestions.)


albullington:
@jtklein removed the unnecessary `useEffect` to avoid performance concerns. Otherwise the same.

albullington:
also changed the branch to merge this into to `main` because i think we decided on that process going forward. tbd.

jtklein (approved):
Changes are good to merge into main.
I am still not sure why we have a useRef that holds the onPress and not just use it directly.

PR #2845, opened 2025-04-09
Display onboarding carousel as react-navigation modal

albullington:
This changes the way we display the OnboardingCarousel. By using react-navigation modal, we eliminate the glitchiness in the screen transition that was present when we needed to first navigate to MyObservations, then check for whether the onboarding screen had previously been displayed, then open the modal. This conditionally displays either Onboarding or MyObservations depending on whether Onboarding has been seen before.


kvangork:
@jtklein regarding your question "I am not sure it is at a stage were I am comfortable with as the first impression to new users though. There is still a flash of white between the two instances of the splash screen." 
Is that a regression from the behavior before this PR? If not, I think we should proceed to merge.

albullington:
afaik (for the reasons mentioned [here](https://linear.app/inaturalist/issue/MOB-606/opening-screen-transition-timing#comment-508d6646)) the white flash is a separate issue related to how the entire React Native bundle loads, and it wasn't something i ever figured out how to fix while working on Seek. maybe something to revisit after Earth Day.

PR #2848, opened 2025-04-09
Ai camera shutter button tappable after first press

angielt:


ios/iNaturalistReactNative.xcodeproj/xcshareddata/xcschemes/iNaturalistReactNative.xcscheme
```diff
@@ -59,7 +59,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
```
albullington, line 62: Let's not change this in this PR.

src/components/Camera/Buttons/TakePhoto.tsx
```diff
@@ -21,12 +22,39 @@ interface Props {
...
+
+    setTimeout( ( ) => {
+      setIsProcessing( false );
+    }, debounceTime );
+  };
+
+  const isDisabled = disabled || ( preventMultipleTaps && isProcessing );
+
   return (
     <Pressable
```
albullington, line 59: Being a tiny bit forward thinking here, can we move this logic into a SharedComponent, maybe called `PressableWithDebounce`? I'm assuming a few things here:

1) we'll want this logic on other buttons throughout the app (maybe even in the camera, on the gallery button for example)
2) it would be nice to be able to quickly swap out a `Pressable` for a `PressableWithDebounce` in case we run into any timing issues with this (here or elsewhere)
3) it's nicer to standardize the debounce time in a single location, rather than making `TakePhoto` responsible for receiving props and knowing how long to debounce

albullington (changes requested):
I unfortunately can't test that this is working, since I haven't been able to tap more than once on the shutter button in the AICamera in the first place. But I still can't tap more than once in this branch, and I stuck with a code-related review.

src/components/SharedComponents/Buttons/PressableWithDebounce.js
```diff
@@ -0,0 +1,77 @@
...
+  onPress,
+  disabled,
+  children,
+  debounceTime = 400,
+  preventMultipleTaps = true
+}: Props ) => {
+  const [isProcessing, setIsProcessing] = useState( false );
+  const onPressRef = useRef( onPress );
+
+  useEffect( ( ) => {
```
albullington, line 37: One small change requested here -- can you try this without the `useEffect`? so just directly set `onPressRef.current = onPress` underneath `onPressRef`. That worked for our related PR here, and should alleviate Johannes' performance concerns: https://github.com/inaturalist/iNaturalistReactNative/pull/2844/files

albullington (changes requested):

albullington:
I'm reassigning this to @jtklein for review since I could never reproduce this bug on my device in the first place -- would be good to see if the issue is actually resolved.

jtklein (changes requested):
I think you are on the right track in terms of how to best control the state of disabling the camera UI while a photo is being processed. I do think the `takingPhoto` state that we already have is a better signal for controlling this behaviour. I would suggest we update the `setTakingPhoto` calls to incorporate the changes you have suggested here. I made a PR to lift your changes up to the `CameraContainer` to update the `takingPhoto` here:   https://github.com/inaturalist/iNaturalistReactNative/pull/2899
Also, I have copied your approach of disabling photo library to the tablet UI.
From my perspective, I would say: if you merge this PR into yours, I would Approve.

jtklein (approved):
Looking good now.

albullington:
> Looking good now. @albullington your requested changes are also done?

Yep!

PR #2849, opened 2025-04-10
Rename iNat Next to iNaturalist

kvangork:
Closes MOB-622
Fixes MOB-672

Forces the rename throughout the untranslatable elements of xcstrings rather than waiting for crowdin to do the same thing.

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,4 +1,4 @@
 {
   "name": "iNaturalistReactNative",
-  "displayName": "iNat Next"
+  "displayName": "iNaturalist"
 }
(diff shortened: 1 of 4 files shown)


albullington (changes requested):
Permissions now say iNaturalist & the name of the app on my iPhone home screen is iNaturalist. Requesting changes because:

1. I'm not sure why the Tagalog translations were dropped from `InfoPlist.xcstrings`.
2. I believe we also need to change `fastlane/metadata/ios/default/name.text` for the name to change in the App Store when we're building with Fastlane.

albullington (approved):

PR #2850, opened 2025-04-10
Cherry pick mob-622-rename-inat-next-to-inaturalist

kvangork:
Rename iNat Next to iNaturalist

### DIFF
--- a/app.json
+++ b/app.json
@@ -1,4 +1,4 @@
 {
   "name": "iNaturalistReactNative",
-  "displayName": "iNat Next"
+  "displayName": "iNaturalist"
 }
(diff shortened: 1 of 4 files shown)


jtklein (approved):
I can approve this in a sense that on installing the app I find everything to be with the correct name. I checked and `fastlane ios build` and e2e locally is also still running.
I can see some outdated references in the Maestro e2e tests that are still looking to start the "iNaturalistReactNative" app. We have never hooked those up into the CI though so, not a blocker I would say.
I can also see the name iNaturalistReactNative still as a label in this file here but I think it is not even used in the app anywhere: https://github.com/inaturalist/iNaturalistReactNative/blob/0f50d4564c39e01fbc8a92762ccfbe893b14a8cc/ios/Launch%20Screen.storyboard#L24-L25
I am not versed enough in native iOS to understand the InfoPlist.xcstrings changes. So, I can only assume that a change from "new" to "needs_review" is correct.

PR #2851, opened 2025-04-10
multicapture camera photos UI is not accounting for iphone (#2827)

angielt:
* Increase space above in multicapture camera to account for notch and dynamic island

* Adjust heights for dynamic island

* Change to using safeareaview
[MOB-612](https://linear.app/inaturalist/issue/MOB-612/multicapture-camera-photos-ui-is-not-accounting-for-iphone-notch)


jtklein (commented):
The `Podfile.lock` change is not really needed for the described UI change. I can see that it was in the cherry-picked PR where it probably should not have been.
By including it here whoever will build the app (@kvangork) for release probably needs to have not-committed changes to this file in order to build from this branch.
As long as we are aware of this and okay with it I can Approve.

The UI change is working as expected though.

PR #2842, opened 2025-04-09
New Crowdin Translations

kueda:
New Crowdin translations by [Crowdin GH Action](https://github.com/crowdin/github-action)

PR #2852, opened 2025-04-10
Cherry pick final launch translations into release branch

kvangork:
New Crowdin translations by GitHub Action (#2842)

Closes MOB-620


fastlane/metadata/ios/fi/description.txt
```diff
@@ -1,5 +1,14 @@
...
+AVAINOMINAISUUDET
+Tunnista lajeja missä tahansa päin maailmaa
+Pidä kirjaa näkemistäsi lajeista
+Edistä tiedettä jakamalla havaintojasi
+Tutki ja suojele luonnon monimuotoisuutta 
+Ole yhteydessä muihin luontoharrastajiin ja opi heiltä
+
+iNaturalistin voittoa tavoittelematon tehtävä on yhdistää ihmiset luontoon ja edistää tiedettä ja luonnonsuojelua. iNaturalist on vapaasti käytettävissä lahjoittajayhteisömme runsaan tuen ansiosta.
+
+Tutustu ja opi vielä enemmän https://www.inaturalist.laji.fi
```
jtklein, line 14: This URL does not work. I don't think they are supposed to be translated actually.

kvangork: I'll hand-fix that for this release once it's been auto-uploaded to the App Store, and have fixed it at the source, which—curiously—was added by a human translator last week.

albullington (approved):
Looks fine to me. I tested by making a dev build, logging in, switching languages in the app, and checking the newest text (Analyzing...) in the loading PreMatch screen in a few different languages. Apparently the first few languages I picked all have the speediest translators (Danish, etc.) and they were already translated. I finally found a language (Turkish) that falls back to English to make sure that was working without crashes too.

kvangork:
> Looks fine to me. I tested by making a dev build, logging in, switching languages in the app, and checking the newest text (Analyzing...) in the loading PreMatch screen in a few different languages. Apparently the first few languages I picked all have the speediest translators (Danish, etc.) and they were already translated. I finally found a language (Turkish) that falls back to English to make sure that was working without crashes too.

Thanks for testing the fallback @albullington, that's helpful. I should have mentioned that another manual test I successfully performed today was to confirm that in regional dialects where we do not have a specific translation, we're correctly falling back to the primary language (i.e.: es-MX uses es-ES strings rather than falling back to en-US)

PR #2855, opened 2025-04-14
Add tracking for Pressable and INatIconButton taps

albullington:
REVISED: see comments below.

ORIGINAL:
Trying out a simple approach to analytics by logging `Button` taps. If we like this approach, the next step would be to extend this to the rest of our tappable components such as `INatIconButton`.

This currently shows up in the logs as something like what's shown below, which also alerts me to one problem with this approach -- we're translating strings before passing them to the Button, which means we'll be logging text in all languages and making it hard to aggregate any data:

`{"level":"info","message":"Button tap: {\"accessibilityHint\":null,\"accessibilityLabel\":\"APP-SPRACHE ÄNDERN\",\"testID\":null,\"text\":\"APP-SPRACHE ÄNDERN\",\"routeName\":\"Settings\"}","context":"Button","timestamp":"2025-04-14T21:36:42.246Z"}`

I'm blanking on the right way to handle finding the original, non-translated version of an i18next string once the Button component receives an already translated version as props, but I'm sure there's a way to do this...

### DIFF
--- a/src/components/styledComponents.js
+++ b/src/components/styledComponents.js
@@ -4,6 +4,8 @@ import { FasterImageView as UnstyledFasterImageView } from "@candlefinance/faste
 import {
   BottomSheetTextInput as StyledBottomSheetTextInput
 } from "@gorhom/bottom-sheet";
+import PressableWithTracking
+  from "components/SharedComponents/Buttons/PressableWithTracking.tsx";
 import { styled } from "nativewind";
 import {
   Image as UnstyledImage,
@@ -11,7 +13,6 @@ import {
   KeyboardAvoidingView as UnstyledKeyboardAvoidingView,
   Modal as UnstyledModal,
   Platform,
-  Pressable as StyledPressable,
   SafeAreaView as UnstyledSafeAreaView,
   ScrollView as UnstyledScrollView,
   Text as UnstyledText,
@@ -37,7 +38,7 @@ const Text = styled( UnstyledText );
 // $FlowIgnore
 const TextInput = styled( UntyledTextInput );
 // $FlowIgnore
-const Pressable = styled( StyledPressable );
+const Pressable = styled( PressableWithTracking );
 // $FlowIgnore
 const Image = styled( UnstyledImage );
 // $FlowIgnore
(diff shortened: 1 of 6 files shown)


albullington:
@kueda tagged you as a reviewer because I'm curious if this is the approach you had in mind & mentioned in our morning meeting. Also curious if this is in a format that's easy to turn into another Grafana dashboard.

kueda (commented):
Close to what I had in mind. If you literally want to log every button tap, you could just pass a unique identifier to every button, e.g. you could make `testID` a required field and ensure every single button receives that prop and then do something like 

```javascript
logger.info( `button tap: ${testID}-currentRoute?.name` )
```

Then if you wanted to know if anyone ever taps the iconic taxa buttons on ObsEdit, you'd look for log events like `iconic-taxa-btn-Animalia-ObsEdit` or something.

Alternative, you could ad a new prop to `<Button>` that specifies a unique logging identifier, and its presence means the button should log taps, like

```javascript
<Button
  logPressWithId="iconic-taxon-btn-Animalia-ObsEdit"
/>
```

and then in Button you just log the press if that identifier is present.

If you want further segmentation of the log data, you could use [the `extra` param Patrick added](https://github.com/inaturalist/iNaturalistAPI/pull/497) to the logging endpoint.

Last note: we started talking about this b/c someone asked if anyone uses the iconic taxon buttons. Logging all button taps would tell us if people tap them, but you could argue that people only *use* them if they actually save an observation with one of them set, which would require logging an event when the user taps SAVE and we know they tapped one of those buttons to add their ID.

albullington:
Neat, I like the simple one line logging statement -- much more elegant than what I was trying to do. I went for a very global solution here and added tracking around all of our Pressables as well as our INatIconButton component. If they have a testID, they'll be tracked.

I'm keen to play around with the `extra` data but will save that for another PR since @jtklein has been nudging us towards smaller PRs with less changes to reason about all at once.

src/components/styledComponents.js
```diff
@@ -37,7 +38,7 @@ const Text = styled( UnstyledText );
 // $FlowIgnore
 const TextInput = styled( UntyledTextInput );
 // $FlowIgnore
-const Pressable = styled( StyledPressable );
+const Pressable = styled( PressableWithTracking );
```
jtklein, line 41: Should we import as UnstyledPressableWithTracking for consistency?

jtklein (approved):
I can see the logs on the server from both types of buttons, so it is working as expected.

I think personally I would have opted for a new `logID` that specifies if we want to log this one specific button. To separate finding something in tests (via testID) from a different function of logging. But I guess either way is working.

src/components/styledComponents.js
albullington: yep, good call

PR #2856, opened 2025-04-15
Consistent margin around user profile stats

jtklein:
Minor UI changes to update user profile to be closer to design specs.

Closes MOB-652.


albullington (approved):
Lgtm. Sometimes with issues like this I find it helpful to do an overall audit of the designs (like, I'm noticing that the margins around the entire `UserProfile` screen are 12px instead of 15px), but it's not necessary.

I noticed the `mt-8` change above PROJECTS because it's a refactor that wasn't requested in this issue and is actually smaller than what's in the designs (32px), but I can't comment on what should happen here -- in some Figma designs that margin is 58px and in some it's 77px.

PR #2858, opened 2025-04-16
Do not access taxon in SuggestionsResult if none

jtklein:
MOB-692 related. Unsure if this is a fix for this issue.
This is the best I could come up with for now. I could not replicate this error locally.
It would be good to get detailed replication steps.
From experience not accessing any properties of the RealmObject is a good way to avoid this error message. `taxon` is the only RealmObject used in this component so maybe this change addresses it.
It's a bit weird because in the CustomFlashList higher in the stack we already access taxon.id which when I am testing with removing taxon directly from the suggestion error out. So, I am unsure how we can access something lower in the tree when not having it higher already should give a different error.
Anyways, I also checked Kibana to gauge what user actions happen before this error appears but couldn't see a pattern. Maybe good idea to release this PR and add a navigation breadcrumbs logging to errors as well.

### DIFF
--- a/src/components/Match/AdditionalSuggestions/SuggestionsResult.js
+++ b/src/components/Match/AdditionalSuggestions/SuggestionsResult.js
@@ -44,6 +44,10 @@ const SuggestionsResult = ( {
     }
   }, [forcedHeight] );
 
+  if ( !taxon ) {
+    return null;
+  }
+
   const accessibleName = accessibleTaxonName( taxon, currentUser, t );
 
   // A representative photo is dependant on the actual image that was scored by computer vision


albullington (approved):
Agree that I'm not sure if it will fix the problem, but it's a solid hypothesis. I think if I were writing this I would move the `null` check a level higher and not render `<SuggestionsResult />` at all if there's no taxon... which would prevent the `useEffect`, `useCurrentUser`, `useTranslation`, and `handleLayout` from running in the background of whatever screen this crash is happening on.

Another thing that occurs to me here is that `useCurrentUser` is also calling Realm here and could be causing issues, though hopefully our `currentUser?.isValid( )` code prevents that.

PR #2859, opened 2025-04-16
Add tracking for any Pressable or INatIconButton with a testID

albullington:
This came out of a Creative Day discussion, where the engagement team wanted to understand whether or not anyone uses the iconic taxon buttons in ObsEdit. This is a first step to understanding that, as it records individual taps to our `/log` endpoint for those buttons as well as every button in the app with a `testID`. My goal here was to see how much data we can collect with a very minimal amount of code changes.

I'm not *quite* sure what we'll end up doing with this data, but I'm picturing a new Grafana dashboard that can help us make sense of these logs. TBD.


albullington:
Closing this because we're likely not using the release branch anymore, so no need for an extra review.

PR #2861, opened 2025-04-17
Add some context to 429 errors

jtklein:
Closes MOB-696
The reason context appeared as null in log calls for some 429 errors is that we do not pass anything as context into the Error constructor. So, now I am changing it to add some information in here.
The issue did not specify as to what information to add, open to alternatives.

### DIFF
--- a/src/api/announcements.js
+++ b/src/api/announcements.js
@@ -25,7 +25,7 @@ const searchAnnouncements = async ( params: Object = {}, opts: Object = {} ): Pr
     );
     return results;
   } catch ( e ) {
-    return handleError( e );
+    return handleError( e, { context: { functionName: "searchAnnouncements", opts } } );
   }
 };
 
@@ -33,7 +33,7 @@ const dismissAnnouncement = async ( params: Object = {}, opts: Object = {} ): Pr
   try {
     return await inatjs.announcements.dismiss( params, opts );
   } catch ( e ) {
-    return handleError( e );
+    return handleError( e, { context: { functionName: "dismissAnnouncement", opts } } );
   }
 };
 
--- a/src/api/comments.js
+++ b/src/api/comments.js
@@ -17,7 +17,7 @@ const createComment = async (
     const { results } = await inatjs.comments.create( { ...PARAMS, ...params }, opts );
     return results;
   } catch ( e ) {
-    return handleError( e );
+    return handleError( e, { context: { functionName: "createComment", opts } } );
   }
 };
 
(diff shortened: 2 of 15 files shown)


albullington (changes requested):
Ah, sorry @jtklein I didn't expect someone to start working on this issue right away. The errorContext I had in mind is the one we already have in `logging.js`: https://github.com/inaturalist/iNaturalistReactNative/blob/cec7caac46a1c2022bf1914944dac4b0cb3e0612/src/sharedHelpers/logging.js#L27

It would also be nice if we could handle this directly in `handleError` instead of updating so many files.

albullington:
@jtklein have you been able to verify that this is working in Kibana? I tried to replicate a 429 error in `fetchRemoteObservation` by replacing our results with the following error, then going to ObsDetails where this fetch is called. However, this never seems to make it to Kibana, because instead of error handling, we're getting an unhandled promise rejection -- this leads me back to my core suspicion here that we're not handling async/await properly whenever we call `handleError` (or at least somewhere along this error handling path).

```
const fetchRemoteObservation = async (
  uuid: string,
  params: Object = {},
  opts: Object = {}
): Promise<?number> => {
  try {
    const error = new Error( "Rate limit exceeded" );
    error.response = { status: 429 };
    throw error;
  } catch ( e ) {
    return handleError( e, { context: { functionName: "fetchRemoteObservation", uuid, opts } } );
  }
};
```

This is the warning I'm seeing when I try to create and propagate a 429 error.
Screenshot 2025-04-29 at 12 38 19 PM

albullington (changes requested):
Requesting another round of changes because the API calls are showing up for me as an unhandled promise rejection, not an actual error that makes its way to Kibana.

jtklein:
Yes, I was checking if these errors turn up in Kibana. I just repeated it. You can check some of the errors from my last session: https://grafana-vpn.inaturalist.org:3000/d/H3LCbBLIz/inat-next-errors?orgId=1&from=now-3h&to=now&timezone=browser&var-Filters=parsed_user_agent.version%7C%3D%7C1.0.2
I changed the version number locally only to filter out the noise.

Logs are only sent to the server in release builds though: https://github.com/inaturalist/iNaturalistReactNative/blob/a5ab8a233937ea0ba6808b33fb4437fbffc23ff2/src/api/log.ts#L28-L31

The unhandled promise rejection in DEV mode I could also see in main as well though. So, I did not want to enlarge the scope of this PR to figure DEV and release mode differences here.

PR #2863, opened 2025-04-18
Decrease TaxonDetailHeader size to match ObsDetails header size

angielt:


jtklein (changes requested):
When I use Inspect Element on my device the header on ObsEdit is 44px but the header on TaxonDetails is 48px.
According to the issue they should be the same, so TaxonDetails should be 44 as well? Or is the header we use from react-navigation directly (on ObsDetails) adapting it's height to the device screen?

jtklein (approved):

PR #2864, opened 2025-04-19
MOB-690 - Add error context to 401 errors

budowski:


src/api/error.js
```diff
@@ -53,6 +64,8 @@ async function handleError( e: Object, options: Object = {} ): Object {
   // matter.
   if ( e.response.status === 429 ) {
     throw new INatApiTooManyRequestsError( context );
+  } else if ( e.response.status === 401 ) {
+    throw new INatApiUnauthorizedError( context );
```
albullington, line 68: one problem we're bumping into in 429 errors is that the context is still showing up as `null` every time. i'm not sure what the core issue is there -- maybe we're not awaiting the ` handleError` function throughout the app and we need to? I don't think this will fix the core issue or help us add context, unfortunately.

src/api/error.js
```diff
@@ -26,6 +26,17 @@ Object.defineProperty( INatApiError.prototype, "name", {
...
+  constructor( context?: Object ) {
+    const errorJson = {
+      error: "Unauthorized",
+      status: 401,
+      context
+    };
+    super( errorJson, 401, context );
+  }
+}
+
```
albullington, line 39: Does this need `Object.defineProperty( INatApiUnauthorizedError.prototype, "name", {
  value: "INatApiUnauthorizedError"
} );`? The other two errors have that defined.

albullington (changes requested):

jtklein:
You might want to rebase/update from main to include these changes to the same files before continuing: https://github.com/inaturalist/iNaturalistReactNative/pull/2861

src/api/error.js
jtklein: Have added

albullington (changes requested):
I'm not sure what happened in this PR, but there are 340 files changed that mostly have nothing to do with this issue, so this isn't ready to merge. My best recommendation here would be to create another branch for this issue from `main`, cherry-pick the actual desired commits, and close this PR.

jtklein:
I also don't know what happened here, I used "Update from main" function in GitHub Desktop but now this branch thinks that all commits that happened already on main are to merge as well. Anyway, here is a copy with only the relevant commits:
https://github.com/inaturalist/iNaturalistReactNative/pull/2921

PR #2865, opened 2025-04-19
MOB703 - changes to obs edit screen identification section

budowski:


albullington (approved):
made one change to the size of the magnifying glass icon; otherwise approved.

PR #2869, opened 2025-04-22
Change calculation of successful upload status stats

jtklein:
Closes MOB-694

Instead of using `numUploadsAttempted` only to show the number of uploaded observations, I have changed to use `numUploadsAttempted - totalUploadErrors` instead. Which in first approximation should be the number of successful uploads, unless I am missing something here. Maybe better in the future would be to have an actual "successful uploads" counter using the actual create/upload response as signal.

I have tested with all obs failing and randomly half of the obs failing (by changing params.observation to {} in observations.js, createObservation which result in a 500 error from server). Both times I think it displays the wanted information now. With all obs failing it only shows the x-failed text. With some failing it shows the x-uploaded and the x-failed test

### DIFF
--- a/src/components/MyObservations/SimpleUploadBanner.js
+++ b/src/components/MyObservations/SimpleUploadBanner.js
@@ -92,7 +92,7 @@ const SimpleUploadBanner = ( {
 
   return (
     <View className="py-2">
-      {status.text !== "" && renderUploadStatusText( )}
+      {( status.text !== "" || !!error ) && renderUploadStatusText( )}
       <UploadProgressBar progress={progress} />
     </View>
   );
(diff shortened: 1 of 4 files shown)


albullington (changes requested):
Could we add tests for the possible states here into `SimpleUploadBannerContainer.test.js`? Something like:

- 1 upload out of 5 succeeded and 4 failed
- 5 uploads out of 5 failed
- 4 uploads out of 5 succeeded and 1 failed

This seems like exactly the type of thing we want to make sure we don't break in a refactor, and I worry that putting the logic directly in the component makes this a little more fragile and less robust as we make code changes. (I'm thinking about how we'll likely need to make the same change in the bottom upload status bar in ObsEdit, and how it would be really nice to be able to pull out variables directly from the store instead of duplicating this logic there...).

jtklein:
Have added unit tests, and refactored the number of uploads without error into the store slice.

albullington (approved):

PR #2870, opened 2025-04-23
Do not show loading state on Android

jtklein:
On Android release builds, imagesLoaded never switched from false to true, and this screen was stuck in a loading state. On iOS it worked as expected. Disabling it now on Android to make a new release possible.


albullington (approved):
Works for me, though it makes me curious if that image loading code is still needed on iOS or if it's unnecessary.

PR #2872, opened 2025-04-23
Maintenance: break upload code into smaller modules -- realm syncing functions

albullington:
Closes Mob-734

This is a sub-issue of a larger issue to start making our upload code easier to maintain. Realm syncing functionality was the most modular and easiest part of the code to break out.

Unit tests were added to make sure functionality didn't change while refactoring. The one exception to this is the very last test, which highlighted that we were swallowing any non-invalidation errors instead of throwing them. I believe throwing these errors is the desired behavior, so made this change & changed the unit test to reflect it.


jtklein:
Sorry, have messed up the original, here is a copy with only the two original commits: https://github.com/inaturalist/iNaturalistReactNative/pull/2883

PR #2880, opened 2025-04-28
Mob 720: Make loading indicator smaller and transparent

jtklein:
So that the map can still be interacted with while the loading indicator is shown.


albullington (approved):
I'm not convinced that we *want* a user to be able to interact with the map while the loading indicator is up, but this isn't a harmful change so approving.

PR #2882, opened 2025-04-29
Add fastlane script to download app reviews

jtklein:


fastlane/Fastfile
```diff
@@ -541,6 +541,58 @@ lane :beta do
...
+lane :download_app_store_reviews do
+  desc "Download App Store reviews and save them to a file"
+
+  UI.message "Logging into App Store Connect..."
+
+  # Use credentials from the Appfile
+  apple_id = CredentialsManager::AppfileConfig.try_fetch_value( :apple_id )
+  itc_team_id = CredentialsManager::AppfileConfig.try_fetch_value( :itc_team_id )
+
+  # Explicitly login with credentials
```
albullington, line 553: I think this could use some error handling. What if login fails?

fastlane/Fastfile
```diff
@@ -541,6 +541,58 @@ lane :beta do
...
+    processed_review = {
+      rating: review.rating.to_i,
+      title: review.title,
+      review_content: review.review,
+      nickname: review.nickname,
+      country: review.store_front,
+      last_modified: review.last_modified,
+      edited: review.edited,
+      responded: review.responded?,
+      developer_response: review.developer_response.response
```
albullington, line 584: Do we know that there is always a developer_response? Will this fail if there's not a developer response?

fastlane/Fastfile
```diff
@@ -541,6 +541,58 @@ lane :beta do
...
+  UI.message "Finding app with bundle ID: #{IOS_BUNDLE_ID}"
+  app = Spaceship::Tunes::Application.find( IOS_BUNDLE_ID )
+
+  unless app
+    UI.abort_with_message! "Could not find app with bundle ID: #{IOS_BUNDLE_ID}"
+  end
+
+  # Fetch reviews
+  UI.message "Fetching reviews for #{app.name}..."
+  reviews = app.ratings.reviews
```
albullington, line 567: are reviews paginated? do we need to consider downloading multiple pages of reviews, or are you only trying to fetch the most recent page of reviews? i'm not sure what this code will ultimately be used for or where this information is going... are you just collecting this locally or planning to share with the broader team?

albullington (approved):
Made some suggestions to make this more robust, but can't be too helpful since I don't know what this is actually going to be used for.

fastlane/Fastfile
jtklein: No, there is not always a response, and in cases there is none it does not fail but save null to the JSON.

fastlane/Fastfile
jtklein: I think that is fine as is, if the login fails the script also just fails with an error message.

fastlane/Fastfile
jtklein: The purpose is to share with the broader team, especially for review analyses.
I have tested the same script in Seek now with all reviews and it did not need any pagination.

jtklein:
Thanks, was helpful to think about these things.

PR #2884, opened 2025-04-29
 Maintenance: break upload code into smaller modules -- data transformation

albullington:
Closes Mob-737

### DIFF
--- a/src/sharedHelpers/uploadObservation.js
+++ b/src/sharedHelpers/uploadObservation.js
@@ -12,87 +12,9 @@ import Observation from "realmModels/Observation";
 import ObservationPhoto from "realmModels/ObservationPhoto";
 import ObservationSound from "realmModels/ObservationSound";
 import emitUploadProgress from "sharedHelpers/emitUploadProgress.ts";
-import safeRealmWrite from "sharedHelpers/safeRealmWrite";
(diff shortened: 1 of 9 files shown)


src/sharedHelpers/uploadObservation.js
```diff
@@ -57,24 +64,8 @@ const uploadEvidence = async (
     return response;
   };
 
-  const responses = await Promise.all( evidence.map( item => {
-    let currentEvidence = item;
-
-    if ( currentEvidence.photo ) {
```
jtklein, line 63: This piece was removed, and I don't see it replaced with a block in another file.
So, just to double-check, this can be removed without replacement?

jtklein (approved):
Other than one question, it looks god to me.

src/sharedHelpers/uploadObservation.js
albullington: yeah, I spent a while puzzling about this and I think it's fine & cleaner code to remove. it's a piece of awkward code where we were first cleaning the null values, but then we were immediately mapping via `mapPhotoForUpload` so we were overwriting any null values we just cleaned anyway.

i'm going to merge, and will note that if we run into issues testing uploads with photos in our internal build this is a great place to look for bugs.

PR #2887, opened 2025-05-04
MOB-752 - use FasterImageView for caching obs/taxon/project images

budowski:


jtklein:
This is not a full review, but I have tested this and to me it "feels" slower than what we currently have.

albullington:
I'm going to hold off on reviewing this for now -- @jtklein had expressed some concerns in Slack about whether or not this is the right approach, and I think this will make more sense to review for MyObservations after https://github.com/inaturalist/iNaturalistReactNative/pull/2919 and https://github.com/inaturalist/iNaturalistReactNative/pull/2922 are merged, since we'll have a better overall sense of how MyObservations performs with this caching.

albullington (approved):
I merged from main to test it with the more performant ObsList, and it looks good to me. We can tweak the cachePolicy as needed in the future, but I think FasterImage is best practice and gives us more control than using the Image cache which we have no control over.

PR #2888, opened 2025-05-05
MOB-708 - suggestions screen - show percentages instead of confidence dots

budowski:


src/components/SharedComponents/TaxonResult.tsx
```diff
@@ -226,12 +229,21 @@ const TaxonResult = ( {
...
             isBackground={false}
           />
-          {!!( confidence && confidencePosition === "photo" ) && (
+          {!!( ( confidence || confidencePercentage ) && confidencePosition === "photo" ) && (
             <View className="absolute -bottom-4 w-full items-center">
-              <ConfidenceInterval
-                confidence={confidence}
-                activeColor={activeColor}
-              />
+              {confidencePercentage
```
jtklein, line 234: The only time we prop in `confidencePercentage` is with `confidencePosition === "text"` at the moment. So this here will never be shown. Can we remove it?

jtklein (changes requested):
Two minor changes.

jtklein (approved):

PR #2889, opened 2025-05-05
MOB-704 - declutter explore obs grid items

budowski:


src/components/Explore/ObservationsView.js
```diff
@@ -135,6 +135,7 @@ const ObservationsView = ( {
         obsListKey="ExploreObservations"
         onEndReached={fetchNextPage}
         showNoResults={!canFetch || totalResults === 0}
+        hideObsUploadStatus
```
jtklein, line 138: This prop is also consumed by the `ObsListItem` [here](https://github.com/inaturalist/iNaturalistReactNative/blob/8ffe02cc98613868141613eb5a4eff3d770a3651/src/components/ObservationsFlashList/ObsPressable.tsx#L77-L78). So, changing like so also removes ObsUploadStatus from the list layout which is not specified in the issue ticket.

src/components/ObservationsFlashList/ObsPressable.tsx
```diff
@@ -18,7 +18,6 @@ type Props = {
   onUploadButtonPress: ( ) => void,
   onItemPress: ( ) => void,
   gridItemStyle: Object,
-  isLargeFontScale: boolean,
```
jtklein, line 21: This is still used as a prop in `ObservationsFlashList` and should be removed there.

jtklein (changes requested):
One small change and one bigger change requested in inline comments.

jtklein (approved):

PR #2890, opened 2025-05-05
Migrate MatchHeader and imported components/functions to TypeScript

jtklein:
Also, few code changes for clarification, hopefully.


src/sharedHelpers/fetchTaxonAndSave.js
```diff
@@ -0,0 +1,23 @@
+import { fetchTaxon } from "api/taxa";
+import { getJWT } from "components/LoginSignUp/AuthenticationService.ts";
+import Taxon from "realmModels/Taxon";
+import safeRealmWrite from "sharedHelpers/safeRealmWrite";
+
+async function fetchTaxonAndSave( id, realm, params = {}, opts = {} ) {
```
jtklein, line 6: This was not imported in the MatchHeader tree so I did not migrate it to TypeScript.

src/sharedHelpers/taxon.ts
```diff
@@ -121,31 +133,36 @@ export const generateTaxonPieces = taxon => {
     ).join( " · " );
   }
 
-  let { name: scientificName } = taxon;
-
-  scientificName = scientificName?.split( " " );
+  const scientificNamePieces = taxon?.name?.split( " " );
   if ( taxon.rank_level < 10 ) {
     if ( taxon.rank === "variety" ) {
-      taxon.rankPiece = "var.";
```
jtklein, line 129: I changed this so that the function does not change the function param.

src/sharedHelpers/taxon.ts
```diff
@@ -101,8 +97,24 @@ export const capitalizeCommonName = name => {
...
-export const generateTaxonPieces = taxon => {
-  const taxonData = {};
+interface Taxon {
+  rank?: string;
+  rank_level: number;
+  preferred_common_name?: string;
+  name?: string;
+}
+
+interface TaxonData {
```
albullington, line 107: This name is a little confusing for me now that it's written into TypeScript -- can we think of a way to name this to differentiate why this is needed vs. why `Taxon` is needed? Maybe this can be called something like `TaxonDisplayData` and/or `Taxon` can be `ApiTaxon`? I'm not sure if we have a naming convention for this already, but it's nice to understand what we've transformed versus what is taken directly from the API.

albullington (changes requested):
One minor naming change requested; otherwise approved.

PR #2891, opened 2025-05-07
Add login button where logout button is on menu

angielt:


src/navigation/CustomDrawerContent.tsx
```diff
@@ -145,6 +145,20 @@ const CustomDrawerContent = ( { state, navigation, descriptors }: Props ) => {
         onPress: ( ) => setShowConfirm( true )
       };
     }
+
+    if ( items.logout === undefined && !currentUser ) {
+      items.login = {
+        label: t( "LOG-IN" ),
+        icon: "door-enter",
+        style: {
+          opacity: 0.5,
```
jtklein, line 154: According to the [designs](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=3773-22107&t=xvLyzfN2P7C9d2d9-11), the icon should be green and the text black.

src/navigation/CustomDrawerContent.tsx
```diff
@@ -145,6 +145,20 @@ const CustomDrawerContent = ( { state, navigation, descriptors }: Props ) => {
         onPress: ( ) => setShowConfirm( true )
       };
     }
+
+    if ( items.logout === undefined && !currentUser ) {
```
jtklein, line 149: Can we not just append an else block to `if ( currentUser ) {` instead of this?

jtklein (changes requested):
Behaviour is correct but design not to spec.

jtklein (approved):
Yes, this looks great now.

PR #2892, opened 2025-05-09
Remove log from script

jtklein:
Only to remove a log that happens for each review while the script is running.


kvangork (approved):
LGTM

PR #2894, opened 2025-05-12
Add a case for one member to string

jtklein:
Comment here: https://inaturalist.slack.com/archives/C02G5AATVGA/p1746214906692239

### DIFF
--- a/src/i18n/l10n/en.ftl
+++ b/src/i18n/l10n/en.ftl
@@ -1408,7 +1408,11 @@ X-Identifiers =
        *[other] { $count } Identifiers
     }
 # Subheader for number of project members screen
-X-MEMBERS = { $count } MEMBERS
+X-MEMBERS =
+    { $count ->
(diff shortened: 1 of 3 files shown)


albullington (changes requested):
@kueda can correct me on this but my understanding is that for all changes to strings, we also need to change the name of the key, otherwise translators won't receive the changed values. I'm assuming that logic holds even in this case where we're just changing variables.

kueda:
Changing the key is ideal in terms of *notifying* translators that there's something they need to do (they get notifications and email updates about new strings but I don't think they get them for changed strings). It's not ideal for keeping Crowdin screenshots up-to-date or for developer sanity. So it's kind of a judgement call. I guess I'd use this heuristic: if it's important that this be translated sooner rather than later, then change the key. If it's not that important, it's ok to leave the key alone.

jtklein:
Decided that it is not important to get translations sooner rather than later.

PR #2902, opened 2025-05-15
Request fewer update fields from API

jtklein:
This came out of a recent user problem where notifications would not load. While the actual error reason was a 500 on the server we are also requesting many more fields here than we actually need for the notifications screen to work.
So, a small cleanup, safeguard, possible speed improvement.
Closes MOB-780


albullington (approved):
Nice, this type of code & performance improvement PR is super helpful. Good example of a small way we're tackling the "app is slow" user problem.

PR #2903, opened 2025-05-15
Update react navigation libraries for performance

albullington:
Closes Mob-757

- Made a wrapper function `navigateToTabStack` in navigationUtils for ease & readability. This addresses the changes to the navigate action where we need to be specific about navigating to nested children
- Removed `unmountOnBlur` since this functionality doesn't have a 1:1 replacement; left a comment about how we can add this back for the Match screen if needed
- Updated props related to displaying the title next to the back button


jtklein (changes requested):
Only one change in the actual code as written in the inline comment.

However, unfortunately, I think this PR is not ready to merge yet in general, as the performance of navigating between screens is not as good to what we currently have.

From what I can see in the react-navigation v7 documentation, all things we should migrate have been considered. So, without looking further into details of the differences between v7 and v6, unfortunately, I can not be more specific as to which changes I would request. I will send a comparison video on Slack. One problem I think is that the screens of the tab navigators are having many more renders now while they are not visible because the app is in another tab.

albullington:
Ready for re-review. This changes our navigation structure ever so slightly by making RootExplore, Notifications, and MyObservations each the root screen of the same tab stack navigator. This preserves (maybe even improves) performance when switching between tabs and maintains the navigation history (for example, if a user goes to Notifications, then ObsDetails, then back, they'll land back on that root screen, Notifications).

PR #2912, opened 2025-05-19
Speed up Explore nearby tile loading

albullington:
Closes MOB-782

- Creates a MapPerformanceTracker to help ballpark the time from when a user first sees the Explore map to the time it takes for tiles to load
- Displays this over the map in Debug mode -- time for tiles to load was previously anywhere from 2.5 to 17 seconds in developer mode, which is expected to be a little slower than production
- Updates the default radius for nearby Explore searches from 50km to 1km, since this was the lowest hanging fruit. This brings average time for tiles to display closer to 1.5 seconds via my internet connection in developer mode, and more like 0.8 seconds in production mode
- Shows a more realistic number of nearby observations in the gray status bar, based on the 1km distance


.eslintrc.js
```diff
@@ -53,7 +53,9 @@ module.exports = {
     "import/extensions": [2, {
       js: "never",
       jsx: "never",
-      json: "always"
+      json: "always",
+      ts: "never",
```
jtklein, line 57: Should changing this now give an error every time we import a .ts file with extension? I do not see any new errors for files where we are already importing with extension.

jtklein (changes requested):
Code changes seem okay.

One requested change:
Currently the "Tiles visible" information is only collected for the map tiles with green obs pins, and not the map tiles with orange squares. Sometimes when navigating to Explore the map starts with orange tiles for me.

One note: I tested in debug and release builds, and I actually do not see any difference in times when using 1 vs 50 radius. I am seeing a drop in loading time comparing the first time I open Explore on a fresh install vs subsequent times (some caching somewhere?) but the average loading time of the non-first times I navigate to Explore is similar with both radiuses. (Should I test on different devices as well?)
So, I guess the decision to make the radius less is more in the product team court, and not a performance recommendation. Which means I am indifferent if we should change it.

.eslintrc.js
albullington: no, it's the opposite. this didn't impact anything locally, but github actions was throwing an error about that file extension. so something is different about our eslint (or tsconfig?) setup in github actions vs. local.

albullington:
It's possible your internet is better than mine? Here's the comparison of how this change impacts my device on my wifi in development mode -- pretty drastic.

I can also confirm that this metric is working for me with orange tiles already, so I don't think there are changes needed here -- I didn't target a particular type of tile in the tracking code. If your device is sometimes showing orange tiles for nearby, that's a separate bug we should report, because we want to consistently zoom the user in far enough that they can tap on the green obs pins.

IMG_0015
IMG_0012
IMG_0034

albullington:
Not as dramatic as a different for me in a release build (as expected), but it pretty consistency gets us closer to 1 second for tile loading which is within best practice for performance, versus up to 4 seconds for tile loading which is not.

IMG_0038
IMG_0042

PR #2918, opened 2025-05-21
Downgrade react-native-screens to v4.4.0

jtklein:
react-native-screens version 4.5.0 increases the react native version that it supports to 0.74.0+ (https://github.com/software-mansion/react-native-screens/pull/2613/files). So, by using 4.4.0 we are still able to build the app on Android as we are still on 0.73 for now (have tested and it works).
Have not seen any effect on the changes related to the latest react-navigation versions.


albullington (approved):
Excellent version sleuthing @jtklein. Confirmed this works as expected in iOS debug/release and that I can get Android running on a local emulator.

PR #2919, opened 2025-05-21
Performance: speed up API calls in MyObs by limiting fields

albullington:
Closes Mob-793

Limit the number of fields when fetching remote observations in MyObservations.

This PR builds on the strategy from [PR: Speed up explore API](https://github.com/inaturalist/iNaturalistReactNative/pull/2912) of showing performance metrics in debug mode, including scroll/fetch events. I wouldn't worry too much about the accuracy of the events themselves as the goal here was to use them to benchmark performance when we change the number of fields fetched via the API in `searchObservations`.

Things to note:
1. I split up the fields needed for default mode and advanced mode, but decided to call the advanced mode API for all users to avoid situations where a user switches into advanced mode & sees that all of their observations have no location. I could imagine a future scenario where we might selectively refetch these observations in advanced mode, but that seemed unnecessarily complex for this PR.
2. I added the loading wheel back into the footer, since that was hidden and made it hard for users (and developers!) to know whether data was being fetched
3. I removed `usePerformance` from MyObservations since it's noisy and I don't believe anyone is actively using it

Ballpark performance improvement:
- In release mode, average fetch time *without* changes to fields is around 3.5 to 5 seconds on my device. This drops squarely to 2-3 seconds with changes.

no changes:
IMG_0064

limited fields:
IMG_0055

### DIFF
--- a/ios/iNaturalistReactNative.xcodeproj/xcshareddata/xcschemes/iNaturalistReactNative.xcscheme
+++ b/ios/iNaturalistReactNative.xcodeproj/xcshareddata/xcschemes/iNaturalistReactNative.xcscheme
@@ -59,7 +59,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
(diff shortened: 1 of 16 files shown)


angielt (changes requested):
MyObs seems faster with the changes for me and adding the loading wheel back is a good change. One thing is I can't load more than 10 species on explore or myObs in this branch. Do you experience the same?

angielt:
PR #2930 seemed to address the species view in Explore and MyObs for me. Would approve if that PR is merged into this one.

PR #2922, opened 2025-05-22
Performance: data transformation in MyObservations

albullington:
Closes Mob-791

### DIFF
--- a/src/realmModels/Comment.js
+++ b/src/realmModels/Comment.js
@@ -14,6 +14,12 @@ class Comment extends Realm.Object {
     user: User && User.FIELDS
   };
 
+  static mapCommentForMyObsAdvancedMode( comment ) {
+    return {
+      uuid: comment.uuid
+    };
+  }
+
   static schema = {
     name: "Comment",
     embedded: true,
--- a/src/realmModels/Identification.js
+++ b/src/realmModels/Identification.js
@@ -41,6 +41,13 @@ class Identification extends Realm.Object {
     return newId;
   }
 
+  static mapIdentificationForMyObsAdvancedMode( id ) {
+    return {
+      uuid: id.uuid,
+      current: id.current
+    };
+  }
+
   static new = attrs => {
     const newIdent = {
       ...attrs,
--- a/src/components/MyObservations/MyObservationsSimple.tsx
+++ b/src/components/MyObservations/MyObservationsSimple.tsx
@@ -196,7 +196,7 @@ const MyObservationsSimple = ( {
 
   const unuploadedObsMissingBasicsIDs = useMemo( () => (
     observations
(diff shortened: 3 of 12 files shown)


src/realmModels/Identification.js
```diff
@@ -33,6 +33,13 @@ class Identification extends Realm.Object {
     };
   }
 
+  static mapIdentificationForMyObsAdvancedMode( id ) {
+    return {
+      uuid: id.uuid,
```
jtklein, line 38: These methods in the other models used ?. chaining.

src/realmModels/Comment.js
```diff
@@ -14,6 +14,12 @@ class Comment extends Realm.Object {
     user: User && User.FIELDS
   };
 
+  static mapCommentForMyObsAdvancedMode( comment ) {
+    return {
+      uuid: comment.uuid
```
jtklein, line 19: Does it need ?. chaining as in the other models?

jtklein (approved):
I did not completely understand the new `handleChange` function in the timebox I was given for this PR review. I am left wondering if since the new map functions return a new object every time they are called, if we are not increasing the amount of renders because of new references. Because I ran out of time I did not double-check any performance changes in production mode.

src/realmModels/Comment.js
albullington: I don't think so because uuid is always defined, and since we're mapping from a Realm object, that object always has all possible properties set to null instead of undefined. But I do think we're starting to see some TypeErrors pop up in very small numbers throughout our JS files where we could probably benefit from more chaining (or more conversion to TS) so it's a good question.

albullington:
> I did not completely understand the new `handleChange` function in the timebox I was given for this PR review. I am left wondering if since the new map functions return a new object every time they are called, if we are not increasing the amount of renders because of new references. Because I ran out of time I did not double-check any performance changes in production mode.

@jtklein a quick run down of the new handleChange function to help you understand it for future debugging:

- It handles changes by *only* updating when there are actual Realm changes (insertions, deletions, modifications) or when specific conditions are met, like changing modes or changes in unsynced count. The goal here is to prevent unnecessary renders which slow down the whole UI.
- When there are new insertions, these get appended to the existing list instead of rebuilding the full list
- Instead of creating a new array every time like we were doing previous, we try to reuse the previous list when possible with spreading
- Using a single ref to track multiple pieces of state ensures that state all stays in sync, derives state like totalResults where possible instead of storing it separately, and limits unnecessary renders from extra state changes
- Caching filtered/sorted results ensures that we're not iterating over the same results twice unnecessarily, and it's more performance to filter, then sort as opposed to sorting, then filtering, then filtering again like we were doing in the previous code

PR #2927, opened 2025-05-28
Show a tooltip to find the obs wheel with all observation options

jtklein:
Closes MOB-784.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "react-native-vector-icons": "^10.2.0",
     "react-native-vision-camera": "4.0.5",
     "react-native-volume-manager": "^1.10.0",
+    "react-native-walkthrough-tooltip": "^1.6.0",
     "react-native-webview": "^13.8.4",
     "react-native-worklets-core": "^1.3.3",
     "realm": "^20.1.0",
--- a/src/components/AddObsModal/AddObsButton.js
+++ b/src/components/AddObsModal/AddObsButton.js
@@ -9,8 +9,8 @@ import { getCurrentRoute } from "navigation/navigationUtils.ts";
 import * as React from "react";
 import Tooltip from "react-native-walkthrough-tooltip";
 import { log } from "sharedHelpers/logger";
-import { useLayoutPrefs } from "sharedHooks";
-import useStore from "stores/useStore";
+import { useCurrentUser, useLayoutPrefs } from "sharedHooks";
+import useStore, { zustandStorage } from "stores/useStore";
 
 const logger = log.extend( "AddObsButton" );
 
@@ -22,15 +22,21 @@ const AddObsButton = ( ): React.Node => {
 
   const { isAllAddObsOptionsMode } = useLayoutPrefs( );
   const currentRoute = getCurrentRoute();
+  const currentUser = useCurrentUser( );
 
   // Controls wether to show the tooltip, and to show it only once to the user
   const showKey = "AddObsButtonTooltip";
   const shownOnce = useStore( state => state.layout.shownOnce );
   const setShownOnce = useStore( state => state.layout.setShownOnce );
+  const numOfUserObservations = zustandStorage.getItem( "numOfUserObservations" );
   // Only show the tooltip if the user has only AI camera as an option in this button.
   // Only show the tooltip on MyObservations screen.
   const triggerCondition = !isAllAddObsOptionsMode
-    && currentRoute?.name === "ObsList";
+    && currentRoute?.name === "ObsList"
+    // If logged out, user should see the tooltip after making their second observation
+    && !currentUser
+    && numOfUserObservations > 1;
+
   // The tooltip should only appear once per app download.
   const tooltipIsVisible = !shownOnce[showKey] && triggerCondition;
 
--- a/src/components/AddObsModal/AddObsButton.js
+++ b/src/components/AddObsModal/AddObsButton.js
@@ -13,7 +13,10 @@ import useStore from "stores/useStore";
 
 const logger = log.extend( "AddObsButton" );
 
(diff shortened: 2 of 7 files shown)


jtklein:
This requires this to be merged first: https://github.com/inaturalist/iNaturalistReactNative/pull/2932
Or if not the right solution to be updated with setting `shownOnce` to {} in all tests that are failing.

package.json
```diff
@@ -130,6 +130,7 @@
     "react-native-vector-icons": "^10.2.0",
     "react-native-vision-camera": "4.0.5",
     "react-native-volume-manager": "^1.10.0",
+    "react-native-walkthrough-tooltip": "^1.6.0",
```
albullington, line 133: I'm increasingly anxious about the vast number of libraries we are struggling to maintain. What's the main reason to use this instead of using the modal library we already have? Positioning? Did you try an implementation with modals?

src/components/AddObsModal/AddObsButton.js
```diff
@@ -2,25 +2,52 @@
...
   const closeModal = React.useCallback( () => setModal( false ), [] );
 
-  const resetObservationFlowSlice = useStore( state => state.resetObservationFlowSlice );
   const { isAllAddObsOptionsMode } = useLayoutPrefs( );
+  const currentRoute = getCurrentRoute( );
+  const currentUser = useCurrentUser( );
+
+  // Controls whether to show the tooltip, and to show it only once to the user
+  const showKey = "AddObsButtonTooltip";
+  const shownOnce = useStore( state => state.layout.shownOnce );
```
albullington, line 29: Can we rename `shownOnce` to be more descriptive about which components are relying on this global state? I can't keep track, but I know it's at least pivot cards and this tooltip. Angie's dismissable banner PR is going to add a similar store called `loginBannerDismissed` which has slightly different viewing conditions but is also shown once, so I'd love to see a name like that: `pivotCardsAndToolTipShown` or something along those lines.

src/components/AddObsModal/AddObsButton.js
```diff
@@ -2,25 +2,52 @@
...
+  const setShownOnce = useStore( state => state.layout.setShownOnce );
+  const numOfUserObservations = zustandStorage.getItem( "numOfUserObservations" );
+  // Only show the tooltip if the user has only AI camera as an option in this button.
+  // Only show the tooltip on MyObservations screen.
+  const triggerCondition = !isAllAddObsOptionsMode
+    && currentRoute?.name === "ObsList"
+    // If logged out, user should see the tooltip after making their second observation
```
albullington, line 36: appreciate the extra comments here

albullington (changes requested):
I'm expecting violent pushback on why we need a tooltip library, so I approve that part even though I'm concerned about it.

Requesting changes just on the naming, because the more things that are relying on `shownOnce`, the more important it is to understand what it's doing.

package.json
jtklein: This library seemed reasonably well maintained. I could try to do the same with existing modals we have but my question would be what is a good cost-benefit analysis here. Building it with the existing codebase has a an additional cost of work, incorporating a new library has a an additional cost of maintenance but where the net benefit is I don't know.

jtklein: The main reason was speed of delivering this feature.

src/components/AddObsModal/AddObsButton.js
jtklein: For sure we can rename this key, however, since it is already shipped to production and persisted on user's devices we would also need to write migration logic to not show the onboarding material again on existing user's devices. So, in essence the name might be vague but there are costs associated with renaming now. In any case what I can already do now is extract the tooltip shown into a boolean similar to `loginBannerDismissed` if that helps. Though I prefer the simplicity of `shownOnce` because is completely agnostic of what component it is used to control. Because we are just adding a key whenever a component is shown it does not matter which component it is as long as we are using the same key.

If we want to rename I would argue for another issue ticket for that.

package.json
kvangork: I've got my own preferences about how to approach tradeoffs like these, but instead of sharing them here I think it'd be best if we have a team conversation then agree to a strategy for future work.
If this code currently works, using the newly added dependency, let's merge it rather than holding and potentially re-implementing based on that discussion (added to next week's agenda).

src/components/AddObsModal/AddObsButton.js
kvangork: In what sense is this key already shipped to production if this PR isn't merged yet?
If it is shared by multiple components, the generic `shownOnce` name might be right as it would prevent future renaming when/if additional components use the key too.
If it's not shared, then I'm confused and would suggest adding a new explicitly named key for each new component (starting with this tooltip).

jtklein: It is shared by multiple components. The key of the persisted store is defined here: [shownOnce](https://github.com/inaturalist/iNaturalistReactNative/blob/7ce12fb66c1a9798a929f53e1dca7e21de4bddd3/src/stores/createLayoutSlice.ts#L63) and is already in production with this https://github.com/inaturalist/iNaturalistReactNative/pull/2667 since v1.0.0 as it also controls wether a pivot card has been shown already.
The basic implementation idea is that persisted under the key `shownOnce` is an object and each component that is only supposed to be shown once is represented as an uniquey key (with a boolean as value but ultimately doesn't matter) while the component key is only being added after it has been shown. So a check for: Does this key exist equals to this component has been shown already.
This object is persisted on user's devices with the key `shownOnce`, so if we were to rename it an existing user's app (since version 1.0.0) will start with an empty object and show all pivot cards again. Unless we migrate the persisted state.

PR #2936, opened 2025-06-03
Fix MOB 725 - reset password form - scroll view so keyboard won't hide field

budowski:


albullington (changes requested):
This approach has much cleaner code, but it unfortunately doesn't work for our use case. We want users to still be able to see the green LOG IN button when they tap on either field on the login screen, and we also want users to be able to see the RESET PASSWORD button when they tap the input field on the forgot password screen. Right now, those buttons are blocked, which introduces a regression on the Login screen from what we currently have in production.

Open to different options here, but I'd prefer a solution that *only* touches the Forgot Password screen and keeps our current Login screen untouched.

jtklein (changes requested):
I have not looked at the code yet in detail.
However, as mentioned here the Sign up screen and the Forgot password screen are noe behaving differently.
Can you update this PR so that the Forgot password screen reuses logic we have implemented in the Sign up screen (if it makes sense). So that both screens behave the same.

PR #2938, opened 2025-06-04
Fix: navigate bottom tabs to the correct root screen

albullington:
Bug found and detailed in [Slack](https://inaturalist.slack.com/archives/C02G5AATVGA/p1749043641077999?thread_ts=1749020630.123149&cid=C02G5AATVGA) while beta testing [v1.0.5](https://github.com/inaturalist/iNaturalistReactNative/releases/tag/v1.0.5%2B165).


jtklein (approved):
I don't actually consider this to be a bug. I think it is standard behaviour for the TabNavigator to remember which screen it is on (i.e. each tab has it's own history to go back to).
Anyways, I understand the current frustration and can Approve this change to make it back to how it was before. But I do think removing the drawer and it's navigator will free us up to do a more detailed review of our navigator stack and possibly revisit best practices again.

albullington:
@jtklein mostly agree with you here, but it's a pretty strange navigation setup in the first place since all of our tab screens need to access the exact same stack. My bigger gripe is that all the screens in the Drawer navigator also need to be part of the stack (it's endlessly confusing to me that some of our Drawer screens have back buttons and some don't), so I'm glad that's going to change in the future.

PR #2939, opened 2025-06-04
Remove block from Podfile that has no effect on being able to build

jtklein:
This came out of PR review of this here: https://github.com/inaturalist/iNaturalistReactNative/pull/2925
A realization that we have some code specific to be able to build on XCode 14 which no one of us uses anymore.
Removing this code block here has no effect on being able to build a Debug or Release build of the app on iOS on my machine.


albullington (approved):
Tested in development mode. This change will be helpful for avoiding unexpected errors in future package updates.

PR #2940, opened 2025-06-04
Remove XCode 14 podfile addition

jtklein:
This came out of PR review of this here: https://github.com/inaturalist/iNaturalistReactNative/pull/2925
A realization that we have some code specific to be able to build on XCode 14 which no one of us uses anymore.
Removing these code blocks here has no effect on being able to build a Debug or Release build of the app on iOS on my machine.


albullington (changes requested):
Screenshot 2025-06-04 at 5 16 33 PM

This one fails instantly and consistently for me with this error message ^, but I suspect this is due to some difference in dev environments or tooling. Any ideas on what I might need to change in my setup to match yours? I'm using the iNaturalist team signing, I'm on XCode 16.2, and I ran this set of commands plus `npm start -- --reset-cache`.

```
rm -rf node_modules/
cd ios
rm -rf Pods/
cd ..
npm i
cd ios
bundle exec pod install
```

PR #2945, opened 2025-06-06
Creative day: drawer and tab navigator dark mode

albullington:
Exploratory and not quite as straightforward as I had hoped, but working. 

Learnings:
- The Drawer menu can't access the `useColorScheme` hook in time for rendering, so we need to pass that as a prop.
- `darkModeStyleWrapper` is the way to get the entire app to know about dark mode so those classNames can be applied automatically. The only problem with this is that most of the examples I picked (`INatIcon`) currently can't accept a className and only accept a color prop, so we still need custom styling
- I needed to use an `isDarkModeEnabled` prop for now to prevent all the icons on other screens from being converted to dark mode before we're fully ready to support those screens

I was hoping we'd be able to upgrade to later versions of Tailwind (4 instead of 3) and Nativewind (4 instead of 2) but it seems Nativewind 4 doesn't yet support Tailwind 4. In theory, Nativewind 4 has better automatic dark mode detection.


albullington:
Design details here: https://inaturalist.slack.com/archives/C02G5AATVGA/p1749833032598929?thread_ts=1749750215.849219&cid=C02G5AATVGA

PR #2947, opened 2025-06-09
Fix: sharing into the app

albullington:
Specify the navigator, per the new React Navigation 7 API.


jtklein (approved):
Looks good to me.
Is this a result of the react-navigation update? Do we need a general audit then though of all navigate calls, to see if they are in line with react-navigation v7?

albullington:
> Looks good to me. Is this a result of the react-navigation update? Do we need a general audit then though of all navigate calls, to see if they are in line with react-navigation v7?

I suspect this is the last call I didn't catch while auditing, because I was searching for `navigation.navigate` and this is the one exception that uses `navigation?.navigate`. But yes, if we run into more issues it might be worth more auditing.

PR #2948, opened 2025-06-09
Testing: unit & integration tests for sharing photos

albullington:
This is core observation flow functionality that we didn't have any tests for, and unsurprisingly, we often run into unexpected bugs with sharing into the app. Here are some unit and integration tests which should help stabilize this part of the app and allow us to fix bugs/make upgrades/refactor as needed without breaking this functionality.


angielt (approved):
Looks good! Not a blocker but wondering if we could test that resetObservationFlowSlice is actually working when sharing photos in succession. A while back we had a bug related to previous shared photo information not being fully reset.

PR #2949, opened 2025-06-10
Update vision camera and plugin

jtklein:
There is no linear ticket for this but it is work that is required. A react-native update to 0.75 is currently not working as it does not compile on Android because of react-native-reanimated, updating react-native-reanimated to support Android breaks the AI camera in a test. So, I think this PR is blocking the RN update.

The packages updated to are exclusive to a specific version (not allowing higher versions). This is supposed to be like that for now because higher versions of vision-camera do not yet work as expected. So, while we are in a transitioning phase to higher versions of RN we have to keep an eye on vision-camera and update to higher versions at a later time.
For this PR, this version of the camera is working as expected on iOS and compiling succesfully on Debug and Release modes.
On Android there is now after this PR a new bug introduced that photos saved are not correctly rotated. I would still prefer to merge it in and come back to that later because this PR is blocking the next RN update, and an Android specific rotation patch is easy to install back in again later.


src/components/Camera/helpers/savePhotoToDocumentsDirectory.ts
```diff
@@ -0,0 +1,31 @@
...
+import { unlink } from "sharedHelpers/util.ts";
+
+const savePhotoToDocumentsDirectory = async (
+  cameraPhoto: PhotoFile
+) => {
+  const path = rotatedOriginalPhotosPath;
+  await RNFS.mkdir( path );
+  // Move the image with ImageResizer (for legacy reasons, because we
+  // used to use it to rotate the photo)
+  const image = await resizeImage(
```
jtklein, line 18: @albullington We used to include the rotation value here to make sure that newly taken photos are rotated to have the correct side up while they are copied to the temporary folder specified here. I have not analysed which steps are taken after this, if the photos used for uploading is the one from this temp folder, or if the photos is first saved to the photo library and uploaded from there?
So, I have not removed this piece that is saving the new photo to a temp folder immediately after taking it without changing rotation or dimensions. But I am wondering if we actually still need it if we are not changing the photo at all. According to the react-native-vision-camera documentation the takePhoto function: "returns a [PhotoFile](https://react-native-vision-camera.com/docs/api/interfaces/PhotoFile) which is stored in a temporary directory". So, depending on how long we want to keep reference to the new photo file we do actually might need to keep this temp folder copying stuff here.
But in any case we might "just" copy the file now instead of using image-resizer (but I would do that if necessary in a separate PR)

albullington: This is an interesting callout, and this is definitely a part of the codebase that always trips me up because it feels like we're doing a ton of extra steps.

I'd say the decision to store photos outside of the temp directory is something we can certainly discuss further. I'm not certain it's necessary for this to happen in the Camera, but it's something we need to do one way or another before the user saves an observation. It's there as a defensive coding strategy -- we don't have any control over when iOS clears the temp directory, so we need to be able to save our own references to the photos in a way that we *do* have control over, that are stored within the app, and can delete on our own terms. Otherwise, when a user saves all of their observations, iOS could clean up all of those photo references before the user ever gets around to uploading their photos, which might be hours, days, or weeks after the observations were taken. (Using references from the photo library is also a no-go because iOS aggressively caches photos and makes a user redownload from iCloud whenever they can, so we don't always have access to those photos when we want them).

albullington (approved):
Approve of the strategy here to get this working in iOS first and make an Android rotation patch at a later date. There are some helpful changes here, since I suspect removing extra photo processing and sharedHooks from children components like the FrameProcessorCamera will improve the camera performance -- this is a good step in the direction of separating business logic from UI concerns.

PR #2950, opened 2025-06-10
Update to react-native 0.75

jtklein:
This PR includes another update with a react native minor version jump to RN 0.75.x ([release notes](https://reactnative.dev/blog/2024/08/12/release-0.75)).
All packages that were blocking compilation on Android have been updated in previous PRs now, and on my machine at least this branch compiles without problems on Debug and Release builds on both platforms.
React v 18.3 is identical to 18.2 but adds warnings for deprecated APIs so there should not be anything to worry about.

Cleaning steps when switching to this branch should work the same as in the last RN update.

One thing that RN 0.75 changes is "Support for Percentage Values in Layout" which we kind of already had to a degree with nativewind. So, there have been two components that I had to update because the UI was just off. I did it quick and dirty knowing that I only want to focus on the RN update here. The changes in src/navigation/BottomTabNavigator/CustomTabBar.js should be stable for now. The changes in the two FloatingButtons.js components should be considered temporary because I don't think we want to have two differing implementations between default and advanced mode, so I would like to roll a future proof styling of these two components into an issue that also unifies these two components.


Gemfile
```diff
@@ -4,10 +4,9 @@ source "https://rubygems.org"
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby ">= 2.6.10"
 
-# Cocoapods 1.15 introduced a bug which break the build. We will remove the upper
-# bound in the template on Cocoapods with next React Native release.
-gem "activesupport", ">= 6.1.7.5", "< 7.1.0"
-gem "cocoapods", ">= 1.13", "< 1.15"
+# Exclude problematic versions of cocoapods and activesupport that causes build failures.
+gem "activesupport", ">= 6.1.7.5", "!= 7.1.0"
+gem "cocoapods", ">= 1.13", "!= 1.15.0", "!= 1.15.1"
```
albullington, line 9: Does this mean we can bump up our version of cocoapods as long as it's not `1.15.0` and not `1.15.1`? If so, I'd love to see you pick an exact version above `1.15.1` & let the mobile team know which version we need to be on going forward. That would let us finally merge [this PR](https://github.com/inaturalist/iNaturalistReactNative/pull/2868) and get dependabot working again instead of ignoring all security updates.

jtklein: Correct. I would like to try to get back to using the latest version of cocoapods but I want to do that in a follow-up PR so that this here is only a react-native update.

PR #2951, opened 2025-06-10
Dismissable banner that opens login/signup 

angielt:


src/components/SharedComponents/DismissableBanner.js
```diff
@@ -0,0 +1,88 @@
+// @flow
```
albullington: Can we make this TypeScript? We had agreed as a team a *long* time ago that all new files should be in TypeScript, and though we haven't been consistent about it, I think it's a good time to restart. We're seeing a lot of undefined TypeErrors in Grafana which show up in our JS files and are hard to track down. Right now they're impacting a low number of users, but this will only grow if we continue to build in JS.

src/components/MyObservations/MyObservationsEmptySimple.js
```diff
@@ -39,13 +44,37 @@ const MyObservationsEmptySimple = ( { currentUser, isConnected }: Props ): Node
...
+  };
+
   return (
     <ViewWrapper>
       {!!currentUser && (
         <View className="flex-start ml-[18px] mt-[26px]">
           <HeaderUser user={currentUser} isConnected={isConnected} />
         </View>
       )}
+      {
```
albullington: For readability, let's move this and the other additions above (useStore, dismissLoginBanner) into their own component called something like `LoginBanner`. When we can, it's good practice to create smaller components and keep functionality tightly coupled together. That way, a future developer who looks at this code will know *all* login banner code is in that component instead of needing to also try to understand the rest of the code on this Empty screen.

src/components/SharedComponents/index.js
```diff
@@ -19,6 +19,7 @@ export { default as CircleDots } from "./CircleDots";
 export { default as ConfidenceInterval } from "./ConfidenceInterval";
 export { default as DateDisplay } from "./DateDisplay";
 export { default as DateTimePicker } from "./DateTimePicker";
+export { default as DismissableBanner } from "./DismissableBanner";
```
albullington: I love the forward thinking of making `DismissableBanner` into its own component that can be shared across the app -- this is really solid engineering thinking -- *and* I'm wondering if this is a component that will actually get reused. Based on your product pod experience, do you foresee this being a one-off component or something that will only be used in this one use case?

src/stores/createLayoutSlice.ts
```diff
@@ -75,6 +75,14 @@ const createLayoutSlice = set => ( {
         ...state.layout,
         shownOnce: {}
       }
+    } ) ),
+    // State to control login/signup banner being only shown once until dismissed by user
+    loginBannerDismissed: false,
```
albullington, line 80: solid naming that makes me want us to rethink the `shownOnce` naming above which doesn't indicate what it's actually used for.

src/components/SharedComponents/index.js
angielt: Good point. I initially made it thinking it could potentially be used on other screens in the future but as far as the login pod is concerned it is temporary. I'll replace it with just a LoginBanner

src/stores/createLayoutSlice.ts
jtklein: See comment here:
https://github.com/inaturalist/iNaturalistReactNative/pull/2927#discussion_r2149737112

If you rename `shownOnce` because it is persisted on current user's devices you probably also need to write migration logic to not show them the onboarding material again. I would argue to do that in another PR if required.

angielt: Thanks for pointing that out. I'll revert the change to shownOnce from this PR and create another PR with those changes.

angielt:
Merging since changes requested were mainly for style

PR #2952, opened 2025-06-10
Fix: isLoggedIn memory leak

albullington:
Closes Mob-808

I tracked down a memory leak where every once in a while, RNSensitiveInfo would fail saying it had been called an excessive amount of times in the console. There were a few problems here:

- `useEffect` was running in `useAuthenticatedQuery` anytime queryOptions or queryKey changed, for a total of 54 immediate calls to RNSensitiveInfo on app start. I changed the dependency array to be empty so this is only called once per component instance (~9 calls).
- `isLoggedIn` was fetching from RNSensitiveInfo every time, even if we already knew the user was logged in or out, so each of those 9 remaining calls would still fetch the same data from RNSensitiveInfo. To address this, I stored the boolean value in a temporary cache, which is cleared when a user logs in or out. If we've already accessed that value within the last 5 seconds (which we often do when a user lands on the home screen and needs lots of data at once, for example), components will receive the cached version.

Result: `isLoggedIn` is still calls RNSensitiveInfo ~7 times when a user first logs in, with another ~10 simultaneous `isLoggedIn` requests returning the cached value instead, for an 87% improvement in the number of rapid-fire fetches to RNSensitiveInfo.


src/sharedHooks/useAuthenticatedQuery.js
```diff
@@ -16,17 +16,19 @@ const useAuthenticatedQuery = (
...
-  // depends on whether the user is signed in. The possible vulnerability
-  // here is that this effect might not run frequently enough to change when
-  // a user signs in or out. The reason we're not using useCurrentUser is it
-  // doesn't tell us whether we know the user's auth state yet, it only
-  // returns null when we don't know OR the user is signed out.
   useEffect( ( ) => {
-    isLoggedIn()
-      .then( result => setUserLoggedIn( result ) )
-      .catch( ( ) => setUserLoggedIn( LOGGED_IN_UNKNOWN ) );
-  }, [queryKey, queryOptions] );
```
jtklein, line 29: While I support the effort to remove useEffects as much as possible from the codebase you did not provide any explanation as to why it is okay to remove those trigger conditions. As a reviewer now in principle I should check every instance of calling `useAuthenticatedQuery.js` to see if we are not relying on those triggers somehow. But that would greatly increase the review time.
Checking to see if `queryKey` is null(ish) and queryOptions do also not make sense for an actual useQuery call might have been an alternative here.

jtklein (approved):
I think these changes are a good way to deal with too many requests to RNSensitiveInfo. I also think though that the library should be robust enough to handle a lot of requests in a short time.
In general it is a good idea to get rid of unnecessary useEffects throughout the codebase.
The authCache adds a layer of complexity that I would like to be able to avoid. But I don't  know what alternative to propose.

src/sharedHooks/useAuthenticatedQuery.js
albullington: Happy to explain further, because I think it's an important lesson in good architecture. The previous logic made an assumption that this `useEffect` couldn't possibly know whether a user had logged in or out of the app, which is why there were dependencies to trigger the effect over and over just in case. With the auth cache, we're intentionally clearing the cache anytime a user signs in or out. When these 9 components that immediately call `useAuthenticatedQuery` in quick succession are mounted, they will receive the latest `isLoggedIn` status directly from sensitive-info since we've cleared the cache. These hooks are remounted when a user changes between stack navigators (i.e. from the LoginStack to the TabNavigatorStack) so they will receive the accurate login status on the first mount.

In general, I prefer more predictable code like this where we know what's expected to happen, rather than relying on side effects like `queryKey` changing where it's near impossible to know what's happening or why an effect is calling this function 54 times in a row. If we're relying on side effects, it's a good indication that we can write more declarative code.

PR #2954, opened 2025-06-11
Update react native reanimated

jtklein:
Version 3.15.x of react-native-reanimated is the minimum version that supports react-native 0.75.x.
This PR is blocking the update to RN 0.75 because in the branch of the RN update the app did not compile on Android with an error message that react-native-reanimated is the culprit. (Have tested to update locally and this PR here actually resolves this issue).


albullington (approved):
Dev and prod builds are working in both iOS and Android for me.

PR #2958, opened 2025-06-13
Fix: sharing into app with multiple photos

albullington:
Closes Mob-812

I could repro this by sharing into the app with multiple photos, then backing out to the home screen. If I did that, then tried to share multiple photos again, nothing would happen. The reason for this was that if a user landed on the PhotoSharing screen a second time in a row, they would automatically be sent back to the home screen because of our navigation listeners.

The fix here is to keep a reference of the previous item. If a new item is shared, we don't immediately navigate backwards. If a user is backing out of this screen from ObsEdit, we still navigate them back.

Also, I noticed that GroupPhotos could be stacked on top of itself, so I applied the navigator reset for that navigation flow to match what we're doing for single photo shares.


ios/.xcode.env
```diff
@@ -1,10 +0,0 @@
-# This `.xcode.env` file is versioned and is used to source the environment
```
jtklein: The removal of this file is unrelated to the linear ticket, why include here?

src/components/PhotoSharing.js
```diff
@@ -141,17 +125,27 @@ const PhotoSharing = ( ): Node => {
...
-  useEffect( ( ) => {
-    const unsubscribe = navigation.addListener( "focus", ( ) => {
-      if ( navigationHandled ) navigation.goBack( );
+  useEffect( () => {
+    const unsubscribe = navigation.addListener( "focus", () => {
+      const isNewShare = item !== previousItem.current;
+
+      // don't navigate backwards if there are new items shared, otherwise
+      // we end up back on the home screen without the user ever seeing their shared items
+      if ( isNewShare ) {
```
jtklein, line 137: This does solve the issue at hand, however, if I understand everything here correctly, I suspect this does mean that the PhotoSharing does not get destroyed and is still kept in the stack?
So in the following navigation
Import photos
-> PhotoSharing
-> GroupPhotos
-> HomeScreen (because of backing out)

I would expect both PhotoSharing and GroupPhotos to be removed from the stack and for the next time of sharing photos a new PhotoSharing to be created.

jtklein (approved):
I am not saying we can't remove the env file, but it seemed unrelated, so preferably in a different PR (with explanation as to why it is necessary to remove it).

PR #2961, opened 2025-06-13
Fix: correct navigation from TaxonSearch while suggesting an ID

albullington:
Closes Mob-819

Repro steps: go to explore, tap an observation, suggest an ID from ObsDetails, land on Suggestions, tap on search, choose "plants" or anything else, screen would crash because navigation sent the user to ObsEdit

Fix: SuggestionsTaxonSearch keeps track of the lastScreen being ObsDetails, so it can navigate back there as expected


jtklein (changes requested):
This does not work for me. When I follow your repo steps I can still get a crash in this PR albeit a different one now:
` ERROR  Warning: TypeError: Cannot read property 'taxon' of undefined

This error is located at:
    in ObsDetailsContainer (created by FadeInObsDetailsContainer)
    in RCTView (created by View)
    in View (created by Animated(View))
    in Animated(View) (created by FadeInView)`

in line 455 of ObsDetailsContainer

jtklein:
No special steps really, I followed the repro steps you outlined above: go to explore, tap an observation, suggest an ID from ObsDetails, land on Suggestions, tap on search, choose "plants" or anything else, screen would crash because navigation sent the user to ObsEdit

I did not see any error though when the observation on Explore was my own but I guess that is expected, as I have the power to edit my own obs.

I can try again tomorrow with additional observations if there was something special about the ones I pressed on by chance.

jtklein:
BB2E45D8-31A8-4683-B07F-C6C07619B7AD_4_5005_c
BB2E45D8-31A8-4683-B07F-C6C07619B7AD_4_5005_c

jtklein (approved):
I can approve with the added safeguard in ObsDetailsContainer.

PR #2962, opened 2025-06-16
Format additional types of dates to not crash projects

albullington:
Closes Mob-816

Fix crash from a different type of time returned from the API: "June 1, 2018" format.


jtklein (approved):
Looking good. I do agree though that it seems odd to receive different date formats in the same key.

PR #2965, opened 2025-06-17
MOB-711 - add location for better IDs change

budowski:


src/components/Match/EmptyMapSection.js
```diff
@@ -16,22 +17,26 @@ const EmptyMapSection = ( {
...
         source={require( "images/topographic-map.png" )}
         accessibilityIgnoresInvertColors
       >
-        <INatIcon name="location" size={40} color={colors.white} />
-        <Subheading2 className="mt-3 text-white px-28 text-center">
-          {t( "Add-location-for-better-identifications" )}
-        </Subheading2>
+        <INatIcon name="map-marker-outline" size={40} color={colors.white} />
+
+        <Pressable
```
jtklein: We should use our `<Button />` component for this.
Please also make sure there is a margin of 15px left and right of this button.

tests/unit/components/Match/Match.test.js
```diff
@@ -49,7 +49,7 @@ describe( "Match", ( ) => {
       />
     );
 
-    const addLocationButton = screen.queryByText( /ADD LOCATION FOR BETTER IDS/i );
```
jtklein, line 52: This is failing in the tests because there are now two buttons with the same text. That makes the test fail.
Can you update this to instead still check for this button text which is what the user would be looking for, but make sure it passes with two identical buttons on the screen.

jtklein (changes requested):
Please use the Button component and update the tests.

tests/unit/components/Match/Match.test.js
budowski: I did exactly the change you requested - it's not supposed to fail - are you using the latest version?

src/components/Match/EmptyMapSection.js
budowski: I used the`Button` component at the beginning, but then there were a bunch of properties it does not allow changing - like the text size and border

jtklein: Here is another screenshot from the design files, the button in the new location section and the one on the bottom of the screenshot are only differing in the border outline. Text size should be the same (Heading 4). The primary version of our Button component should not have a border, so you should be fine to use it: https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=3302-20109&t=ClrNJ8P5c3KrZfGW-4

Screenshot 2025-06-20 at 12 38 39

tests/unit/components/Match/Match.test.js
jtklein: Sorry, I meant to say the reason why it was failing for you before you did this change was that now there are two buttons with this text "ADD LOCATION FOR BETTER IDS" instead of one. In such a case the test fails. Instead of making the test search for a different label I would prefer to keep searching for the button text (as this is what the user sees) but make it pass with two buttons instead.

src/components/Match/EmptyMapSection.js
jtklein: Thanks, that's perfect.
I only moved the view wrapper around a bit so that the button always fills the entire screen (independent of text size) as do the other buttons down below.

tests/unit/components/Match/Match.test.js
jtklein: Thanks, works as expected now.

PR #2980, opened 2025-06-26
Add conditions for which the Obswheel tooltip is shown

jtklein:
The ticket for this work is MOB-785.

Most conditions given in the ticket are functioning as expected. For reference I ticked off everything that works as expected in the linear ticket, at least when I am testing. Two points in the checklist work a bit different than what is written in the ticket:
In cases were the tooltip would collide with a pivot card the linear ticket outlined that the tooltip should be shown immediately after the pivot card is dismissed. I was not able to make this happen in a reasonable timeframe (my estimate of reasonable).
So, what is happening in those two cases instead is that when a user dismisses the pivot card, then closes app or moves away from ObsList, then opens app or moves back to ObsList the tooltip is shown. I think this is acceptable to close the ticket.


tests/unit/components/AddObsModal/AddObsButton.test.js
```diff
@@ -0,0 +1,69 @@
...
+    renderComponent( <AddObsButton /> );
+
+    const tooltipText = screen.queryByText(
+      "Press and hold to view more options"
+    );
+    expect( tooltipText ).toBeFalsy();
+  } );
+} );
+
+describe( "shows tooltip", () => {
```
albullington, line 47: It looks like you're skipping a test for the `justFinishedSignup` state here. Any reason for that?

albullington (approved):
Nice to have would be full coverage for all conditions when the tooltip is shown, but tests can always be added in the future.

PR #2987, opened 2025-06-27
Update react-native-reanimated

jtklein:
This PR updates react-native-reanimated to the currently latest version as it is [compatible](https://docs.swmansion.com/react-native-reanimated/docs/guides/compatibility/) with RN 0.75 (currently on main).

I have compiled and tested on Debug and Release on both platforms and found no issues so far.


albullington (approved):
Approved once it passes e2e tests.

PR #2988, opened 2025-06-27
Update react native gesture handler 2.22.1

jtklein:
Update of a package in preparation for RN 0.77, react-native-gesture-handler v2.22.0 introduces support for RN 0.77.x.

Have compiled for Debug and Release and tested the camera and zooming into a photo in the media viewer which are the two components I can see this library being imported.


albullington (approved):
Approved once tests pass.

PR #2996, opened 2025-07-01
Performance: speed up Notifications

albullington:
Closes MOB-853

- Fixes a memory leak where `useEffect` would run an infinite amount of times when returning from ObsDetails by running an unsubscribe cleanup on the add focus listener
- Fetch only the fields needed from API (photos, sounds, user_id) instead of fetching and passing around very large observation objects

### DIFF
--- a/src/components/Notifications/NotificationsContainer.tsx
+++ b/src/components/Notifications/NotificationsContainer.tsx
@@ -44,11 +44,12 @@ const NotificationsContainer = ( {
   }
 
   useEffect( ( ) => {
-    navigation.addListener( "focus", ( ) => {
+    const unsubscribe = navigation.addListener( "focus", ( ) => {
       if ( isConnected && currentUser ) {
         refetch();
       }
     } );
+    return unsubscribe;
   }, [isConnected, currentUser, navigation, refetch] );
 
   const onRefresh = async () => {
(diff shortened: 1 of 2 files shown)


src/sharedHooks/useInfiniteNotificationsScroll.ts
```diff
@@ -67,7 +67,22 @@ async function fetchObsByUUIDs(
   // TODO convert api/observations to TS
   const observations: ApiObservation[] | null = await fetchRemoteObservations(
     uuids,
-    { fields: Observation.FIELDS },
+    {
```
jtklein, line 70: Does it make sense to turn these field sets into constants, similar to `Observation.ADVANCED_MODE_LIST_FIELDS` that you created for observation fetching.

src/components/Notifications/NotificationsContainer.tsx
```diff
@@ -44,11 +44,12 @@ const NotificationsContainer = ( {
   }
 
   useEffect( ( ) => {
-    navigation.addListener( "focus", ( ) => {
+    const unsubscribe = navigation.addListener( "focus", ( ) => {
```
jtklein, line 47: Is this a universal pattern? I have found two other instances in the codebase where we do not return the addListener result.

albullington: Yes, this often leads to memory leaks if there is no cleanup -- it's one of the first things to look for as far as performance issues in a React Native app.

src/sharedHooks/useInfiniteNotificationsScroll.ts
albullington: No strong preference here. It seems reasonable, and also I don't think these fields are being used anywhere else in the app so it's fine for it to live here.

PR #2997, opened 2025-07-02
Performance: use only API fields needed for project lists

albullington:
didn't make an issue for this, but falls under the general app is slow bucket

### DIFF
--- a/src/components/Projects/hooks/useInfiniteProjectsScroll.ts
+++ b/src/components/Projects/hooks/useInfiniteProjectsScroll.ts
@@ -7,10 +7,19 @@ import {
 const useInfiniteProjectsScroll = ( { params: newInputParams, enabled }: object ): object => {
   const baseParams = {
     ...newInputParams,
-    per_page: 50,
+    per_page: 20,
     ttl: -1,
     rule_details: true,
-    fields: "all"
+    fields: {
+      id: true,
+      project_type: true,
+      title: true,
+      icon: true,
+      rule_preferences: {
+        field: true,
+        value: true
+      }
+    }
   };
 
   const { ...queryKeyParams } = baseParams;


src/components/Projects/hooks/useInfiniteProjectsScroll.ts
```diff
@@ -7,10 +7,19 @@ import {
 const useInfiniteProjectsScroll = ( { params: newInputParams, enabled }: object ): object => {
   const baseParams = {
     ...newInputParams,
-    per_page: 50,
+    per_page: 20,
```
jtklein, line 10: Does this actually have an effect on performance, like is the time it takes to retrieve one such request smaller?
I would assume that overall if we have to load a large list of projects needing to splice the dataset into smaller chunks (as with 50) is better for performance.

src/components/Projects/hooks/useInfiniteProjectsScroll.ts
```diff
@@ -7,10 +7,19 @@ import {
...
     ttl: -1,
     rule_details: true,
-    fields: "all"
+    fields: {
```
jtklein, line 13: Does it make sense to turn these field sets into constants, similar to `Observation.ADVANCED_MODE_LIST_FIELDS` that you created for observation fetching.

jtklein (approved):
Only questions, but I am not convinced that changing from 50 to 20 makes faster API call turnaround (but does require more calls for longer lists).

src/components/Projects/hooks/useInfiniteProjectsScroll.ts
albullington: You're welcome to experiment with that! I made the `PerformanceDebugView` thinking it would be helpful for future developers to try out theories like this. Generally though, all API calls should reflect what we actually want to show the user -- in this case 20 is already about double the number of project items they'll see on first load on an iPhone device with the current UI, so it seems more than sufficient.

PR #2998, opened 2025-07-02
Performance: fetch only needed API fields for Project Details

albullington:
also re: app is slow


src/components/ProjectDetails/ProjectDetailsContainer.js
```diff
@@ -30,7 +30,15 @@ const ProjectDetailsContainer = ( ): Node => {
   const { data: project } = useAuthenticatedQuery(
     fetchProjectsQueryKey,
     optsWithAuth => fetchProjects( id, {
-      fields: "all"
+      fields: {
+        id: true,
```
jtklein, line 34: Does it make sense to turn these field sets into constants, similar to `Observation.ADVANCED_MODE_LIST_FIELDS` that you created for observation fetching.

PR #2999, opened 2025-07-02
Performance: limit API fields for Explore observations fetch

albullington:
Limit Explore Observation fields to what's displayed in list view. Re: app is slow


jtklein (approved):
LGTM

PR #3000, opened 2025-07-02
Update react-native-safe-area-context to latest

jtklein:
Update of a package in preparation for RN 0.77, react-native-safe-area-context v5.1.0 introduces support for RN 0.77.x. on Android. But the latest version also compiles.

Have compiled for Debug and Release on iOS and for Debug on Android (having some build issues on Android but that should not matter for this PR) and tested the app in places where a notched phone should show correct UI.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1406,7 +1406,7 @@ PODS:
     - React-Core
   - react-native-restart (0.0.27):
     - React-Core
-  - react-native-safe-area-context (4.14.1):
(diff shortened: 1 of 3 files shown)


angielt (approved):
The notches look good to me on iOS and Android. I did run into this when running it on android and not sure if it is a regression or not. It looked like this for a brief moment but returned back to normal when opening and closing the modal. I think the tooltip appeared like this initially. If we are just caring about notches for this PR than its fine. 

image

PR #3008, opened 2025-07-08
Assert that observationPhoto position is a number

jtklein:
Closes MOB-844.
I was not able to replicate MOB-844 in a real production environment, but I was able to trigger the same error in development by setting `position` to null or undefined when changing observation photo positions in the ObsEdit screen.

So, to tackle this I first started migrating the relevant files to TypeScript to get an understanding of how anything other than a number can be sent to the API. Therefore, this PR includes a lot of TS migrations, and also quite a bit of cleanup from the current state: some function params were actually not used, many types not given.
I have limited myself to a "reasonable" amount of time spent on migrating the functions in `ObservationPhoto.ts` and `Photo.ts` that I do not actually needed for this bugfix. So, as a result there are still some TS warnings present in those files. But fixing those requires a deeper understanding of the places where those functions are used than I could gather in this time.

The actual fix in the code to avoid this error was pretty minor though:
In ObservationPhoto.ts, `mapPhotoForUpdating` returns an object with defined shape, so even if position itself is undefined the key is added. So in `prepareMediaForUpload` we have to check the shape of the object sent to the API (and remove the key if the value is not a number).

This does resolve one kind of error happening when trying to upload an observation that blocks the observation from ever being uploaded. So the the priority for this PR is "urgent".


src/realmModels/ObservationPhoto.ts
```diff
@@ -61,17 +75,23 @@ class ObservationPhoto extends Realm.Object {
...
+  // but it is not important for my current linear ticket so I'll skip typing it more
+  static mapObservationPhotoForMyObsDefaultMode( observationPhoto: {
+    photo?: { url?: string, localFilePath?: string },
+    uuid?: string
+  } ) {
     return {
       photo: {
-        url: obsPhoto?.photo?.url,
-        localFilePath: obsPhoto?.photo?.localFilePath
+        url: observationPhoto?.photo?.url,
```
jtklein, line 87: Renamed for consistency.

angielt (approved):
I'm seeing the importance of using Typescript a lot more when reviewing this PR. 
Similar to you, I was never able to replicate the error as mentioned in the issue in a production environment but I tested it using the method you described and it seems good. It'll help to get this in to see if the errors lessen since I saw a higher frequency of these types of errors in grafana.

PR #3013, opened 2025-07-09
Fix: Observations shown without data

jtklein:
Closes MOB-866
Observations that were fetched from server on the notifications screen were inserted into the local database with limited data, because we started requesting less data from the API. The major problem with that was that they were interpreted as having a "Missing date" which on the Me tab sorts them to the top of the list. While not being a data corruption issue per se it looks like something is wrong with the data.
I switched back to request more fields again for observations loaded from notifications. Instead of all fields, we now request the "Advanced fields" which is the max of data we show anywhere in the app.

### DIFF
--- a/src/sharedHooks/useInfiniteNotificationsScroll.ts
+++ b/src/sharedHooks/useInfiniteNotificationsScroll.ts
@@ -70,7 +70,8 @@ async function fetchObsByUUIDs(
     {
       fields: {
         user: {
-          id: true
+          id: true,
+          login: true
         },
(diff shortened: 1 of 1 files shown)


angielt (approved):
I ran into a bug when reviewing that showed unknown in notifications for the current user. I saw that the change in the notifications fetch fields to ADVANCED_MODE_LIST_FIELDS was missing a user field so I added that back into the notifications fetch. 
Please review if you think it is needed but otherwise it works fine with these changes. 
Simulator Screenshot - iPhone SE (3rd generation) - 2025-07-09 at 10 39 32

PR #3014, opened 2025-07-09
Update @react-native-community/geolocation to latest

jtklein:
After updating this library, I am no longer able to replicate these two location-fetching related issues: MOB-815, MOB-758.

So I believe this closes those issues.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -1343,7 +1343,7 @@ PODS:
     - React-Core
   - react-native-geocoder-reborn (0.9.0):
     - React
-  - react-native-geolocation (3.2.1):
+  - react-native-geolocation (3.4.0):
     - DoubleConversion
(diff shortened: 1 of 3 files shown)


angielt (changes requested):
I'm seeing this package update fix most of the issues listed so I am open to adding it to the build this week. 

However, I did encounter this during the testing but It might be related to this [issue](https://linear.app/inaturalist/issue/MOB-714/immediately-after-granting-location-permissions-add-location-is-not)
1. I took a photo and backed out of match screen before processing was done.
2. took another photo and when I got to the match screen and I allowed location permissions nothing happened. 
3. backed out of that screen and took another photo and location appeared

https://github.com/user-attachments/assets/c1b54a0c-2db0-4f9c-bf91-cd1ac930f74b

PR #3015, opened 2025-07-11
Recovery from observations not uploading

jtklein:
We added a debug button to the app, to log in again. This has been tested by some user, and it does resolve the error of not being able to upload their observations. So, in this PR I am adding a recovery option whenever we hit the error of trying to upload an observation without api token (which will surely fail server side).
As recovery:
Navigate to login screen to login again (which replenishes the api token).
In addition, stop all ongoing uploads if multiple have been queued because they for sure also would error out.

I would not say this PR solves the ticket MOB-851, because it is still possible for the app to arrive at the error state of user not being able to upload. However, I think this workaround is enough to remove the "urgent" label for now.

### DIFF
--- a/src/components/StartupService.js
+++ b/src/components/StartupService.js
@@ -90,6 +90,9 @@ const StartupService = ( ) => {
       };
       // don't remove this logger.info statement: it's used for internal metrics
       logger.info( "pickup" );
+      logger.info(
+        `App version: ${DeviceInfo.getVersion()}, build: ${DeviceInfo.getBuildNumber()}`
+      );
 
       try {
         await checkForSignedInUser( );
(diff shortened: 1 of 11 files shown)


src/components/StartupService.js
```diff
@@ -90,6 +90,9 @@ const StartupService = ( ) => {
       };
       // don't remove this logger.info statement: it's used for internal metrics
       logger.info( "pickup" );
+      logger.info(
+        `App version: ${DeviceInfo.getVersion()}, build: ${DeviceInfo.getBuildNumber()}`
```
jtklein, line 94: This makes looking at log files a bit easier, so we know what version of the code they are actually running.

angielt (approved):

PR #3020, opened 2025-07-16
Remove suggestions screens when navigating back after choosing a taxon

angielt:
[MOB-846](https://linear.app/inaturalist/issue/MOB-846/swiping-back-after-adding-id-in-explore-doesnt-work-as-expected)

This seemed to be happening to me when adding an ID in explore but also adding an ID on an observation from MyObs too in both default and advance mode.


angielt:
@jtklein Weird. After the changes I am not seeing it go back to suggestions after returning to ObsDetails. Can you give me steps to reproduce? 

https://github.com/user-attachments/assets/75aa98eb-0e06-4181-adab-de77843dd986

jtklein (approved):
LGTM

PR #3022, opened 2025-07-17
Repair ObsWheel tooltip

jtklein:
Almost closes MOB-785 (see below).

Basically, I think one of the reasons this tooltip was buggy before was that we have a lot of modals opening and closing at the same time. In the case of a user logging in with >50 obs, they would see AccountCreation onboarding modal, tooltip modal, AddObs modal, all in tight sequence.
So, since the only option to leave the tooltip should be to make a long press on the GradientButton that then directly opens the ObsWheel modal, I decided to move the tooltip into the ObsWheel modal itself, and a long-press just switches to the other content. This way we can remove the need for one npm package, and get rid of one modal to modal transition.
Overall, this fixes all bugs I have previously encountered with the tooltip.

One side note though: While I was testing the condition of a user hat just signs up I encountered a bug that the Account creation pivot card is not showing when the user signs up before making one observation. This messes with the trigger condition for the tooltip which should show directly after closing the account creation pivot card. So, I have for now not implemented the tooltip for user that just signed up.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -96,7 +96,6 @@
         "react-native-vector-icons": "^10.2.0",
         "react-native-vision-camera": "^4.7.0",
         "react-native-volume-manager": "^1.10.0",
-        "react-native-walkthrough-tooltip": "^1.6.0",
         "react-native-webview": "^13.15.0",
         "react-native-worklets-core": "1.5.0",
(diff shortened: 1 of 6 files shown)


angielt (approved):
Looks good to me

PR #3030, opened 2025-07-27
MOB-870 - fix - account creation pivot card not showing up immediately, if no observations were made

budowski:


jtklein:
Looks good to me

PR #3072, opened 2025-08-25
MOB-835 - show info icon and link in taxon search results when searching from explore filters modal

budowski:


jtklein:
Hey Yaron, the change is correct towards the actual ticket. The info button does now appear when the search is accessed from the Explore Filters Modal.
However, I think this is not mergeable like so because pressing the "i" button does not seem to do something at first glance. It does actually navigate to Taxon details, but the navigation is happening behind the opened modal. So, I think we need input from Product/Design on how to proceed with this one. Or, what I would prefer, to not fix this ticket at all.

PR #3074, opened 2025-08-26
Migrate `src/sharedHelpers/breakpoint` to TypeScript.

frewsxcv:
I was reading through some of the code that relies on this file and improved some of the typing along the way.


jtklein:
Thanks for your PR. Migrating from flow types to TS is something we do gradually at the moment, so any help there is greatly appreciated.
First, could you make sure your git history shows that the file was moved, and not deleted and another one created.
`git mv src/sharedHelpers/breakpoint.js src/sharedHelpers/breakpoint.ts` should do the trick. Then the diff is easier to digest for a reviewer.
Second, why are you changing the `BREAKPOINTS` in the ways you did? Is there any reasoning behind these changes, are they necessary? I have not looked into the changes in detail yet. Before I spend some time on it, I would like to see a rationale. (I am thinking: "If it ain't broke, don't fix it.")

frewsxcv:
Thanks for taking a look!

> First, could you make sure your git history shows that the file was moved, and not deleted and another one created.
`git mv src/sharedHelpers/breakpoint.js src/sharedHelpers/breakpoint.ts` should do the trick.

From what I understand about git operations and the index, `git mv` is equivalent to `git rm N && git add N`, so if the diff isn't rendering, that's a GitHub UI thing. For situations like this I usually use things like https://mergely.com/ to see a diff.

> Second, why are you changing the BREAKPOINTS in the ways you did? Is there any reasoning behind these changes, are they necessary? I have not looked into the changes in detail yet. Before I spend some time on it, I would like to see a rationale. (I am thinking: "If it ain't broke, don't fix it.")

Fair! I'll add some pull request comments right now to go into more detail. If you're worried about the functionality changing I can add a unit test to confirm the result is the same since there doesn't appear to be any tests right now.

src/sharedHelpers/breakpoint.ts
```diff
@@ -0,0 +1,29 @@
+import screens from "styles/tailwindScreens";
+
+export const BREAKPOINTS
+  = Object.fromEntries(
+    Object.entries( screens || {} ).map( ( [breakpoint, widthString] ) => {
+      if ( typeof widthString !== "string" ) {
+        throw new Error( `Unexpected breakpoint value: ${widthString}` );
+      }
```
frewsxcv: Image

The typing of `screens` is somewhat complicated because Tailwind allows so many different formats for the width value of the breakpoint config. And because `parseInt` requires a `string` be passed in, we need to ensure that `widthString` is indeed a `string`, hence this line.

frewsxcv:
@jtklein I restored the `_.reduce` usage which cleaned up the diff, and now GitHub renders the diff for that file.

jtklein:
Thanks for cleaning this up. To just expand a little bit, yes `git mv` is equivalent to `git rm N && git add N`. I will keep preferring and suggesting it though because it forces you to do the file name change before you do any code changes. I also mostly then commit just the file name change. If you do a `git rm N && git add N` after you have already made substantial changes to the code you rely on the diffing algorithm to pick up on the act that it is a filename change as well.

PR #3079, opened 2025-08-27
Update `LoginSignUpWrapper` to gracefully handle screen size changes.

frewsxcv:
Prior to this pull request, the `LoginSignUpWrapper` would fetch the screen size _once_, and use that to determine the minimum height of the scroll view content. Because screen sizes are dynamic (e.g. rotation, or resizing apps), this can cause isses with the layout math.

After this pull request, this component will dynamically recalculate the screen size math whenever the screen dimensions change.


src/components/LoginSignUp/LoginSignUpWrapper.tsx
```diff
@@ -36,6 +30,8 @@ const LoginSignupWrapper = ( {
 }: Props ) => {
   const scrollViewRef = useRef( null );
   const navigation = useNavigation( );
+  const { height: windowHeight } = useWindowDimensions( );
+  const scrollViewMinHeightRef = useRef<number>( windowHeight * SCROLL_VIEW_SCALE );
```
jtklein, line 34: Please remove the useRef and useEffect parts.
You can follow the example in this file `FlashListEmptyWrapper.tsx`.

src/components/LoginSignUp/LoginSignUpWrapper.tsx
```diff
@@ -6,15 +6,13 @@ import {
 import React, {
   PropsWithChildren, useEffect, useRef
 } from "react";
-import type {
+import {
```
jtklein, line 9: Please leave the imports of types separated.

jtklein (changes requested):
Thanks you for this contribution, this is not something I see as a current priority though because we are locking the orientation of this screen when rotating the phone, and we do not support tablets at the moment. Curious were did you come across this being an issue?

PR #3080, opened 2025-08-27
Remove "Always" location permission string

kvangork:
- Remove NSLocationAlwaysAndWhenInUseUsageDescription from Info.plist
- Remove corresponding localization strings from InfoPlist.xcstrings
- Prevents "Always" option from appearing in iOS location settings

Closes MOB-745

Result:
image


jtklein (approved):
LGTM

PR #3086, opened 2025-08-31
Limit Jest usage to test files; conslidate ESLint config.

frewsxcv:
Prior to this commit, technically Jest could have been imported in production files and ESLint would have been okay with that since it was allowed across the codebase.

After this commit, ESLint will only allow Jest usage in test files. In addition, the 3-line test-specific ESLint config was consolidated into the main config.


jtklein:
Thanks for cleaning this up a bit.

I have run e2e tests locally, so this LGTM.

PR #3088, opened 2025-08-31
Disable spellcheck functionality in login form fields.

frewsxcv:
Also resolved an ESLint warning:

```
  21:30  warning  The `Function` type accepts any function-like value.
Prefer explicitly defining any function parameters and return type  @typescript-eslint/no-unsafe-function-type

✖ 1 problem (0 errors, 1 warning)
```

# Before

Screenshot 2025-08-31 at 12 50 04 PM

# After

Screenshot 2025-08-31 at 12 49 29 PM

### DIFF
--- a/src/components/LoginSignUp/LoginSignUpInputField.tsx
+++ b/src/components/LoginSignUp/LoginSignUpInputField.tsx
@@ -18,7 +18,7 @@ const OUTLINE_STYLE = {
   borderRadius: 8
 };
 
-const LoginSignUpInputField: Function = forwardRef( ( {
+const LoginSignUpInputField = forwardRef( ( {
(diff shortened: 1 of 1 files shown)


frewsxcv:
Spell checking is for finding typos, and users are trained to treat red underlining as problematic. Because usernames, emails, first names, etc. are not words we would find in dictionaries, we shouldn't expect the spellchecking to helpful functionality here. Separate from that, there are also [privacy considerations](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/spellcheck#security_and_privacy_concerns) with spell checkers, and the general recommendation from product and designers I've worked with is to disable it on login and signup form field inputs.

PR #3110, opened 2025-09-22
MOB-860 - empty states for user's followers/following/projects screens

budowski:


jtklein (changes requested):
The project screen looks as it should.
However, the follower/following lists show the new "empty" content but also the activity indicator at the same time. Once the list fetching is complete and there are no data to display the indicator should disappear and the "empty" content should be presented.

src/components/ProjectList/ProjectListContainer.tsx
```diff
@@ -19,7 +22,16 @@ const ProjectListContainer = ( ) => {
   return (
     <ViewWrapper>
       <View className="border-b border-lightGray mt-5" />
-      <ProjectList projects={projects} />
+      <ProjectList
+        projects={projects}
+        ListEmptyComponent={(
```
FLGMwt, line 27: we don't need a fetching ternary here like the others because ProjectListContainer gets the `projects` from the params, is that right?

src/i18n/strings.ftl
```diff
@@ -1254,6 +1254,9 @@ This-observation-needs-more-identifications = This observation needs more identi
 This-observation-needs-more-identifications-to-become-research-grade = This observation needs more identifications to become research grade.
 This-observer-has-opted-out-of-the-Community-Taxon = This observer has opted out of the Community Taxon
 This-organism-was-placed-by-humans = This organism was placed in this location by humans. This applies to things like garden plants, pets, and zoo animals.
+This-user-has-no-followers = This user has no followers.
```
FLGMwt, line 1257: for the process of adding translation keys: is this the only file that we would hand-edit? and the translate script edits the `en.ftl` and `en.ftl.json` files?

jtklein: Correct. There is some explanation about this process here: https://github.com/inaturalist/iNaturalistReactNative/blob/3c036d7560d6b9305d8dc0a8c770e709fadde9ec/README.md#L96-L97

src/components/ProjectList/ProjectListContainer.tsx
jtklein: For some reason the fetching of the projects data is happening on the previous screen (and the project list is sent to this detail screen here).
So, we do not have info on `isFetching` on this screen.
I have not looked into detail, why the fetching is happening one screen earlier, but even sending the entire project list over the navigation param to the next screen is something I think we should avoid. Navigation params should ideally be slim and JSON-serializable.

FLGMwt: 👍 I agree with you there, it'd certainly be odd to have something that heavy captured in route or query params for a web url. It smells to me like it was done out of convenience but should probably be in zustand or context. I'll make a ticket to investigate this pattern throughout the app. EDIT: MOB-964 (iNat internal) created "Stub: Investigate use of React Navigation "params" for non-trivial data".

PR #3112, opened 2025-09-23
fix developer warning on login screen, provide react key

FLGMwt:
### Issue [dev only]: "LoginForm" load logs React "key" prop warning

On loading the login screen, we get a warning log from React about a missing `key` prop on the children of a list. The culprit is `<Trans components={ [<A />, <B />] />` which I'm guessing doesn't get any help from the linter. 

I just satisfied them with key props that are in line with their interpolation components in the i18n https://github.com/inaturalist/iNaturalistReactNative/blob/main/src/i18n/strings.ftl#L445. Happy to change them to anything else : )

before (on login screen load in dev): 

image
image

after (on login screen load in dev):

image

### DIFF
--- a/src/components/ObsDetails/Sheets/PotentialDisagreementSheet.tsx
+++ b/src/components/ObsDetails/Sheets/PotentialDisagreementSheet.tsx
@@ -1,3 +1,4 @@
+import type { ApiTaxon } from "api/types";
 import {
   Body1,
   DisplayTaxon,
@@ -7,13 +8,14 @@ import {
 import { View } from "components/styledComponents";
 import React from "react";
 import { Trans } from "react-i18next";
+import type { RealmTaxon } from "realmModels/types";
 import { useCurrentUser, useTranslation } from "sharedHooks";
 
 interface Props {
   onPressClose: () => void,
   onPotentialDisagreePressed: ( _checkedValue: string ) => void;
-  newTaxon: object,
-  oldTaxon: object
+  newTaxon: RealmTaxon | ApiTaxon,
+  oldTaxon: RealmTaxon | ApiTaxon
 }
 
 const PotentialDisagreementSheet = ( {


frewsxcv (approved):

jtklein (approved):
I think this is perfectly fine as is. However, we probably need to do the same in all the other places we use the <Trans> component, right?
If we do get the same warning on the other instances of using this component as well, I think you can add the same approach to the other cases in this PR.

src/components/ObsDetails/ActivityTab/DisagreementText.tsx
```diff
@@ -54,7 +36,17 @@ const DisagreementText = ( { taxon, username, withdrawn }: Props ) => {
...
+            key="0"
+            layout="horizontal"
+            prefersCommonNames={currentUser?.prefers_common_names}
+            removeStyling
+            scientificNameFirst={currentUser?.prefers_scientific_name_first}
+            small
+            taxon={taxon}
+            topTextComponent={Body4}
+            withdrawn={withdrawn}
+          />
```
FLGMwt, line 49: inlined this from the render helper

jtklein: I do prefer this.

src/components/ObsDetails/Sheets/PotentialDisagreementSheet.tsx
```diff
@@ -1,35 +1,33 @@
...
 import { useCurrentUser, useTranslation } from "sharedHooks";
 
 interface Props {
-  onPressClose: Function,
-  onPotentialDisagreePressed: Function,
-  newTaxon: Object,
-  oldTaxon: Object
+  onPressClose: () => void,
+  onPotentialDisagreePressed: ( _checkedValue: string ) => void;
+  newTaxon: object,
```
jtklein: Can we do better here than `object`? Should be `taxon: RealmTaxon | ApiTaxon;` because that is expected in `DisplayTaxonName.tsx`.

src/components/ObsDetails/Sheets/PotentialDisagreementSheet.tsx
```diff
@@ -1,35 +1,33 @@
...
   onPotentialDisagreePressed,
   newTaxon,
   oldTaxon
-}: Props ): Node => {
+}: Props ) => {
   const { t } = useTranslation( );
   const currentUser = useCurrentUser( );
 
-  const showTaxonName = ( taxon, fontComponent ) => (
+  const showTaxonName = ( reactKey: string, taxon: object, fontComponent: React.FC ) => (
```
jtklein, line 30: Should we refactor same as above? So that instead of a function we have a JSX tag in each array.

FLGMwt: The other helper was entirely unnecessary, but this actually does serve to reduce some prop duplication. I think this pattern is pervasive enough in the app that I'd like to talk more about it before doing too much "fixing" of it.

PR #3116, opened 2025-09-30
MOB-576 (iOS) fix crash on first launch

FLGMwt:
https://www.i18next.com/principles/fallback#:~:text=fallbackLng%3A%20(code)%20%3D%3E%20%7B

looks like we weren't getting passed anything for the language code. mirroring the sample docs, I'm adding an eager return with our `'en'` default

this one was harder to notice in dev because the dev server sometimes "helps" by reloading so you don't get the LogBox stack

aside: any concerns about Cocoapods version bump?

Recording before and after:

https://github.com/user-attachments/assets/483d6f67-168e-4798-acd5-c4475444f4c0


jtklein (approved):
Code change looks fine. Not sure it is related to the ticket referenced though.

PR #3117, opened 2025-09-30
apply link styles to UserText links

FLGMwt:
Adds `a` styling via  `tagsStyles` https://meliorence.github.io/react-native-render-html/docs/guides/styling#mixed-style-records

|Before|After|
|---|---|
|Fimmvörduháls Trail, Hvolsvöllur,|•Maps lane|


jtklein (approved):
When I did the same styling work in Seek, I used `"underline" as const` but my editor at least seems to infer it correctly as `textDecorationLine: "underline";`.

PR #3127, opened 2025-10-10
Add firebase dependency, and basic logging

jtklein:
Closes MOB-890.

We need two different sets of keys for staging and production. In Android the folder structure takes care of importing the correct config file. In iOS we have a linking script that we already use for our cv and geomodel, so I extended this script. It either pulls the staging in Debug, or the production keys in Release mode builds into the app.

We also provide example files for OSS contributors to copy into the places we’d expect those config files.

I can probably remove the e2e setup step to use XCode 16.2 once updated from main after this is merged: https://github.com/inaturalist/iNaturalistReactNative/pull/3133

Some screenshots of the dashboard [here](https://inaturalist.slack.com/archives/C02G5AATVGA/p1760397109101609).

### DIFF
--- a/.github/workflows/e2e_ios.yml
+++ b/.github/workflows/e2e_ios.yml
@@ -36,10 +36,6 @@ jobs:
     timeout-minutes: 120
 
     steps:
-    - uses: maxim-lobanov/setup-xcode@v1
-      with:
-        xcode-version: '^16.2'
-  
     - name: Check out Git repository
       uses: actions/checkout@v4
       with:


ios/PrivacyInfo.xcprivacy
kvangork, line 1: Reason codes look good.

ios/link-inat-model-files.sh
kvangork, line 1: Suggestion: rename this file in the future to reflect all the jobs it does

README.md
```diff
@@ -18,6 +18,9 @@ See [CONTRIBUTING](CONTRIBUTING.md) for guidelines on contributing to this proje
 1. Run `npx pod-install` or `cd ios && pod install` from the root directory
 1. `cp env.example .env.staging` for staging and `cp env.example .env` for production and fill in appropriate values. This is not part of the code repo (contains secrets, such as OAuth client ID).
 1. To run on Android, do this `cp android/example-keystore.properties android/keystore.properties`. Fill in the relevant values. If you are a member of iNat staff, get them from another member of iNat Staff.
+1. Firebase is optional, you can remove it if you don't need it. If you do want to use it, you have to add platform-specific config files that are not part of the code repo.
```
frewsxcv, line 21: > 1. Firebase is optional, you can remove it if you don't need it.

Do you have a recommendation for how to easily remove it? I'm running into this locally not able to build the project without removing all the Firebase initialization lines. If you share how I can update the README to add in those instructions.

PR #3129, opened 2025-10-11
922 share sheet changes

jtklein:
Closes MOB-922

Following the setup [here](https://github.com/inaturalist/react-native-share-menu/blob/f5805e92082fd79f251cefbe86759201589f6e9a/SHARE_EXTENSION_VIEW.md) I have enabled the share extension to display a custom sheet. It uses a separate React instance and entry point that is registered in `index.share.js`. It is not set up to use anything from the main app, like nativewind or styled components.

Previous:
D284FCED-33C1-4978-B998-4018E86428A9_4_5005_c

Now:
5523CAA0-D7A1-4121-A319-FA9776D35C3A_4_5005_c

### DIFF
--- a/src/components/ShareSheet/ShareSheet.tsx
+++ b/src/components/ShareSheet/ShareSheet.tsx
@@ -38,7 +38,7 @@ const ShareSheet = () => {
   return (
     <View style={container}>
       <Text style={text}>
-        {`Share ${sharedData.length} photos with iNaturalist`}
+        {`Share ${sharedData.length} photos with iNaturalist?`}
       </Text>
       <View style={buttonGroup}>
         <Button


src/components/ShareSheet/ShareSheet.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const {
+    container, text, buttonGroup, destructive
+  // eslint-disable-next-line no-use-before-define
+  } = styles;
+
+  return (
+    <View style={container}>
+      <Text style={text}>
+        {`Share ${sharedData.length} photos with iNaturalist?`}
```
FLGMwt, line 41: should this be a translation?

jtklein: I would assume it should be translated at one point. It was not mentioned in the ticket for it though.
Adding translation would mean to enable i18n for this mini React app with how we do it in the main app which I'd personally defer to whenever product comes back to us with this new requirement.

PR #3134, opened 2025-10-14
Update ios maestro flows

FLGMwt:
linkIntoApp:
https://github.com/user-attachments/assets/ee1c3474-1ca9-44fb-8f09-c42196573473

shareIntoApp:
https://github.com/user-attachments/assets/080a9daf-897d-4481-a214-44c04e5da8d1

Main change here is updating shareIntoApp to not rely on an existing image. We download an sample image w/ exif and share that from photos.

If we end up doing any more Maestro, I think we'll probably want to extract a "before all" flow to go and dismiss all all the FTUEs (photobooth has one, in addition to our app). I have a Google APIs version of shareInto in another branch and _man_ google loves them some FTUEs.

### DIFF
--- a/e2e/maestro/ios/shareIntoApp.yaml
+++ b/e2e/maestro/ios/shareIntoApp.yaml
@@ -24,10 +24,9 @@ appId: org.inaturalist.iNaturalistMobile
 - longPressOn: Canon_40D
 - tapOn: Share
 - tapOn: iNaturalist
-- tapOn: iNaturalist
 
 # iNaturalist custom Share Extension
-- tapOn: "Yes" # quotes necessary to distinguish `"Yes"` string from `Yes` which Maestro treats as a bool
+- tapOn: "Yes" # <- quotes are necessary to distinguish `"Yes"` string from `Yes` which Maestro treats as a bool
 
 # iNaturalist app
 - scrollUntilVisible: # For clarify, center the "OBSERVED AT" header which should have the time right after


e2e/maestro/ios/shareIntoApp.yaml
```diff
@@ -1,18 +1,33 @@
...
+    when:
+      visible: Continue
+    commands:
+      - tapOn: Continue
+
+- tapOn: Select # we can _see_ our photo in the gallery, but can't target it w/o focusing "Select"
+- tapOn: Photo, May 30, 2008, 3:56 PM
+- tapOn: Share
+- tapOn: iNaturalist
+- tapOn: Yes # <- doesn't seem to work. "No" doesn't either. Click works fine in simulator.
```
FLGMwt: FYI @jtklein, on the new Share Extension. This is not currently working when Maestro taps on it, I was wondering if you had an idea. The Maestro Studio inspector sees it for sure, but the tap doesn't work.

e2e/maestro/ios/shareIntoApp.yaml
```diff
@@ -1,18 +1,33 @@
 appId: org.inaturalist.iNaturalistMobile
 ---
+# Download sample image with exif and stable public link
+- launchApp:
+    appId: com.apple.mobilesafari
+- tapOn: Address
+- inputText: "https://raw.githubusercontent.com/ianare/exif-samples/refs/heads/master/jpg/Canon_40D.jpg"
+- pressKey: Enter
+- tapOn: Share
```
jtklein, line 9: On an iPhone 17 Pro OS 26 simulator the test fails at this step here for me.

e2e/maestro/ios/shareIntoApp.yaml
```diff
@@ -1,18 +1,33 @@
...
+
+# dismiss iOS Photos FTUE
+- runFlow:
+    when:
+      visible: Continue
+    commands:
+      - tapOn: Continue
+
+- tapOn: Select # we can _see_ our photo in the gallery, but can't target it w/o focusing "Select"
+- tapOn: Photo, May 30, 2008, 3:56 PM
```
jtklein: On an iPhone 16 Pro simulator OS 18.6 the test fails here for me:
 ║    ❌   Tap on "Photo, May 30, 2008, 3:56 PM"

FLGMwt: I'm successfully getting past this to the share extension step on a fresh iPhone 16 Pro, 18.6 sim. Could you share a screenshot / screencap of the phone at the failure point?

e2e/maestro/ios/shareIntoApp.yaml
FLGMwt: Mmmmmm, yeah, probably worth talking about how OS-resilient we want our Maestro tests to be at this point. I've been testing on iPhone 16 Pro, 18.6. I want to be able to run them on different devices but I also don't think we want a lot of OS-conditional code, _especially_ because conditionals add a few seconds each (as they wait for a potential element).

e2e/maestro/ios/shareIntoApp.yaml
jtklein: Uploading Simulator Screen Recording - iPhone 16 Pro - 2025-10-14 at 21.39.50.mov…

e2e/maestro/ios/shareIntoApp.yaml
jtklein: `Yes` is boolean in yaml: https://yaml.org/type/bool.html
Changed it to `- tapOn: "Yes"` locally and it works for me.

e2e/maestro/ios/shareIntoApp.yaml
FLGMwt: ^we discovered that this was an issue with the simulator labeling the photo based on the simulator's local timezone. Unlike region/location, simulator Time Zone can _only_ be changed by changing the time zone on the host machine 😒

I rewrote this to save to files and share from the Files app which should be more deterministic across more OS versions anyhow.

e2e/maestro/ios/shareIntoApp.yaml
FLGMwt: brilliant! thank you

PR #3138, opened 2025-10-14
Update react-native-gesture-handler

jtklein:
Closes MOB-932

Compiled and tested in Debug and Release mode on iOS, only Debug mode on Android.


FLGMwt:
I can test a release build on device tonight

PR #3140, opened 2025-10-14
Add a set of recommended VS Code extensions

jtklein:
I would like to add a set of recommended VS Code extensions for any developer that in future works on this project.

"dbaeumer.vscode-eslint" eslint integrates very nicely with VS code and we should recommend it being on to see eslint errors in the file before committing new code
"usernamehw.errorlens" just screams at you when you have errors in files
"wix.vscode-import-cost" because library import sizes are important for performance on mobile. The lower the bundle size, less memory usage, better for phones.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,6 @@
 {
   "recommendations": [
     "dbaeumer.vscode-eslint",
-    "sysoev.vscode-open-in-github",
     "usernamehw.errorlens",
     "wix.vscode-import-cost"
   ]


.gitignore
```diff
@@ -95,7 +95,7 @@ artifacts/
 *.log
 
 # VisualStudioCode #
-.vscode
+.vscode/settings.json
```
FLGMwt, line 98: @jtklein I was thinking it'd be good to add the following `settings.json` to facilitate new devs in the spirit of the extension recommendations, do you have any objection to that? 

```json
{
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": "explicit"
  }
}
```

PR #3142, opened 2025-10-15
Show offline profile picture in header

jtklein:
Closes MOB-926.

Checking if the user to display is the current user and if so show the profile picture we have also while offline.

### DIFF
--- a/tests/unit/components/SharedComponents/InlineUser/InlineUser.test.js
+++ b/tests/unit/components/SharedComponents/InlineUser/InlineUser.test.js
@@ -17,7 +17,7 @@ jest.mock( "@react-navigation/native", ( ) => {
 const mockUser = factory( "RemoteUser" );
 const mockUserWithoutImage = factory( "RemoteUser", { icon_url: null } );
 
-const snapshotUser = { login: "some_login", icon_url: "some_icon_url" };
+const snapshotUser = { login: "some_login", icon_url: "some_icon_url", id: 1 };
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/InlineUser/InlineUserBase.tsx
```diff
@@ -8,45 +6,41 @@ import UserIcon from "components/SharedComponents/UserIcon";
...
   const userImgUri = User.uri( user );
   const userHandle = user?.login;
+  const currentUser = useCurrentUser();
+  const isCurrentUser = userHandle === currentUser?.login;
 
   const { t } = useTranslation( );
 
   const renderUserIcon = () => {
-    if ( !userImgUri || !isConnected ) {
+    if ( !userImgUri || ( !isConnected && !isCurrentUser ) ) {
```
FLGMwt, line 43: nit (and i know this is from existing code): I think w/ the conditional logic it's more readable to start with the positive condition of "we have a picture" and _then_ fall back to the fallback

FLGMwt (approved):
LGTM

src/components/SharedComponents/InlineUser/InlineUserBase.tsx
jtklein: I think it is more common to return early for edge cases (negative UI, error states, loading, etc.) and let the “positive” be the main return at the bottom.

PR #3146, opened 2025-10-17
Teach TypeScript about SVG imports.

frewsxcv:
See https://stackoverflow.com/questions/58726319/typescript-cannot-find-module-when-import-svg-file.

# Before

Screenshot 2025-10-17 at 12 30 57 AM

# After

Screenshot 2025-10-17 at 12 30 20 AM

### DIFF
--- a/src/types/react-navigation.d.ts
+++ b/src/types/react-navigation.d.ts
@@ -1,0 +1,7 @@
+import { RootStackParamList } from "@react-navigation/native";
+
+declare global {
+  namespace ReactNavigation {
+    type RootParamList = RootStackParamList
+  }
+}
--- a/src/types/svg.d.ts
+++ b/src/types/svg.d.ts
@@ -1,5 +1,3 @@
-import { RootStackParamList } from "@react-navigation/native";
-
 // Using typescript with react-native-svg-transformer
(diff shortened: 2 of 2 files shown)


src/svg.d.ts
```diff
@@ -0,0 +1,13 @@
+// Provides a TypeScript type definitiont for SVG files imported as React components.
```
frewsxcv:
```suggestion
// Provides a TypeScript type definition for SVG files imported as React components.
```

frewsxcv:
@jtklein Thanks for the review. I just pushed a change that removes my type definition and fixes the existing one

jtklein:
Is it working for you locally? I can now see `Cannot find module 'images/svg/casual_grade.svg' or its corresponding type declarations.ts(2307)` also in this branch.

frewsxcv:
The latest commit fixed it for me, though I'm not sure why. Something to do with both of those type declarations sharing the same file.
Screenshot 2025-10-21 at 12 01 55 PM
Screenshot 2025-10-21 at 12 01 51 PM

jtklein:
Yes! Works for me now as well. Thanks for getting back to this one so quickly. Will merge now.

PR #3147, opened 2025-10-17
Migrate `ForgotPassword` component to TypeScript.

frewsxcv:


src/components/LoginSignUp/ForgotPassword.tsx
```diff
@@ -15,11 +13,16 @@ import ForgotPasswordForm from "./ForgotPasswordForm";
 import Header from "./Header";
 import LoginSignUpWrapper from "./LoginSignUpWrapper";
 
-const ForgotPassword = ( ): Node => {
+type RenderProps = {
+  // eslint-disable-next-line react/no-unused-prop-types
```
frewsxcv, line 17: Unsure why this lint is firing, so I silenced it. I think it has to do with the atypical design of the `children` prop on `LoginSignUpWrapper`, which could be addressed in the future.

src/components/LoginSignUp/ForgotPassword.tsx
```diff
@@ -44,6 +47,7 @@ const ForgotPassword = ( ): Node => {
             navigation.navigate( "LoginStackNavigator", { screen: "Login" } );
           }}
           buttonType="focus"
+          loading={false}
```
frewsxcv, line 50: `loading` is a required prop on `WarningSheet`. Previously it was silently passed as `undefined` and now it's being passed `false`, which are both false-y values, so it should be identical.

jtklein: I agree

src/components/LoginSignUp/ForgotPassword.tsx
jtklein: You are right, this children setup does feel a bit off. I have not looked into detail yet but there might be some future changes here. Anyway, for now your types here seem okay.

src/components/LoginSignUp/ForgotPassword.tsx
```diff
@@ -15,11 +13,16 @@ import ForgotPasswordForm from "./ForgotPasswordForm";
 import Header from "./Header";
 import LoginSignUpWrapper from "./LoginSignUpWrapper";
 
-const ForgotPassword = ( ): Node => {
+type RenderProps = {
+  // eslint-disable-next-line react/no-unused-prop-types
+  scrollViewRef: { current: null | React.Ref<typeof ScrollView> }
+};
+
+const ForgotPassword = ( ): React.ReactElement => {
```
jtklein: One change: I would like us to use type inference for the return type of functional components.

jtklein (changes requested):
Thanks for that! I can see only one change, see inline comment.

src/components/LoginSignUp/ForgotPassword.tsx
frewsxcv: Fixed in https://github.com/inaturalist/iNaturalistReactNative/pull/3147/commits/2cfb2078fb44ccab824e2b532f6e8b7c8e54bed1

PR #3149, opened 2025-10-18
(android) limit image picker to 100 max

FLGMwt:
Looks like the Android picker can only support 100 images.

Stack trace doesn't have the actual number, but it fails at 101, works at 100. 

Note: sharing from photo library on Android is still blocked by an existing issue, but this addresses the new separate regression

closes MOB-959


jtklein (approved):
Thanks for finding this. LGTM

PR #3150, opened 2025-10-18
Migrate `RotatingINatIconButton` component to TypeScript.

frewsxcv:


src/components/SharedComponents/Buttons/RotatingINatIconButton.js
```diff
@@ -112,7 +109,6 @@ const RotatingINatIconButton = ( {
         icon={icon}
         mode={mode}
         onPress={onPress}
-        rotating={rotating}
```
frewsxcv, line 115: This prop does not exist on the `INatIconButton` component

jtklein: Correct. Seems good to be removed

jtklein:
Thanks. I changed a few thing to be more in line with how we use TypeScript in the other components.

PR #3151, opened 2025-10-18
Migrate `StartupService` components to TypeScript.

frewsxcv:


src/components/StartupService.tsx
```diff
@@ -60,7 +60,8 @@ const StartupService = ( ) => {
   const realm = useRealm( );
   const currentUser = realm.objects( "User" ).filtered( "signedIn == true" )[0]?.isValid( );
   const { loadTime } = usePerformance( {
-    screenName: "StartupService"
+    screenName: "StartupService",
+    isLoading: false
```
frewsxcv, line 64: Required parameter to `usePerformance`

src/components/StartupService.tsx
```diff
@@ -47,7 +47,7 @@ const checkForPreviousCrash = async ( ) => {
   try {
     const crashData = zustandStorage.getItem( "LAST_CRASH_DATA" );
     if ( crashData ) {
-      const parsedData = JSON.parse( crashData );
+      const parsedData = JSON.parse( crashData.toString( ) );
```
frewsxcv, line 50: `JSON.parse` takes a `string`, and `crashData` was previously `string|number`

jtklein: Yes, that seems to me just based on how the zustand store is currently typed. The only time we write to this item is [here](https://github.com/frewsxcv/iNaturalistReactNative/blob/d4650735b4ef6857f6e77afdd6a3478a0e08affb/index.js#L111-L112) and should be a string only.

PR #3152, opened 2025-10-20
feat(build): Unify JS config for i18n and scripts

frewsxcv:
This change introduces `babel-node` to the development workflow, allowing the project to use modern JavaScript features like ES modules (`import`/`export`) in our build scripts and i18n command-line interface.

Previously, scripts in the `i18n` directory and the `scripts` directory were executed directly with `node`, which required them to use CommonJS (`require`/`module.exports`) and prevented the use of other Babel transformations that are standard across the rest of the project.

By switching to `babel-node`, this project now has unified JavaScript tooling, and now developers are no longer required to know which files need to use CommonJS (none anymore!).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,6 +108,7 @@
       },
       "devDependencies": {
         "@babel/core": "^7.25.2",
+        "@babel/node": "^7.28.0",
         "@babel/preset-env": "^7.25.3",
         "@babel/preset-react": "^7.24.1",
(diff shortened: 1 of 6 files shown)


FLGMwt (approved):
LGTM, thank you for this suggestion! I like the idea of removing the small context-shift annoyance on projects with minimal scripts. Would you mind pulling in main and resolving the package-lock.json?

PR #3155, opened 2025-10-21
MOB-948 add programmatic test user reset step to e2e

FLGMwt:
We encountered an issue where e2e tests were failing due to a combination of the sometimes-there announcement banner and the user's more-than-expected number of observations. 

This PR aims to set more consistent testing conditions by dismissing announcements and cleaning up (deleting) observations programmatically before each run. This is done directly against the APIs for reliability and so we can keep the detox use focused on the testing critical path (instead of test setup).

The new script / function relies on duplicating some AuthenticationService logic which would need to be decoupled from RN-specific persistence to reuse here.

### DIFF
--- a/e2e/sharedFlows/resetUserForTesting.js
+++ b/e2e/sharedFlows/resetUserForTesting.js
@@ -90,7 +90,6 @@ export default async function resetUserForTesting() {
       opts
     );
   } ) );
-
   const usersEditResponse = await apiClient.get(
     "/users/edit.json",
     {},
@@ -133,7 +132,10 @@ export default async function resetUserForTesting() {
 
   console.log( "Creating sample observation" );
   const sampleObservationParams = {
-    observation: sampleObservation
+    observation: {
+      latitude: sampleObservation.latitude,
+      longitude: sampleObservation.longitude
+    }
   };
   await inatjs.observations.create(
     sampleObservationParams,


.github/workflows/e2e_ios.yml
```diff
@@ -186,15 +186,22 @@ jobs:
     - name: Set MOCK_MODE to e2e
       run: echo "MOCK_MODE=e2e" >> "$GITHUB_ENV"
 
-    # Generate the secret files needed for the test run
+    # Generate the secret files needed for a release build
```
FLGMwt, line 189: I had trimmed the test step's .env generation command when I split build/test since the test _run_ previously only needed the UN/PW. With this new reset script, we're using the api at test runtime so we need the oath stuff again. I've restored it to being _exactly_ the same as build for clarity. Unfortunately, there's no _great_ way to reduce the duplication here in GitHub Actions without compromising secret detection / redaction.

e2e/helpers.js
```diff
@@ -109,6 +113,8 @@ export function terminateApp( deviceId, bundleId ) {
 }
 
 export async function iNatE2eAfterEach( device ) {
+  await resetUserForTesting();
```
frewsxcv, line 116: Is it redundant resetting before and after each test? Not a big deal though if `resetUserForTesting` is a quick operation

FLGMwt: good callout! and I did give this some thought. it's absolutely redundant, but my thinking was to be annoyingly sure that "the user always looks like this", even if it's overkill. This does the job for now, but I'd probably like to do something to call it _external_ to the detox orchestration itself.

e2e/sharedFlows/resetUserForTesting.js
```diff
@@ -0,0 +1,159 @@
...
+
+const userAgent = "iNaturalistRN/e2e";
+const installId = uuid.v4( ).toString( );
+
+inatjs.setConfig( {
+  apiURL: Config.API_URL,
+  writeApiURL: Config.API_URL,
+  userAgent,
+  headers: {
+    "X-Installation-ID": installId
```
jtklein: We should remove this. It is not required for making an authenticated request, and it does inform product decisions, so we should minimize adding false data here.

e2e/sharedFlows/resetUserForTesting.js
```diff
@@ -0,0 +1,159 @@
...
+    const message = "This e2e test deletes observations of the user under test."
+    + "Add this username to the `testUsernameAllowlist` if that's really what you want.";
+    throw new Error( message );
+  }
+
+  const apiClient = create( {
+    baseURL: apiHost,
+    headers: {
+      "User-Agent": userAgent,
+      "X-Installation-ID": installId
```
jtklein: Same as above.

e2e/sharedFlows/resetUserForTesting.js
```diff
@@ -0,0 +1,159 @@
...
+  console.log( `Dismissing ${announcementIdsToDismiss.length} announcements` );
+
+  await Promise.all( announcementIdsToDismiss.map( async id => {
+    await inatjs.announcements.dismiss(
+      { id },
+      opts
+    );
+  } ) );
+
+  const usersEditResponse = await apiClient.get(
```
jtklein, line 93: That is such an obscure call I had to double check.
Looking at it with a fresh perspective (i.e., what call would I make to get the user ID) I would probably have chosen the documented `users/me` [endpoint](https://api.inaturalist.org/v2/docs/#/Users/get_users_me). But I do see now that we are using the same call in the app to get the username after someone uses third-party auth as an account creation mechanism. So, seems valid but possibly arcane.

e2e/sharedFlows/resetUserForTesting.js
```diff
@@ -0,0 +1,159 @@
...
+      opts
+    );
+  } ) );
+
+  console.log( "Creating sample observation" );
+  const sampleObservationParams = {
+    observation: {
+      captive_flag: false,
+      geoprivacy: "open",
+      latitude: 52.52,
```
jtklein: Let's re-use the same data as we use when making an observation during the test: src/appConstants/e2e.ts, so we introduce less fake data.

e2e/sharedFlows/resetUserForTesting.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+  );
+
+  const userId = usersEditResponse.data.id;
+
+  const observationResponse = await inatjs.observations.search( { user_id: userId }, opts );
+
+  // is this _really_ an e2e test user if they have more than expected observations?
+  const suspiciousObservationThreshold = 10;
+  if ( typeof observationResponse.total_results !== "number"
```
jtklein: I would make a different error message for `typeof observationResponse.total_results !== "number" `. I can imagine someone seeing the error message and without looking at the code trying to solve this by going to the website and not seeing any observations there.

e2e/sharedFlows/resetUserForTesting.js
FLGMwt: Yeah, I mirrored the existing calls. I'll make sure that we update this call to match the app's call if we end up updating that to the documented one.

e2e/sharedFlows/resetUserForTesting.js
FLGMwt: sounds good, thanks for the context

e2e/sharedFlows/resetUserForTesting.js
FLGMwt: Oh, yeah, good call, I duplicated a sample obs I made and logged in the app, but reusing the e2e is better.

PR #3156, opened 2025-10-23
MOB-961 - taxon details - when navigating the taxonomy tree, preserve the navigation stack

budowski:


src/components/TaxonDetails/Taxonomy.tsx
```diff
@@ -23,15 +23,13 @@ const Taxonomy = ( { taxon: currentTaxon, hideNavButtons }: Props ) => {
   const scientificNameFirst = currentUser?.prefers_scientific_name_first;
 
   const navigateToTaxonDetails = ( taxonId: number ) => (
-    navigation.navigate( {
+    navigation.push( "TaxonDetails", {
       // Ensure button mashing doesn't open multiple TaxonDetails instances
       key: `${route.key}-Taxonomy-TaxonDetails-${taxonId}`,
```
jtklein: I think `name` has no effect in screen params because push takes the name as the first argument. When I am testing your branch `key` also does no longer what it is intended to do, i.e. now it is possible to quickly smash on a taxon link and open the same screen multiple times. (This was not possible in main but it might also be the case that key didn't actually matter on main because of the faulty behaviour of the navigation not stacking multiple screens anyway.)

src/components/TaxonDetails/TaxonomyTaxon.tsx
```diff
@@ -58,7 +59,7 @@ const TaxonomyTaxon = ( {
   );
 
   return (
-    <Pressable
+    <PressableWithDebounce
```
abbeycampbell, line 62: ooh good to know this exists

PR #3157, opened 2025-10-24
MOB-925 - new designs for add-obs button sheet

budowski:
[MOB-925](https://linear.app/inaturalist/issue/MOB-925/replace-the-obs-wheel-and-dismissible-card-with-a-new-bottom-sheet)
[MOB-980](https://linear.app/inaturalist/issue/MOB-980/obs-wheel-tooltip-not-showing-up)

### DIFF
--- a/tests/unit/components/AddObsBottomSheet/AddObsButton.test.js
+++ b/tests/unit/components/AddObsBottomSheet/AddObsButton.test.js
@@ -61,11 +61,10 @@ describe( "shows tooltip", () => {
 
     renderComponent( <AddObsButton /> );
 
-    // Temporarily disabled the tooltip for new users, as it is freezing the app in some cases.
-    // const tooltipText = await screen.findByText(
-    //   "Press and hold to view more options"
-    // );
-    // expect( tooltipText ).toBeTruthy();
+    const tooltipText = await screen.findByText(
+      "Press and hold to view more options"
+    );
+    expect( tooltipText ).toBeTruthy();
   } );
 
   it( "to new users only after they dismissed the account creation card", async () => {
@@ -87,11 +86,10 @@ describe( "shows tooltip", () => {
       }
     } );
 
-    // Temporarily disabled the tooltip for new users, as it is freezing the app in some cases.
-    // const tooltipTextAfter = await screen.findByText(
-    //   "Press and hold to view more options"
-    // );
-    // expect( tooltipTextAfter ).toBeTruthy();
+    const tooltipTextAfter = await screen.findByText(
+      "Press and hold to view more options"
+    );
+    expect( tooltipTextAfter ).toBeTruthy();
   } );
 
   it( "to logged in users with more than 50 observations after card dismissal", async () => {
@@ -110,10 +108,9 @@ describe( "shows tooltip", () => {
       }
     } );
 
-    // Temporarily disabled the tooltip for new users, as it is freezing the app in some cases.
-    // const tooltipTextAfter = await screen.findByText(
-    //   "Press and hold to view more options"
-    // );
-    // expect( tooltipTextAfter ).toBeTruthy();
+    const tooltipTextAfter = await screen.findByText(
+      "Press and hold to view more options"
+    );
+    expect( tooltipTextAfter ).toBeTruthy();
   } );
 } );


src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
```diff
@@ -0,0 +1,169 @@
+import {
+  Body3, BottomSheet, INatIcon, INatIconButton
+} from "components/SharedComponents";
+import { View } from "components/styledComponents";
+import React, { useMemo } from "react";
+import { Platform, Pressable } from "react-native";
```
jtklein: We need to import `Pressable` from `styledComponents`instead. There we have overrides for all base react-native components.

src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
```diff
@@ -0,0 +1,169 @@
...
+interface Props {
+  closeBottomSheet: ( ) => void;
+  navAndCloseBottomSheet: ( screen: string, params?: {
+    camera?: string
+  } ) => void;
+  hidden: boolean;
+}
+
+type ObsCreateItem = {
+    text?: string,
```
jtklein: Do we want text to be undefined here?

src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+interface Props {
+  closeBottomSheet: ( ) => void;
+  navAndCloseBottomSheet: ( screen: string, params?: {
+    camera?: string
+  } ) => void;
+  hidden: boolean;
+}
+
+type ObsCreateItem = {
```
jtklein, line 20: Wrong indentation.

src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
```diff
@@ -0,0 +1,169 @@
...
+    accessibilityHint,
+    accessibilityLabel,
+    icon,
+    onPress,
+    testID,
+    text
+  }: ObsCreateItem ) => (
+    <Pressable
+      className="bg-white w-1/2 flex-column items-center py-4 rounded-lg flex-1 shadow-sm
+      shadow-black/25 active:opacity-50"
```
jtklein, line 104: Please make sure that these still work when we import from `styledComponent`: shadow-... active:...

src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
```diff
@@ -0,0 +1,169 @@
...
+        </View>
+
+        <View className={ROW_CLASS}>
+          {renderAddObsIcon( obsCreateItems.soundRecorder )}
+          {AI_CAMERA_SUPPORTED && renderAddObsIcon( obsCreateItems.aiCamera )}
+        </View>
+
+        <Pressable
+          className="bg-mediumGray w-full flex-row items-center py-[10px] px-5 rounded-lg
+          active:opacity-50"
```
jtklein: How did you land on `-50`? Is there a paper trail to that decision? Is that something we should add to `Pressable` coming from `styledComponents` so that all our Pressables behave the same way?

Not in this PR though.

src/components/AddObsBottomSheet/AddObsButton.js
```diff
@@ -0,0 +1,200 @@
...
+      setCurrentRoute( current );
+    } );
+
+    return unsubscribe;
+  }, [] );
+
+  return (
+    <>
+      <AddObsTooltip isVisible={showTooltip} dismissTooltip={dismissTooltip} />
+      {/* match the animation timing on FadeInView.tsx */}
```
jtklein: Remove this comment.

src/components/AddObsBottomSheet/AddObsButton.js
```diff
@@ -0,0 +1,200 @@
...
+
+  // Controls whether to show the tooltip, and to show it only once to the user
+  const showKey = "AddObsButtonTooltip";
+  const shownOnce = useStore( state => state.layout.shownOnce );
+  const setShownOnce = useStore( state => state.layout.setShownOnce );
+  const justFinishedSignup = useStore( state => state.layout.justFinishedSignup );
+  const numOfUserObservations = zustandStorage.getItem( "numOfUserObservations" );
+  const obsCountLoaded = typeof numOfUserObservations === "number";
+
+  React.useEffect( () => {
```
jtklein, line 38: In general I'd like to see less use of `useEffect` and not more. But we do have an upcoming ticket to audit the use of useEffects in the app, so this is fine for here.

src/components/MyObservations/MyObservationsEmptySimple.js
```diff
@@ -77,15 +76,10 @@ const MyObservationsEmptySimple = ( { currentUser, isConnected, justFinishedSign
             />
           </View>
         </View>
-        <Modal
-          showModal={showModal}
+        <AddObsBottomSheet
```
jtklein: I wanted to comment that we have to update the names of the props here (to xx BottomSheet) but after testing I am pretty sure this instance is actually non-functional. The "AI camera" button that is centered in the Me tab when there is no data to show used to also open the AddObsModal on long press but does no longer do so.
So, I removed it and send a PR into yours, please review and accept before merging yours into main.

jtklein: https://github.com/inaturalist/iNaturalistReactNative/pull/3245

src/components/AddObsBottomSheet/AddObsTooltip.tsx
```diff
@@ -0,0 +1,46 @@
+import classNames from "classnames";
+import { Body2 } from "components/SharedComponents";
+import GradientButton from "components/SharedComponents/Buttons/GradientButton";
+import React from "react";
+import { Modal, View } from "react-native";
```
jtklein: Please import `View` from `styledComponents`.

src/components/AddObsBottomSheet/AddObsTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useTranslation } from "sharedHooks";
+
+type Props = {
```
jtklein: I'd prefer Props to be an interface in all of our functional components.

src/components/AddObsBottomSheet/AddObsTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+    isVisible: boolean;
```
jtklein: Wrong indentation.

src/components/AddObsBottomSheet/AddObsTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+    dismissTooltip: () => void;
```
jtklein: According to our code style there should be a space in empty parentheses.

src/components/AddObsBottomSheet/AddObsTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+};
+
+const AddObsTooltip = ( { isVisible, dismissTooltip }: Props ): React.ReactNode => {
```
jtklein: I'd prefer the return type of functional components to be inferred.

src/components/AddObsBottomSheet/AddObsTooltip.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { t } = useTranslation();
+  return (
+    <Modal
+      visible={isVisible}
+      animationType="fade"
```
jtklein: Better to keep the animationTiming values here as they were for the <Modal in AddObsButton, or else provide a reason as to why we can remove it:
```
        animationIn="fadeIn"
        animationOut="fadeOut"
        animationInTiming={250}
        animationOutTiming={250}        
```

src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -29,14 +32,16 @@ interface Props {
...
+  onLayout?: ( event: object ) => void;
   // Callback when the user presses the close button or backdrop, not whenever the sheet
   // closes
-  onPressClose?: Function;
+  onPressClose?: () => void;
   snapPoints?: Array<string>;
   insideModal?: boolean;
   keyboardShouldPersistTaps?: string;
   testID?: string;
+  additionalClasses?: string;
```
jtklein: I think `containerClass` would be more apt. `additional` makes me wonder additional to what? We are using `containerClass` in several places but also `containerClasses` in one place. As long as we are not expecting an array here maybe class is better.

src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -29,14 +32,16 @@ interface Props {
   hidden?: boolean;
   hideCloseButton?: boolean;
   headerText?: string;
-  onLayout?: Function;
+  onLayout?: ( event: object ) => void;
   // Callback when the user presses the close button or backdrop, not whenever the sheet
   // closes
-  onPressClose?: Function;
+  onPressClose?: () => void;
```
jtklein: Missing space in parentheses.

src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -109,27 +116,31 @@ const StandardBottomSheet = ( {
...
         >
-          <View className="mx-12 flex">
-            <Heading4
-              testID="bottom-sheet-header"
-              className="w-full text-center"
-            >
-              {headerText}
-            </Heading4>
-          </View>
+          {headerText && (
```
jtklein: `{ condition && <JSX>}` is more problematic than `{ !condition ? null : <JSX>}` because if condition is `false` we get an error of trying to put something in a Text that is not a string (can't remember the actual message).

src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
abbeycampbell: This was `TouchableOpacity` before; I changed to a `Pressable` because they are a bit more flexible, the RN docs recommend them over `TouchableOpacity` (calling them future-proof), and a quick search through our code base showed we use Pressables far more often than `TouchableOpacity`. 

I added the .5 opacity on press because that's default behavior with touchables, but not with pressables, and I assumed we'd want to keep the feedback experience here since it's oft-used. Let me know if you think this is something we actually don't want, or I can ask Abhas about it--in the meantime, I'm changing it to `-75` since that's the active opacity we use in our `SharedComponents/Button`--more consistent with what we already do.

I think it's probably unnecessary to add this behavior to our styled `Pressable`, but if we do we also might want to build in a prop to turn it off, since part of what's cool about `Pressables` is the ability to customize them.

src/components/AddObsBottomSheet/AddObsTooltip.tsx
abbeycampbell: weird, I was triggering errors for this rule while working so I'm not sure why this one (and the indentation stuff) wasn't caught

src/components/AddObsBottomSheet/AddObsTooltip.tsx
abbeycampbell: I was using `Modal` from react-native, I've switched to using the one from `SharedComponents` since that seems to be the one originally used here and just about everywhere else. 

I will say I don't think this was working quite as expected before, because our shared components modal doesn't take `animationInTiming` or `animationOutTiming` props, it just takes a `noAnimation` prop (with a default value of false) that sets the animation timing to 1 or 300 miliseconds. Not going to futz with adding that functionality in this PR, so, I've opted to leave that value as false here, so we end up with a slower fade-in/out.

src/components/SharedComponents/Sheets/BottomSheet.tsx
abbeycampbell: ah yeah "text must be rendered inside `<Text>`" or something like that

PR #3162, opened 2025-10-26
Migrate `NetworkService.js` to TypeScript.

frewsxcv:


src/components/NetworkService.ts
```diff
@@ -12,7 +12,8 @@ const logger = log.extend( "NetworkService" );
 
 const NetworkService = ( ) => {
   const { loadTime } = usePerformance( {
-    screenName: "NetworkService"
+    screenName: "NetworkService",
+    isLoading: false
```
frewsxcv, line 16: `isLoading` is a required parameter. Previously implicitly `undefined` now explicitly `false`

jtklein: Agreed

PR #3166, opened 2025-10-29
fully cutover to react-native-safe-area-context and fix safe area issues

FLGMwt:
MOB-951

Fully removes references to the soon-to-be-removed `SafeAreaView` from `react-native` in favor of the community alternative `react-native-safe-area-context`. With the necessary exception of the `ErrorBoundary` class component, I've used the `useSafeAreaInsets` according to guidance from React Navigation https://reactnavigation.org/docs/handling-safe-area/ For what it's worth, this seems to have been necessary in some places where the view component had no effect (namely some modals).

In some cases, we were unnecessarily and to no effect using SafeAreaView where parents were already handling safe area. 

Since this affects most of the app, we'll need some extensive testing of all the different screens and states, including some I probably don't know about yet.

### DIFF
--- a/src/components/LoginSignUp/LoginSignUpWrapper.tsx
+++ b/src/components/LoginSignUp/LoginSignUpWrapper.tsx
@@ -80,8 +80,7 @@ const LoginSignupWrapper = ( {
     return unsubscribe;
   }, [navigation] );
 
-  // const fitContentWithinScreenStyle = { height: windowHeight * 0.85 };
-  const fitContentWithinScreenStyle = { };
+  const fitContentWithinScreenStyle = { height: windowHeight * 0.85 };
 
   return (
     <ImageBackground


index.js
```diff
@@ -147,21 +147,21 @@ const AppWithProviders = ( ) => {
   return (
     <QueryClientProvider client={queryClient}>
       <RealmProvider>
-        <SafeAreaProvider>
```
jtklein, line 150: Please revert this change if it is not necessary. So that we have one less variable of introduced change.

package.json
```diff
@@ -124,7 +124,7 @@
     "react-native-reanimated-carousel": "^3.5.1",
     "react-native-render-html": "^6.3.4",
     "react-native-restart": "^0.0.27",
-    "react-native-safe-area-context": "^5.5.1",
+    "react-native-safe-area-context": "^5.6.1",
```
jtklein: As discussed offline this should be a separate PR to update the package.

src/components/Camera/StandardCamera/StandardCamera.js
```diff
@@ -198,7 +197,7 @@ const StandardCamera = ( {
   }, [handleTakePhoto, initialVolume] );
 
   return (
-    <SafeAreaView className={classnames( containerClasses )}>
+    <View className={classnames( containerClasses )} style={{ paddingTop: insets.top }}>
```
jtklein: This screen also needs bottom padding from `insets`.

src/components/LoginSignUp/LoginSignUpWrapper.tsx
```diff
@@ -86,17 +87,7 @@ const LoginSignupWrapper = ( {
...
-          "w-full",
-          "h-full",
-          // In LoginStackNavigator we set `headerTransparent: true`, but this
-          // makes content inside the SafeAreaView stay there in iOS but flow
-          // up under the header area in Android. Barring a better solution,
-          // this makes sure content stay below the header on Android
-          Platform.OS === "android" && "pt-[60px]"
-        )}
-      >
+      <View style={{ paddingTop: insets.top }}>
```
jtklein, line 91: The padding of the main content has shifted now, a before and after screenshot:
48DC3525-8E08-40FD-B421-312B827FF2BA_4_5005_c
220C0A7B-EA9F-4867-A944-B35BE1CFE426_4_5005_c

src/components/MediaViewer/MediaViewer.js
```diff
@@ -111,7 +113,7 @@ const MediaViewer = ( {
   ] );
 
   return (
-    <SafeAreaView className="flex-1 bg-black" testID="MediaViewer">
+    <View className="flex-1 bg-black" style={{ paddingTop: insets.top }} testID="MediaViewer">
```
jtklein: This needs bottom padding from `insets` as well.

src/components/Notifications/Notifications.tsx
```diff
@@ -1,9 +1,7 @@
 import { NotificationOnboarding } from "components/OnboardingModal/PivotCards";
-import {
-  Tabs,
-  ViewWrapper
-} from "components/SharedComponents";
+import { Tabs } from "components/SharedComponents";
 import React, { useState } from "react";
+import { View } from "react-native";
```
jtklein: We should import from `styledComponents` here.

src/components/TaxonDetails/TaxonDetails.js
```diff
@@ -398,8 +397,9 @@ const TaxonDetails = ( ): Node => {
   }
 
   return (
-    <SafeAreaView
+    <View
       className="flex-1 bg-black"
+      style={{ paddingTop: insets.top, paddingBottom: insets.top }}
```
jtklein: There is an extra area of blackness now at the bottom. So, we probably only have to remove the bottom padding here.

src/components/SharedComponents/ViewWrapper.tsx
```diff
@@ -1,9 +1,9 @@
 import classnames from "classnames";
 import { Body1 } from "components/SharedComponents";
-import { SafeAreaView } from "components/styledComponents";
 import type { PropsWithChildren } from "react";
 import * as React from "react";
-import { StatusBar } from "react-native";
+import { StatusBar, View } from "react-native";
```
jtklein: We should import View from `styledComponents` here.

FLGMwt: fixed

src/components/TaxonDetails/TaxonDetails.js
FLGMwt: removed

src/components/Notifications/Notifications.tsx
FLGMwt: fixed

src/components/MediaViewer/MediaViewer.js
FLGMwt: fixed

src/components/LoginSignUp/LoginSignUpWrapper.tsx
FLGMwt: good spot!

It looks like this was previously getting an extra padding from our old safe area which by chance made it look correct. This seemed to be the reason for the Android-specific padding. I've moved that explicit 60px padding to the scroll content container style which addresses the issue on both platforms so we don't need the Android-specific style anymore.

index.js
FLGMwt: added

src/components/LoginSignUp/LoginSignUpWrapper.tsx
```diff
@@ -78,25 +80,16 @@ const LoginSignupWrapper = ( {
     return unsubscribe;
   }, [navigation] );
 
-  const fitContentWithinScreenStyle = { height: windowHeight * 0.85 };
+  // const fitContentWithinScreenStyle = { height: windowHeight * 0.85 };
```
jtklein: This seems to be still required because now there is too little spacing between the "A global community ..." header and the first text input.

FLGMwt: ahhhh, I missed reverting this from testing. It's restored now.

PR #3167, opened 2025-10-30
Resolve all type errors in `src/sharedHooks/useCurrentUser.ts`.

frewsxcv:
Resolves all type errors that currently exist in this file:

Screenshot 2025-10-30 at 10 36 01 AM

### DIFF
--- a/src/sharedHooks/useCurrentUser.ts
+++ b/src/sharedHooks/useCurrentUser.ts
@@ -5,15 +5,6 @@ import User from "realmModels/User";
 
 const { useRealm } = RealmContext;
 
-function emptyChangeSet( ): Realm.CollectionChangeSet {
-  return {
-    insertions: [],
-    deletions: [],
-    newModifications: [],
-    oldModifications: []
-  };
-}
-
 const useCurrentUser = ( ): User | null => {
   const realm = useRealm( );
   const [currentUser, setCurrentUser] = useState<User | null>( null );
@@ -36,7 +27,12 @@ const useCurrentUser = ( ): User | null => {
     };
 
     // Run the listener on the results for the first time
-    listener( realmResults, emptyChangeSet( ) );
+    listener( realmResults, {
+      insertions: [],
+      deletions: [],
+      newModifications: [],
+      oldModifications: []
+    } );
 
     // Add the listener
     realmResults?.addListener( listener );


src/sharedHooks/useCurrentUser.ts
```diff
@@ -1,23 +1,23 @@
...
 const { useRealm } = RealmContext;
 
-const useCurrentUser = ( ): RealmUser | null => {
+const useCurrentUser = ( ): User | null => {
   const realm = useRealm( );
-  const [currentUser, setCurrentUser] = useState<RealmUser | null>( null );
+  const [currentUser, setCurrentUser] = useState<User | null>( null );
   useEffect( ( ) => {
-    const realmResults = realm.objects( "User" ).filtered( "signedIn == true" );
+    const realmResults = realm.objects( User ).filtered( "signedIn == true" );
```
frewsxcv, line 12: By specifying the model class constructor here instead of a string, it results in stronger returned typing incorporating that model class, rather than the generic `DefaultObject`.

Before:

Image

After:

Image

src/sharedHooks/useCurrentUser.ts
```diff
@@ -27,7 +27,7 @@ const useCurrentUser = ( ): RealmUser | null => {
     };
 
     // Run the listener on the results for the first time
-    listener( realmResults, {} );
```
FLGMwt, line 30: I think I'd much rather have a noop changeset argument for a sort of "exceptional" call than to double the arity of the listener. Was there another reason for making the listener argument optional?

frewsxcv: I made this change because it resulted in the smallest diff. But I can restore the arity. How would you like the second parameter specified? Restoring the `{}`? Or an empty `undefined` or `null` value?

src/sharedHooks/useCurrentUser.ts
FLGMwt: yeah, this is tasty, good spot

src/sharedHooks/useCurrentUser.ts
FLGMwt: I'd prefer the `{}`. The existing contract is "please give me an object with fields for everything that's changed" which an object with _no_ fields answers. Thank you by the way!

frewsxcv: @FLGMwt How does https://github.com/inaturalist/iNaturalistReactNative/pull/3167/commits/f49b64761112fa5342bd1e6f7cf50880c1843f69 look?

FLGMwt: Oh, I might have misunderstood your type error impetus for this. What makes `Partial` necessary here? If it's just the deletions field, let's make the argument `{ deletions: [] }`

frewsxcv: It's `Partial` because all of the fields in a `CollectionChangeSet` are required.

```
export type CollectionChangeSet = {
    /**
     * The indices in the collection where objects were inserted.
     */
    insertions: number[];
    /**
     * The indices in the collection where objects were deleted.
     */
    deletions: number[];
    /**
     * The indices in the new state of the collection where objects were modified.
     */
    newModifications: number[];
    /**
     * The indices in the old state of the collection where objects were modified.
     */
    oldModifications: number[];
};
```

Sure I'll push the change you're suggesting

frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3167/commits/0561b9dcb2100d056dfaa1b8e52d5aa63e7775b6

FLGMwt: alrighty, yeah, let's load that argument up:

```
{
  insertions: [],
  deletions: [],
  newModifications: [],
  oldModifications: []
}
```

frewsxcv: Is there any reason why you prefer that over using the type alias which has all that already specified?

Screenshot 2025-10-30 at 11 20 31 AM

frewsxcv: Oh I see. You would rather it be passed an object matching the shape with the required fields. Got it I'll push that now

FLGMwt: Here's my thinking:

The primary responsibility of `listener` is to call `realmResults.addListener(listener)`. We expect `realm` will call `listener` with `Realm.CollectionChangeSet` and we should expect to handle that shape of argument. Additionally, we are making a one-off "special" call to listener for initialization. If it's not too burdensome to call that with a noop argument that satisfies the type it will normally be called with, I'd prefer that over changing the functionality or signature of `listener` to cater to the one-off call.

I'm sorry to be so long winded on this ; D

frewsxcv: @FLGMwt Makes sense. This look good? https://github.com/inaturalist/iNaturalistReactNative/pull/3167/commits/cd28b748df4c5fde189d6a50b4d7eb1cb4087931

src/sharedHooks/useCurrentUser.ts
```diff
@@ -27,7 +36,7 @@ const useCurrentUser = ( ): RealmUser | null => {
...
+    listener( realmResults, emptyChangeSet( ) );
```
FLGMwt: yeah, if that satisfies the realm type, that looks great!

Could you please inline the argument? I don't think it warrants a helper function.

frewsxcv: I opted for the function because it felt more descriptive than a magic literal object with some weird fields but in any case https://github.com/inaturalist/iNaturalistReactNative/pull/3167/commits/6e066e86a7f3d9b252b0fbe76cd0bd1e8ea9c91f

src/sharedHooks/useCurrentUser.ts
```diff
@@ -1,19 +1,19 @@
 import { RealmContext } from "providers/contexts";
 import { useEffect, useState } from "react";
 import Realm from "realm";
-import type { RealmUser } from "realmModels/types";
```
frewsxcv, line 4: Random question to anyone who may know the answer– why do these `realmModels/types` type definitions exist? It seems like we could (and should?) always use the types from the actual models. Maybe a legacy artifact?

FLGMwt: we have some open questions about typing as we navigate the gradual migration from js and flow to typescript. we probably won't have guidelines or strong opinions about project-level typing decisions until we're closer to eliminating js/flow. For now, we're fairly comfortable keeping most types loose just to make the js=>ts file transitions. 

That said, @frewsxcv , your js=>ts contributions are a good help in that direction of getting to a point where we can make more holistic efforts for type strategy so thank you!

PR #3168, opened 2025-10-30
MOB-921 fix explore grid layout

sepeterson:
[MOB-921](https://linear.app/inaturalist/issue/MOB-921/grid-view-explore-bug)

Fixed an issue where Explore was showing List layout items in a Grid layout. I believe what was happening here is the `layout` prop was causing the underlying Flashlist to update the UI layout, but nothing was prompting renderItem to be called again as neither the `data` nor `extraData` props were changing.

### DIFF
--- a/src/components/ObservationsFlashList/ObservationsFlashList.js
+++ b/src/components/ObservationsFlashList/ObservationsFlashList.js
@@ -251,7 +251,8 @@ const ObservationsFlashList: Function = forwardRef( ( {
 
   const extraData = {
     gridItemWidth,
-    numColumns
+    numColumns,
+    layout
   };
 
   // only used id as a fallback key because after upload


src/components/ObservationsFlashList/ObservationsFlashList.js
```diff
@@ -251,7 +251,8 @@ const ObservationsFlashList: Function = forwardRef( ( {
 
   const extraData = {
     gridItemWidth,
-    numColumns
+    numColumns,
+    layout
```
FLGMwt, line 255: awwwwwwyeah, I love a good #OLC (one line commit) fix ; D

LGTM but it's maybe worth testing with Johannes's flash list update? https://github.com/inaturalist/iNaturalistReactNative/pull/3163

FLGMwt: er, actually I went ahead and merged that so could you pull main into this and test it again?

sepeterson: I can't recreate this issue after the flash list update so I'm going to close this PR and the ticket. Good looking out!

FLGMwt: closed as completed! ; D

PR #3170, opened 2025-10-31
Fix type errors where `forEach` does not exist on generator objects.

frewsxcv:
Screenshot 2025-10-31 at 11 43 45 AM

```
Property 'forEach' does not exist on type 'Generator<number, any, unknown>'.
```

Even though the method exists in practice, there's no guarantee this method exists based off the ECMAScript spec. <https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.keys>

To resolve this, we'll utilize a `for...of` loop.

### DIFF
--- a/src/realmModels/index.ts
+++ b/src/realmModels/index.ts
@@ -44,27 +44,27 @@ export default {
     if ( oldRealm.schemaVersion < 59 ) {
       const oldTaxa = oldRealm.objects( "Taxon" );
       const newTaxa = newRealm.objects( "Taxon" );
-      newTaxa.forEach( ( _, objectIndex: number ) => {
+      for ( const objectIndex of newTaxa.keys() ) {
         const newTaxon = newTaxa[objectIndex];
         const oldTaxon = oldTaxa[objectIndex];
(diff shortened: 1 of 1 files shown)


frewsxcv:
@FLGMwt Here's what I get when I try that:

```
Type 'Generator<number, any, unknown>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.ts(2802)
```
Screenshot 2025-11-03 at 11 05 26 AM

frewsxcv:
I'll try again because it might be resolved now. It was `es2015` until recently:

https://github.com/facebook/react-native/commit/5370347f54719f318a4e032aba6cbf2269e7c3d7

And the React Native TypeScript config was updated in:

https://github.com/inaturalist/iNaturalistReactNative/commit/e82ec31ed9ccd531121cc0f62a9f45b7a7dd01af

FLGMwt:
> Yep things are good now [a220cb0](https://github.com/inaturalist/iNaturalistReactNative/commit/a220cb02835c600f421cc81353ede34d61a54e72)

oh, sweet, thanks so much for figuring that out! I missed the older RN version on the PR branch before.

Closes MOB-1002

PR #3173, opened 2025-10-31
MOB-978 turn off restricted-syntax rule to allow forof loops

FLGMwt:
The eslint rule disallowing `forof` loops is from the airbnb config but is [based on outdated optimization guidance](https://stackoverflow.com/a/72072751/31670731). There are some cases where this forcing us to use "[].forEach" where I think this is actually harmful when it's not obvious that the async callback is unawaited, such as https://github.com/inaturalist/iNaturalistReactNative/blob/362ed537ee7e13df9175dc406f00b6a65058087d/src/sharedHelpers/sentinelFiles.ts#L74-L82

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -84,6 +84,7 @@ module.exports = {
     "no-var": 1,
     "operator-linebreak": [2, "before"],
     "prefer-const": [2, { destructuring: "all" }],
+    "no-restricted-syntax": 0,
     // "react/forbid-prop-types": 0,
     "react/prop-types": 0,
     "react/destructuring-assignment": 0,


.eslintrc.js
```diff
@@ -84,6 +84,7 @@ module.exports = {
     "no-var": 1,
     "operator-linebreak": [2, "before"],
     "prefer-const": [2, { destructuring: "all" }],
+    "no-restricted-syntax": 0,
```
FLGMwt, line 87: afaict, we can't targetedly configure these from airbnb except by explicitly providing the three other rules that we aren't addressing here https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L340-L358, but I don't think that's worth muddying our config with. I've never seen anyone use the `with` keyword ; D

PR #3175, opened 2025-10-31
Migrate `fetchUTFGridData.ts` to TypeScript.

frewsxcv:


src/sharedHelpers/fetchUTFGridData.ts
```diff
@@ -73,13 +83,17 @@ const getKeyForPixelExpansive = ( x, y, json ) => {
...
+  y: number,
+  json: UTFGrid | null | undefined
+): unknown | null => {
   if ( !json || !json?.data || !json?.grid ) return null;
 
   const key = getKeyForPixelExpansive( x, y, json );
 
   // Non existent key
-  if ( !json.data[key] ) { return null; }
+  if ( key === null || !json.data[key] ) { return null; }
```
frewsxcv, line 96: Without this change:

```
Type 'null' cannot be used as an index type.ts(2538)
```

PR #3176, opened 2025-10-31
typescriptify Developer.js and refactor size calculation

FLGMwt:
The primary changes here are:
- adding more types to `useAppSize`
- separating the `AppSizes` calculation from the hook and fixing the wonky promise / mutable state logic
- moving the AppSize display in Developer.js to an independent component instead of render helper functions in the main component

MOB-986

### DIFF
--- a/src/components/Developer/Developer.tsx
+++ b/src/components/Developer/Developer.tsx
@@ -10,7 +10,7 @@ import {
 } from "components/SharedComponents";
 import { fontMonoClass, View } from "components/styledComponents";
 import { t } from "i18next";
-import React, { PropsWithChildren } from "react";
+import React, { type PropsWithChildren } from "react";
 import { I18nManager, Platform, Text } from "react-native";
 import Config from "react-native-config";
 import RNFS from "react-native-fs";
--- a/src/components/Developer/hooks/useAppSize.ts
+++ b/src/components/Developer/hooks/useAppSize.ts
@@ -124,19 +124,19 @@ type AppSize = {
   [directoryName: string]: DirectoryEntrySize[]
 }
 
-export async function fetchAppSize(): Promise<AppSize> {
-  const existingDirectories = await Promise.all(
+async function fetchAppSize(): Promise<AppSize> {
+  const maybeExistingDirectories = await Promise.all(
     directories.map( async directory => ( {
       directory,
       exists: await RNFS.exists( directory.path )
     } ) )
   );
-  const filteredDirectories = existingDirectories
+  const existingDirectories = maybeExistingDirectories
     .filter( dir => dir.exists )
     .map( dir => dir.directory );
 
   const directoryToDirectorySizesKvps = await Promise.all(
-    filteredDirectories.map( async dir => {
+    existingDirectories.map( async dir => {
       const directoryEntrySizes = await getDirectoryEntrySizes( dir.path );
       return [dir.directoryName, directoryEntrySizes] as [string, DirectoryEntrySize[]];
     } )


src/components/Developer/Developer.tsx
```diff
@@ -49,48 +48,62 @@ const geomodelFileName = Platform.select( {
...
-              <CODE optionalClassName={boldClassname( line )}>
-                {line}
-              </CODE>
-            </P>
-          );
-        } )}
-      </View>
-    );
-  } ), [fileSizes] );
+/* eslint-disable i18next/no-literal-string */
```
frewsxcv: It doesn't look like this is being used in this component

```suggestion
```

src/components/Developer/hooks/useAppSize.ts
```diff
@@ -89,15 +95,15 @@ const androidDirectories = [
...
 
-export const directories = Platform.OS === "android"
-  ? sharedDirectories.concat( androidDirectories )
-  : iOSDirectories.concat( sharedDirectories );
+export const directories = sharedDirectories.concat( Platform.OS === "android"
+  ? androidDirectories
+  : iOSDirectories );
 
-export function formatAppSizeString( name, size ) {
+export function formatAppSizeString( name:string, size: number ): string {
```
frewsxcv:
```suggestion
export function formatAppSizeString( name: string, size: number ): string {
```

src/components/Developer/hooks/useAppSize.ts
```diff
@@ -89,15 +95,15 @@ const androidDirectories = [
...
   return `${name}: ${formatSizeUnits( size )}`;
 }
 
-export async function getDirectoryContentSizes( directory ) {
+export async function getDirectoryContentSizes( directory: string ): Promise<DirectorySize[]> {
```
frewsxcv: I'm not positive but I think if it's an `async` function you don't need to specify that it returns a `Promise`, that's implied

```suggestion
export async function getDirectoryContentSizes( directory: string ): DirectorySize[] {
```

src/components/Developer/hooks/useAppSize.ts
```diff
@@ -106,41 +112,49 @@ export async function getDirectoryContentSizes( directory ) {
...
+  const filteredDirectories = existingDirectories
+    .filter( dir => dir.exists )
+    .map( dir => dir.directory );
 
-  return fileSizes;
-};
+  const directoryToDirectorySizesKvps = await Promise.all(
+    filteredDirectories.map( async dir => {
+      const contentSizes = await getDirectoryContentSizes( dir.path );
+      return [dir.directoryName, contentSizes] as [string, DirectorySize[]];
```
frewsxcv: Is there a way to write this without the casting?

frewsxcv (approved):
👏🏻

src/components/Developer/hooks/useAppSize.ts
FLGMwt: the return type is inferred as `Promise<inferred type of return statement>`, but if you provide an explicit return type annotation, it needs the explicit outer `Promise`. 

image

src/components/Developer/hooks/useAppSize.ts
FLGMwt: We need at least some type hint to refine the type of things we're handing to `fromEntries` Without anything, the inferred type is `(string | Bloop)[]` where we want the tuple `[string, Bloop]`. the main alternative would be a return type annotation for the mapping function, but I think that this is a little more readable.

src/components/Developer/Developer.tsx
FLGMwt: aye, you're right. We were previously only doing it once just before the component where it mattered, I've updated it with the split out components to do the same.

src/components/Developer/hooks/useAppSize.ts
FLGMwt: oo, don't know why lint didn't catch this.

src/components/Developer/hooks/useAppSize.ts
frewsxcv: Ahh TIL, thanks for the info. I think I'm mixing this up with Rust

src/components/Developer/Developer.tsx
```diff
@@ -10,22 +10,38 @@ import {
 } from "components/SharedComponents";
 import { fontMonoClass, View } from "components/styledComponents";
 import { t } from "i18next";
-import type { Node } from "react";
-import React, { useCallback } from "react";
+import React, { PropsWithChildren } from "react";
```
jtklein: At first I wanted to comment that a separate line of `import type { PropsWithChildren } from "react";` would be consistent with our codebase. Then I checked our codebase and can see that it depended on the past developer which one we use. I do prefer a separate import type. Do you mind which one we use?

src/components/Developer/hooks/useAppSize.ts
```diff
@@ -89,58 +95,66 @@ const androidDirectories = [
...
-        tempFileSizes[directoryName] = contentSizes;
-        return tempFileSizes;
-      } ) );
-      if ( !_.isEmpty( size[0] ) ) {
-        setFileSizes( size[0] );
-      }
-    };
-    fetchAppSize( );
-  }, [] );
+export async function fetchAppSize(): Promise<AppSize> {
```
jtklein: Not imported anywhere. The function can be local only.

src/components/Developer/hooks/useAppSize.ts
```diff
@@ -89,58 +95,66 @@ const androidDirectories = [
   }
 ];
 
-export const directories = Platform.OS === "android"
-  ? sharedDirectories.concat( androidDirectories )
-  : iOSDirectories.concat( sharedDirectories );
+export const directories = sharedDirectories.concat( Platform.OS === "android"
+  ? androidDirectories
+  : iOSDirectories );
```
jtklein, line 100: This changes the order of items in directories on iOS, is that fine, or do we need the specific order for something?

src/components/Developer/hooks/useAppSize.ts
```diff
@@ -89,58 +95,66 @@ const androidDirectories = [
...
-    fetchAppSize( );
-  }, [] );
+export async function fetchAppSize(): Promise<AppSize> {
+  const existingDirectories = await Promise.all(
+    directories.map( async directory => ( {
+      directory,
+      exists: await RNFS.exists( directory.path )
+    } ) )
+  );
+  const filteredDirectories = existingDirectories
```
jtklein: I would argue that the result of `.filter( dir => dir.exists )` are the existingDirectories. And the previous const could have another name.

src/components/Developer/Developer.tsx
```diff
@@ -49,48 +65,66 @@ const geomodelFileName = Platform.select( {
...
-              <CODE optionalClassName={boldClassname( line )}>
-                {line}
-              </CODE>
-            </P>
-          );
-        } )}
-      </View>
-    );
-  } ), [fileSizes] );
+const AppFileSizes = () => {
```
jtklein, line 109: Do you prefer this, i.e. having `<AppFileSizes />` in the return tree, over a
```
{ 
!appSize ? null : 
    <>
      {Object.entries( appSize ).map( ( [directoryName, directoryEntrySizes] ) => (
        <DirectoryFileSizes
          key={directoryName}
          directoryName={directoryName}
          directoryEntrySizes={directoryEntrySizes}
        />
      ) )}
    </>
}
```

src/components/Developer/Developer.js
```diff
@@ -49,48 +65,66 @@ const geomodelFileName = Platform.select( {
...
+              {line}
+            </CODE>
+          </P>
+        );
+      } )}
+    </View>
   );
+};
 
-  const displayFileSizes = useCallback( ( ) => Object.keys( fileSizes ).map( directory => {
```
jtklein, line 64: Could you make more granular commits here, so that the process of changes becomes easier to review. The refactor of this code block into standalone component, i.e. copying it over into the new component, could have been one commit. So that the changes in this one commit are only copy paste and indentation changes. This makes it easier for me to see which pieces of the code are actually changed and which ones just copied over.

jtklein (approved):
Thanks, this looks good. Left a few inline comments.
Overall a change this big is easier to review if split into smaller pieces. Multiple PRs if possible, but even more granular commits make it easier to review.
The process of gradual code changes (and commit messages) - step by step how you arrived at the diff - can already explain some of the changes without having to write a comment on the PR.

src/components/Developer/Developer.tsx
FLGMwt: oh, yeah, I hadn't noticed that. I created a MOB ticket for the team to figure out which we want to use, add the appropriate lint config, and autofix to make the repo consistent.

Between the three(?) options:
```
import { value } from "some-module";
import type { Type } from "some-module";
OR
import { Type, value } from "some-module";
OR
import { type Type, value } from "some-module";
```
I don't much care which one, but it'd be nice to be linted / autofixable.

src/components/Developer/Developer.tsx
FLGMwt: Yeah, I think this has a pretty strong case for being its own separate component. It's the only consumer of `useAppSize` and it handles a very focused concern.

src/components/Developer/Developer.js
FLGMwt: We chatted about this, but reiterating: I'm trying to be more mindful of this going forward : )

PR #3179, opened 2025-11-02
Update to react-native 0.80.x

jtklein:
This PR includes another update with a react-native minor version jump to RN 0.80.x ([release notes](https://reactnative.dev/blog/2025/06/12/react-native-0.80)).

Two packages that were 1) crashing the app during start on Android (realm) and 2) presenting a broken UI (flashlist) have been updated in previous PRs now, so there is nothing blocking this one.

Compiled and tested in Debug and Release mode on both platforms.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -9,7 +9,7 @@ PODS:
   - BVLinearGradient (2.8.3):
     - React-Core
   - DoubleConversion (1.1.6)
-  - fast_float (6.1.4)
+  - fast_float (8.0.0)
   - FasterImage (1.7.2):
     - FasterImage/Nuke (= 1.7.2)
     - FasterImage/NukeUI (= 1.7.2)
@@ -36,7 +36,7 @@ PODS:
     - React-Core
     - React-RCTFabric
     - ReactCommon/turbomodule/core
-  - FBLazyVector (0.79.5)
+  - FBLazyVector (0.80.2)
   - Firebase/CoreOnly (12.3.0):
     - FirebaseCore (~> 12.3.0)
   - FirebaseAnalytics/Core (12.3.0):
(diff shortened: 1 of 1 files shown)


ios/iNaturalistReactNative/Info.plist
```diff
@@ -73,6 +73,8 @@
 	<string>Export iNaturalist photos to your library.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Export and import iNaturalist photos to and from your library.</string>
+	<key>RCTNewArchEnabled</key>
+	<true/>
```
FLGMwt, line 77: what's the significance of this? I might have misunderstood but I thought we were already on new arch

android/app/src/main/java/com/inaturalistreactnative/MainApplication.kt
```diff
@@ -35,10 +33,6 @@ class MainApplication : Application(), ReactApplication {
 
   override fun onCreate() {
     super.onCreate()
-    SoLoader.init(this, OpenSourceMergedSoMapping)
-    if (BuildConfig.IS_NEW_ARCHITECTURE_ENABLED) {
-      // If you opted-in for the New Architecture, we load the native entry point for this app.
-      load()
-    }
```
FLGMwt, line 42: Is this now handled by `loadReactNative`?

FLGMwt (approved):
LGTM, assuming the new arch was already enabled on iOS and this isn't _newly_ enabling it.

android/app/src/main/java/com/inaturalistreactnative/MainApplication.kt
jtklein: Yes.

ios/iNaturalistReactNative/Info.plist
jtklein: Yes, we were on the New Architecture, it is on by default. From this version on the flag is in Info.plist (and no longer in Podfile, I believe).

PR #3180, opened 2025-11-03
Migrate `DetailsMap` to TypeScript.

frewsxcv:
* [`src/components/SharedComponents/Map/DetailsMap.tsx`](diffhunk://#diff-14c0b3d26b78aac5fd5b82674fa549bc1d9eb39f5d65399df1aa82d5b344f972L1-L2): Migrated the component from JavaScript to TypeScript, replacing Flow types with TypeScript interfaces and type annotations for props and internal types. [[1]](diffhunk://#diff-14c0b3d26b78aac5fd5b82674fa549bc1d9eb39f5d65399df1aa82d5b344f972L1-L2) [[2]](diffhunk://#diff-14c0b3d26b78aac5fd5b82674fa549bc1d9eb39f5d65399df1aa82d5b344f972R12-R48)
* [`src/components/SharedComponents/Map/Map.tsx`](diffhunk://#diff-387cf1ada3798a4a4128a00edcbf9d330d774e4f50c9e376531d6624a03a64e7R54-R59): Added a new `TileMapParams` interface to define the expected shape of `tileMapParams` and updated its usage in the `Props` interface. [[1]](diffhunk://#diff-387cf1ada3798a4a4128a00edcbf9d330d774e4f50c9e376531d6624a03a64e7R54-R59) [[2]](diffhunk://#diff-387cf1ada3798a4a4128a00edcbf9d330d774e4f50c9e376531d6624a03a64e7L79-R85)
* [`src/components/SharedComponents/Map/DetailsMap.tsx`](diffhunk://#diff-14c0b3d26b78aac5fd5b82674fa549bc1d9eb39f5d65399df1aa82d5b344f972L72-R75): Removed the unused `showLocationIndicator` prop from both the component signature and its usage. [[1]](diffhunk://#diff-14c0b3d26b78aac5fd5b82674fa549bc1d9eb39f5d65399df1aa82d5b344f972L72-R75) [[2]](diffhunk://#diff-14c0b3d26b78aac5fd5b82674fa549bc1d9eb39f5d65399df1aa82d5b344f972L117)

### DIFF
--- a/src/components/SharedComponents/Map/DetailsMap.tsx
+++ b/src/components/SharedComponents/Map/DetailsMap.tsx
@@ -25,6 +25,7 @@ interface Props {
   closeModal: () => void,
   coordinateString?: string,
   headerTitle?: React.ReactNode,
+  // TODO MOB-1038: reconcile the type issues here requiring the intersection
   observation?: Observation & RealmObservation,
   region?: Region,
   tileMapParams: Record<string, string> | null,


src/styles/tailwindColors.d.ts
```diff
@@ -0,0 +1,30 @@
+declare module "styles/tailwindColors" {
```
frewsxcv, line 1: If you prefer, I can pull this out to a separate pull request. Prior to this pull request, references to Tailwind colors in TypeScript would always error.

src/components/SharedComponents/Map/DetailsMap.js
```diff
@@ -114,7 +115,6 @@ const DetailsMap = ( {
           observation={observation}
           region={region}
           showCurrentLocationButton
-          showLocationIndicator={showLocationIndicator}
```
frewsxcv, line 118: This component prop doesn't exist.

src/components/SharedComponents/Map/Map.tsx
```diff
@@ -51,6 +51,12 @@ const getDefaultRegion = ( initialLatitude, initialLongitude ) => ( {
     : 100
 } );
 
+export interface TileMapParams {
+  order: unknown;
+  order_by: unknown;
+  per_page: unknown;
```
abbeycampbell: Is it possible to go ahead and type these more explicitly? Or would that balloon this PR too much? (just getting a sense of how we're handling typescript migration as we go)

frewsxcv: I just looked at this code again and I read it wrong the first time. These three fields aren't just getting fetched, they're getting deleted:

https://github.com/inaturalist/iNaturalistReactNative/blob/0bdc53bd3c4387d2ab252998be141e3c8c081ac2/src/components/SharedComponents/Map/Map.tsx#L199-L205

It actually seems like the user can pass any params here since they just get stringified onto the URL:

https://github.com/inaturalist/iNaturalistReactNative/blob/0bdc53bd3c4387d2ab252998be141e3c8c081ac2/src/components/SharedComponents/Map/Map.tsx#L319

I just pushed https://github.com/inaturalist/iNaturalistReactNative/pull/3180/commits/fa0083db974820d3140d9413331842c39b39f7ac. In the future it might make sense to look up the relevant API endpoint and add an interface for the expected params to the endpoint, but that seems sufficiently outside the scope of this pull request (since that's a different file which has many other type errors).

src/components/SharedComponents/Map/SwitchMapTypeButton.tsx
```diff
@@ -12,7 +12,7 @@ const DROP_SHADOW = getShadow( );
 interface Props {
   currentMapType?: string;
   mapType?: string;
-  setCurrentMapType: Function;
```
frewsxcv, line 15: Commit hook was failing about this so just added it in

src/styles/tailwindColors.d.ts
jtklein: The smaller the PR the easier it is to review (and revert). In general I would prefer PRs to be as atomic as possible. Lines of separation can be: different features, different concerns (e.g. TS migrations vs. logic changes vs. renaming vs. refactors), different files.
So, here I can see 5 files touched independent from one another, so this can be 5 PRs. Haven't looked closely but if the DetailsMap refactor needs any of the other four than they can be treated as prerequisites and the DetailsMap started as soon as the others are merged.

frewsxcv: Creating five pull requests for this tiny changeset seems excessive to me:

Screenshot 2025-11-07 at 9 24 00 AM

But I don't have any power or say in this project so if that's what's necessary then I'll do it.

Would you like me to create the four other pull requests now and close this one?

frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3201

frewsxcv: #3201 merged. And I just merged in `main` and the diff for this pull request is now smaller.

src/components/SharedComponents/Map/DetailsMap.tsx
```diff
@@ -13,35 +11,38 @@ import {
...
-    privateLatitude?: number,
-    longitude?: number,
-    privateLongitude?: number,
-    obscured?: boolean
-  },
-  region?: Object,
-  showLocationIndicator: boolean,
-  tileMapParams: Object,
+  headerTitle?: React.ReactNode,
+  observation?: Observation & RealmObservation,
```
FLGMwt, line 29:
```suggestion
  // TODO MOB-1038: reconcile the type issues here requiring the intersection
  observation?: Observation & RealmObservation,
```

Discussed this w/ team and we decided to defer a type reconciliation, I'm just adding a comment here to reference that

PR #3182, opened 2025-11-03
MOB-674 upload status text to 2 lines if necessary

sepeterson:
[MOB-674](https://linear.app/inaturalist/issue/MOB-674/upload-status-text-should-wrap)

When the upload status text gets too long, it should go to a 2nd line instead of overflowing the screen. I couldn't actually recreate this bug in English but it was present for other languages

image


jtklein (changes requested):
This seems to be working correctly while observations are uploading, and in case there are errors during the upload (according to your screenshot, although I was unable to test this locally.
However, this introduces changes in other states as well:
When the upload is finished and some errors happened the only text visible is shifted to the right:
EA7F3537-F19C-43D6-8195-B1F7914ED25B_4_5005_c

When all was successful the checkmark icon is shifted to the right:
Deletions:
3C5BB09F-F18A-4B5F-A7DE-BC6A552D6ED6_1_102_o
Uploads:
6B741F19-7771-4313-B85E-01B059488CE7_1_102_o


Ideally, the change would only affect the text.status while the app is uploading without affecting any other pieces of the `SimpleUploadBanner`.

sepeterson:
See screenshots under these sections:
<details>

<summary>With a mock error state to create overflow</summary>

Simulator Screenshot - iPhone 13 Pro - 2025-11-06 at 11 05 06Simulator Screenshot - iPhone 13 Pro - 2025-11-06 at 11 12 48

</details>

<details>

<summary>And testing with live data</summary>

IMG_2279
IMG_2280
IMG_2281


</details>

src/components/MyObservations/SimpleUploadBanner.js
```diff
@@ -51,38 +52,44 @@ const SimpleUploadBanner = ( {
...
-        <Pressable
-          onPress={handleSyncButtonPress}
-          accessibilityRole="button"
-          disabled={syncDisabled}
-        >
-          <Body2
-            className={classnames( "text-darkGray py-3", {
-              "text-white": status.styling === "white-on-green"
-            } )}
+        {status.text && (
```
jtklein, line 76: Good catch! Because `status.text` can be an empty string this would render as a views without content, correct?

src/components/MyObservations/SimpleUploadBanner.js
```diff
@@ -51,38 +52,44 @@ const SimpleUploadBanner = ( {
...
+            accessibilityRole="button"
+            disabled={syncDisabled}
+            className="shrink-[2] ml-2"
           >
-            {status.text}
-          </Body2>
-        </Pressable>
+            <Body2
+              className={classnames( "text-darkGray py-3 text-center", {
+                "text-white": status.styling === "white-on-green"
```
jtklein: Please, reuse `isWhiteOnGreenStyling` here.

src/components/MyObservations/SimpleUploadBanner.js
```diff
@@ -51,38 +52,44 @@ const SimpleUploadBanner = ( {
...
       <INatIcon name="checkmark" size={11} color={colors.inatGreen} />
     </View>
   );
 
   const renderErrorText = ( ) => (
-    <Body2 className="color-white">
-      {separator}
-      <Body2 className="color-warningRed">{error}</Body2>
+    <Body2 className="color-white flex-2 text-center ml-1">
+      {isWhiteOnGreenStyling
```
jtklein, line 62: I was surprised to see this boolean here. It is indicating which background styling mode we should display, so I was wondering how can i also influence what kind of separator we want. So, I checked the current designs and realized that there are many more things to change in this banner than only making the text wrap as you did. So, for now this is okay, even though the name is a bit confusing.

jtklein (approved):
While looking into this review, I checked the current designs and realized that there are many more things to change in this banner than only making the text wrap as you did.
Anyway, I don't want to scope creep this ticket, as it is appropriately dealt with in this PR. So, I'll approve now as I can see it working as it should according to the ticket.
Only one minor change, so you can merge after it on your own time.

While looking at the designs I did push one minor change on your branch. Let me know if you would have preferred a PR into your branch.

src/components/MyObservations/SimpleUploadBanner.js
sepeterson: Yep, that's correct!

PR #3183, opened 2025-11-04
Fix type error and unnecessary rendering in `LocationIndicator`.

frewsxcv:
Screenshot 2025-11-03 at 7 51 58 PM

This pull request resolves the type error above.

In addition, one functional difference is it prevents the `Circle` component from rendering in the document tree when `positionalAccuracy` is `0` since `!!positionalAccuracy` is false-y. Despite that difference, there should be no difference visually since a circle with no radius wouldn't have ever be visible.

### DIFF
--- a/src/components/SharedComponents/Map/LocationIndicator.tsx
+++ b/src/components/SharedComponents/Map/LocationIndicator.tsx
@@ -18,16 +18,15 @@ const LocationIndicator = ( {
   positionalAccuracy
 }: Props ) => (
   <>
-    {!!positionalAccuracy
-      && (
-        <Circle
-          center={{ latitude, longitude }}
-          radius={positionalAccuracy}
-          strokeWidth={2}
-          strokeColor={colors.inatGreen}
-          fillColor="rgba( 116, 172, 0, 0.2 )"
-        />
-      )}
+    {!!positionalAccuracy && (
+      <Circle
+        center={{ latitude, longitude }}
+        radius={positionalAccuracy}
+        strokeWidth={2}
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/Map/LocationIndicator.tsx
```diff
@@ -18,13 +18,17 @@ const LocationIndicator = ( {
   positionalAccuracy
 }: Props ) => (
   <>
-    <Circle
-      center={{ latitude, longitude }}
-      radius={positionalAccuracy}
-      strokeWidth={2}
-      strokeColor={colors.inatGreen}
-      fillColor="rgba( 116, 172, 0, 0.2 )"
-    />
```
FLGMwt, line 27: Since this conditional render is a functional change, would you mind describing how this impacts the rendering of the LocationIndicator component?

Also this is a nit, but the codebase tends to prefer `{!!positionalAccuracy && <Circle />}` for boolean conditional renders.

frewsxcv: > Since this conditional render is a functional change, would you mind describing how this impacts the rendering of the LocationIndicator component?

Yep, I tried to articulate it in the pull request description, but I'll add more details to the description now.

> Also this is a nit, but the codebase tends to prefer `{!!positionalAccuracy && <Circle />}` for boolean conditional renders.

Pushed a commit for this

FLGMwt: I'm sorry, I entirely missed this in the description. Thank you!

PR #3185, opened 2025-11-04
Remove unused accessibility props on `INatIconButton` usage.

frewsxcv:


src/components/SharedComponents/Map/SwitchMapTypeButton.tsx
```diff
@@ -48,12 +48,6 @@ const SwitchMapTypeButton = ( {
       )}
       style={DROP_SHADOW}
       accessibilityLabel={t( "Toggle-map-type" )}
-      accessibilityRole="button"
```
frewsxcv, line 51: `accessibilityRole` is not a prop on `INatIconButton`, and it already has `accessibilityRole="button"` specified in the underlying `Pressable` element

src/components/SharedComponents/Map/SwitchMapTypeButton.tsx
```diff
@@ -48,12 +48,6 @@ const SwitchMapTypeButton = ( {
...
-      accessibilityState={
-        currentMapType === "standard"
-          ? t( "Standard--map-type" )
-          : t( "Satellite--map-type" )
-      }
```
frewsxcv, line 56: `accessibilityState` is not a prop on `INatIconButton`. `accessibilityState` doesn't take a translated string so I'm not sure what the original intention was here

FLGMwt (approved):
Good spot! Thanks as always ; D

PR #3186, opened 2025-11-05
Resolve all type errors in `WarningSheet` component.

frewsxcv:
This resolves the three type errors in `WarningSheet`.

### DIFF
--- a/src/components/SharedComponents/ButtonBar.tsx
+++ b/src/components/SharedComponents/ButtonBar.tsx
@@ -8,11 +8,14 @@ const DROP_SHADOW = getShadow( {
   offsetHeight: -2
 } );
 
-interface ButtonConfiguration {
+export interface ButtonConfiguration {
   title: string;
   onPress: ( ) => void;
   isPrimary: boolean;
   className?: string;
+  disabled?: boolean;
+  level?: string;
+  loading?: boolean;
 }
 
 interface Props extends PropsWithChildren {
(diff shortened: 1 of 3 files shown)


src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -35,7 +35,7 @@ interface Props {
   onPressClose?: Function;
   snapPoints?: Array<string>;
   insideModal?: boolean;
-  keyboardShouldPersistTaps: string;
+  keyboardShouldPersistTaps?: string;
```
frewsxcv, line 38: Ten or so lines below:

```
keyboardShouldPersistTaps = "never",
```

It has a default value specified, so this is just updating the type definition to reflect that.

src/components/SharedComponents/Sheets/WarningSheet.tsx
```diff
@@ -35,13 +36,14 @@ const WarningSheet = ( {
...
 }: Props ) => {
-  const buttons = [
+  const buttons: ButtonConfiguration[] = [
     {
       title: buttonText,
       onPress: confirm,
       disabled: loading,
       level: buttonType || "warning",
       loading,
+      isPrimary: false,
```
frewsxcv, line 46: `isPrimary` is a required property of the configuration. It was previously implicitly `undefined` false-y, and now it's actually specified.

src/components/SharedComponents/ButtonBar.tsx
```diff
@@ -8,11 +8,14 @@ const DROP_SHADOW = getShadow( {
...
 
-interface ButtonConfiguration {
+export interface ButtonConfiguration {
   title: string;
   onPress: ( ) => void;
   isPrimary: boolean;
   className?: string;
+  disabled?: boolean;
+  level?: string;
+  loading?: boolean;
```
frewsxcv, line 18: These are all specified in `WarningSheet`'s usage of `ButtonConfiguration` and are just passed through to the underlying `Button` component

PR #3188, opened 2025-11-05
Migrate `MasonryLayout` component to TypeScript.

frewsxcv:
This component was a bit trickier to migrate because of the extra object fields it adds on the fly to existing Realm objects, hence all the type aliases in this pull request. I'll add in-line pull request comments for additional context.


src/components/ObsDetails/MasonryLayout.tsx
```diff
@@ -1,40 +1,66 @@
...
 const spacing = 6;
 
-const photoUrl = photo => Photo.displayLocalOrRemoteLargePhoto( photo );
+const photoUrl = ( photo: PhotoItem ) => {
+  const url = Photo.displayLocalOrRemoteLargePhoto( photo );
+  if ( !url ) {
+    throw new Error( "No photo URL found" );
+  }
+  return url;
+};
```
frewsxcv, line 29: Prior to this pull request, the `photoUrl` function was typed to return `string | null`. This causes an issue later in the file because of this:

```
const imageDimensions = await getImageDimensions( photoUrl( item ) );
```

`getImageDimensions` is expected to never be passed a `null` value, so presumably `photoUrl` in practice is always returning a `string`. So to safely alter the return type (without casting), I added the `throw` here. If you'd prefer casting I can change it to that.

src/components/ObsDetails/MasonryLayout.tsx
```diff
@@ -1,40 +1,66 @@
...
+
+// Photos have a `url` property, sounds have a `file_url` property
+const isSound = ( item: Item ): item is SoundItem => (
+  "file_url" in item
+);
```
frewsxcv, line 34: This is a new [assertion function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) that is used to narrow the type from `PhotoItem | SoundItem` into `SoundItem`. It replaces these usages in the old code and should be functionally equivalent:

```
if ( item.file_url ) {
```

src/components/ObsDetails/MasonryLayout.js
```diff
@@ -1,40 +1,66 @@
...
+      const newColumns: HydratedItem[][] = ( Array.from( { length: numColumns }, () => [] ) );
 
       const itemPromises = items.map( async item => {
         // If a sound, just return it
-        if ( item.file_url ) {
+        if ( isSound( item ) ) {
           return item;
         }
         const imageDimensions = await getImageDimensions( photoUrl( item ) );
-        return { ...item, ...imageDimensions };
```
frewsxcv: This old code was potentially problematic, which TypeScript caught. The spread operator does _not_ copy instance methods.

```
> class Cat {
|     age = 10;
|     meow() {}
| }
undefined

> cat = new Cat()
Cat { age: 10 }

> { ...cat, name: "Kevin" }
{ age: 10, name: 'Kevin' }

> { ...cat, name: "Kevin" }.meow()
Uncaught TypeError: {(intermediate value)(intermediate value)}.meow is not a function

> Object.assign(cat, { name: "Kevin" }).meow()
undefined
```


So Realm methods on `item` would not be copied here, which is presumably not what we want. To remedy this, I switched this to use the `Object.assign` function which should do what we want.

src/components/ObsDetails/MasonryLayout.tsx
```diff
@@ -59,16 +89,21 @@ const MasonryLayout = ( { items, onImagePress } ) => {
...
   );
 
-  const renderSound = ( item, index, column ) => (
+  const renderSound = ( item: SoundItem, index: number, column: number ) => (
     <SoundContainer
       key={`MasonryLayout.column${column}.sound_${index}`}
       sizeClass="w-full aspect-square"
       sound={item}
       isVisible
+      autoPlay={false}
```
frewsxcv, line 99: Required and previously implicitly false-y

src/components/ObsDetails/MasonryLayout.js
frewsxcv: I extracted this to https://github.com/inaturalist/iNaturalistReactNative/pull/3261 in an effort to make this pull request a little smaller and more manageable.

src/components/ObsDetails/MasonryLayout.tsx
FLGMwt: I quite like this pattern for very explicit type narrowing. I think there's a lot of places where we would benefit from this. That said, since it's a new pattern / "tool" being introduced to the codebase, even though it's "correct", I'd like to bring this up with the team to discuss internally.

FLGMwt: followup because I forgot to say this: no action expected @frewsxcv , I just think this change might be in a holding pattern for a moment until we talk about this.

PR #3189, opened 2025-11-05
remove misleading suggestion checkmark state

abbeycampbell:
[MOB-797](https://linear.app/inaturalist/issue/MOB-797/change-the-green-checkmark-under-top-id-suggestion-on-the-suggestions)

Removes anything that makes the checkmark on a suggestion indicate a selected state, since this is misleading. Props `isTopSuggestion` and `checkmarkFocused` seemed to be used only to determine when the green checkmark should be rendered, so we can say goodbye to those.

### DIFF
--- a/src/components/SharedComponents/TaxonResult.tsx
+++ b/src/components/SharedComponents/TaxonResult.tsx
@@ -164,22 +164,6 @@ const TaxonResult = ( {
   // useTaxon could return null, and it's at least remotely possible taxonProp is null
   if ( !usableTaxon ) return null;
 
-  const renderCheckmark = () => (
-    <INatIconButton
-      className="ml-2"
-      icon="checkmark-circle-outline"
-      size={40}
-      color={String(
-        clearBackground
-          ? colors?.white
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/TaxonResult.tsx
```diff
@@ -166,36 +164,21 @@ const TaxonResult = ( {
...
-            ? colors?.white
-            : colors?.darkGray
-        )}
-        onPress={() => handleCheckmarkPress( usableTaxon )}
-        accessibilityLabel={accessibilityLabel}
-        testID={`${testID}.checkmark`}
-      />
-    );
-  };
+  const renderCheckmark = () => (
```
jtklein: Since this function now has no logic anymore, let us add the JSX component directly into the return tree that should be a bit clearer.

abbeycampbell: of course! And good to know about shipping deprecated icons, I'll keep that in mind

PR #3153, opened 2025-10-20
Fix spurious test user by setting `currentUser`.

frewsxcv:
The spurious failure was seen here and in other pull requests:

https://github.com/inaturalist/iNaturalistReactNative/actions/runs/18662797244/job/53207075939?pr=3152

```
ObsEdit › displays the number of photos in global state obsPhotos

    User tried to edit observation they do not own

      102 |   // This should never, ever happen
      103 |   if ( currentObservation?.user && currentUser && currentUser.id !== currentObservation.user.id ) {
    > 104 |     throw new Error( "User tried to edit observation they do not own" );
          |           ^
      105 |   }
      106 |
      107 |   return (

      at ObsEdit (src/components/ObsEdit/ObsEdit.js:104:11)
      at Object.Component [as react-stack-bottom-frame] (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12679:20)
      at callComponentInDEV (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:3622:25)
      at renderWithHooks (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:5635:19)
      at updateFunctionComponent (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:7086:18)
      at callback (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:540:16)
      at runWithFiberInDEV (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10925:22)
      at performUnitOfWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10763:41)
      at workLoopSync (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10744:11)
      at renderRootSync (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10299:39)
      at performWorkOnRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:1894:7)
      at performSyncWorkOnRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:1761:21)
      at flushSyncWorkAcrossRoots_impl (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:11290:13)
      at flushPassiveEffects (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:11159:11)
      at callback (node_modules/react/cjs/react.development.js:862:34)
      at flushActQueue (node_modules/react/cjs/react.development.js:1151:10)
[...]

PR #3190, opened 2025-11-06
MOB-1000 mock user to fix unit test flake

FLGMwt:
Credit for this fix goes to @frewsxcv who implemented it in https://github.com/inaturalist/iNaturalistReactNative/pull/3153/files. I should have been more clear, I didn't mean to expect an investigation before merging. We've seen this crop up a couple more times so I'm recreating their PR. Thanks for this fix @frewsxcv !

### DIFF
--- a/tests/unit/components/ObsEdit/ObsEdit.test.js
+++ b/tests/unit/components/ObsEdit/ObsEdit.test.js
@@ -2,7 +2,6 @@ import { screen, waitFor } from "@testing-library/react-native";
 import ObsEdit from "components/ObsEdit/ObsEdit";
 import React from "react";
 import { View } from "react-native";
-import * as useCurrentUser from "sharedHooks/useCurrentUser";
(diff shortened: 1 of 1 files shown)


tests/unit/components/ObsEdit/ObsEdit.test.js
```diff
@@ -48,6 +49,7 @@ const mockObservation = factory( "LocalObservation", {
 
 describe( "ObsEdit", () => {
   beforeEach( ( ) => {
+    jest.spyOn( useCurrentUser, "default" ).mockImplementation( ( ) => mockUser );
```
jtklein: Why not use a `jest.mock` here? `useCurrentUser` has no other exports than default and we do not update this mock anywhere else in this file.

FLGMwt: Aye, that's better

PR #3192, opened 2025-11-06
Migrate `src/api/error.js` to TypeScript.

frewsxcv:


src/api/error.ts
```diff
@@ -1,23 +1,25 @@
...
-  constructor( json: Object, status?: number, context?: Object ) {
+  constructor(
+    json: Record<string, unknown> & { status: string | number },
+    status?: number,
+    context?: Record<string, unknown> | null
+  ) {
     super( JSON.stringify( json ) );
     this.json = json;
-    this.status = status || json.status;
+    this.status = status || +json.status;
```
frewsxcv: Functional change 1: This ensures the string statuses returned by the backend are always stored as number properties

src/api/error.ts
```diff
@@ -112,8 +152,8 @@ async function handleError( e: Object, options: Object = {} ): Object {
   let errorJson;
   try {
     errorJson = await e.response.json( );
-  } catch ( jsonError ) {
-    if ( jsonError.message.match( /JSON Parse error/ ) ) {
+  } catch ( jsonError: unknown ) {
+    if ( jsonError instanceof Error && jsonError.message.match( /JSON Parse error/ ) ) {
```
frewsxcv, line 156: Functional change 2: If a non `Error` is raised, then `jsonError.message` won't exist

src/api/error.ts
FLGMwt: Oh nice, great fix. I checked other places, but yeah, afaict, confirmed that we always treat these as numbers: 

Image

nit: I think the `+` operator coercion can be hard to notice and the codebase tends to use the `Number` constructor.

```suggestion
    this.status = status || Number(json.status);
```

src/api/error.ts
FLGMwt: Sounds good, thanks for the callout.

src/api/error.ts
frewsxcv: Hmm I wonder how those checks were working in the past. Was there some string to number conversion happening elsewhere? IIRC, all error response statuses from the API return strings

src/api/error.ts
```diff
@@ -1,23 +1,25 @@
...
-  constructor( json: Object, status?: number, context?: Object ) {
+  constructor(
+    json: Record<string, unknown> & { status: string | number },
+    status?: number,
+    context?: Record<string, unknown> | null
+  ) {
     super( JSON.stringify( json ) );
     this.json = json;
-    this.status = status || json.status;
+    this.status = status || Number(json.status);
```
FLGMwt:
```suggestion
    this.status = status || Number( json.status );
```

PR #3200, opened 2025-11-10
Remove unneeded `getImageDimensions` helper function.

frewsxcv:
The built-in React Native function has an `async` API that we can call directly.


src/components/Match/PhotosSection.js
```diff
@@ -8,8 +8,8 @@ import {
 } from "components/styledComponents";
 import _, { compact } from "lodash";
 import React, { useEffect, useState } from "react";
+import ReactNative from "react-native";
```
jtklein: Can we do `import { Image } from "react-native";` here instead?

src/components/ObsDetails/MasonryLayout.js
```diff
@@ -1,7 +1,7 @@
 import { ScrollView, View } from "components/styledComponents";
 import React, { useEffect, useState } from "react";
+import ReactNative from "react-native";
```
jtklein: Same as above?

src/components/Match/PhotosSection.js
```diff
@@ -69,7 +69,7 @@ const PhotosSection = ( {
   useEffect( ( ) => {
     const checkImageOrientation = async ( ) => {
       if ( observationPhoto ) {
-        const imageDimensions = await getImageDimensions( observationPhoto );
+        const imageDimensions = await ReactNative.Image.getSize( observationPhoto );
```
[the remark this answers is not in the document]
frewsxcv: > The React Native `Image.getSize` API does not support async/await directly.

🤔 

https://github.com/facebook/react-native/blob/d314e5f4e4b72a0cfa4cfb8a5216995a3d453aea/packages/react-native/Libraries/Image/Image.d.ts#L343

jtklein (changes requested):
Can we import { Image }?

src/components/Match/PhotosSection.js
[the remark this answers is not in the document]
jtklein: Wow, sorry for the noise. Am still only experimenting with the Copilot reviews.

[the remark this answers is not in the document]
jtklein: @copilot
`if (typeof success !== 'function') {
    return promise;
  }`

src/components/Match/PhotosSection.js
```diff
@@ -8,8 +8,8 @@ import {
 } from "components/styledComponents";
 import _, { compact } from "lodash";
 import React, { useEffect, useState } from "react";
+import { Image as RNImage } from "react-native";
```
frewsxcv, line 14: `Image` is already an import

src/components/ObsDetails/MasonryLayout.js
frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3200/commits/dafd61ea8773153d7cf642a5cca8e56df177f8fe

src/components/Match/PhotosSection.js
frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3200/commits/dafd61ea8773153d7cf642a5cca8e56df177f8fe

src/components/Match/PhotosSection.js
jtklein: Then we should be able to use `Image` as it is imported already. Coming from styledComponents it is just a wrapper around RN's Image.

frewsxcv: Good call https://github.com/inaturalist/iNaturalistReactNative/pull/3200/commits/b00bd1a5b19b09b8207459784dd613aaebeb1d7c

frewsxcv: @jtklein Doesn't appear to work:

```
FAIL tests/unit/components/Match/Match.test.js
  ● Match › should show location permissions button if permissions not granted

    TypeError: _styledComponents.Image.getSize is not a function

      69 |     const checkImageOrientation = async ( ) => {
      70 |       if ( observationPhoto ) {
    > 71 |         const imageDimensions = await Image.getSize( observationPhoto );
         |                                       ^
      72 |         if ( imageDimensions.width < imageDimensions.height ) {
      73 |           setDisplayPortraitLayout( true );
      74 |         } else {
```

jtklein: Is this not just a test failure because the mock doesn't have it?

frewsxcv: Where is it being mocked in that test?

jtklein: Yes, you are right it doesn't seem to be. Then we should remove the mock from `ObsMedia.test.js` as well though.

PR #3207, opened 2025-11-11
Fix type error in `src/components/SharedComponents/Buttons/GradientButton.tsx`.

frewsxcv:


src/components/SharedComponents/Buttons/GradientButton.tsx
```diff
@@ -25,9 +25,9 @@ const GradientButton = ( {
   iconName,
   iconSize
 }: Props ) => {
-  const handleLongPress = _event => {
```
frewsxcv, line 28:
```
Parameter '_event' implicitly has an 'any' type.ts(7006)
```

sepeterson (approved):

PR #3209, opened 2025-11-11
Fix type errors in `src/components/SharedComponents/Modal.tsx` and potential Android back button bug.

frewsxcv:


src/components/SharedComponents/Modal.tsx
```diff
@@ -12,8 +12,8 @@ interface Props {
   fullScreen?: boolean;
   onModalHide?: () => void,
   style?: ViewStyle,
-  animationIn?: string;
-  animationOut?: string;
```
frewsxcv, line 16: Image

src/components/SharedComponents/Modal.tsx
```diff
@@ -53,7 +53,7 @@ const Modal = ( {
       onBackdropPress={closeModal}
       onSwipeComplete={closeModal}
       // Always close the modal when Android back button pressed
-      onRequestClose={closeModal}
```
frewsxcv, line 56: `onRequestClose` is not a known component prop of `RNModal`.

Image

Here is the relevant code for reference:

https://github.com/react-native-modal/react-native-modal/blob/v14.0.0-rc.0/src/modal.tsx

PR #3214, opened 2025-11-11
Remove unused `logStage` option on `useWatchPosition` function.

frewsxcv:


FLGMwt (approved):
LGTM, acknowledging that this is a functional change. This is for sure an unused `effect` dep

PR #3216, opened 2025-11-11
Mob 980 obs wheel tooltip not showing up

abbeycampbell:
[MOB-980](https://linear.app/inaturalist/issue/MOB-980/obs-wheel-tooltip-not-showing-up)

The problem here was that our evaluation of the currentRoute name never matched the 'ObsList' we're looking for--it stays stuck as 'TabNavigator', so we never trigger the correct conditions for the tooltip to show. My solution for this is to make a new piece of state that uses the built in getCurrentRoute method to get the currently focused navigation route properly for this use case.

I didn't want to modify our getCurrentRoute() util since we have other files that use it for analytics reasons.

Resetting the navigation stack (line 128) still works correctly, keeping the user on the Obs screen after opening multiple observation wheel methods.


jtklein:
Can this be closed?

PR #3217, opened 2025-11-12
MOB-936 introduce new Android CI and Maestro smoke test

FLGMwt:
This PR introduces a reworked `e2e_Android` GitHub Actions workflow that:
- Builds the Android app in release mode
- Starts an emulator and installs the app on it
- Runs a Maestro test that starts the app and confirms the FTUE landing page

It:
- For now can only be triggered manually (not on push or pull request)
- Does not run Detox, does not have coverage parity with our old `e2e_Android` or  active `e2e_iOS` workflows

We're wanting to get Android turned back on in CI now that we're ready to start working towards platform parity. Some factors such as RN upgrades + New Architecture, key package updates, and changes to the app have made this quite a bit more involved than the "update the tests, turn it back on" I was naively hoping for ; D

We are currently teetering over the edge of GitHub hosted runner resource constraints (disk space, RAM) that we were already dangerously close to. This “rework” alleviates that and gives us a bit of headroom. We don’t wanna pay for larger runners if we don’t have to, but while we’re manual-run-only, the emulator-based `test` step is currently using a 4 core ubuntu “Larger Runner”.

I have progress elsewhere on getting Detox tests back and running on Android but I've deferred that work to try to land on a solid, general "Android CI for iNatRN" which I hope this gets close to.


.github/workflows/e2e_android.new.yml
FLGMwt, line 1: This is essentially a rewrite of the previous workflow, but it's _juuuuuuust_ similar enough that I'm not able to convince git that it's a rename. Since the diff view is meaningless for review, I've added the rewrite as a new file, but I'll have a follow-up PR to remove the old one and rename this one.

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+    
+jobs:
+  build: 
+    runs-on: ubuntu-24.04
```
FLGMwt: This is the free GitHub hosted runner. The "stuff" about it is defined as part of the [runner-images](https://github.com/actions/runner-images) repo, specifically [here](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md).

It is the blueprint for the machine that everything in the `build` step runs on.

notably, it's different from the runner used on `test`, `ubuntu-24.04-m` (see later)

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+    - name: Generate release keystore
+      env:
+        ANDROID_ALIAS: ${{ secrets.ANDROID_ALIAS }}
+        ANDROID_KEY_STORE_PASSWORD: ${{ secrets.ANDROID_KEY_STORE_PASSWORD }}
+        ANDROID_KEY_PASSWORD: ${{ secrets.ANDROID_KEY_PASSWORD }}
+      run: |
+        keytool -genkeypair -v -noprompt -storetype PKCS12 -keystore release.keystore -alias "$ANDROID_ALIAS" -keyalg RSA -keysize 2048 -validity 10000 -storepass "$ANDROID_KEY_STORE_PASSWORD" -keypass "$ANDROID_KEY_PASSWORD" -dname "CN=mqttserver.ibm.com, OU=ID, O=IBM, L=Hursley, S=Hants, C=GB"
+
+    - name: Move keystore
+      run: mv release.keystore android/app/release.keystore
```
FLGMwt, line 65: secret / "juicy" file generation is unchanged from previous workflow

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+
+    - uses: actions/setup-java@v5
+      with:
+        distribution: 'zulu'
+        java-version: '17'
+
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@v5
+      with:
+        gradle-version: '8.14.1'
```
FLGMwt, line 80: these "setup" actions handle a number of environment and caching concerns for their respective tools. the gradle one in particular is hugely impactful on the build step.

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+        # git status
+
+    - name: Download the small example cv and geomodel
+      run: |
+        npm run add-example-model -- -f=main
+
+    - name: Android Build
+      run: |
+        cd android
+        ./gradlew build -PreactNativeArchitectures=x86 -x lint -x lintVitalRelease
```
FLGMwt, line 95: this mirrors the gradle invocation we use for release builds in our [fastlane android build](https://github.com/inaturalist/iNaturalistReactNative/blob/main/fastlane/Fastfile#L323)

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+    - name: Android Build
+      run: |
+        cd android
+        ./gradlew build -PreactNativeArchitectures=x86 -x lint -x lintVitalRelease 
+
+    - name: Upload APK
+      uses: actions/upload-artifact@v5
+      with:
+        name: release-apk
+        path: android/app/build/outputs/apk/release/*-release.apk
```
FLGMwt, line 104: We glob here to be agnostic to our base name which includes dynamic version numbers: https://github.com/inaturalist/iNaturalistReactNative/blob/main/android/app/build.gradle#L111

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+    
+  test:
+    runs-on: ubuntu-24.04-m
```
FLGMwt, line 110: We use the runner label `ubuntu-24.04-m` here. This is a "label" or "profile" that's been configured for this iNat repo to use which uses the same [ubuntu 24.04 runner image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) that the free test runner uses. It's configured for the ["4 core" ubuntu specification.](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/runners/larger-runners#specifications-for-general-larger-runners). Honestly, I think the tiny bit of extra disk space is the main benefit ; D but we do bump our emulator config to use the 4 cores. This could be made dynamic if we want to continue exploring 2/4/8 size runners.

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+    needs: build
+
+    steps:
+      - uses: actions/checkout@v2
```
FLGMwt, line 115: note! unlike our iOS [separate test step](https://github.com/inaturalist/iNaturalistReactNative/blob/main/.github/workflows/e2e_ios.yml#L138), we don't have to do an npm install for the packages dependent on detox, jest, etc.

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+
+      - name: Download APK
+        uses: actions/download-artifact@v6
+        with:
+          name: release-apk
+          path: e2e-build
```
FLGMwt, line 122: we provide a new directory here so we can expect that after this, there's a brand new directory that only has `e2e-build/myBuild.apk`, making us feel a little safer about globbing the `adb install`

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
```
FLGMwt, line 138: This and the emulator action below are honestly the game changes on being able to run our emu tests on GHA at all: https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+
+      - name: Install Maestro
+        run: |
+          export MAESTRO_VERSION="2.0.8"; curl -Ls "https://get.maestro.mobile.dev" | bash
```
FLGMwt: Would love to know how anyone prefers avoiding curlbashes in CI

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+          echo "$HOME/.maestro/bin" >> "$GITHUB_PATH"
```
FLGMwt: neat learning, this is the magical way to get `$PATH` updates to span workflow steps w/in a job https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-of-adding-a-system-path

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+          script: echo "Generated AVD snapshot for caching."
+
+      - name: Save AVD cache
+        if: steps.avd-cache-restore.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-29
```
FLGMwt, line 168: This, along with the actual _usage_ step that follows, is informed by the [avd cache guidance outlined](https://github.com/ReactiveCircus/android-emulator-runner/tree/v2#:~:text=We%20can%20significantly%20reduce%20emulator%20startup%20time%20by%20setting%20up%20AVD%20snapshot%20caching%3A) by the `android-emulator-runner` action that we use for wrapping our emulator use.

A key distinction here is that we use the relatively newly public _separate_ save/restore actions in `actions/cache` for only saving when we create a new avd cache after a miss and doing it so immediately. This is intentionally different than the typically-understood behavior of `actions/cache` saving as a [`post` step](https://github.com/actions/cache/blob/0057852bfaa89a56745cba8c7296529d2fc39830/action.yml#L42-L43) of a successful job. We don't want to save an avd cache of the workspace we've mucked with!

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,189 @@
...
+        with:
+          cores: 4
+          api-level: 29
+          target: google_apis
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
+          script: |
+            adb install -r e2e-build/*.apk
+            maestro test --no-ansi e2e/maestro/android/signedOut.yaml
```
FLGMwt, line 188: This is our first and only "real" use of [reactivecircus/android-emulator-runner](https://github.com/ReactiveCircus/android-emulator-runner/tree/v2) which orchestrates an emulator available to our `script` step that installs our apk to the emulator and runs our single maestro test against it.

This action (along with it's [apparently associated GH-paired kvm affordance ](https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/)) made a _huge_ difference in even making the emulator _possible_ to run on the 2/4 core ubuntu runners, let alone execute tests.

e2e/maestro/android/signedOut.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - stopRecording
+---
+- launchApp:
+    appId: org.inaturalist.iNaturalistMobile
+    clearState: true
+
+- extendedWaitUntil:
+    visible: "Close"
+    timeout: 30000
+- waitForAnimationToEnd
```
FLGMwt, line 18: for a first time launch, especially if this is the "first step" / FTUE dismiss before potential subsequent tests, I'm comfortable being generous with these timeouts for now. It's terribly annoying when a 30m build fails due to the first actions on an emulator being slow.

.github/workflows/e2e_android.new.yml
FLGMwt: This is heavily annotated so if you want to just read, heads up about "Minimize comments"

Image

.github/workflows/e2e_android.new.yml
FLGMwt: oh, except for the fact that we add `-PreactNativeArchitectures=x86` to drastically minimize our build time by only targeting one arch, namely, the arch of the emu we're installing it on. https://reactnative.dev/docs/build-speed#build-only-one-abi-during-development-android-only

e2e/maestro/android/signedOut.yaml
```diff
@@ -0,0 +1,22 @@
+appId: org.inaturalist.iNaturalistMobile
+# FLGMwt: I would prefer to be able to configure setup/teardown
+# in workspace config, but this is not supported
+# https://github.com/mobile-dev-inc/Maestro/issues/2063
+onFlowStart:
+    - runFlow: onFlowStart.yaml
+onFlowComplete:
+    # noop if not recording
+    - stopRecording
```
FLGMwt, line 9: This is an important callout: the "standard" method for recording a full flow run would be to invoke `maestro record --local myFlow.yaml` instead of `maestro test myFlow.yaml`. An alternative is to, inside of a flow definition, wrap the specific commands you want to record in `- startRecording`/`- stopRecording` directives.

`record` produces a "nicer" video that shows the flow steps & progress alongside the emulator recording. `- startRecording`/`- endRecording` only captures the emulator during the wrapped steps.

I'm not using `record` for a few reasons:
- **Mainly** because `record` is _painfully_ slow. The `record` video is "nicer" but is also a much larger video to render. I don't think the runtime cost is worth the value of the side-by-side steps being included.
- `record` doesn't respect its documented `--no-ansi` whereas `test` does: https://github.com/mobile-dev-inc/Maestro/issues/1009. not a big deal, but annoying.
- `record` only supports invoking a single flow. `test` supports multiple flow paths / flow directories. also not a big deal, but it's nice to avoid a necessary orchestration when we can.

In the [RN sample experimentation repo](https://github.com/FLGMwt/react-native-maestro-gha) I did most of this work on:

| Command | Example Run | `time` result | Recording Size | Screenshot of video "style" midrun |
|---------|-------------|---------------|-----------|-----------------------------|
|`maestro record --local maestro/Launch.android.yaml Launch.android.mp4`| [see "run tests" step](https://github.com/FLGMwt/react-native-maestro-gha/actions/runs/19302453833/job/55200786462)|`4:37.92`|`25.5MB`|Image|
|`MAESTRO_RECORD=true maestro test maestro/Launch.android.yaml`|[see "run tests" step](https://github.com/FLGMwt/react-native-maestro-gha/actions/runs/19310391952/job/55229120108)|`0:54.24`|`474KB`|Image|

_Note: the `MAESTRO_RECORD` env var is not understood by maestro, it's only used to flag the flow behavior_

FLGMwt: Other Maestro limitations I've been stewing over having spent time with it:

- these hooks would be _great_ to include in the [supported `.maestro/config.yml`](https://docs.maestro.dev/api-reference/configuration/workspace-configuration#maestro-configuration). https://github.com/mobile-dev-inc/Maestro/issues/2063
- Maestro doesn't support arbitrary shell execution in flows _or_ any configuration hooks for them. At the very least, we'd want to run our [test user reset script](https://github.com/inaturalist/iNaturalistReactNative/blob/main/e2e/sharedFlows/resetUserForTesting.js) between flows. Maestro supports running js in a [_very_ limited engine](https://docs.maestro.dev/advanced/javascript/run-javascript#others), but that'd be a pain to rely on. I'd much rather write write an orchestration script wrapping individual `maestro test myFlow` calls to handle this (overriding my `record` grievance above ; D). https://github.com/mobile-dev-inc/Maestro/issues/1682

Running Maestro in Maestro Cloud is appealing and very fast, but we'd need to evaluate the costs and tradeoffs introducing a new service.

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,188 @@
...
+        cache: 'npm'
+
+    - uses: actions/setup-java@v5
+      with:
+        distribution: 'zulu'
+        java-version: '17'
+
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@v5
+      with:
```
jtklein, line 79: Can the version we need be read from the repository? What will happen if we update RN and hence gradle version. Will this step error out, or the build later?

.github/workflows/e2e_android.new.yml
jtklein: Thanks that is helpful context, like many of the comments here. I think these comments should also be added to the source code directly.

.github/workflows/e2e_android.new.yml
jtklein: Do we need to keep the installs on the iOS side?

.github/workflows/e2e_android.new.yml
jtklein: I don't think we have to avoid them. Or is this not just purely style?

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,187 @@
...
+      - name: Attempt to restore AVD cache
+        uses: actions/cache/restore@v4
+        id: avd-cache-restore
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-29
+
+      - name: Create AVD and generate snapshot for caching
```
jtklein, line 150: Do you know what is the actual key to decide when we have a cached emulator and when to make a new one? If the default is somehow branch related, I think we could also do something like the combined hash of the two build.gradle files. Kind of as a first approximation that we don't need to use a different SDK and tests should be stable without any required emulator updates.

jtklein: Even the firebase inclusion which didn't require an SDK update but forced us to switch from AOSP to google-apis required an update in build.gradle files.

.github/workflows/e2e_android.new.yml
```diff
@@ -0,0 +1,187 @@
...
+          script: echo "Generated AVD snapshot for caching."
+
+      - name: Save AVD cache
+        if: steps.avd-cache-restore.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-29
```
jtklein, line 168: So, with this key can we later reuse the same avd cache for two different test runs, maestro and detox?

e2e/maestro/android/signedOut.yaml
jtklein: ok

e2e/maestro/android/signedOut.yaml
jtklein: I like this explanation. It should be available to other developers. I am not sure where, but first place I'd put some of this (not exhaustively, probably only "How to use `record` instead of `test`:" is the README.
I also don't think we should split this up into two versions, where the local variant uses `record` so that while debugging the flows we have the "better" video. It would add too much confusion probably (and provide different test circumstances between local and CI).

.github/workflows/e2e_android.new.yml
FLGMwt: For the `test` job? Yes we do, the main difference being Maestro vs Detox. If we _do_ eventually end up fixing Android Detox and swapping that in here, we _will_ need the install step here.

The only dependency for Maestro to run is the executable (installed in the `Install Maestro` step). Detox relies on some (but not all) of the project tooling / deps defined in our `package.json` such as the `detox` npm package and `jest`. For the iOS (and eventual Android) test steps, we _could_ replace `npm install` with just those minimal packages we need, but I don't know if that's worth the cognitive burden / maintenance concern of an exceptional way of doing things.

PR #3221, opened 2025-11-12
Migrate `tryToReplaceWithLocalTaxon` to TypeScript.

frewsxcv:


src/components/Match/helpers/tryToReplaceWithLocalTaxon.ts
```diff
@@ -1,4 +1,10 @@
-const tryToReplaceWithLocalTaxon = ( localTaxa, suggestion ) => {
+import Taxon from "realmModels/Taxon";
+import type { RealmTaxon } from "realmModels/types";
+
+const tryToReplaceWithLocalTaxon = (
+  localTaxa: ( Taxon & RealmTaxon )[],
+  suggestion: { taxon: { id: number } }
```
frewsxcv, line 6: Is there an existing type representing a "suggestion" somewhere?

sepeterson: Doesn't look like it. This is good for now!

PR #3226, opened 2025-11-14
Migrate `ObsDetailsDefaultMode` to TypeScript.

frewsxcv:


src/components/ObsDetailsDefaultMode/ObsDetailsDefaultMode.js
```diff
@@ -64,8 +64,8 @@ const ObsDetailsDefaultMode = ( {
   targetActivityItemID,
   wasSynced,
   uuid
-}: Props ): Node => {
-  const scrollViewRef = useRef( );
```
frewsxcv, line 68: The first (and only) parameter to `useRef` is required

sepeterson (approved):

PR #3229, opened 2025-11-14
Migrate `src/sharedHelpers/parseExif` to TypeScript.

frewsxcv:


src/sharedHelpers/dateAndTime.ts
```diff
@@ -215,7 +215,7 @@ function dateFnsLocale( i18nextLanguage: string ) {
   }
 }
 
-function formatISONoTimezone( date: Date ) {
+function formatISONoTimezone( date: Date | null | undefined ) {
```
frewsxcv, line 218: In `parseExif`, a `null | undefined` value is being passed here, which this function gracefully handles on the line below

src/sharedHelpers/parseExif.ts
```diff
@@ -54,31 +55,40 @@ export const writeExifToFile = async ( photoUri: ?string, exif: ExifToWrite ): P
...
-
-  allExifPhotos.filter( x => x ).forEach(
+  const allExifPhotos: ( ExifData | null )[] = responses
+    .map( r => ( r.status === "fulfilled"
+      ? r.value
+      : null ) );
+
+  allExifPhotos.forEach(
     currentPhotoExif => {
+      if ( !currentPhotoExif ) return;
```
frewsxcv, line 94: This is identical to the old line above:

```
.filter( x => x )
```

src/sharedHelpers/parseExif.ts
```diff
@@ -54,31 +55,40 @@ export const writeExifToFile = async ( photoUri: ?string, exif: ExifToWrite ): P
...
-  // Flow will complain that value is undefined, but the filter call ensures
-  // that it isn't
-  // $FlowIgnore
-  }> = responses.filter( r => r.value ).map( r => r.value );
-
-  allExifPhotos.filter( x => x ).forEach(
+  const allExifPhotos: ( ExifData | null )[] = responses
+    .map( r => ( r.status === "fulfilled"
+      ? r.value
+      : null ) );
```
frewsxcv, line 90: I had to add a `r.status === "fulfilled"` check in here because otherwise `r.status` could be `"rejected"`, in which case there is now `r.value`:

Image

src/sharedHelpers/parseExif.ts
```diff
@@ -1,18 +1,30 @@
...
 import { formatISONoTimezone } from "sharedHelpers/dateAndTime";
 
 class UsePhotoExifDateFormatError extends Error {}
 
+interface ExifData {
+  latitude?: number;
+  longitude?: number;
+  positional_accuracy?: number;
+  date?: string;
+}
```
frewsxcv, line 12: For reference:

https://github.com/inaturalist/react-native-exif-reader/blob/17780909f4b1899ce21524119c23412ac28c51d8/README.md?plain=1#L17

frewsxcv:
Migrating to draft until I have time to address conflicts

PR #3230, opened 2025-11-14
MOB-991 MOB-993 MOB-989 add firebase performance monitoring and a couple events

sepeterson:
[MOB-993](https://linear.app/inaturalist/issue/MOB-993/firebase-dashboard-for-fresh-installs) [MOB-991](https://linear.app/inaturalist/issue/MOB-991/firebase-record-whether-or-not-the-user-has-given-location-permissions) [MOB-989](https://linear.app/inaturalist/issue/MOB-989/firebase-record-the-amount-of-time-it-takes-to-load-the-match-screen)

- add firebase performance monitoring library
- event for AI camera shutter tap
- events for onboarding buttons
- global performance trace solution in zustand
- add trace for AI camera -> match screen flow

To avoid memory leaks, I added a timeout to cleanup unclosed traces. One side effect is that if a trace fails to be stopped, it gets sent as taking the full timeout duration so we want to be sure traces are always correctly stopped

### DIFF
--- a/src/stores/createFirebaseTraceSlice.ts
+++ b/src/stores/createFirebaseTraceSlice.ts
@@ -1,5 +1,6 @@
-import { FirebasePerformanceTypes, getPerformance } from "@react-native-firebase/perf";
-import { StateCreator } from "zustand";
+import type { FirebasePerformanceTypes } from "@react-native-firebase/perf";
+import { getPerformance } from "@react-native-firebase/perf";
+import type { StateCreator } from "zustand";
 
 import { log } from "../../react-native-logs.config";
 


src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export enum FIREBASE_TRACE_ATTRIBUTES {
+  ONLINE = "online",
+}
+
+interface TraceData {
+  trace: FirebasePerformanceTypes.Trace;
+  // eslint-disable-next-line no-undef
+  timeoutId: NodeJS.Timeout;
```
frewsxcv: You may be able to do:

```suggestion
  timeoutId: ReturnType<setTimeout>;
```

To avoid the `eslint-disable-next-line no-undef` warning

frewsxcv: Ah right `ReturnType<typeof setTimeout>`

android/app/build.gradle
```diff
@@ -2,6 +2,7 @@ apply plugin: "com.android.application"
 apply plugin: "org.jetbrains.kotlin.android"
 apply plugin: "com.facebook.react"
 apply plugin: "com.google.gms.google-services"
+apply plugin: 'com.google.firebase.firebase-perf'
```
jtklein: We should match the style of the above.

android/build.gradle
```diff
@@ -22,6 +22,7 @@ buildscript {
         classpath("com.facebook.react:react-native-gradle-plugin")
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
         classpath("com.google.gms:google-services:4.4.3")
+        classpath 'com.google.firebase:perf-plugin:2.0.1'
```
jtklein: We should match the style of the above.

src/sharedHelpers/tracking.ts
```diff
@@ -0,0 +1,18 @@
...
+// eslint-disable-next-line import/prefer-default-export
+export const logFirebaseEvent = (
+  eventId: string,
+  parameters: FirebaseParameters
+) => {
+  // eslint-disable-next-line no-undef
+  if ( __DEV__ ) {
+    logger.info( `Firebase event: ${eventId} ${JSON.stringify( parameters )}` );
+  }
+  void analytics().logEvent( eventId, parameters );
```
jtklein: This call is giving me a deprecation warning in devtools:
```
This method is deprecated (as well as all React Native Firebase namespaced API) and will be removed in the next major release as part of move to match Firebase Web modular SDK API. Please see migration guide for more details: https://rnfirebase.io/migrating-to-v22 Please use `getApp()` instead. Error Stack:
    at RCTView (<anonymous>)
anonymous @ console.js:654
overrideMethod @ backend.js:17416
anonymous @ setUpDeveloperTools.js:40
registerWarning @ LogBox.js:171
anonymous @ LogBox.js:84
warnIfNotModularCall @ index.js:770
getApp @ app.js:89
firebaseModuleWithApp @ namespace.js:145
logFirebaseEvent @ tracking.ts:17
t37 @ OnboardingCarousel.js:322
handlePress @ Button.tsx:215
t0 @ PressableWithTracking.tsx:27
_performTransitionSideEffects @ Pressability.js:757
_receiveSignal @ Pressability.js:696
onResponderRelease @ Pressability.js:519
executeDispatch @ ReactFabric-dev.js:717
runWithFiberInDEV @ ReactFabric-dev.js:683
executeDispatchesAndReleaseTopLevel @ ReactFabric-dev.js:1492
forEachAccumulated @ ReactFabric-dev.js:958
anonymous @ ReactFabric-dev.js:1536
batchedUpdatesImpl @ ReactFabric-dev.js:16604
batchedUpdates$1 @ ReactFabric-dev.js:1461
dispatchEvent @ ReactFabric-dev.js:1511
Show 19 more frames
Show less
```

I was working through that deprecation notice when I first introduced firebase into the app and landed on this solution here: https://github.com/inaturalist/iNaturalistReactNative/blob/38686bfdc2e34fb94550664af764575536fc2d43/src/components/SharedComponents/Buttons/PressableWithTracking.tsx#L19-L24

Can you reuse that snippet?

src/sharedHelpers/tracking.ts
```diff
@@ -0,0 +1,18 @@
+import analytics from "@react-native-firebase/analytics";
+import { log } from "sharedHelpers/logger";
+
+const logger = log.extend( "tracing.ts" );
```
jtklein: `tracking.ts` here or rename file?

src/sharedHelpers/tracking.ts
jtklein: Could you also please refactor the logEvent calls in PressableWithTracking and OfflineNavigationGuard to import this function here and use it. I did opt to track all button taps with the same eventID and also all screen transitions with the same eventID at the time but if you think that using the testID + screen combo as unique eventID is better in terms of analyzing the data feel free to change it to that.

src/sharedHelpers/tracking.ts
```diff
@@ -0,0 +1,18 @@
...
+
+type FirebaseParameters = Record<string, string | number | string[]>
+
+// eslint-disable-next-line import/prefer-default-export
+export const logFirebaseEvent = (
+  eventId: string,
+  parameters: FirebaseParameters
+) => {
+  // eslint-disable-next-line no-undef
+  if ( __DEV__ ) {
```
jtklein: If this is only debug related maybe we don't need to keep it in.

src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export interface FirebaseTraceSlice {
+  activeTraces: Record<string, TraceData>;
+  startTrace: ( traceId: string ) => Promise<void>;
+  stopTrace: ( traceId: string ) => Promise<void>;
+}
+
+const createFirebaseTraceSlice: StateCreator<FirebaseTraceSlice>
+= ( set, get ) => ( {
+  activeTraces: {},
```
jtklein: Zustand mushes all separate slices into one big store where names of keys can collide. So better to be more descriptive than less in our state keys. `activeFirebaseTraces`, `startFirebaseTrace` and `stopFirebaseTrace` come to my mind.

src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  startTrace: async ( traceId: string, attributes: Record<string, string> = {} ) => {
+    const trace = await perf().startTrace( traceId );
```
jtklein: This call also gives me a deprecation warning in devtools. I have not worked with this but do we need to re-use the same syntax as linked to in the inline comment in tracking.ts?

src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const timeoutId = setTimeout( () => {
+      get().stopTrace( traceId );
+    }, TRACE_TIMEOUT );
+
+    try {
+      Object.entries( attributes ).forEach( ( [key, value] ) => {
+        trace.putAttribute( key, value );
```
jtklein: There is also `putMetric` for numbers. Do we want to already implement that `attributes`' values are parsed into strings and numbers and the added either via putAttribute or putMetric? E.g. we do already log the average time it takes to run computer vision per frame in the ai camera. That is a metric I would like to see in the AI camera trace at one point (not now in this PR though)

src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const createFirebaseTraceSlice: StateCreator<FirebaseTraceSlice>
+= ( set, get ) => ( {
+  activeTraces: {},
+
+  startTrace: async ( traceId: string, attributes: Record<string, string> = {} ) => {
+    const trace = await perf().startTrace( traceId );
+
+    const timeoutId = setTimeout( () => {
+      get().stopTrace( traceId );
```
jtklein: Should we add an attribute that this trace was stopped via timeout.

src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+    }, TRACE_TIMEOUT );
+
+    try {
+      Object.entries( attributes ).forEach( ( [key, value] ) => {
+        trace.putAttribute( key, value );
+      } );
+    } catch ( _ ) {
+      /* this can error for a few reasons like value being a non-string
```
jtklein: Don't know if this still applies after refactoring to fix the deprecation notice, but if you catch an error here we should at least log it to console.

src/components/Camera/AICamera/AICamera.js
```diff
@@ -16,13 +16,15 @@ import { VolumeManager } from "react-native-volume-manager";
...
 import { deleteSentinelFile, logStage } from "sharedHelpers/sentinelFiles";
+import { logFirebaseEvent } from "sharedHelpers/tracking";
 import {
   useDebugMode,
   useLayoutPrefs,
   usePerformance,
   useTranslation
 } from "sharedHooks";
 import { isDebugMode } from "sharedHooks/useDebugMode";
+import { FIREBASE_TRACES } from "stores/createFirebaseTraceSlice";
```
jtklein: Love the use of enums here.

src/stores/createFirebaseTraceSlice.ts
```diff
@@ -0,0 +1,79 @@
...
+  stopTrace: async ( traceId: string, attributes: Record<string, string> = {} ) => {
+    const { activeTraces } = get();
+    const traceData = activeTraces[traceId];
+    if ( traceData ) {
+      clearTimeout( traceData.timeoutId );
+
+      const { trace } = traceData;
+      try {
+        Object.entries( attributes ).forEach( ( [key, value] ) => {
+          trace.putAttribute( key, value );
```
jtklein: The doc comment for this function reads:
```
    /**
     * Sets a String value for the specified attribute. Updates the value of the attribute if it already exists.
     * The maximum number of attributes that can be added is 5.
     *
     * #### Example
     *
     * ```js
     * trace.putAttribute('userId', '123456789');
     * ```
     *
     * @param attribute Name of the attribute. Max length is 40 chars.
     * @param value Value of the attribute. Max length is 100 chars.
     */
```
in the [docs](https://rnfirebase.io/analytics/usage#event-parameters) for `logEvent` I have seen that validation of these strings happens server side and the request to log an event does silently fail if we fail these criteria like max 100 chars. So, I think we should validate these attributes before sending them to the server here with `stopTrace`.

src/sharedHelpers/tracking.ts
jtklein: We should catch any errors from the `logEvent` function call. For now we can log all errors to Kibana using `logger.error`.

src/components/Onboarding/OnboardingCarousel.js
```diff
@@ -314,10 +318,19 @@ const OnboardingCarousel = ( ) => {
                 level="primary"
                 forceDark
                 text={t( "CONTINUE" )}
-                onPress={() => (
-                  carouselRef.current?.getCurrentIndex() >= ONBOARDING_SLIDES.length - 1
-                    ? closeModal()
-                    : carouselRef.current?.scrollTo( { count: 1, animated: true } ) )}
+                onPress={() => {
+                  logFirebaseEvent(
```
jtklein, line 322: This is a `Button` which already has a `Pressable` wrapper imported from styledComponents, which is an instance of `PressableWithTracking`. So, we actually are sending already a logEvent when this on Press function is called without adding it here. So, adding it will send two logEvents for this particular button being pressed. I wonder what we can do to avoid sending two logs, giving the additional attribute as prop into the `PressableWithTracking` which then forwards it to one `logEvent` call.
Or from your research into Firebase, do you think it makes more sense to send two different logs, one which indicates "this button was pressed" and one for "this was the current slide while button was pressed"?
One problem with using only the logEvent inside of PressableWithTracking would be that it uses the testID as eventID and would then have a quite different name than `onboarding_close_pressed`.

src/components/Camera/AICamera/AICamera.js
```diff
@@ -172,6 +176,8 @@ const AICamera = ( {
   const handleTakePhoto = useCallback( async ( ) => {
     await logStage( sentinelFileName, "take_photo_start" );
     setHasTakenPhoto( true );
+    logFirebaseEvent( "ai_camera_shutter_tap", { hasLocationPermissions } );
+    startTrace( FIREBASE_TRACES.AI_CAMERA_TO_MATCH );
```
jtklein: Looking at the issue description of MOB-989, I think this is fine for now.
Ideally we should expand the trace with some milestones, is that possible?
Every time we call `await logStage` is basically the end of a block of work that needs to get done on the way to Match screen, and it would be great if in Firebase we can see the performance trace broken down by those milestones.

jtklein (changes requested):
Left quite a few change requests in inline comments.

My main question is: in cases where the actual button press already extends functionality of PressableWithTracking, do we need to send two logEvents or refactor to prop any attributes into the respective PressableWithTracking that then sends one log only.

src/sharedHelpers/tracking.ts
sepeterson: I set up a screen tracking function following the [example in the docs](https://rnfirebase.io/analytics/screen-tracking#react-navigation)

For the button tracking, I would actually propose that we remove the logEvent in PressableWithTracking. I think it would be better to decide user events we want to track and add logEvents to track them rather than capturing a bunch of button presses we may not use. It doesn't state this explicitly in the docs but they seem to take this approach too. I am open to discussion here but it would solve the problem of doubling up as well

src/stores/createFirebaseTraceSlice.ts
sepeterson: Good to know. Thank you!

src/components/Camera/AICamera/AICamera.js
sepeterson: 😎

src/stores/createFirebaseTraceSlice.ts
sepeterson: Yes! Excellent idea -- added

src/components/Camera/AICamera/AICamera.js
sepeterson: As far as I can see, it doesn't support milestones out of the box but I see a couple ways we could implement this: either using metrics or multiple simultaneous traces

src/stores/createFirebaseTraceSlice.ts
sepeterson: Added extra logic to support putMetric

src/sharedHelpers/tracking.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const logFirebaseEvent = (
+  eventId: string,
+  parameters?: FirebaseParameters
+) => {
+  try {
+    const analytics = getAnalytics();
+    logEvent( analytics, eventId, parameters );
+  } catch ( error ) {
+    logger.error( "Error logging firebase event", error );
```
jtklein: You would need to wrap `error` here in `JSON.stringify( error )` or it will appear as `[object Object]` in the server logs and not as the actual content of the error object.

src/sharedHelpers/tracking.ts
jtklein: For me it is fine to remove the `logEvent` part from `PressableWithTracking`. I added it a few weeks ago to just any kind of data coming into Firebase. But your approach of just adding what we need is also fine. I will resolve this thread because the other things here are taken care of already.

src/components/Camera/AICamera/AICamera.js
jtklein: OK, definitely not a requirement for this PR.

src/components/Camera/CameraContainer.tsx
```diff
@@ -199,6 +201,11 @@ const CameraContainer = ( ) => {
     visionResult: StoredResult | null
   ) => {
     if ( !showPhotoPermissionsGate ) {
+      if ( cameraType === "AI" ) {
+        // We won't start the trace if the user hasn't granted photo permissions
+        // because the time it takes to request permissions would be included in the trace
+        startFirebaseTrace( FIREBASE_TRACES.AI_CAMERA_TO_MATCH );
```
jtklein: To complicate things a bit: we do allow users to use the app without us ever giving photo permission. So, it is possible for a user to always hit the `else` path here. So, to be consistent with your idea of removing the time spent on the permission gate from the trace, we'd need to move the start of the trace into the `handleNavigation` function call I think. This would take care of tracing the photo-permission-denied cases, and also the case of the very first instance of a Ai camera to Match screen navigation right after allowing the photo permission which is handled in the `onPermissionGranted` callback of `renderSavePhotoPermissionGate`. 

However, we would compare apples to oranges a bit with this approach because without photo permission we actually do not have a "save to gallery" file operation and so these traces would always be faster. I had a quick chat with Abhas about that, and product teams preferred way of seeing this would be a boolean flag in the trace wether the user has Save Photo Permission or not.

Can you move the trace start, and add the boolean attribute to your PR?

sepeterson: Makes sense. Thanks for following up on this with Abhas! I moved the start as requested and added the permissions attribute. I also changed the screen view tracking to use logEvent as the other func was throwing a deprecation warning.

jtklein (approved):
Thanks for following up on the latest requested changes. I think this is in a good spot now. The error in CI is related to the new eslint-role introduced a few days ago. It should be auto-fixable in the `createFirebaseTraceSlice` file.
If you are done with that small fix, please fell free to merge in your own time.

PR #3232, opened 2025-11-15
Migrate `INatPaperProvider` to TypeScript.

frewsxcv:


src/providers/INatPaperProvider.tsx
```diff
@@ -21,17 +18,14 @@ const theme = {
...
-// eslint-disable-next-line react/jsx-props-no-spreading
-const renderCustomIcon = props => <INatIcon {...props} />;
+} as const;
 
-type Props = {
-  // $FlowIgnore
-  children: unknown
-}
+const renderCustomIcon = (
+  // eslint-disable-next-line react/jsx-props-no-spreading
```
FLGMwt, line 24: in approval, calling this out only because the delta is a little disconnected: this eslint-disable is consistent with the code as written

FLGMwt (approved):
LGTM, thanks yet again!

PR #3234, opened 2025-11-15
Start running `tsc` in CI by comparing against a baseline.

frewsxcv:
Prior to this pull request, TypeScript was not run in CI, presumably because of the _many_ errors that currently exist when running it (`npm run lint:tsc`).

After this pull request, TypeScript will be run in CI, through a baseline system. The main change is the addition of a script that tracks existing TypeScript errors and fails the linting process if new errors are introduced. This helps ensure that the codebase does not accumulate additional TypeScript issues over time. This will also make it easier to see progress as we chip away at the TypeScript errors over time.


FLGMwt:
This suggestion and generous implementation is appreciated, but I do not think that we want this in the pipeline _at the moment_. The gradual intention of TypeScript migration is in consideration with each change, but checking / coverage is not necessarily a priority right now in or current work. We're still working towards progressive migration but it's not an immediate mandate so we may intentionally regress at points.

That said, I don't think we want to include this change at the moment, but this may desired in future.

PR #3242, opened 2025-11-19
Migrate `SaveDiscardButtons` to TypeScript.

frewsxcv:


src/components/Match/SaveDiscardButtons.tsx
```diff
@@ -13,12 +12,12 @@ const DROP_SHADOW = getShadow( {
 } );
 
 type Props = {
```
jtklein: I'd prefer for Props to be an interface in all components.

src/components/Match/SaveDiscardButtons.tsx
```diff
@@ -13,12 +12,12 @@ const DROP_SHADOW = getShadow( {
...
-  handlePress: Function
+  handlePress: ( action: "save" | "discard" ) => void;
```
jtklein: I'd prefer for `save` and `discard` to be two values of an enum to avoid having to keep the same string in several places in this file. I'd prefer an enum here instead of two consts because the two states are mutually exclusive.
At a later point I'd want this enum to be exported and imported wherever we create the `handlePress` function that is passed in here to avoid breaking the contract of this component. For this PR it is okay to export the enum and not import it anywhere.

jtklein (changes requested):
Thanks for this! Only minor changes as outlined in the inline comments.

src/components/Match/SaveDiscardButtons.tsx
jtklein: Like for example here: https://github.com/frewsxcv/iNaturalistReactNative/blob/7ecb572a4a3c61efbc9000147db733754f923c96/src/components/Projects/ProjectsContainer.tsx#L13-L14

frewsxcv: >  I'd prefer an enum here instead of two consts because the two states are mutually exclusive.

I'm having trouble following what you're saying. A union of types is also mutually exclusive.

Doing `export type Action = "save" | "discard"` should alleviate your concerns here about being able to export it, and doesn't have the overhead of an `enum`.

src/components/Match/SaveDiscardButtons.tsx
frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3242/commits/ede958307ebffe380a109f32a2266859d4a77164

frewsxcv: Could this be enforced by the style checker if it's a style requirement?

jtklein: Yes, we are working towards that. Ideally we'd like to end up in a state where pressing "Save" in the editor takes care of most of these stylistic issues.

src/components/Match/SaveDiscardButtons.tsx
jtklein: I am more concerned with making the contract between two developers (one working on this component, the other one creating an instance of this) less brittle by using a single source of truth for these two strings. So, instead of "save" I want to see a reference to a definition in all places it is used in this file here, but also the same reference being imported in places we create an instance of this component.
Since there are multiple mutually exclusive options of this action an enum makes more sense to me than two constants.

frewsxcv: I understand the concern about brittleness. The good news is that a string union (`export type Action = ...`) provides the exact same 'source of truth' safety as an `enum`. If we change the string in the type definition later, TypeScript will flag every usage of the old string as an error across the app.

The main reason I prefer unions here is that TypeScript `enum`s are 'nominal', meaning a consumer must import the `enum` just to use the component. Unions are 'structural,' so consumers can just pass the string 'save' and TypeScript validates it against the exported type automatically. It's much lower friction for the same level of safety.

PR #3246, opened 2025-11-20
Migrate `Settings` component to TypeScript.

frewsxcv:


FLGMwt (approved):
in the most generous way possible: that'll do!

thanks again ; D

PR #3248, opened 2025-11-20
lint config for consistent import style

FLGMwt:
MOB-1008

Adds [`typescript-eslint/consistent-type-imports`](https://typescript-eslint.io/rules/consistent-type-imports/) rule to enforce the following style for type imports w/ separate import lines for types and non-types:

```js
import type { ErrorWithResponse, INatApiError } from "api/error";
import handleError from "api/error";
```

This config change was limited to `.eslintrc.js` in the first commit.

All other file edits are the result of applying the rule's autofix on the codebase. There were two things that needed to be fixed manually:
- `ForgotPasswordForm.tsx` mistakenly had a `flow` tag and flow did not like the result of the fix (subtly different syntax than flow supports). This was fixed by removing the flow tag.
- I added `eslint-disable-next-line @typescript-eslint/no-unsafe-function-type` to a few files caught up in the changes that the failed the commit check. I opted to ignore them now to limit non-automatic changes to this PR

I manually checked each file for issues and everything seems good! And noodling around writing imports, it doesn't seem to conflict with any other autofix like the import grouping

### DIFF
--- a/src/components/SharedComponents/UploadStatus/FadeOutIcon.tsx
+++ b/src/components/SharedComponents/UploadStatus/FadeOutIcon.tsx
@@ -1,12 +1,11 @@
 import { View } from "components/styledComponents";
-import type { ReactNode } from "react";
+import type { PropsWithChildren } from "react";
 import React from "react";
 import Reanimated, {
   Keyframe
 } from "react-native-reanimated";
 
-type Props = {
-  children: ReactNode;
+interface Props extends PropsWithChildren {
   uniqueKey: string;
 }
 const AnimatedView = Reanimated.createAnimatedComponent( View );


frewsxcv (approved):
Nice!

jtklein (approved):
Yes! Thank you for this. This is really helpful to tackle those inconsistencies in the codebase with eslint rules.
Only comment I have is it would also be good to actually type the functions that now have the `eslint-disable-next-line @typescript-eslint/no-unsafe-function-type` comment. But in another PR: https://linear.app/inaturalist/issue/MOB-1042/type-functions-that-have-an-eslint-warning-of

PR #3249, opened 2025-11-21
MOB-512 saved match screen experience

sepeterson:
[MOB-512](https://linear.app/inaturalist/issue/MOB-512/implement-saved-match-screen)

When a user has observations that are saved locally but not uploaded, we show them an observation detail screen with similar UI to the match screen.

- add SavedMatch screen component and corresponding container
- add ObsDetailsDefaultModeScreensWrapper to select what screen to show
- some ts conversions

### DIFF
--- a/src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatchHeaderRight.tsx
+++ b/src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatchHeaderRight.tsx
@@ -25,7 +25,6 @@ const SavedMatchHeaderRight = ( {
     ( ) => (
       <INatIconButton
         testID="SavedMatch.editButton"
-        // TODO remove cast when useLocalObservation converted to typescript
         onPress={() => navigateToObsEdit( observation )}
         icon="pencil"
         color={String( colors?.darkGray )}
--- a/src/components/Match/PhotosSection.js
+++ b/src/components/Match/PhotosSection.js
@@ -1,3 +1,5 @@
+// @flow
+
 import classnames from "classnames";
 import MediaViewerModal from "components/MediaViewer/MediaViewerModal";
 import {
@@ -7,6 +9,7 @@ import {
   Image, Pressable, View
 } from "components/styledComponents";
 import _, { compact } from "lodash";
+import type { Node } from "react";
 import React, { useEffect, useState } from "react";
 import Photo from "realmModels/Photo";
 import getImageDimensions from "sharedHelpers/getImageDimensions";
@@ -25,7 +28,7 @@ const PhotosSection = ( {
   obsPhotos,
   navToTaxonDetails,
   hideTaxonPhotos
-}: Props ) => {
+}: Props ): Node => {
   const [displayPortraitLayout, setDisplayPortraitLayout] = useState( null );
   const [mediaViewerVisible, setMediaViewerVisible] = useState( false );
 
--- a/tests/unit/components/DisplayTaxonName.test.js
+++ b/tests/unit/components/DisplayTaxonName.test.js
@@ -275,6 +275,14 @@ describe( "DisplayTaxonName", ( ) => {
     } );
   } );
 
+  describe( "when taxon is undefined", ( ) => {
+    it( "it displays fallback text", ( ) => {
+      let taxon;
+      render( <DisplayTaxonName taxon={taxon} /> );
+      expect( screen.getByText( /Unknown/ ) ).toBeTruthy( );
+    } );
+  } );
+
   describe( "when displayed as plain text within a Trans component", ( ) => {
     it( "it displays common name followed by scientific name", async ( ) => {
       render( <DisplayTaxonName taxon={subspeciesTaxon} removeStyling layout="horizontal" /> );


src/sharedHooks/useRemoteObservation.ts
```diff
@@ -10,7 +9,35 @@ const { useRealm } = RealmContext;
...
+  hidden?: boolean;
+  [key: string]: unknown;
+}
+
+interface IdentificationWithHidden {
+  hidden?: boolean;
+  [key: string]: unknown;
+}
+
+interface RemoteObservation {
```
jtklein: Liking this gradual migration to TS!
We already have started to created interfaces for all data that is coming from the API. For this function you should re-use `ApiObservation` from [here](https://github.com/inaturalist/iNaturalistReactNative/blob/7acdbb768ebd0183b6bc5483b3da26963adbe920/src/api/types.d.ts#L131-L132). We started creating them with the keys we needed on the fly, so please add the missing keys for comments and IDs there, using `ApiComment` and `ApiIdentification`. And also add `hidden` to those two interfaces.

.eslintrc.js
```diff
@@ -164,6 +164,12 @@ module.exports = {
         "@typescript-eslint/no-require-imports": "off"
       }
     },
+    {
```
jtklein, line 171: Could have been a separate PR (as a prerequisite of this work).

src/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper.tsx
```diff
@@ -0,0 +1,100 @@
...
+      remoteObservation={remoteObservation}
+      remoteObsWasDeleted={remoteObsWasDeleted}
+      setRemoteObsWasDeleted={setRemoteObsWasDeleted}
+      fetchRemoteObservationError={fetchRemoteObservationError}
+      currentUser={currentUser}
+      belongsToCurrentUser={belongsToCurrentUser}
+      isRefetching={isRefetching}
+      refetchRemoteObservation={refetchRemoteObservation}
+      isConnected={isConnected}
+      isSimpleMode={showSavedMatch}
```
jtklein: According to your new early return this will never be `true` (because then we show the SavedMatch screen). This is essentially the toggle that either displays the full ObsDetailsDefaultMode screen, or the simple version of the ObsDetailsDefaultMode. Think of the isSimpleMode here as the first prototype of the "SavedMatch" screen. Now that we have the SavedMatch screen implemented we can remove the "simple mode" of ObsDetailsDefaultMode. The simple mode corresponds to this design [here](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=24990-30561&t=PUCoI5zdLPkTRAMx-4) which I have asked @abhasinat to remove from Figma to avoid confusion.

src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatchHeaderRight.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: Props ) => {
+  const navigation = useNavigation( );
+  const { localObservation } = useLocalObservation( uuid );
+  const { t } = useTranslation( );
+  const navigateToObsEdit = useNavigateToObsEdit( );
+
+  const headerRight = useCallback(
+    ( ) => (
+      <INatIconButton
+        testID="ObsDetail.editButton"
```
jtklein: SavedMatch.editButton

src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatch.tsx
```diff
@@ -0,0 +1,69 @@
...
+}: Props ) => {
+  const { t } = useTranslation( );
+  const { isConnected } = useNetInfo( );
+
+  const latitude = observation?.privateLatitude || observation?.latitude;
+  const { taxon } = observation;
+
+  return (
+    <ScrollViewWrapper>
+      <SavedMatchHeaderRight uuid={observation.uuid} />
```
jtklein: `SavedMatch` has the entire observation as prop which comes from [here](https://github.com/inaturalist/iNaturalistReactNative/blob/7acdbb768ebd0183b6bc5483b3da26963adbe920/src/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper.tsx#L50-L51). In SavedMatchHeaderRight the uuid is again used to get the same localObservation with the useLocalObservation as in the Wrapper. So, why not prop in the entire observation?

src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatchContainer.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const navigation = useNavigation<NativeStackNavigationProp<Record<string, { id?: number }>>>( );
+
+  const navToTaxonDetails
+  = ( photo?: { id?: number; isRepresentativeButOtherTaxon?: boolean } ) => {
+    const navParams: {
+        id?: number;
+        firstPhotoID?: number;
+        representativePhoto?: unknown;
+    } = { id: observation.taxon?.id };
+    if ( !photo?.isRepresentativeButOtherTaxon ) {
```
jtklein: This entire if..else can be removed. The information about the representative_photo is only transiently available in the AI camera -> Match screen ( ->  Taxon Details) flow. It is not persisted to the realm DB and not available for such saved observations that we want to show the SavedMatch screen for. So, the only param we can send to TaxonDetails here is the taxon ID. And you can remove the photo argument.

src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatchContainer.tsx
```diff
@@ -0,0 +1,40 @@
+import { useNavigation } from "@react-navigation/native";
+import { NativeStackNavigationProp } from "@react-navigation/native-stack";
+import SavedMatch from "components/ObsDetailsDefaultMode/SavedMatch/SavedMatch";
+import _ from "lodash";
+import React from "react";
+import type { RealmObservation } from "realmModels/types";
+
+type Props = {
```
jtklein: I'd prefer Props to be an interface.

src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatch.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { Button, ScrollViewWrapper } from "components/SharedComponents";
+import { View } from "components/styledComponents";
+import _ from "lodash";
+import React from "react";
+import type { RealmObservation } from "realmModels/types";
+import { useTranslation } from "sharedHooks";
+
+import SavedMatchHeaderRight from "./SavedMatchHeaderRight";
+
+type Props = {
```
jtklein: I'd prefer Props to be an interface.

src/realmModels/types.d.ts
```diff
@@ -123,6 +123,7 @@ export interface RealmObservationPojo {
   observed_time_zone?: string;
   owners_identification_from_vision: boolean | null;
   place_guess: string | null;
+  privateLatitude?: number;
```
jtklein: I understand you only needed this for your new TS components but it does look a bit lonely without it's brother `privateLongitude` see [here](https://github.com/inaturalist/iNaturalistReactNative/blob/7acdbb768ebd0183b6bc5483b3da26963adbe920/src/realmModels/Observation.js#L555).

src/components/ObsDetailsDefaultMode/SavedMatch/SavedMatch.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { taxon } = observation;
+
+  return (
+    <ScrollViewWrapper>
+      <SavedMatchHeaderRight uuid={observation.uuid} />
+      <View className={matchCardClassTop}>
+        <MatchHeader hideObservationStatus topSuggestion={observation} />
+      </View>
+      <PhotosSection
+        representativePhoto={taxon?.representative_photo}
```
jtklein: `representative_photo` is not persisted to the DB and hence not available here for this taxon.

jtklein (changes requested):
All right, I have looked through all the code, and tested this in the app and it's looking mostly good. Some UI feedback:
The screen header (not `MatchHeader`) is not quite to [spec](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=23655-66107&t=PUCoI5zdLPkTRAMx-4).
The pencil icon should be 22x22.
Screenshot 2025-11-28 at 15 43 09
The padding between the two icons and the card top is 10 but should be 20px.

Also, when navigating to the new saved Match sceen for an observation without an ID I can see a crash. To reproduce:
1. Take a photo that gives no suggestion in the AI camera.
2. Make sure you see the "The AI is not confident ..." text on the Match screen.
3. "SAVE" the obs without pressing one of the "It might be one of these" suggestions.
4. See an "Unknown" obs in the list
5. Press on this "Unknown" obs.
6. Crash with "Cannot read property 'rank_level' of null

In general I would prefer a PR to also remove dead code it creates. So, I would say removing the `isSimpleMode` version (see inline comment) should be part of this PR.

Can you add a test for the new ScreenWrapper component that checks that the correct version of ObsDetails vs SavedMatch  is shown please.
Can you also add an integration test for the SavedMatch screen, or at least make a ticket for it if you think that it would be easier to copy over an integration test for it once the Match screen [integration test ticket](https://linear.app/inaturalist/issue/MOB-723/testing-match-screen-integration-test) is done. If you opt for adding it here, maybe update the ticket with a link to the new test.

src/realmModels/types.d.ts
sepeterson: 🫂

tests/unit/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper.test.js
```diff
@@ -0,0 +1,149 @@
...
+const mockParams = {
+  uuid: "test-123"
+};
+
+jest.mock( "@react-native-community/netinfo", () => ( {
+  useNetInfo: () => ( { isConnected: true } ),
+  configure: jest.fn( )
+} ) );
+
+jest.mock( "sharedHooks/useCurrentUser", () => jest.fn() );
```
jtklein: If we don't need to change the return of this mock during the test, which we don't here, we should mock it once at the beginning of this file, like for example is done here: https://github.com/inaturalist/iNaturalistReactNative/blob/2a51f5ad9c41f56696ea96af66a2b1e2fe6863f3/tests/unit/components/ObsDetails/ObsDetails.test.js#L87

Then you can remove the part in beforeEach and the import.

tests/unit/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper.test.js
```diff
@@ -0,0 +1,149 @@
...
+jest.mock( "sharedHooks/useLocalObservation", () => jest.fn() );
+jest.mock( "sharedHooks/useRemoteObservation", () => jest.fn() );
+
+describe( "ObsDetailsDefaultModeScreensWrapper", ( ) => {
+  beforeEach( ( ) => {
+    useRoute.mockReturnValue( {
```
jtklein: Actually, if you follow the same mocks setup as we have in `tests/unit/components/ObsDetails/ObsDetails.test.js` you should be able to remove these here from the beforeEach, and remove the afterEach.

`useLocalObservation` is anyways changed for each test.

tests/unit/components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper.test.js
```diff
@@ -0,0 +1,149 @@
...
+  describe( "when showSavedMatch is true", ( ) => {
+    it(
+      "renders SavedMatchContainer when observation belongs to current user and is not synced",
+      ( ) => {
+        const unsyncedLocalObservation = {
+          ...mockLocalObservation,
+          wasSynced: jest.fn( () => false )
+        };
+
+        useLocalObservation.default.mockReturnValue( {
```
jtklein: `ObsDetails.test.js` also has example of mocking and changing the return of the mock for a single test: https://github.com/inaturalist/iNaturalistReactNative/blob/2a51f5ad9c41f56696ea96af66a2b1e2fe6863f3/tests/unit/components/ObsDetails/ObsDetails.test.js#L80-L81

I think we could safely copy the setup from there.

tests/integration/SavedMatch.test.js
```diff
@@ -0,0 +1,116 @@
...
+describe( "SavedMatch", ( ) => {
+  beforeEach( async ( ) => {
+    jest.clearAllMocks( );
+  } );
+
+  it( "should call navigation.push on learn more button press", async ( ) => {
+    const actor = userEvent.setup( );
+    renderAppWithComponent( <SavedMatchContainer observation={mockObservation} /> );
+    const learnMoreButton = await screen.findByText( t( "LEARN-MORE-ABOUT-THIS-SPECIES" ) );
+    await act( async ( ) => actor.press( learnMoreButton ) );
```
jtklein: That doesn't look right. `actor.press` should be wrapped internally in act already, I think. And removing here does not make the test fail.

jtklein (approved):
No big changes, for the ObsDetailsDefaultModeScreensWrapper.test.js I think you could largely follow the setup of tests/unit/components/ObsDetails/ObsDetails.test.js this could remove the beforeEach and afterEach hooks, and some of the hooks imports.

I haven't looked into the SavedMatch test too deeply, but everything seems alright at first glance. I would also rather get this merged in now than brood over tests.

Checked all changes in all other files, and everything seems okay to me.

PR #3250, opened 2025-11-21
Migrate `useAuthenticatedInfiniteQuery` to TypeScript.

frewsxcv:


src/sharedHooks/useAuthenticatedInfiniteQuery.ts
```diff
@@ -1,29 +1,51 @@
...
   // Use locale in case there is no user session
   const locale = i18n?.language ?? "en";
 
   return useInfiniteQuery( {
-    queryKey: [...queryKey, queryOptions.allowAnonymousJWT, currentUser],
-    queryFn: async params => {
-      if ( !currentUser ) {
-        params.locale = locale;
-      }
+    queryKey: [...queryKey, queryOptions.allowAnonymousJWT, currentUser, locale],
```
frewsxcv: Functional change 2:

Image

src/sharedHooks/useAuthenticatedInfiniteQuery.ts
```diff
@@ -1,29 +1,51 @@
...
+    queryFn: async ( { pageParam } ) => {
+      const params = {
+        pageParam,
+        ...( !currentUser && { locale } )
+      };
```
frewsxcv: Functional change 3: The closure parameter doesn't have a `locale` property, so we need to define a new object which potentially includes it, and then pass it along.

src/sharedHooks/useAuthenticatedInfiniteQuery.ts
```diff
@@ -1,29 +1,51 @@
...
+>, "queryKey"> & {
+  allowAnonymousJWT?: boolean;
+}
+
 // Should work like React Query's useInfiniteQuery with our custom reactQueryRetry
 // and authentication
-const useAuthenticatedInfiniteQuery = (
-  queryKey: Array<string>,
-  queryFunction: Function,
-  queryOptions: object = {}
```
frewsxcv: Functional change 1: Removal of ` = {}` because `getNextPageParam` is a required option parameter, so `{}` is invalid. I confirmed all callers pass `queryOptions`.

FLGMwt (changes requested):
I definitely appreciate that you've intentionally identified in the PR the functional changes in contrast to the non-functional TS changes.

This is a critical path of a few parts of the app so I think it does warrant a couple passes for safety:

I'm 👍 on the interfaces/types/parameter changes, including the "Functional Change" #.1 to the `queryOptions` which affirms our existing usage. If you'd be comfortable PRing that set of changes on it's own, it'd be much appreciated : )

For the "Functional Changes" "2" and "3", I think I'd maybe want some more exploration into that before introducing. I'm guessing the `exhaustive-deps` lint warning is probably correct, but this impacts a lot of the app so I think we want to be intentional about this change. Have you seen any issues in testing the app with this change?

PR #3251, opened 2025-11-21
Migrate `AppStateListener` to TypeScript.

frewsxcv:


src/components/AppStateListener.ts
```diff
@@ -12,7 +12,8 @@ const logger = log.extend( "AppStateListener" );
 
 const AppStateListener = ( ) => {
   const { loadTime } = usePerformance( {
-    screenName: "AppStateListener"
+    screenName: "AppStateListener",
+    isLoading: false
```
frewsxcv, line 16: Required

FLGMwt: 👍 

This in mind, feel free to call out any booleans in _our_ code that you think oughta be typed as explicitly required. I know a lot of bool params/props are defaulted. That sometimes fine, but it's nice when the _really_ key bool params are explicit. Don't want to put any expectation on you, just letting you know it's welcome while you work on some files that may not have as much recent attention < 3

FLGMwt (approved):
woooo! LGTM! this is my favorite kind of file delta >: D

PR #3252, opened 2025-11-21
Reuse `QueryKey` type from `@tanstack/react-query`.

frewsxcv:
Prior to this, query key was typed to be `string[]`, even though it can be other types.

https://tanstack.com/query/v5/docs/framework/react/guides/query-keys


src/sharedHooks/useAuthenticatedInfiniteQuery.ts
```diff
@@ -8,7 +8,7 @@ import { useCurrentUser } from "sharedHooks";
 // Should work like React Query's useInfiniteQuery with our custom reactQueryRetry
 // and authentication
 const useAuthenticatedInfiniteQuery = (
-  queryKey: Array<string>,
+  queryKey: QueryKey,
```
FLGMwt, line 12: maybe this type "correction" is just highlighting an erroneous caller of this code but.... we spread this here https://github.com/inaturalist/iNaturalistReactNative/pull/3252/files#diff-76b47bbee105159db467d028840da68c4b5dd0dabe8a57db2fb2fce94a9a1f28R21.

Just looking at the later change https://github.com/inaturalist/iNaturalistReactNative/pull/3252/files#diff-9059d29458ca045e30c8ebcbc99e0c41e7317961c0c2a47c70a6c9c2b36051f3L7-R7

```
  queryKey: Array<string>,
  queryKey: QueryKey[],
```

we go from `string[]` to `QK[]`. Do you have any insight as to if we're using this wrong in either place?

frewsxcv: `QueryKey` is typed to be an array, so it shouldn't be `QueryKey[]`, good catch! If there was type checking enabled in CI it would have caught that. Hopefully in the future!

PR #3254, opened 2025-11-23
Resolve all type errors in `IconicTaxonChooser` component.

frewsxcv:


src/components/SharedComponents/IconicTaxonChooser.tsx
```diff
@@ -71,7 +70,9 @@ const IconicTaxonChooser = ( {
           onPress={( ) => {
             onTaxonChosen( iconicTaxonName );
           }}
-          color={isSelected && colors.white}
```
FLGMwt, line 74: wait why was this grumpy? : p I much prefer the former

FLGMwt: OH hah, we shouldn't have been using the `<View color={...} />` prop here. Would you mind switching our wrong code to a conditional tailwind classname?

frewsxcv: The issue was that `isSelected && colors.white` can return `false`, which isn't an acceptable value for `color`. And yep I can switch it to a class

frewsxcv: @FLGMwt Actually, what are thinking with respect to the classname? There's no className prop and I'm unsure how that which component it would get propagated through to

FLGMwt: Oh, whoops, I'm sorry, I fell victim to a line-squashing red-herring in the PR review UI, I thought this was on a `<View />` but it is not, it's on the internal `Icon...` component. That's embarrassing of me and I'm sorry for the distraction.

I'll merge this in as proposed so no changes requested @frewsxcv, thank you again : )

That said, this has highlighted a wonky API for a common component so I've created an internal ticket to address this from the component API perspective and it's varied uses that seem to take varied strategies to workaround this.

PR #3257, opened 2025-12-02
MOB-988 types supporting online/offline experiment flag

FLGMwt:
This was intended to preface #3258, but in the interest of getting that functional change in, I'm PR'ing it in isolation.

In addition to a handful of  `js => ts` and type annotations, it includes the addition of `sharedHooks/useSuggestions/types.ts` for shared types that are consumed by `sharedHooks/useSuggestions/`. 

**_I'm calling this out_ because it's a bit new in our TS migration strategy**, aside from the more platform-level api/types and realmModels. Reasoning for this:

- `Suggestion`-related types are defined and redefined in a varied manner across `useSuggestions`'s multi-file implementation, and its consumers, `MatchContainer` (for AI Camera) and `SuggestionsContainer` (for photo library). Sometimes even circularly (`useSuggestions/filterSuggestions` importing `SuggestionsContainer`) or across concerns (`MatchContainer` importing `SuggestionsContainer`)
- `useSuggestions`'s components, `useOnlineSuggestions` and `useOfflineSuggestions` have very small but important differences in their APIs which are not obvious since they don't share type definitions.
- ^these factors in mind, I think we can improve the readability and maintainability of this area of code w/ types explicitly intended for reuse. 

These shared types are not (yet*) intended to represent the entire codebase's understanding of a `Suggestion`, but rather a shared understanding _only_ within the implementation of `useSuggestions`, hence the verbose namespacing. *:As a separate effort, we might decide it makes sense to "lift" these types out of the hook-implementation-only context into a more generally shared type definition and drop the namespace.

^I'm being extra explicit here because I want to confirm if y'all think that this is in line with our "from the leaves" type definition strategy we talked about @abbeycampbell @jtklein @sepeterson.

### DIFF
--- a/src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
+++ b/src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
@@ -34,7 +34,7 @@ interface UseOnlineSuggestionsResponse {
   dataUpdatedAt: number;
   onlineSuggestions?: {
     results: UseSuggestionsOnlineSuggestion[];
-    common_ancestor: UseSuggestionsOnlineSuggestion;
+    common_ancestor?: UseSuggestionsOnlineSuggestion;
   };
   timedOut: boolean;
   error: Error | null;
@@ -44,7 +44,7 @@ interface UseOnlineSuggestionsResponse {
 
 interface OnlineSuggestionsQueryResponse {
   results: UseSuggestionsOnlineSuggestion[];
-  common_ancestor: UseSuggestionsOnlineSuggestion;
+  common_ancestor?: UseSuggestionsOnlineSuggestion;
 }
 
 const useOnlineSuggestions = (


src/sharedHooks/useSuggestions/useSuggestions.ts
```diff
@@ -101,6 +109,8 @@ export const useSuggestions = ( photoUri, options ) => {
     ]
   );
 
+  console.log( { offlineSuggestions, onlineSuggestions } );
```
abbeycampbell: remove log?

src/sharedHooks/useSuggestions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+    name: string;
+    rank_level: number;
+    iconic_taxon_name?: string;
+  };
+}
+
+export type TopSuggestionType = string;
+
+export type UseSuggestionsSuggestion =
+    UseSuggestionsOnlineSuggestion | UseSuggestionsOfflineSuggestion;
```
abbeycampbell, line 35: this is definitely a nitpick but I don't love this name, it kind of wigs my brain out lol. what about like `UseNormalizedSuggestion`? 
again not at all a blocker

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -104,7 +123,7 @@ const useOnlineSuggestions = (
     // we're already getting all this taxon information anytime we make this API
     // call, so we might as well store it in realm immediately instead of waiting
     // for useTaxon to fetch individual taxon results
-    const mappedTaxa = onlineSuggestions?.results?.map(
+    const mappedTaxa = onlineSuggestions!.results!.map(
```
abbeycampbell: I'd really love to avoid usage of the non-null assertion, since it can lead to silent errors and kind of defeats the point of static type-checking in the first place. This can get a little fiddly and you've spent a lot longer messing around with this code than me, so ultimately I'll defer to your judgement though. 
I see that for this one, we already handle the `onlineSuggestions === undefined` case above anyways, but we could also just add an early return to this memoized callback like:
`if ( !onlineSuggestions ) return;`
that would keep us from having to ignore ts errors.

src/sharedHooks/useSuggestions/filterSuggestions.ts
```diff
@@ -62,18 +71,22 @@ const filterSuggestions = ( suggestionsToFilter, commonAncestor ) => {
     ).at( 0 );
     return {
       ...newSuggestions,
-      topSuggestion: firstSuggestion,
+      topSuggestion: firstSuggestion!,
```
abbeycampbell: do we need to assert this is non-null?

src/sharedHooks/useSuggestions/filterSuggestions.js
```diff
@@ -1,5 +1,4 @@
 import {
-  initialSuggestions,
```
jtklein, line 2: I agree this makes no sense being imported from `SuggestionsContainer`. But I think we can import in `SuggestionsContainer` from here then, instead of initialising the same object in two places.

Or import both from `useSuggestions`?

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
jtklein: I agree.

src/sharedHooks/useSuggestions/types.ts
jtklein: I understand the intensive namespace here, but only if we intend to arrive at a unified Suggestion type later on.

src/sharedHooks/useSuggestions/useSuggestions.ts
```diff
@@ -1,12 +1,21 @@
...
+import type { ScoreImageParams, UseSuggestionsResult } from "./types";
 import useOfflineSuggestions from "./useOfflineSuggestions";
 import useOnlineSuggestions from "./useOnlineSuggestions";
 
-export const useSuggestions = ( photoUri, options ) => {
+interface Options {
+  shouldFetchOnlineSuggestions: boolean;
+  onFetchError: ( options: { isOnline: boolean } ) => void;
+  onFetched: ( options: { isOnline: boolean } ) => void;
+  scoreImageParams: ScoreImageParams;
```
jtklein: Can be `null`.

src/sharedHooks/useSuggestions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+    onlineSuggestions?: {
+        results: UseSuggestionsOnlineSuggestion[];
+        commonAncestor?: UseSuggestionsOnlineSuggestion;
+    };
+    timedOut: boolean;
+    resetTimeout: () => void;
+}
+
+export interface ScoreImageParams {
+  lat: number;
```
jtklein: Can be `undefined`.

src/sharedHooks/useSuggestions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  lng: number;
```
jtklein: Can be `undefined`.

src/sharedHooks/useSuggestions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  image: {
+    uri: string,
```
jtklein: semicolon

src/sharedHooks/useSuggestions/useSuggestions.ts
```diff
@@ -1,12 +1,21 @@
...
+  queryKey: string[];
```
jtklein: This is propped down to `useQuery` from `@tanstack/react-query`  so in principle it can be anything that is accepted there as a type. E.g. in our `useAuthenticatedQuery.ts` we infer it as `(string | boolean | null | undefined)[]`. However, in the `MatchContainer` and `SuggestionsContainer`, we even provide an object in this array which is so far not included in any type down to useQuery. See also: https://github.com/inaturalist/iNaturalistReactNative/pull/3252

Maybe I am a bit unsure what you want the scope of this PR to be. From the top we do get other things than `string` propped in here. In all the files you touched here we are only sending it down into the tips. So, in the scope of this PR it could be `any` as far as we know. The tip usage of it (in useQuery) is defined as `string[]` so I can understand why you'd just want to prop this back up.

src/sharedHooks/useSuggestions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export type TopSuggestionType = string;
+
+export type UseSuggestionsSuggestion =
+    UseSuggestionsOnlineSuggestion | UseSuggestionsOfflineSuggestion;
+
+export interface UseSuggestionsResult {
+  suggestions: {
+  otherSuggestions: UseSuggestionsSuggestion[];
```
jtklein: Indentation is off for this interface

src/sharedHooks/useSuggestions/filterSuggestions.ts
jtklein: dito. Removing it doesn't seem to change the type.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
jtklein: I checked through the codebase and this syntax has not been used before. Personally, I am also not familiar with it. I checked the TypeScript documentation and I am not sure when to use it because I would imagine that in general I would rather have a null check before I'd use whatever I am asserting here as not being null.
So, in a component setting I'd expect an early return or fallback if I don't have the data I need.

src/sharedHooks/useSuggestions/filterSuggestions.ts
```diff
@@ -62,18 +71,22 @@ const filterSuggestions = ( suggestionsToFilter, commonAncestor ) => {
...
 
   // online or offline common ancestor
   if ( commonAncestor ) {
     const sortableCommonAncestor = {
-      ...commonAncestor,
-      // temp patch to let calling code sort online common ancestor with other suggestions
-      combined_score: commonAncestor.combined_score ?? commonAncestor.score
+      ...commonAncestor
     };
+    if ( "score" in commonAncestor ) {
```
jtklein: That is not the same as before. Before: We always use `commonAncestor.combined_score` if present. Only if not present we use score.
After: We use score over combined_score. Even if both are present in commonAncestor.
I would say to change this back to as it was, or provide an explanation as to why this is necessary.

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
```diff
@@ -53,7 +45,9 @@ const useOfflineSuggestions = (
     let rawPredictions = [];
     let commonAncestor;
     try {
-      const location = { latitude, longitude };
+      const location = ( typeof latitude === "number" && typeof longitude === "number" )
```
jtklein, line 49: Directionally this is correct, because the offline cv model also works fine without a location given. Could you just also update the `predictImage` function to accept `undefined` as type.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -3,35 +3,54 @@ import {
...
 
-interface OnlineSuggestionsResponse {
-  dataUpdatedAt: Date;
-  onlineSuggestions: object;
-  loadingOnlineSuggestions: boolean;
+interface OnlineSuggestionOptions {
+  onFetchError: ( options: { isOnline: boolean } ) => void;
+  onFetched: ( options: { isOnline: boolean } ) => void;
+  scoreImageParams: ScoreImageParams;
+  queryKey: string[];
```
jtklein, line 29: Same comment as in useSuggestions

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -3,35 +3,54 @@ import {
...
+  shouldFetchOnlineSuggestions: boolean;
+}
+
+interface UseOnlineSuggestionsResponse {
+  dataUpdatedAt: number;
+  onlineSuggestions?: {
+    results: UseSuggestionsOnlineSuggestion[],
+    common_ancestor: UseSuggestionsOnlineSuggestion
```
jtklein: Can this be `undefined`?

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -3,35 +3,54 @@ import {
...
-  error: object;
+  error: Error | null;
   resetTimeout: () => void;
-  isRefetching: boolean;
+  refetch: () => void;
+}
+
+interface OnlineSuggestionsQueryResponse {
+  results: UseSuggestionsOnlineSuggestion[];
+  common_ancestor: UseSuggestionsOnlineSuggestion;
```
jtklein: Can this be `undefined`?

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -115,9 +134,9 @@ const useOnlineSuggestions = (
     safeRealmWrite( realm, ( ) => {
       mappedTaxa.forEach( remoteTaxon => {
         realm.create(
-          "Taxon",
+          Taxon,
```
jtklein: This is a new pattern that hasn't been used in the codebase before. I would like to be included in the learning process as to why this is also possible to use (or if it is better than the existing pattern).

src/sharedHooks/useSuggestions/useSuggestions.ts
```diff
@@ -70,18 +79,17 @@ export const useSuggestions = ( photoUri, options ) => {
...
     }
   };
+  const hasOnlineSuggestionResults = ( onlineSuggestions?.results?.length || 0 ) > 0;
+  const hasOfflineSuggestionResults = ( offlineSuggestions?.results?.length || 0 ) > 0;
 
-  const usingOfflineSuggestions = tryOfflineSuggestions || (
-    offlineSuggestions?.results?.length > 0
-      && ( !onlineSuggestions || onlineSuggestions?.results?.length === 0 )
+  const usingOfflineSuggestions = fallingBackToOfflineSuggestions || (
+    hasOfflineSuggestionResults && !hasOnlineSuggestionResults && onlineSuggestionsAttempted
```
jtklein: Why include `onlineSuggestionsAttempted` here?

src/sharedHooks/useSuggestions/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  combined_score: number;
+  taxon: {
+    id: number;
+    name: string;
+    rank_level: number;
+    iconic_taxon_name?: string;
+  };
+}
+
+export type TopSuggestionType = string;
```
jtklein: This is nowhere imported. But I do think it deserves a central type. It is not a string though, more a set of mutually exclusive values defined [here](https://github.com/inaturalist/iNaturalistReactNative/blob/e9ee6752e4d3265bec32a6e5c6545c4fef945a5e/src/components/Suggestions/SuggestionsContainer.tsx#L41-L42):
```
export const TOP_SUGGESTION_NONE = "none";
export const TOP_SUGGESTION_HUMAN = "human";
export const TOP_SUGGESTION_ABOVE_THRESHOLD = "above-threshold";
export const TOP_SUGGESTION_COMMON_ANCESTOR = "common-ancestor";
export const TOP_SUGGESTION_NOT_CONFIDENT = "not-confident";
```
I would prefer to use an enum in such cases, as they are all mutually exclusive states.

So, I would say we refactor these states here and import in `SuggestionsContainer`, and also use them down below in `UseSuggestionsResult.suggestions.topSuggestionType`

jtklein (changes requested):
Requested a few changes here and there. There are a few functional changes in here as well which I am not sure about the implications off and so if we actually want them to be included. I think I left questions were I don't necessarily see the need for a functional change.
There are also two new patterns introduced here blabla! and using the Realm class as argument in .create. If anything such an introduction should take the entire mobile team along as to why this is better or needed. And in the case of the realm.create if it is a better pattern create a follow up ticket to get to uniform usage.
So, in my view this warrants a separate PR and being reasoned about before we make a ticket.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
frewsxcv: See also https://github.com/inaturalist/iNaturalistReactNative/pull/3167#discussion_r2478375265

And for what it's worth it's already in use in the codebase:

Screenshot 2025-12-11 at 10 44 34 AM

src/sharedHooks/useSuggestions/types.ts
FLGMwt: Yeah, the goal was to make this explicitly scoped to this directory for now to differentiate it from locally-defined ones in other files. My intent is that we'd eventually move this to a codebase-level "official" type that anything dealing with a "Suggestion" would use this type, at which point we'd drop the namespace

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
FLGMwt: so sorry! taking out my `!`s. I took for granted my previous teams' use of `!`. we used it as kind of a similar type system todo like `any` / `unknown` / etc. in this case, I just wanted to keep existing behavior, but document it as a "hey, we're fighting the type system here". we would have NRE'd had we passed it an undef before or after this change. I think it's nice to highlight that, even with an NRE instead of your caller not be sure why your function didn't work (or run?).

this had me look at it a bit more and I realized I could inline it into its caller so this is OBE!

src/sharedHooks/useSuggestions/filterSuggestions.ts
FLGMwt: yarp, artifact from when this was looser as I was adding types. the fact that TS didn't call it out as unnecessary is further souring me on `!`. fixed.

src/sharedHooks/useSuggestions/filterSuggestions.js
FLGMwt: for sure, fixed.

this was one of the ones I was on the fence about because I didn't want to dig _too_ much into the non-hook realm of circular dependencies / coupling (MOB-1081 created for TODO comment). I decided to make this an export from the hooks w/ a `internal` namespace and a TODO comment to the effect of "we're changing this, you shouldn't use it" to prevent any further functional changes.

src/sharedHooks/useSuggestions/types.ts
FLGMwt: switched to a TS discriminated union for this: https://github.com/inaturalist/iNaturalistReactNative/blob/fa6791d30fcb4a89e40f9b81d789ee4a6ed4f75b/src/sharedHooks/useSuggestions/types.ts#L56. they're both still `number?`, but ts will know that if one's one way, the other will be that way.

let me know if you think this is too much. I was tempted to push them into a `location?` property where lat/long are required, but that required _way_ more changes. didn't make sense to tackle that now

src/sharedHooks/useSuggestions/useSuggestions.ts
FLGMwt: yeah, my thinking was to align with the types coming  _from_ `uAQ` which defines `queryKey: string[]` as an arg. The `(string | boolean | null | undefined)[]` you mentioned is internal to `uAQ` and is what we pass `uQ`. I didn't want to take this moment to challenge "hey uAQ should take / is actually taking more than `string[]`" since my changes aren't _creating_ any qks (or type problems : P). That is: useSuggestions doesn't _know_ it has a problem with the qK type.

and yeah, uAQ should technically accept `AnySerializableType[]`.

src/sharedHooks/useSuggestions/types.ts
FLGMwt: oops, fixed.

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
FLGMwt: aye, fixed. similar to above, it would be nice to have a `location` property to avoid the one-off / two-off checks, but that turned out to be too much change surface area.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
FLGMwt: mmmm, yeah, I didn't realize that was totally new at that time. I've pulled this back for now and added as a team topic. I'm pro, it's just not something we've done before.

src/sharedHooks/useSuggestions/types.ts
FLGMwt: yup, yup, added as enum. that was part of the reckoning of "should I touch SuggestionsContainer / MatchContainer w/ this"? but I agree, it's cleaner to do here.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -65,9 +75,15 @@ const useOnlineSuggestions = (
...
+  } = useAuthenticatedQuery<OnlineSuggestionsQueryResponse>(
     queryKey,
-    queryFn,
+    async optsWithAuth => {
+      const params = {
+        ...scoreImageParams,
+        ...( !currentUser && { locale } ),
+      };
+      return scoreImage( params, optsWithAuth ) as Promise<OnlineSuggestionsQueryResponse>;
+    },
```
FLGMwt, line 86: inlining this to infer `optsWithAuth` & fix _this_ type error. We export `ApiOpts` that's technically meant `ApiOpts` and `useAuthenticatedQuery` disagree about null vs undef. I'm just making that not-suggestion's-problem ; D

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
```diff
@@ -67,30 +62,21 @@ const useOfflineSuggestions = (
...
     // but we're offline so we only need the local list from realm
     // and don't need to fetch taxon from the API
-    const iconicTaxa = realm?.objects( "Taxon" ).filtered( "isIconic = true" );
-    const iconicTaxaIds = iconicTaxa.map( t => t.id );
-    const iconicTaxaLookup: {
-      [key: number]: string;
-    } = iconicTaxa.reduce( ( acc, t ) => {
-      acc[t.id] = t.name;
-      return acc;
-    }, { } );
```
FLGMwt, line 77: Change on the right fixes the _last_ type error in `sharedHooks/useSuggestions`. I found the pattern https://github.com/inaturalist/iNaturalistReactNative/blob/1411a9a5238f85a83661adb9f61b40b457ddff0e/src/components/MyObservations/MyObservationsContainer.tsx#L278-L279 we use for type info from `realm.object`. Using `Object.entries` and `in` [below here](https://github.com/inaturalist/iNaturalistReactNative/pull/3257/changes#diff-059a5fd6a2ae97e57d7407fd2d7c501a159177d4af8a782f232e4397d214c688R79) reduces the amount of type definition we have to do.

FLGMwt (commented):
Since we did our walkthrough earlier @jtklein, I went ahead and fix those two last type errors that I was considering leaving outstanding, just wanted to FYI the delta.

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
jtklein: This seems alright, but it does surface that we use two different kinds of syntax in the codebase:
`realm.objects( User )`
`realm.objects<RealmObservation>( "Observation" )`
maybe not mutually exclusive and possibly with different concerns, albeit it looks to me hat `RealmObservation` is the type we expect to get back from the client-side database.

jtklein (approved):
I looked at the difference of commits to the latest I reviewed and I think all requested changes are taken care of and all additions are fine, as discussed offline.
I glanced only briefly at the complete changes introduced in this PR and I think it is good to go.
Thanks for the detailed discussion about new TS patterns and how we can distribute the updates in the team's shared knowledge about these. I hope we have follow-up tickets and notes that turn into actions exhaustively of the things we discussed.

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
FLGMwt: yeah, I think this is a similar thing to what we discussed w/ `.create` in terms of codebase consistency. MOB-1085 created for this concern. https://linear.app/inaturalist/issue/MOB-1085/stub-how-even-do-we-type-realmobjects-realmcreate-figure-it-out-and

Not adding a todo comment to this line since the ticket should catch this in codebase search. Since using this style codifying and enforces the existing use of the realm response, I'll stick to this style for this PR and expect it may be updated based on the result of that ticket.

FLGMwt:
> I looked at the difference of commits to the latest I reviewed and I think all requested changes are taken care of and all additions are fine, as discussed offline. I glanced only briefly at the complete changes introduced in this PR and I think it is good to go. Thanks for the detailed discussion about new TS patterns and how we can distribute the updates in the team's shared knowledge about these. I hope we have follow-up tickets and notes that turn into actions exhaustively of the things we discussed.

Thanks again for the detailed reviews and thoughtful discussion! I didn't intend to bring up as many discussions but I think it was really productive

PR #3258, opened 2025-12-02
MOB-988 setup to run unconditional offline experiment

FLGMwt:
Hat in hand, I got overambitious trying to take on a bigger refactor of `useSuggestions` that got away from me, I should have focused on the minimal implementation. I moved this ahead of the [related TypeScript changes](https://github.com/inaturalist/iNaturalistReactNative/pull/3257) as well so we can pull this in in isolation.

Hard-coded boolean in code for now per ticket. I'll need to clarify if we wanted it on or off on merge, but PR'd as off to be safe.

I know the name is verbose but damn if it isn't accurate ; D

### DIFF
--- a/src/sharedHooks/useSuggestions/useShouldFetchOfflineSuggestionsUnconditionally.ts
+++ b/src/sharedHooks/useSuggestions/useShouldFetchOfflineSuggestionsUnconditionally.ts
@@ -1,0 +1,22 @@
+import { useMemo } from "react";
+
+const simultaneousOnlineOfflineSuggestionsEnabled = false;
+const simultaneousOnlineOfflinePercentageChance = 1;
+
+const useShouldFetchOfflineSuggestionsUnconditionally = ( ) => {
+  const fetchOfflineSuggestionsUnconditionally = useMemo(
+    () => {
+      if ( !simultaneousOnlineOfflineSuggestionsEnabled ) {
+        return false;
+      }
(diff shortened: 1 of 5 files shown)


src/sharedHooks/useSuggestions/useShouldFetchOfflineSuggestionsUnconditionally.ts
```diff
@@ -0,0 +1,22 @@
...
+const simultaneousOnlineOfflineSuggestionsEnabled = false;
+const simultaneousOnlineOfflinePercentageChance = 1;
+
+const useShouldFetchOfflineSuggestionsUnconditionally = ( ) => {
+  const fetchOfflineSuggestionsUnconditionally = useMemo(
+    () => {
+      if ( !simultaneousOnlineOfflineSuggestionsEnabled ) {
+        return false;
+      }
+      return Math.random() <= simultaneousOnlineOfflinePercentageChance;
```
jtklein, line 12: `Math.random() * 100` ? Although I am also fine with using 0.01 for `simultaneousOnlineOfflinePercentageChance` even though it is labelled as percentage.

FLGMwt: Yes, you're correct, thank you! I renamed / changed the "concept" of the constant (on the 1 vs 100 scale) after testing and just before push ing and I wasn't careful enough with it.

PR #3259, opened 2025-12-02
use Subheading2 over Heading2 for location name

FLGMwt:
This was one of the two areas mentioned in MOB-744 that needed fixing, but it looks like the other instance in for the Taxon name is already corrected for the named uses.

### DIFF
--- a/src/components/SharedComponents/SimpleObservationLocation.tsx
+++ b/src/components/SharedComponents/SimpleObservationLocation.tsx
@@ -4,7 +4,9 @@ import React from "react";
 import { useTranslation } from "sharedHooks";
 
 interface Props {
-  observation: object;
+  observation: {
+    private_place_guess?: string
+  };
 }
 
 const SimpleObservationLocation = ( {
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/SimpleObservationLocation.tsx
```diff
@@ -11,23 +11,19 @@ const SimpleObservationLocation = ( {
...
-      observation.private_place_guess
-        ? "privatePlaceGuess"
-        : "placeGuess"
-    ),
-    [observation]
+  const displayLocation = checkCamelAndSnakeCase(
+    observation,
+    "private_place_guess" in observation
+      ? "privatePlaceGuess"
+      : "placeGuess"
```
FLGMwt, line 20: Two things:
- removing unnecessary memoization of inexpensive operation
- fix property access type error by replacing truthy check on `observation.private_place_guess` w/ the more explicit boolean challenge of `"private_place_guess" in observation`

I don't think it makes sense to refine the `observation: object` type in this changeset as this is something we'll probably want to address holistically. Otherwise we're looking at `observation: { private_place_guess: string } | { privatePlaceGuess: string } | { place_guess: string } | { placeGuess: string }` ; D

PR #3261, opened 2025-12-02
Replace potentially problematic splatting behavior with `Object.assign`.

frewsxcv:
Extracted from https://github.com/inaturalist/iNaturalistReactNative/pull/3188, and see [this comment](https://github.com/inaturalist/iNaturalistReactNative/pull/3188#discussion_r2496101477).

Splatting does not copy the prototypal inheritance, which means you won't be able to call instance methods on the new splatted Realm Models.


FLGMwt (approved):
LGTM, nice that ts caught this. Doesn't appear to affect any problematic use now, but nice avoid any future assumptions.

PR #3263, opened 2025-12-02
Fix the only two type errors in `PhotoContainer` component.

frewsxcv:


src/components/ObsDetails/PhotoContainer.tsx
```diff
@@ -73,7 +73,7 @@ const PhotoContainer = ( { photo, onPress, style }: Props ) => {
   const renderLoadingIndicator = ( ) => {
     if ( loadSuccess === null ) {
       return (
-        <ActivityIndicator className="absolute" />
```
frewsxcv, line 76:
```
Type '{ className: string; }' is not assignable to type 'IntrinsicAttributes & Props'.
  Property 'className' does not exist on type 'IntrinsicAttributes & Props'.ts(2322)
```

`ActivityIndicator` did not have a `className` prop, so this presumably had no effect.

src/components/ObsDetails/PhotoContainer.tsx
```diff
@@ -23,7 +23,7 @@ const PhotoContainer = ( { photo, onPress, style }: Props ) => {
 
   const imageSources = [];
   if ( photo.localFilePath ) {
-    imageSources.push( { uri: Photo.getLocalPhotoUri( photo.localFilePath ) } );
+    imageSources.push( { uri: Photo.getLocalPhotoUri( photo.localFilePath ) || undefined } );
```
frewsxcv:
```
Type '({ uri: string | null; } | { uri: string; width: number; height: number; })[]' is not assignable to type 'ImageSourcePropType | undefined'.
  Type '({ uri: string | null; } | { uri: string; width: number; height: number; })[]' is not assignable to type 'ImageURISource[]'.
    Type '{ uri: string | null; } | { uri: string; width: number; height: number; }' is not assignable to type 'ImageURISource'.
      Type '{ uri: string | null; }' is not assignable to type 'ImageURISource'.
        Types of property 'uri' are incompatible.
          Type 'string | null' is not assignable to type 'string | undefined'.
            Type 'null' is not assignable to type 'string | undefined'.ts(2322)
Image.d.ts(233, 3): The expected type comes from property 'source' which is declared here on type 'IntrinsicAttributes & ImageProps & { className?: string | undefined; tw?: string | undefined
```

`uri` is expected to be `string | undefined`, not `string | null`.

FLGMwt: I think we should be instead pushing conditionally based on the result of `getLocalPhotoUri` rather than (now explicitly) pushing null/undef

frewsxcv: Makes sense. Resolved in https://github.com/inaturalist/iNaturalistReactNative/pull/3263/commits/c133d05fb2849484ee748834092cf1cbb2533a56

src/components/ObsDetails/PhotoContainer.tsx
frewsxcv: See also https://github.com/inaturalist/iNaturalistReactNative/pull/3280#discussion_r2607469256

frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3280#discussion_r2614242234

PR #3264, opened 2025-12-03
Migrate `navigateToObsDetails` to TypeScript.

frewsxcv:


src/components/ObsDetails/helpers/navigateToObsDetails.ts
```diff
@@ -1,8 +1,11 @@
-import { CommonActions } from "@react-navigation/native";
+import { CommonActions, type NavigationProp, type ParamListBase } from "@react-navigation/native";
```
FLGMwt, line 1: ooooh, this called out the fact that our recent lint rule for enforcing consistent import type was incomplete. I'll merge this in, fix that lint rule, and make sure this is corrected with the autofix

FLGMwt (approved):
LGTM

src/components/ObsDetails/helpers/navigateToObsDetails.ts
FLGMwt: here's that followup: https://github.com/inaturalist/iNaturalistReactNative/pull/3265/files#diff-5bcc463b5b7a703da01337b40971b6868cb130cf8a837a65ad5a5dcc8707c626R1-R2

PR #3266, opened 2025-12-04
add stylistic eslint plugin package dependency

FLGMwt:
MOB-1039 1/2 (`@stylistic` plugin package only)

Our objective is to add lint rules to enforce common coding styles / patterns so we don't have to worry about making them consistent manually.

These types of rules have sometimes been a part of eslint / tslint _but_ in recent years, those ecosystems decided to separate style concerns from general linting concerns: https://v4.eslint.style/guide/why

For example, we want to make type delimiters consistent (`;` vs `,` in interface definitions) which was _previously_ done here: https://typescript-eslint.io/rules/member-delimiter-style/ <- but this notes that that rule has moved to @stylistic _and has been removed from tslint_: https://eslint.style/rules/member-delimiter-style#member-delimiter-style

This PR simply adds the `@stylistic/eslint-plugin` package and adds it to our configured eslint plugins. It has no rules enabled by default so this PR is a noop.

Following PRs will add new rules and autofixes.

NOTE: our project uses the "legacy" eslint configuration style (`module.exports`). @stylistic adoption guidance points us to use @stylistic v3 for this reason: https://v4.eslint.style/guide/migration#:~:text=If%20you%20are%20still%20using%20legacy%20config

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,8 @@ module.exports = {
     "react-native",
     "simple-import-sort",
     "@tanstack/query",
-    "@typescript-eslint"
+    "@typescript-eslint",
+    "@stylistic"
   ],
   rules: {
     "arrow-parens": [2, "as-needed"],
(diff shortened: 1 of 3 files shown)


sepeterson (approved):
🆒

PR #3267, opened 2025-12-04
Migrate `useStoredLayout` to TypeScript.

frewsxcv:


jtklein (approved):
Thanks for that, this is actually quite timely! We need this for some re-working of the Explore tab very soon.
E2e tests are passing for me locally, so I'll merge.

PR #3268, opened 2025-12-04
Migrate `useLogs` to TypeScript.

frewsxcv:


src/sharedHooks/useLogs.ts
```diff
@@ -22,7 +22,7 @@ async function getLogContents() {
     const contents = await RNFS.readFile( logFilePath );
     return contents.split( "\n" ).join( "\n" );
   } catch ( readFileError ) {
-    if ( readFileError.message.match( /no such file/ ) ) {
+    if ( readFileError instanceof Error && readFileError.message.match( /no such file/ ) ) {
```
frewsxcv, line 25: `.message` can't exist without an `instanceof Error` check

src/sharedHooks/useLogs.ts
```diff
@@ -71,7 +71,7 @@ const useLogs = ( ) => {
             },
             {
               text: t( "Share" ),
-              onPress: () => shareLogFile( emailParams )
```
frewsxcv, line 74: `shareLogFile` does not take any parameters

src/sharedHooks/useLogs.ts
FLGMwt: Yup, makes sense to me!

Mentioning for the sake of narrative: there were a couple recent instances of us being a little more guarded about type refinement strategies that might be somewhat new to the codebase (https://github.com/inaturalist/iNaturalistReactNative/pull/3259#issuecomment-3603013058 / https://github.com/inaturalist/iNaturalistReactNative/pull/3188#discussion_r2496084403), but I think this is distinct from those in that it has a few examples already in the codebase _and_ this `catch` / `instanceof Error` is a much more pervasive js pattern in the ecosystem than our more local ts concerns.

FLGMwt:
oops, posthoc LGTM

PR #3269, opened 2025-12-04
Avoid lint suppressions in `useEffect` by making returns consistent.

frewsxcv:
`useEffect` expects all returns to either return `undefined`, or destructor functions. In some spots, we are return a mix, so there were lint suppressions added to ignore those cases. Instead of encouraging lint suppressions, this replaces those with return `noop` functions so we comply with the `useEffect` typing.

### DIFF
--- a/src/components/hooks/useWorkQueue.ts
+++ b/src/components/hooks/useWorkQueue.ts
@@ -1,6 +1,5 @@
 import { fetchUserMe, updateUsers } from "api/users";
 import { getJWT } from "components/LoginSignUp/AuthenticationService";
-import { noop } from "lodash";
 import { RealmContext } from "providers/contexts";
 import { useEffect, useState } from "react";
 import QueueItem from "realmModels/QueueItem";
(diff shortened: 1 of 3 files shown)


src/components/hooks/useWorkQueue.ts
```diff
@@ -64,7 +65,7 @@ const useWorkQueue = ( ) => {
 
   useEffect( ( ) => {
     if ( realm === null || realm.isClosed ) {
-      return;
+      return noop;
```
FLGMwt: I personally think it may be clearer to just define the noop `() => {}` inline where we need it. I don't think there's a material performance concern and it makes it makes functions more "straight-down" readable w/o having to figure out what the `noop` var is. Maybe with a comment clarifying that it's to satisfy the effect cleanup consistency?

I don't feel that strongly about it, what do you think?

frewsxcv: Yep sounds good to me. I was on the fence myself

FLGMwt: 👍 thanks for the update. I brought this up with the team as well and we also landed on the inline noop being preferable.

I want to say again that I appreciate your patience and responsiveness with some of these minor (sometimes pedantic :p) style quibbles. Some context for this is: the team grew recently and with that, we've identified an opportunity to make some shared decisions about coding style to help with consistency and focus changes / reviews on the functional parts of code. This includes strategies for TypeScript and in truth, a number of your contributions have surfaced good style questions that have helped our discussions, decisions, and linting rules! That is to say, yet again, thank you! 🙏

PR #3270, opened 2025-12-05
1048 explore redesign debug mode

jtklein:
Closes MOB-1048

This introduces a copy of the `Explore` component that is returned when the app is in debug mode. Contrary to the ticket description I opted to have the return from the two container that return the `Explore` component and not as an early return in `Explore` itself. I did not have the code at hand when writing the ticket description and I think this way here is a bit clearer. In both container no changes have been made to the actual Explore, only indentation was changed.

Copied over are only the parts that are not re-designed in Figma. The parts that will look different are removed for now, so that we can start from scratch with bringing the new UI in and taking the logic of the existing UI in a new place. The logic is not supposed to massively change with this re-design.

All steps in the Explore re-design will be laid out by follow-up tickets.

### DIFF
--- a/src/components/Explore/ExploreV2.tsx
+++ b/src/components/Explore/ExploreV2.tsx
@@ -37,6 +37,12 @@ enum EXPLORE_VIEW {
   SPECIES = "species"
 }
 
+enum EXPLORE_OBSERVATIONS_LAYOUT {
+  GRID = "grid",
+  LIST = "list",
+  MAP = "map"
+}
+
 interface Props {
   canFetch?: boolean;
   currentExploreView: EXPLORE_VIEW;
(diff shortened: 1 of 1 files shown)


abbeycampbell (approved):
This looks good to me! 

My only thing is we could go ahead and make `ExploreRedesigned` a `tsx` file, but I also understand wanting to keep this PR constrained and only about surfacing the redesigned screen on debug mode.

src/components/Explore/ExploreV2.tsx
FLGMwt, line 1: Would you be open to a different name for this File/Component? From my past experience and in our codebase as well, I think this is the type of file / code unit that tends to stick around longer than expected. My concern is that "Redesigned", though we currently understand that in the context of this redesign effort, might not be clear in the future. I might suggest something like ExploreV2 or Explore[RedesignProjectCodeName] to pin it to a specific effort, in this case [MOB-1048](https://linear.app/inaturalist/issue/MOB-1048) and its related tickets. I know the ideal would be that we'd be able to delete the existing Explore and rename ExploreRedesigned or whatever back to Explore once the effort is done, but there might be a long tail to that.

src/components/Explore/ExploreV2.tsx
```diff
@@ -0,0 +1,180 @@
...
+            accessibilityHint={t( "Opens-location-permission-prompt" )}
+            level="focus"
+            onPress={( ) => requestLocationPermissions()}
+          />
+        </View>
+      );
+    }
+    return (
+      <View className="flex-1">
+        {currentExploreView === "observations" && (
```
FLGMwt: non-blocker, and I know this is from the previous implementation, just taking the opportunity to ask style:

what do we think about these conditional renders since they're mutually exclusive? like in a non-markup equivalent, I might not immediately notice that we'll never see two outputs:

```js
if (currentExploreView === "observations") {
  console.log(1);
}
if (currentExploreView === "species") {
  console.log(2);
}
```

If we're keeping it inline we could do a map, something like:

```js
{
  {
    observations: <ObservationsView ... />,
    species: <SpeciesView ... />.
    ...
  }[currentExploreView]
}

PR #3271, opened 2025-12-05
Mob 963 change slideover menu to be a static page

abbeycampbell:
[MOB-963](https://linear.app/inaturalist/issue/MOB-963/change-slideover-menu-to-be-a-static-page)

- Most of what was in `CustomDrawerContent.tsx` is now our regular `Menu.tsx` screen, with the requisite UI updates.
- I removed the `rootDrawerNavigator` (now a native stack navigator: `RootStackNavigator`) and any lingering code, copy, and libraries related to drawer functionality.

There remain typescript issues in the new Menu file related to reconciling type `User` with `RealmUser | ApiUser` and with navigation (when it can't infer a type, it defaults to never, and we get a type mismatch when we try to pass a route param). These are app-wide issues so I did not try to solve them in this PR. 

One thing that was a bit of a mystery to me was a persistent delay when navigating from `<Menu />` to a menu item screen such as `<About />` and back--there'd be a beat of blankness after the stack pop, even though `<Menu />` did not re-render. Eventually I determined this seemed to be the result of the `<FadeInView />` that wraps many of our components. It seems like, reading [Amanda's note](https://github.com/inaturalist/iNaturalistReactNative/blob/4ca1fcea7e98be16b5aee4dd0119c20c16e29fce/src/navigation/StackNavigators/TabStackNavigator.js#L125), we probably don't want/need this wrapper now that we're on react nav 7. I didn't remove `<FadeInView />` completely in this PR (since I'm not totally sure how we want the fade animation to work across screens in the app, and it's not really the business of this ticket anyways), but I did unwrap our tab screens and apply a fade animation to them via the `Tab.Navigator` which fixed the issue. ~~This also required a flag to switch the animation off for testing, since native animations update state after the test cycle and break tests.~~

### DIFF
--- a/src/navigation/RootStackNavigator.tsx
+++ b/src/navigation/RootStackNavigator.tsx
@@ -6,7 +6,7 @@ import * as React from "react";
 import { useOnboardingShown } from "sharedHelpers/installData";
 
 import BottomTabNavigator from "./BottomTabNavigator";
-import { hideHeader } from "./navigationOptions";
+import { hideHeader, preventSwipeToGoBack } from "./navigationOptions";
 
 const Stack = createNativeStackNavigator( );
 
@@ -18,7 +18,7 @@ const RootStackNavigator = ( ) => {
   const [onboardingShown] = useOnboardingShown( );
 
   return (
-    <Stack.Navigator screenOptions={{ ...hideHeader }}>
+    <Stack.Navigator screenOptions={{ ...hideHeader, ...preventSwipeToGoBack, animation: "none" }}>
       {!onboardingShown
         ? (
           <Stack.Screen
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -231,7 +231,12 @@ const AICamera = ( {
 
   const handleClose = async ( ) => {
     await deleteSentinelFile( sentinelFileName );
-    navigation.goBack( );
+    navigation.navigate( "TabNavigator", {
+      screen: "ObservationsTab",
+      params: {
+        screen: "ObsList"
+      }
+    } );
   };
 
   return (


e2e/sharedFlows/switchPowerMode.js
```diff
@@ -1,17 +1,22 @@
...
-  await waitFor( drawerButton ).toBeVisible().withTimeout( TIMEOUT );
-  await drawerButton.tap( { x: 0, y: 0 } );
-  // Tap the settings drawer menu item
-  const settingsDrawerMenuItem = element( by.id( "settings" ) );
-  await waitFor( settingsDrawerMenuItem ).toBeVisible().withTimeout( TIMEOUT );
-  await settingsDrawerMenuItem.tap();
+  const menuButton = element( by.id( "OPEN_MENU" ) );
+  // Temporarily disable synchronization to work around fade animation timing
+  // issues that can cause Detox to wait indefinitely for animations to complete
+  await device.disableSynchronization();
```
FLGMwt: oooooooooooooooo was this the trickyness?

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const queryClient = useQueryClient( );
+  const currentUser = useCurrentUser( );
+  const { t } = useTranslation( );
+
+  const { isConnected } = useNetInfo( );
+
+  const [showConfirm, setShowConfirm] = useState( false );
+  const [showFeedback, setShowFeedback] = useState( false );
+
+  const menuItems = useMemo( ( ) => {
```
FLGMwt: we can drop the `useMemo` here, we're not calculating anything expensive and react will optimize this out

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,263 @@
...
+    return items;
+  }, [currentUser, isConnected, isDebug, navigation, t] );
+
+  const onSignOut = async ( ) => {
+    await signOut( { realm, clearRealm: true, queryClient } );
+    setShowConfirm( false );
+
+    // TODO might be necessary to restart the app at this point. We just
+    // deleted the realm file on disk, but the RealmProvider may still have a
+    // copy of realm in local state
```
FLGMwt, line 154: Was this meant to be addressed / investigated in this PR? seems a bit spooky

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,263 @@
...
+              onPress={() => {
+                if ( item.navigation ) {
+                  navigation.navigate( "TabNavigator", {
+                    screen: "MenuTab",
+                    params: {
+                      screen: menuItems[key].navigation
+                    }
+                  } );
+                }
+                item.onPress?.();
```
FLGMwt, line 249: Is there any concern about having the onPress _after_ the navigate?

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,263 @@
...
+      {showFeedback && (
+        <TextInputSheet
+          buttonText={t( "SUBMIT" )}
+          onPressClose={() => setShowFeedback( false )}
+          headerText={t( "FEEDBACK" )}
+          confirm={onSubmitFeedback}
+          description={t( "Thanks-for-using-any-suggestions" )}
+          maxLength={1000}
+        />
+      )}
```
FLGMwt: Is there any worry that these might both be rendered?

package.json
```diff
@@ -65,7 +65,6 @@
     "@react-native-picker/picker": "^2.11.1",
     "@react-native-vector-icons/common": "^12.3.0",
     "@react-navigation/bottom-tabs": "^7.4.6",
-    "@react-navigation/drawer": "^7.5.7",
```
FLGMwt, line 68: 🥳

e2e/sharedFlows/switchPowerMode.js
abbeycampbell: This is the trickyness but my trick did not work unfortunately

src/components/Menu/Menu.tsx
abbeycampbell: This is a comment that I just copied over from `CustomDrawerContent.tsx`. Logging out seems to work just fine, but I left it in because I really didn't change anything about the code here, so it may still be relevant/provide a useful clue to someone someday.

src/components/Menu/Menu.tsx
abbeycampbell: I don't think so, since no menu item has both `navigation` and an `onPress`

src/components/Menu/Menu.tsx
abbeycampbell: I don't think so, since both booleans are each only set in one place: the onPress of a specific menu item. Closing the modal or dismissing the bottom sheet sets it's own state back to false, so it shouldn't be possible that these ever both become true at once.

abbeycampbell:
Only after Johannes gives me permission to move on do I figure it out 🙃 (kudos to Ryan for devoting some time to troubleshooting as well!)

My issue was that I was not understanding the e2e test output. It wasn't hanging on the animation like Jest did, I had just forgotten to surface the testID in `MenuItem`, so the test just could not find the "Settings" menu option at all. 

To round up all my back-and-forth: This PR _does_ remove the fade wrapper from Bottom Tabs in favor of a native fade animation. This does away with any navigation delays in the Menu screen.

src/components/Menu/Menu.tsx
FLGMwt: Ah, I see, maybe we can enforce them as mutually exclusive at the `Props` type level then? I'm just considering a future dev experiencing undefined behavior.

src/components/Menu/Menu.tsx
FLGMwt: Oh got it! I promise I know more than one word (mutex) but: I think if we have the expectation of mutual exclusivity, this might be a good opportunity to turn that concern into an enum-y status var rather than separate bools. It seems similar to the earlier concern of: we might not have the possibility for now, but we can guard against future uses that have undefined behavior.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,273 @@
...
+      navigation: "Settings",
+      icon: "gear"
+    },
+
+    feedback: {
+      label: t( "FEEDBACK" ),
+      icon: "feedback",
+      onPress: () => {
+        if ( isConnected ) {
+          setModalState( "provideFeedback" );
```
jtklein: I'd prefer a single source of truth for these strings. Personally, I would use an enum here because the modal states are mutually exclusive.

src/components/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,34 @@
...
+  item: MenuOption;
+  onPress: ( ) => void;
+} ) => (
+  <Pressable
+    testID={item.testID}
+    className={classNames(
+      item.isLogout
+        ? "opacity-50"
+        : "",
+      "flex-row items-center pl-10 py-5 border-b border-lightGray"
```
jtklein: Instead of py-5 which is 20px the designs have 22px for py here (`py-[22px]`).

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,273 @@
...
+        {/* Menu Items */}
+        <View>
+          {Object.entries( menuItems ).map( ( [key, item] ) => (
+            <MenuItem
+              key={key}
+              item={item}
+              onPress={() => {
+                if ( item.navigation ) {
+                  navigation.navigate( "TabNavigator", {
+                    screen: "MenuTab",
```
jtklein, line 243: This is a change in function here, because it used to navigate to `ObservationsTab` here. This is not mentioned in the ticket, so just to make sure this is not an oversight. You did intend to change the navigation here of the menu items to stay in the new Menu tab? I do think this is more suitable now to stay in the Menu tab because the menu is the entry screen.
But we could also make this PR without this change and ask for more directions product-wise.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,273 @@
...
+      ? "DEFAULT:"
+      : "ADVANCED:";
+    feedbackLogger.info( mode, text );
+    Alert.alert( t( "Feedback-Submitted" ), t( "Thank-you-for-sharing-your-feedback" ) );
+    setModalState( null );
+    return true;
+  }, [isConnected, t] );
+
+  return (
+    <View className="bg-white h-full">
```
jtklein: The drawer used to be scrollable. I never had a phone so small that there were items beyond the screen, so probably this is okay.

src/i18n/strings.ftl
```diff
@@ -751,6 +751,8 @@ Navigates-to-AI-camera = Navigates to AI camera
 Navigates-to-bulk-importer = Navigates to bulk importer
 Navigates-to-camera = Navigates to camera
 Navigates-to-explore = Navigates to explore
+Navigates-to-log-in-screen = Navigates to log in screen
```
jtklein, line 754: It is always good to provide context to strings in this file, especially for strings that are not directly user-facing, so everything a11y related. E.g.:
`# Accessibility hint for ...` one line above.

These comments appear in Crowdin were we crowdsource our translations. So, translators will see this context as well.

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,273 @@
...
+        <Pressable
+          testID="menu-header"
+          accessible
+          accessibilityRole="button"
+          accessibilityHint={
+            currentUser
+              ? t( "Navigates-to-user-profile" )
+              : t( "Navigates-to-log-in-screen" )
+          }
+          className="px-[26px] pt-[68px] pb-[31px] border-b border-lightGray"
```
jtklein, line 189: The padding top is the correct pixel value, however, the entire screen does not take insets into account. So, the distance of 68px should be to the inset and not top of screen.
Anyways, I think this is not blocking but we should make a new ticket to make this new screen insets aware after this PR is merged.

src/navigation/RootStackNavigator.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+// DEVELOPERS: do you need to add any screens here? This is the RootStack.
+// All the rest of our screens live in:
+// NoBottomTabStackNavigator, TabStackNavigator, OnboardingStackNavigator, or LoginStackNavigator
+
+const RootStackNavigator = ( ) => {
+  const [onboardingShown] = useOnboardingShown( );
+
+  return (
+    <Stack.Navigator screenOptions={{ headerShown: false }}>
```
jtklein: Do you want to import `hideHeader` here, as it was before?

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
```diff
@@ -57,14 +58,16 @@ const CustomTabBarContainer: React.FC<Props> = ( { navigation, state } ) => {
   const tabs: TabConfig[] = useMemo( ( ) => ( [
     {
       icon: "hamburger-menu",
-      testID: DRAWER_ID,
+      testID: MENU_ID,
```
jtklein: Why not use `SCREEN_NAME_MENU` here like the other tabs?

tests/jest.setup.js
```diff
@@ -175,6 +175,30 @@ jest.mock( "components/Camera/FadeInOutView", () => {
   return jest.fn( ( ) => React.createElement( View, null ) );
 } );
 
+// Mock @react-navigation/bottom-tabs to disable animations in Jest tests
+// This prevents the act() warnings caused by fade animations triggering state updates
+jest.mock( "@react-navigation/bottom-tabs", () => {
```
jtklein, line 180: That looks awfully complicated but I am okay if it works for us. Commenting this out makes one test fail so something is needed here.
react-navigation recommends to not mock but use fake timers I think. Have you tried something along the lines of this? https://reactnavigation.org/docs/testing/#navigation-between-tabs

src/components/Menu/Menu.tsx
abbeycampbell: This was intentional; I just made an assumption here that we'd want to go back to the Menu screen, but I should've asked to be sure. 
Going to keep this behavior as discussed in stand-up today. I also added a note about it in the ticket.

src/components/Menu/Menu.tsx
abbeycampbell: I tried this on an 12 mini which is one of the smallest iPhones someone could feasibly have, and it looked fine which is why I decided to remove it. However, with the proper padding at the top of the screen and the slightly beefier padding on the menu items, debug mode will be slightly cut off on this size of device (Still able to tap/interact with it though). I'm going to add back a ScrollView just to catch any edge cases.

src/components/Menu/Menu.tsx
abbeycampbell: oh got it! That's actually how I built it on my first pass, but then I thought maybe all that space was meant to accommodate the notch. Happy to just go ahead and add back now.

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
abbeycampbell: I was just following the pattern we already had, where there is a separate constant that I guess is more declarative about what the test is doing? But there's not reason we can't use `SCREEN_NAME_MENU`, and it lets us lose a duplicative id.

src/components/Menu/MenuItem.tsx
abbeycampbell: this is just a minor design thing that irks me slightly but for cases like this where there is a built-in tailwind class that is very close to what we want (2 pixels of difference), I kinda wish the designs would just default to the standard, unless there's a specific reason not to.

tests/jest.setup.js
abbeycampbell: I haven't. Since this works for now, I'm going to punt investigating this to a different ticket.

abbeycampbell:
> 
> However, I was seeing some big changes in behaviour of the navigation. E.g. when starting an observation by: 1.) + button 2.) "Take photos" I can see the camera screen navigating into view with an animation to slide in from the right whereas on current main this screen fades into existence without sliding. Also when pressing X on the camera the navigation back to MyObs slides in from the right which feels wrong at this point. This PR fixes this specific issue: #3282, it seems to me the default for Stack is slide-in whereas the default for drawer was none maybe.
> 
> There is one more requested change though: when pressing X in the AI camera the app does nothing with a dev warning of 'GO_BACK' not being handled. So, that is a fix we would need in this PR. Maybe also a difference in Drawer vs Stack defaults because this would be a navigation event between the root "screen"s.
> 

Thanks for catching these issues! I opted not to merge your PR since I'd already touched that file and needed to pull in `preventSwipeToGoBack` as well. That issue should be resolved. 

The issue with the closing the AI camera is resolved as well, by navigating to `'ObsList'` inside the `TabNavigator`. This feels a little bit slow, but it preserves the original behavior of always returning to `ObsList` regardless of what screen you opened the camera from; it's also how we do it for the sound recorder. 

I don't think it's ideal though that the `NoBottomTabs` stack (taking/uploading photos, sound recording) is a sibling of The main `BottomTabNavigator`. Do you know if there is particular reason we needed to do it this way? 
It seems like we could use stack [groups](https://reactnavigation.org/docs/group/) to handle these screens like modals, inside `TabStackNavigator`. I think this would also let us get rid of the `SharedStackScreens` group--all the app screens become siblings in one main navigator.

tests/jest.setup.js
abbeycampbell: [MOB-1067](https://linear.app/inaturalist/issue/MOB-1067/investigate-alternative-way-to-fix-failing-test-related-to-fade)

src/components/Menu/MenuItem.tsx
jtklein: I am with you on this one, but our current process flow is that the design files are the source of truth. So, we'd need to communicate back to the design files first.

PR #3275, opened 2025-12-08
Resolve all type errors in `MapView` component.

frewsxcv:


src/components/Explore/MapView.tsx
```diff
@@ -120,7 +122,11 @@ const MapView = ( {
       // but per user feedback, we want to show users a more zoomed in view
       // when they're looking at NEARBY view
       if ( nearbyRegion.latitude !== undefined && nearbyRegion.longitude !== undefined ) {
-        mapRef.current.animateToRegion( nearbyRegion );
+        mapRef.current.animateToRegion( {
+          ...nearbyRegion,
+          latitude: nearbyRegion.latitude,
+          longitude: nearbyRegion.longitude
+        } );
```
frewsxcv, line 129:
```
Type '{ latitude: number | undefined; longitude: number | undefined; latitudeDelta: number; longitudeDelta: number; }' is not assignable to type 'Region'.
  Type '{ latitude: number | undefined; longitude: number | undefined; latitudeDelta: number; longitudeDelta: number; }' is not assignable to type 'LatLng'.
    Types of property 'latitude' are incompatible.
      Type 'number | undefined' is not assignable to type 'number'.
        Type 'undefined' is not assignable to type 'number'.ts(2322)
```

The type of `nearbyRegion.latitude` is getting narrowed with the `nearbyRegion.latitude !== undefined` check above, but TypeScript isn't smart enough to do that narrowing in the `nearbyRegion` type. So we need to specify those fields with the narrowed values here (and later in the file). There shouldn't be any functional change here in runtime.

src/components/Explore/MapView.tsx
```diff
@@ -146,6 +152,7 @@ const MapView = ( {
   const handleRedoSearch = async ( ) => {
     setShowRedoSearchButton( false );
     const currentBounds = await mapRef?.current?.getMapBoundaries( );
+    if ( !currentBounds ) { return; }
```
frewsxcv, line 155: `currentBounds` can be `undefined`, so just return early. Presumably the `dispatch` calls below wouldn't work anyways if it was `undefined`.

PR #3276, opened 2025-12-08
MOB-722 match components ts conversion

sepeterson:
Supporting [MOB-722](https://linear.app/inaturalist/issue/MOB-722/testing-match-screen-unit-tests)

I attempted to keep the scope of changes to components under the Match directory, with 1 or 2 exceptions.

Additionally, I attempted to minimize functional changes. There were a few spots where ts revealed we should have been doing type conversions or checking for truthiness that I went ahead with.

Quite a few type errors remain but it's a step in the right direction.

### DIFF
--- a/src/components/Match/PhotosSection.tsx
+++ b/src/components/Match/PhotosSection.tsx
@@ -33,7 +33,7 @@ const PhotosSection = ( {
 
   const localTaxonPhotos = taxon?.taxonPhotos;
   const observationPhoto = obsPhotos?.[0]?.photo?.url
-  || obsPhotos?.[0]?.photo?.localFilePath;
+  || Photo.getLocalPhotoUri( obsPhotos?.[0]?.photo?.localFilePath );
 
   const taxonPhotos = compact(
     localTaxonPhotos


src/components/Match/MatchContainer.tsx
```diff
@@ -35,13 +38,44 @@ import tryToReplaceWithLocalTaxon from "./helpers/tryToReplaceWithLocalTaxon";
...
+  shouldUseEvidenceLocation: boolean;
+  orderedSuggestions: ApiSuggestion[];
+}
+
+type ActionType =
+  | { type: "SET_UPLOAD_PARAMS"; scoreImageParams: ImageParamsType }
+  | { type: "SET_ONLINE_FETCH_STATUS"; onlineFetchStatus: string }
+  | { type: "SET_OFFLINE_FETCH_STATUS"; offlineFetchStatus: string }
+  | { type: "SET_LOCATION"; scoreImageParams: ImageParamsType; shouldUseEvidenceLocation: boolean }
+  | { type: "ORDER_SUGGESTIONS"; orderedSuggestions: ApiSuggestion[] };
```
FLGMwt: yaaaaaas, I love a discriminated union, especially for actions

src/components/Match/MatchContainer.tsx
```diff
@@ -225,7 +262,7 @@ const MatchContainer = ( ) => {
   const [needLocation, setNeedLocation] = useState(
     shouldFetchObservationLocation( currentObservation )
   );
-  const shouldFetchLocation = hasPermissions && needLocation;
+  const shouldFetchLocation = !!( hasPermissions && needLocation );
```
FLGMwt, line 265: 👍 I'm a fan of this instead of changing to looser types for downstream falsys

src/components/Match/MatchHeader.tsx
```diff
@@ -36,10 +32,12 @@ const MatchHeader = ( { topSuggestion, hideObservationStatus }: Props ) => {
...
-    if ( confidence >= 93 ) {
-      confidenceType = "observed";
-    } else if ( confidence >= 50 && confidence < 93 ) {
-      confidenceType = "likely_observed";
+    if ( confidence ) {
+      if ( confidence >= 93 ) {
+        confidenceType = "observed";
+      } else if ( confidence >= 50 && confidence < 93 ) {
+        confidenceType = "likely_observed";
+      }
```
FLGMwt, line 45: Do you have an idea what drives `93` and `50`? If it's known, maybe we could move these to descriptively named file-level constants?

sepeterson: I don't know why those numbers specifically, but I threw some constants in just for fun

jtklein: The numbers are from here (but the text in Figma is outdated it seems)
https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=23127-92653&t=FKfMR44IGsHayrfF-4

FLGMwt: awesome, thanks y'all! Totally makes sense to name them after the confidenceType enums they're driving

src/components/Match/MatchContainer.tsx
```diff
@@ -50,7 +84,7 @@ const initialState = {
   orderedSuggestions: []
 };
 
-const reducer = ( state, action ) => {
+const reducer = ( state: StateType, action: ActionType ): StateType => {
```
jtklein: I think it's okay to call this State and Action, like [here](https://github.com/inaturalist/iNaturalistReactNative/blob/aab6bf6aa163a859b59c3a32ec8f29d12d151b0c/src/providers/ExploreContext.tsx#L389-L390) if you want.

src/components/Match/MatchContainer.tsx
```diff
@@ -35,13 +38,44 @@ import tryToReplaceWithLocalTaxon from "./helpers/tryToReplaceWithLocalTaxon";
...
+}
+
+interface NavParams {
+  id?: number | string;
+  firstPhotoID?: number | string;
+  representativePhoto?: { isRepresentativeButOtherTaxon?: boolean; id?: number | string };
+}
+
+interface StateType {
+  onlineFetchStatus: string;
```
jtklein: More like mutually exclusive states in the form of a few strings and not the general string type. So, maybe you want to refactor to an enum.

src/components/Match/MatchContainer.tsx
```diff
@@ -35,13 +38,44 @@ import tryToReplaceWithLocalTaxon from "./helpers/tryToReplaceWithLocalTaxon";
...
+  offlineFetchStatus: string;
```
jtklein: More like mutually exclusive states in the form of a few strings and not the general string type. So, maybe you want to refactor to an enum.

src/components/Match/MatchContainer.tsx
```diff
@@ -35,13 +38,44 @@ import tryToReplaceWithLocalTaxon from "./helpers/tryToReplaceWithLocalTaxon";
 import Match from "./Match";
 import PreMatchLoadingScreen from "./PreMatchLoadingScreen";
 
-const setQueryKey = ( selectedPhotoUri, shouldUseEvidenceLocation ) => [
+interface ImageParamsType {
+  uri?: string;
+  image: {
+    uri: string;
```
jtklein, line 39: Also, `name: string; type: string;` I think within image.

src/components/Match/MatchContainer.tsx
```diff
@@ -35,13 +38,44 @@ import tryToReplaceWithLocalTaxon from "./helpers/tryToReplaceWithLocalTaxon";
...
+  representativePhoto?: { isRepresentativeButOtherTaxon?: boolean; id?: number | string };
+}
+
+interface StateType {
+  onlineFetchStatus: string;
+  offlineFetchStatus: string;
+  scoreImageParams: ImageParamsType | null;
+  queryKey: ( string | { shouldUseEvidenceLocation: boolean } )[];
+  shouldUseEvidenceLocation: boolean;
+  orderedSuggestions: ApiSuggestion[];
```
jtklein, line 59: As first approximation this is okay. Technically, either offline cv or online vs succeeds in sending suggestions out of the useSuggestion, so the type here would be either `ApiSuggestion` or OfflineSuggestion (not implemented yet). But I can see @FLGMwt implementing TS for the useSuggestions hook in another PR. So, I would say we can leave this here and consolidate after both PRs have been merged. (https://github.com/inaturalist/iNaturalistReactNative/pull/3257)

src/components/Match/MatchContainer.tsx
```diff
@@ -117,9 +151,12 @@ const MatchContainer = ( ) => {
 
   const evidenceHasLocation = !!currentObservation?.latitude;
 
-  const [topSuggestion, setTopSuggestion] = useState( );
-  const [iconicTaxon, setIconicTaxon] = useState( );
-  const [currentUserLocation, setCurrentUserLocation] = useState( null );
+  const [topSuggestion, setTopSuggestion] = useState<ApiSuggestion>( );
+  const [iconicTaxon, setIconicTaxon] = useState<ApiTaxon>( );
```
jtklein: Should be `RealmTaxon` according to `IconicSuggestionsScroll`.

src/components/Match/IconicSuggestions/IconicSuggestionsScroll.tsx
```diff
@@ -42,7 +48,7 @@ const IconicSuggestionsScroll = ( {
         ListHeaderComponent={renderHeader}
         horizontal
         renderItem={renderItem}
-        keyExtractor={item => item?.id}
+        keyExtractor={( item: ApiTaxon ) => `${item?.id}`}
         data={iconicTaxa}
```
jtklein, line 52: Data source is Realm for these items, so instead of `ApiTaxon` we should use `RealmTaxon` here I think. (And in higher order Props)

src/components/Match/SaveDiscardButtons.tsx
```diff
@@ -13,12 +12,12 @@ const DROP_SHADOW = getShadow( {
 } );
 
 type Props = {
-  handlePress: Function
+  handlePress: ( action: string ) => void
```
jtklein: `action: "save" | "discard"`, here and in higher order Props

src/components/Match/MatchContainer.tsx
```diff
@@ -451,17 +488,18 @@ const MatchContainer = ( ) => {
...
-      navParams.firstPhotoID = photo.id;
+      navParams.firstPhotoID = photo?.id;
     } else {
       navParams.representativePhoto = photo;
     }
     navigation.push( "TaxonDetails", navParams );
   };
 
-  const handleSaveOrDiscardPress = async action => {
+  const handleSaveOrDiscardPress = async ( action: string ) => {
```
jtklein: "save" | "discard"

src/components/Match/Match.tsx
```diff
@@ -25,19 +28,19 @@ export const matchCardClassBottom
   = "rounded-b-2xl border-lightGray border-[2px] pb-3 border-t-0 -mt-0.5 mb-[30px]";
 
 type Props = {
-  observation: Object,
-  obsPhotos: Array<Object>,
-  handleSaveOrDiscardPress: ( ) => void,
-  navToTaxonDetails: ( ) => void,
+  observation: RealmObservation,
+  obsPhotos: RealmObservationPhoto[],
+  handleSaveOrDiscardPress: ( action: string ) => void,
```
jtklein: "save" | "discard"

src/components/Match/PreMatchLoadingScreen.tsx
```diff
@@ -16,7 +16,7 @@ const fade = ( value: number ) => ( {
 
 interface Props {
   isLoading: boolean;
-  onSkip: ( ) => void;
+  onSkip?: ( ) => void;
```
jtklein, line 20: Looking at this component I see that we actually never prop this in anywhere. So, we probably should remove this vestige `onSkip` prop and whatever depends on it here.
Does not need to be this PR but if not please make a new ticket for this removal.

src/components/Match/PhotosSection.tsx
```diff
@@ -9,16 +8,16 @@ import {
...
 
 type Props = {
-  representativePhoto?: Object,
-  taxon: Object,
-  obsPhotos: Array<Object>,
-  navToTaxonDetails: ( photo: Object ) => void,
+  representativePhoto: ApiPhoto,
+  taxon?: ApiTaxon | RealmTaxon,
+  obsPhotos: RealmObservationPhoto[],
+  navToTaxonDetails: ( photo: ApiPhoto ) => void,
```
jtklein: ApiPhoto | RealmPhoto, also in higher order Props

src/components/Match/Match.tsx
```diff
@@ -25,19 +28,19 @@ export const matchCardClassBottom
...
+  navToTaxonDetails: ( photo?: ApiPhoto ) => void,
```
jtklein: ApiPhoto | RealmPhoto

src/components/Match/MatchContainer.tsx
```diff
@@ -451,17 +488,18 @@ const MatchContainer = ( ) => {
   const otherSuggestions = orderedSuggestions
     .filter( suggestion => suggestion.taxon.id !== taxonId );
 
-  const navToTaxonDetails = photo => {
-    const navParams = { id: taxonId };
+  const navToTaxonDetails
+  = ( photo?: { isRepresentativeButOtherTaxon?: boolean; id?: number } ) => {
```
jtklein: `photo?: ApiPhoto | RealmPhoto`, according to `PhotosSection`

src/components/Match/Match.tsx
```diff
@@ -25,19 +28,19 @@ export const matchCardClassBottom
...
+  topSuggestion?: ApiSuggestion,
+  otherSuggestions: ApiSuggestion[],
   suggestionsLoading: boolean,
-  onSuggestionChosen: ( ) => void,
-  scrollRef: Object,
-  iconicTaxon: Object,
-  setIconicTaxon: ( ) => void
+  onSuggestionChosen: ( suggestion: ApiSuggestion ) => void,
+  scrollRef: React.RefObject<ScrollView | null>,
+  iconicTaxon?: ApiTaxon,
```
jtklein: RealmTaxon

src/components/Match/Match.tsx
```diff
@@ -25,19 +28,19 @@ export const matchCardClassBottom
...
+  setIconicTaxon: ( taxon: ApiTaxon ) => void
```
jtklein: RealmTaxon

src/components/Match/PhotosSection.tsx
```diff
@@ -9,16 +8,16 @@ import {
...
 import Photo from "realmModels/Photo";
+import type { RealmObservationPhoto, RealmTaxon } from "realmModels/types";
 import getImageDimensions from "sharedHelpers/getImageDimensions";
 
 type Props = {
-  representativePhoto?: Object,
-  taxon: Object,
-  obsPhotos: Array<Object>,
-  navToTaxonDetails: ( photo: Object ) => void,
+  representativePhoto: ApiPhoto,
```
jtklein: Can be `undefined`

src/components/Match/Match.tsx
```diff
@@ -80,7 +83,6 @@ const Match = ( {
             }
           </View>
           <PhotosSection
-            representativePhoto={topSuggestion?.taxon?.representative_photo}
```
jtklein, line 83: Good spot! In those two special cases there will be no taxon photos displayed.

jtklein (approved):
Overall, only few minor changes. One mention is that useSuggestions returns either online or offline suggestions. So, the type is not only `ApiSuggestion` almost wherever you use it here. But @FLGMwt is also working on migrating [useSuggestions to TS](https://github.com/inaturalist/iNaturalistReactNative/pull/3257), so I would not request those changes here. So, I would say we don't need to increase the scope of this here, but we do need to consolidate at one point when the other PR is merged. (The redundancy here is not surprising as the Match screen is basically a revamped fork of the SuggestionsScreen, which reminds me we might need some more tickets for the consolidation of these two screens in general now that the Match screen is supposed to stay).

Approved with minor changes I would call this. So, whenever those minor changes are made feel free to merge in your own time and if you could please also take care of the follow-up tickets mentioned that would be splendid.

src/components/Match/PreMatchLoadingScreen.tsx
sepeterson: https://linear.app/inaturalist/issue/MOB-1068/remove-unused-onskip-prop

PR #3277, opened 2025-12-09
Migrate `checkCamelAndSnakeCase` to TypeScript.

frewsxcv:


src/components/ObsDetails/helpers/checkCamelAndSnakeCase.ts
```diff
@@ -1,6 +1,8 @@
-// @flow
-
-const checkCamelAndSnakeCase = ( object: Object, camelCaseKey: string ): ?string => {
+const checkCamelAndSnakeCase = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  object: Record<string, any>,
```
frewsxcv, line 3: There's definitely an opportunity for stronger typing here, but that'd be a much larger change.

FLGMwt: haha, yeah, we could do some _nasty_ stuff w/ template literal types https://stackoverflow.com/a/65015868/31670731 but I agree, let's not do that here ; D

I'd personally rather we remove the need for this and normalize snake case objects before they pass the API layer, but that's again a broader concern.

src/components/ObsDetails/helpers/checkCamelAndSnakeCase.ts
```diff
@@ -1,6 +1,8 @@
...
+  camelCaseKey: string
+): string|undefined => {
```
FLGMwt: This should not be string. We're getting _some_ key that exists on `Record<string, any>` and using that to index from it, so this should be any.

frewsxcv: Sounds good, addressed in https://github.com/inaturalist/iNaturalistReactNative/pull/3277/commits/27f3d75202b482b17cf4e021bdaf2f355a77b7d5

For context I had `string|undefined` here because that was the previous typing, but I completely understand why that doesn't make sense for this function

FLGMwt: fair enough, makes sense w/ the old type. but thanks for addressing that! LGTM

PR #3279, opened 2025-12-09
Migrate `DataQualityAssessment` component to TypeScript.

frewsxcv:


src/components/ObsDetails/DataQualityAssessment.tsx
```diff
@@ -40,20 +38,22 @@ const titleDescription = option => {
       return t( "Data-quality-assessment-description-casual" );
   }
 };
-type Props = {
-  checkTest: Function,
-  ifMajorityAgree: Function,
+interface Props {
+  checkTest: ( metric: string ) => boolean | null,
```
jtklein: checkTest does only return boolean if I see this correctly here: https://github.com/frewsxcv/iNaturalistReactNative/blob/66a2bec46462bb5404f340496650a0e5b25ccce0/src/components/ObsDetails/DQAContainer.js#L216-L217

frewsxcv: Good catch, addressed in https://github.com/inaturalist/iNaturalistReactNative/pull/3279/commits/6c0a7acffcc878d6135be188c8afda7ed431bc18

frewsxcv:
> In general, we'd like to try to migrate from tip nodes of the "DOM" tree upwards. SO, in this case migrating DQAVoteButtons first and then bubbling up the types later would have been our preferred approach.

Thanks for letting me know. I'll keep this in mind going forward.

PR #3280, opened 2025-12-10
Remove only type error in `Button` component.

frewsxcv:
`className` is not a prop on this component.


src/components/SharedComponents/Buttons/Button.tsx
```diff
@@ -238,7 +238,6 @@ const Button = ( {
       {loading && (
         <ActivityIndicator
           size={18}
-          className="mr-3 absolute right-0"
```
FLGMwt, line 241: This makes me nervous that we _wanted_ this style at some point, but now I'm seeing that we don't actually have any usages of `<Button />` that use the loading prop except for the style library https://github.com/inaturalist/iNaturalistReactNative/blob/8b0b9e9d3be9fc837a6f936c41aaf9ce315ad081/src/components/Developer/UiLibrary/Buttons.js#L69

What do you think @jtklein, can we just remove the loading prop & use entirely?

jtklein: That doesn't sound right. At least in the login form we have a loading button: https://github.com/inaturalist/iNaturalistReactNative/blob/e9ee6752e4d3265bec32a6e5c6545c4fef945a5e/src/components/LoginSignUp/LoginForm.tsx#L289-L290
And the indicator is in the correct spot, I think, from just looking at the UI. So, in other words the className prop works only TS doesn't know it exists.

frewsxcv:
The underlying component doesn't actually read the className value, right?
So it's not just a TypeScript difference

src/components/SharedComponents/Buttons/Button.tsx
frewsxcv: The underlying component doesn't actually read the className value, right? So it's not just a TypeScript difference

https://github.com/inaturalist/iNaturalistReactNative/blob/8f97ae197f3c79d2350ed13b1c8f3a83da94a999/src/components/SharedComponents/ActivityIndicator.tsx#L17-L35

frewsxcv: This also shows up in https://github.com/inaturalist/iNaturalistReactNative/pull/3284

jtklein: First, it clearly works if you compare main and your branch here:
Main:
FF9CBDB9-2F46-442A-B743-1972AE461E42_4_5005_c
Branch:
B7CE9FFE-216A-4767-9EA8-8553BD241CC8_4_5005_c

jtklein: From the [docs](https://www.nativewind.dev/v5/guides/third-party-components):
"Nativewind works by passing the className prop to the core React Native components."
And the `ActivityIndicator` from "react-native-paper" does use props spreading to the View component as well.
So, I think the issue is rather somewhere else, that our TS setup doesn't know how nativewind works. For example, all the components exported from `styledComponents` after we use the outdated `styled()`wrapper accept a typed className prop.

frewsxcv: I think I figured it out. Like you said the underlying react-native-paper `ActivityIndicator` _does_ use prop spreading, but the iNaturalist `ActivityIndicator` does _not_, so something else is responsible.

The nativewind project includes a Babel plugin [which iNaturalist uses](https://github.com/inaturalist/iNaturalistReactNative/blob/1e38e21521dfbdb156f2df9a24ad6468f3d8f861/babel.config.js#L7). The way that plugin works is it transforms [plain React components](https://github.com/nativewind/nativewind/blob/e9e5f6d3ca2b84e935e1cad3c5b211e09b81e2aa/packages/nativewind/__tests__/babel/content-directory/code.tsx) into those same components [wrapped by `StyledComponent`](https://github.com/nativewind/nativewind/blob/nativewind%402.0.11/packages/nativewind/__tests__/babel/content-directory/output.tsx)s, which include a `className` prop. And the `StyledComponent` [wraps the underlying component in `styled(...)`](https://github.com/nativewind/nativewind/blob/nativewind%402.0.11/packages/nativewind/src/styled-component.tsx#L10), like you mentioned. 

So effectively, every single component in this iNaturalist project is being wrapped by Nativewind's `styled` function, even if we don't do it explicitly. Which is very tricky! In the future we may want to consider (somehow) silencing all the "`className` is not a valid prop" TypeScript errors.

Given all of this, I will mark this pull request as draft and when I get the time I will undo the `className` removals.

PR #3281, opened 2025-12-10
MOB-1039 apply member delimiter style and autofix

FLGMwt:
This adds the `@stylistic/member-delimiter-style` eslint rule (https://eslint.style/rules/member-delimiter-style) with the default settings, enforcing semicolon `;` as the property delimiter for all TS interface and types. It applies `eslint --fix` to fix all existing violations.


FLGMwt:
locally merged main into this before merging PR expecting new `fix` results, but there were none! since that's not necessary to reflect latest main, going ahead w/ merge.

PR #3283, opened 2025-12-11
Add CI step to verify there are no lockfile changes after `npm install`.

frewsxcv:


frewsxcv:
Hmm running `npm install` on `main` for me results in lockfile changes. I opened this to see if that would be the case in the CI environment too, but that does not appear to be the case. Must be due to mismatched npm versions? In any case, closing this.

PR #3284, opened 2025-12-11
Migrate `OverviewCounts` to TypeScript.

frewsxcv:


src/components/SharedComponents/OverviewCounts.tsx
```diff
@@ -71,14 +78,14 @@ const CountPressable = ( {
     </View>
     {typeof count === "number"
       ? <Body2 className="mt-2">{t( "Intl-number", { val: count } )}</Body2>
-      : <ActivityIndicator className="mt-2" size={20} />}
```
frewsxcv, line 74: See the convo here: https://github.com/inaturalist/iNaturalistReactNative/pull/3280#discussion_r2607469256

frewsxcv: https://github.com/inaturalist/iNaturalistReactNative/pull/3280#discussion_r2614242234

frewsxcv: Restored `className`, as per the above linked discussion https://github.com/inaturalist/iNaturalistReactNative/pull/3284/commits/bd4c65d5cef1c6045987220a99636221a670ab28

src/components/SharedComponents/OverviewCounts.tsx
```diff
@@ -1,32 +1,39 @@
...
-  count: number,
-  icon: string,
-  label: string,
-  onPress?: Function
+interface CountPressableProps {
+  accessibilityLabel: string;
+  count: number;
+  icon: string;
+  label: string;
+  onPress?: PressableProps["onPress"];
```
jtklein, line 36: This is a new pattern to the codebase it seems to me.
We have several places where we propagate onPress down to a Pressable. We do use `( _event: GestureResponderEvent ) => void` in several places but also `() => void` if we are sure the caller does not need `event`.
Is this here preferable overall? Or is it easier to use this intermittently, in the sense of while we are typing from the tips of the tree, so that we don't have to check if the prop can actually be null or undefined? I would assume in the places where we typed it as a function we took care to check that neither null nor undefined is propped in.

frewsxcv: [Indexed access types](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html) aren't a new pattern in the codebase:

Screenshot 2025-12-24 at 6 12 14 AM

I only did it because it's terser than writing out `null | ((event: GestureResponderEvent) => void) | undefined;` here and it's a basic feature of TypeScript so I didn't think there would be any controversy.

frewsxcv: Thanks for the update!

abbeycampbell: My bad, I deleted my comment. Adding back for posterity:

We discussed this internally and decided that while there's nothing wrong with this pattern and it absolutely has it's uses, we think in this specific case it makes the API a bit confusing and the props unnecessarily nullable.
We're going to opt for () => void instead.
This is a pretty minor difference, so in the interest of getting this PR merged, I'm just going to do so and follow-up with a PR that reflects the slack discussion.

PR #3285, opened 2025-12-11
Validate `INatIcon` name (or alias) with stronger typing.

frewsxcv:
The main benefit of the new typing is it will validate that the icon with the provided name actually exists, rather than finding out at runtime.

# Before

Screenshot 2025-12-11 at 1 18 22 PM

# After

Screenshot 2025-12-11 at 1 18 14 PM

### DIFF
--- a/src/components/SharedComponents/INatIcon/index.tsx
+++ b/src/components/SharedComponents/INatIcon/index.tsx
@@ -7,15 +7,6 @@ import colors from "styles/tailwindColors";
 import type { INatIconName } from "./INatIcon";
 import Icon from "./INatIcon";
 
-interface Props {
-  testID?: string;
-  name: string;
-  color?: string;
-  size?: number;
-  dropShadow?: boolean;
-  isDarkModeEnabled?: boolean;
-}
-
 // Most of these are names for these icons used in design mapped to more
 // consistent and deduped filenames. We might also put aliases of convenience
 // here, e.g. "speech" and "chat" might both map to "comments" if we find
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/INatIcon/index.tsx
```diff
@@ -4,30 +4,19 @@ import {
...
 // here, e.g. "speech" and "chat" might both map to "comments" if we find
 // ourselves forgetting the name "comments"
-const ALIASES: Aliases = {
+const aliasMap = {
   addevidence: "plus",
   "evidence-add": "plus-bold",
   addid: "id-agree",
   back: "chevron-left",
-  captive: "pot",
+  captive: "pot" as INatIconNames, // Casting because "pot" does not exist in the glyphmap.
```
frewsxcv: Because `"pot"` and `"globe"` don't appear in the glyphmap, it makes me wonder if these aliases are currently broken. Yay stronger typing for finding this.

src/components/SharedComponents/INatIcon/index.tsx
```diff
@@ -4,30 +4,19 @@ import {
 } from "react-native";
 import colors from "styles/tailwindColors";
 
+import type { INatIconNames } from "./INatIcon";
 import Icon from "./INatIcon";
 
-interface Props {
```
frewsxcv, line 9: Moved to later in the file now that it references the aliases map typing

src/components/SharedComponents/INatIcon/INatIcon.ts
```diff
@@ -2,6 +2,8 @@ import { createIconSet } from "@react-native-vector-icons/common";
 
 import glyphmap from "./glyphmap.json";
 
+export type INatIconName = keyof typeof glyphmap;
```
frewsxcv, line 5: For reference:

Image

src/components/SharedComponents/INatIcon/index.tsx
```diff
@@ -16,16 +17,17 @@ interface Props {
 }
 
 type Aliases = Record<string, string>;
+
```
frewsxcv:
```suggestion
```

PR #3289, opened 2025-12-12
Use Photo.getLocalPhotoUri helper to display local photo

jtklein:
Closes MOB-1064

We introduced this helper function a while ago because on iOS between app updates the folder names for cached photos can change. This was not an issue on the Match screen because you couldn't access it for an observation that was saved with a previous app version. But it is required here now.

### DIFF
--- a/src/components/Match/PhotosSection.js
+++ b/src/components/Match/PhotosSection.js
@@ -34,7 +34,7 @@ const PhotosSection = ( {
 
   const localTaxonPhotos = taxon?.taxonPhotos;
   const observationPhoto = obsPhotos?.[0]?.photo?.url
-  || obsPhotos?.[0]?.photo?.localFilePath;
+  || Photo.getLocalPhotoUri( obsPhotos?.[0]?.photo?.localFilePath );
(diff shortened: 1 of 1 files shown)


src/components/Match/PhotosSection.js
```diff
@@ -34,7 +34,7 @@ const PhotosSection = ( {
 
   const localTaxonPhotos = taxon?.taxonPhotos;
   const observationPhoto = obsPhotos?.[0]?.photo?.url
-  || obsPhotos?.[0]?.photo?.localFilePath;
+  || Photo.getLocalPhotoUri( obsPhotos?.[0]?.photo?.localFilePath );
```
FLGMwt, line 37: non-blocker for merge, just wanted to know if you knew whether this affects Android any differently

FLGMwt (approved):

src/components/Match/PhotosSection.js
jtklein: I don't remember any issues with this helper on Android.

PR #3291, opened 2025-12-12
Add key for components resulting from a .map

jtklein:
Fixes a dev build warning:

Each child in a list should have a unique "key" prop.
Check the render method of `View`. It was passed a child from AddObsBottomSheet.


jtklein:
E2e tests are passing for me locally.

PR #3294, opened 2025-12-15
(CI) run detox once (leaner flags) then retry once on failure with recordings / verbose logs

FLGMwt:
We regularly have false negative failures on e2e ios, which we manually rerun. We can add an inline retry so we don't have to manually re-run as often. 

We can also limit the trace-level logs and recordings to only running on the retry to minimize the load on the first run. For this reason, I'm adding the retry as an explicit separate step instead of using Detox CLI's retry flag.

### DIFF
--- a/.github/workflows/e2e_ios.yml
+++ b/.github/workflows/e2e_ios.yml
@@ -213,6 +213,7 @@ jobs:
         brew install applesimutils
 
     - name: Ensure servers are running
+      id: status_check
       run: |
         # is rails running?
         curl -I --fail "https://staging.inaturalist.org/ping"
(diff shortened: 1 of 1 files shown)


jtklein (approved):
Great idea.
Does `if: failure()` mean this step will run even if the failure is somewhere higher up? So, if `Ensure servers are running` fails would this run? When we know the servers are down we do not want the retry step to run. So, do we need to only make this run when the first "Run e2e test" fails?
Also, do we need to increase the `timeout-minutes` to accomodate for two test runs? Maybe not because we did not decrease the value when splitting this up into build and test.

FLGMwt:
> Great idea. Does `if: failure()` mean this step will run even if the failure is somewhere higher up? So, if `Ensure servers are running` fails would this run? When we know the servers are down we do not want the retry step to run. So, do we need to only make this run when the first "Run e2e test" fails? Also, do we need to increase the `timeout-minutes` to accomodate for two test runs? Maybe not because we did not decrease the value when splitting this up into build and test.

Good catch on both of those! 

For the timeout, yeah, I suppose it would have made sense to halve that when I split it, so it's funny that that's "fixed" now. I think the main value of those is that we don't go for the full 8 hour default timeout, heh

for the status check: oh, yup, this was also the case for the existing "upload artifacts", but for sure, the extraneous `detox test` would be much more annoying. I've added an id to the status check so we can _explicitly_ not do those two steps if it's the thing that failed.

PR #3297, opened 2025-12-15
Mob 1072 fix issues with loading projects

abbeycampbell:
[MOB-1072](https://linear.app/inaturalist/issue/MOB-1072/fix-issues-with-loading-projects)

Spun off of creative day exploration in 12/2025, but this is really just a bugfix--fixes issues with janky and confusing loading indicators/repeated fetches on the projects screen. I opted not to push up any code related to skeleton UI, since I think we should get product/design input before actually adding that kind of thing, and input from engineers re: technical implementation. However, I did add our standard activity indicator to the initial loading state, which is an improvement over showing a blank screen for a beat.

un-alphabetizing the 'Nearby' tab will be a separate ticket.

### DIFF
--- a/src/components/Projects/hooks/useInfiniteProjectsScroll.ts
+++ b/src/components/Projects/hooks/useInfiniteProjectsScroll.ts
@@ -51,10 +51,12 @@ const useInfiniteProjectsScroll = ( { params: newInputParams, enabled }: object 
     /* eslint-disable consistent-return */
     {
       getNextPageParam: lastPage => {
-        if ( !lastPage || lastPage.results.length < ITEMS_PER_PAGE ) {
-          return undefined;
-        }
-        return lastPage.page + 1;
+        if ( !lastPage ) return undefined;
+        const totalProjectCount = lastPage.total_results;
+        const totalFetchedCount = lastPage.page * ITEMS_PER_PAGE;
+        return totalFetchedCount < totalProjectCount
+          ? lastPage.page + 1
+          : undefined;
       },
       enabled
     }


src/components/Projects/hooks/useInfiniteProjectsScroll.ts
```diff
@@ -45,12 +45,18 @@ const useInfiniteProjectsScroll = ( { params: newInputParams, enabled }: object
...
+    /* eslint-disable consistent-return */
     {
-      getNextPageParam: lastPage => ( lastPage
-        ? lastPage.page + 1
-        : 1 ),
+      getNextPageParam: lastPage => {
+        if ( !lastPage || lastPage.results.length < 20 ) {
+          return undefined;
+        }
+        return lastPage.page + 1;
```
FLGMwt: Is there any issue querying an "extra page" for results in multiples of 20? Like if we ask for "page 3" when there's 40 results. I'm guessing we just get an empty result which this handles, just want to be sure.

src/components/Projects/hooks/useInfiniteProjectsScroll.ts
```diff
@@ -45,12 +45,18 @@ const useInfiniteProjectsScroll = ( { params: newInputParams, enabled }: object
...
       return searchProjects( params, optsWithAuth );
     },
+    // TO DO: we need to properly type queryOptions in useAuthenticatedInfiniteQuery
+    /* eslint-disable consistent-return */
     {
-      getNextPageParam: lastPage => ( lastPage
-        ? lastPage.page + 1
-        : 1 ),
+      getNextPageParam: lastPage => {
+        if ( !lastPage || lastPage.results.length < 20 ) {
```
FLGMwt: non-blocker: This is based on `baseParams.per_page`, right? Can we use that here over the literal `20`? or a file-level constant they both use.

src/components/Projects/hooks/useInfiniteProjectsScroll.ts
abbeycampbell: in that case, we'd make one extra request, get `results: []`, and stop making further requests. I figured that was fine, but you're question prompted me to look at it again and I decided to re-work it to use total_results instead... that should let us stop making requests as soon as the fetched results equals the total.

FLGMwt: oh nice! I was looking at the wrong API docs, I didn't think we had this

PR #3304, opened 2025-12-17
MOB-535: hide gradient for group photo view and tweak icon drop shadows

sepeterson:
[MOB-535](https://linear.app/inaturalist/issue/MOB-535/remove-gradient-from-grid-items-on-group-photos-screen)

- No gradient on group photos
- add shadow to photo count icon
- slightly intensify shadow to more closely resemble designs

<details>
<summary>screenshot</summary>
IMG_2475
</details>

### DIFF
--- a/src/components/ObservationsFlashList/ObsImagePreview.tsx
+++ b/src/components/ObservationsFlashList/ObsImagePreview.tsx
@@ -11,7 +11,9 @@ import colors from "styles/tailwindColors";
 import ObsImage from "./ObsImage";
 
 const ICON_DROP_SHADOW = getShadow( {
-  offsetHeight: 1
+  offsetHeight: 1,
+  shadowOpacity: 1,
+  shadowRadius: 1
 } );
 
 interface Props extends PropsWithChildren {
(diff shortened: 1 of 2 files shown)


src/components/ObservationsFlashList/ObsImagePreview.tsx
```diff
@@ -139,6 +145,7 @@ const ObsImagePreview = ( {
   }, [selectable, selected] );
```
FLGMwt, line 145: [not at all suggesting changes to this PR, just noticed about this file's "render helpers".]

sanity check because I might be missing it: does anyone know any reason or benefit to having these `useCallback`'d? 

I'd be less skeptical if these were used as props and we had a reason to want a sable function reference, **but** they're only called in the markup. I don't think there's a reason these shouldn't just be defined as the arrow function.

My guess is that it's a misunderstanding / overapplication of React memoization? to clarify: we don't save any function allocations since we're always making a new function ever render for the first argument, it just happens that `useCallback` sometimes throws that away and passes us a saved one.

PR #3309, opened 2025-12-18
Migrate `ExploreFiltersModal` to TypeScript.

frewsxcv:


src/components/Explore/Modals/ExploreFiltersModal.tsx
```diff
@@ -1,22 +1,21 @@
...
-  filterByIconicTaxonUnknown: Function,
-  renderLocationPermissionsGate: Function,
-  requestLocationPermissions: Function,
-  updateTaxon: Function,
-  updateLocation: Function,
-  updateUser: Function,
-  updateProject: Function
-};
+interface Props extends Pick<
+  React.ComponentProps<typeof FilterModal>,
```
FLGMwt, line 7: I'm not sure we necessarily want to do this type of type manipulation here to migrate this file. This definitely helps identify that we have an opportunity for type reuse among these similar / related components in this file / directory. That said, I'd say let's go for duplication _at the moment_ because it's in line with the other files.

We're still talking through how we want to handle shared type definition / shared type extraction but I think in general, that will be more holistically addressed after we have just fewer `js` files.

FLGMwt: which! not related to this PR's changes but still related related and I wanted to FYI: I ran this the other day to check in on JS/TS metrics. 

progress on JS => TS in iNatRN since Sept 30

Then:
- js: 245
- ts: 360

Now:
- js: 191
- ts: 427

Image

@frewsxcv you've been a _big_ reason for this great delta so just wanted to say thanks : )

PR #3310, opened 2025-12-18
Mob 1043 lint rule for enforcing consistent array type definition

abbeycampbell:
[MOB-1043](https://linear.app/inaturalist/issue/MOB-1043/lint-rule-for-enforcing-consistent-array-type-definition-style)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -130,7 +130,7 @@ module.exports = {
     "no-undef": "error",
 
     "@typescript-eslint/array-type": ["error", {
-      default: "array"
+      default: "array",
     }],
     "@typescript-eslint/no-unused-vars": [
       "error",


.eslintrc.js
```diff
@@ -127,6 +127,9 @@ module.exports = {
     // https://eslint.org/docs/latest/rules/no-undef#handled_by_typescript
     "no-undef": "error",
 
+    "@typescript-eslint/array-type": ["error", {
+      default: "array"
```
FLGMwt: oh nice, thank you for adding this explicitly! 

I realized that I have been inconsistent about this, but I think it makes sense to have the option block provided explicitly, even if it matches the default, when it's not obvious from the rule name what our "choice" is.

FLGMwt (approved):
wooo! I think this is one of the bigger day-to-day "I'm not certain which way to use" rules so it's great we won't have to think about it anymore : )

.eslintrc.js
abbeycampbell: yeah that was my thought. I don't think we need to go back and be explicit about the others tho, unless you really want to. Hopefully we're setting these and never really thinking about them again :)

FLGMwt (approved):
fwiw, double approve w/ latest merged in before PR merge 🎉

PR #3312, opened 2025-12-22
MOB-1077: simple navigation.navigate for ObsEdit

sepeterson:
[MOB-1077](https://linear.app/inaturalist/issue/MOB-1077/cannot-go-back-from-edit-saved-observation-screen)

As near as I can tell, for every situation where we navigate to ObsEdit, this also works just fine. I am concluding that a year of updates made resetting the navigator unnecessary.


sepeterson:
I tested these scenarios that use this hook:

Default mode
- ObsDetails -> Edit icon
- My observations grid view with missing data -> Edit icon
- My observations list view with missing data -> Edit icon

Advanced mode
- ObsDetails -> Edit icon
- My observations grid view with missing data -> grid pressable
- My observations list view with missing data -> list pressable

Lmk if I overlooked anything

PR #3315, opened 2026-01-07
integrate rozenite devtool base config & plugins

FLGMwt:
This adds base configuration for Rozenite, a plugin-management tool for plugins built for React Native's DevTools. It also includes the packages and configuration for two relevant Rozenite plugins addressing key DX concerns: network tracing and React Query (TanStack Query). 

We would likely benefit from additional available _rozenite-official_ plugins supporting React Navigation, MMKV, Redux, but as those involve more configuration those will be deferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ See [CONTRIBUTING](CONTRIBUTING.md) for guidelines on contributing to this proje
 
 ### Rozenite (React Native DevTools plugins)
 
-You may optionally, as an environment variable, set the Rozenite env var to enable the Rozenite React Native DevTools plugins: `WITH_ROZENITE=true npm start`. This makes the installed Rozentite plugins available in React Native DevTools in their respective tabs.
+You may optionally, as an environment variable, set the Rozenite env var to enable the Rozenite React Native DevTools plugins: `WITH_ROZENITE=true npm start`. This makes the installed Rozenite plugins available in React Native DevTools in their respective tabs.
 
 ### Running with staging environment
 


README.md
```diff
@@ -39,6 +39,10 @@ See [CONTRIBUTING](CONTRIBUTING.md) for guidelines on contributing to this proje
 1. Run `npm start -- --reset-cache` (`npm start` works too, but resetting the cache each time makes for a lot less build issues)
 2. Run `npm run ios` or `npm run android`
 
+### Rozenite (React Native DevTools plugins)
+
+You may optionally, as an environment variable, set the Rozenite env var to enable the Rozenite React Native DevTools plugins: `WITH_ROZENITE=true npm start`. This makes the installed Rozentite plugins available in React Native DevTools in their respective tabs.
```
FLGMwt: For now I'm following, as closely as possible, the Rozenite docs' [suggestion for config/enablement](https://www.rozenite.dev/docs/getting-started#metro-configuration). If we want the rozenite-empowered devtool goodies, we need to invoke our bundler-initiating command w/ the `WITH_ROZENITE` env var. `WITH_ROZENITE` doesn't mean anything special to the package, it's just the documentation convention what's concern terminates in `metro.config.js` where we decide whether to enable rozenite or not. 

In future, I think we can explore things like baking in `"start": "WITH_ROZENITE=true react-native start",` or an alt `npm run start:rozenite`, but I think we should figure out what we want out of rozenite for a beat before we do that.

jtklein: Typo in `Rozentite`

package.json
```diff
@@ -69,6 +69,9 @@
     "@react-navigation/native": "^7.1.17",
     "@react-navigation/native-stack": "^7.3.25",
     "@realm/react": "^0.20.0",
+    "@rozenite/metro": "^1.1.0",
```
jtklein, line 72: They can not be dev dependencies?

FLGMwt: yeah, I'm not certain about this. I have them in regular deps since they're _imported_ in app code, even if they're noops in prod. Fwiw, I don't think it impacts the app/build since they'll get treeshaken out for non-dev builds.

PR #3317, opened 2026-01-08
add & config mmkv rozenite devtool plugin

FLGMwt:
This PR adds the MMKV Rozenite plugin for React Native Devtools

image

image

Because we're now importing the mmkv storage instance in app root in support of the devtool, I extracted the mmkv storage instance initiation into a separate file/module to avoid circular dependency issues.

There are no functional changes.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,7 @@ import Config from "react-native-config";
 import { setJSExceptionHandler, setNativeExceptionHandler } from "react-native-exception-handler";
 import { GestureHandlerRootView } from "react-native-gesture-handler";
 import { SafeAreaProvider } from "react-native-safe-area-context";
-import { getInstallID } from "sharedHelpers/installData";
+import { getInstallID, store as installDataMMKVStorage } from "sharedHelpers/installData";
 import { reactQueryRetry } from "sharedHelpers/logging";
 import DeviceInfo from "react-native-device-info";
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -147,6 +149,11 @@ const AppWithProviders = ( ) => {
   // note: automatically disabled in Production builds
   useTanStackQueryDevTools( queryClient );
   useNetworkActivityDevTools();
+  useMMKVDevTools( {
+    storages: {
+      "zustand-storage": zustandMMKVBackingStorage,
```
jtklein: We should re-use the id `persisted-zustand` from here: https://github.com/inaturalist/iNaturalistReactNative/blob/5b6f65904e7cdd9d4ec2ca62177d48e895da5f94/src/stores/useStore.js#L75-L76

index.js
```diff
@@ -147,6 +149,11 @@ const AppWithProviders = ( ) => {
   // note: automatically disabled in Production builds
   useTanStackQueryDevTools( queryClient );
   useNetworkActivityDevTools();
+  useMMKVDevTools( {
+    storages: {
```
jtklein, line 153: Can you also add the `install-data` store from `installData`.

FLGMwt:
> LGTM, but please also add the install-data store from installData.

good spots! both addressed and confirmed installData availability in the devtool tab (storage dropdown in top right toggles between the mmkv stores):
image

PR #3320, opened 2026-01-12
MOB-968 Edit icon from Match Screen

sepeterson:
Closes [MOB-968](https://linear.app/inaturalist/issue/MOB-1077/cannot-go-back-from-edit-saved-observation-screen)

In this PR:

- refactor match screen to only have one return block
- add edit icon to match screen
- reuse component for saved match edit button
- handle case where user edits in-flight match. For now, we'll use essentially the advanced mode experience where [navigating back discards the observation](https://linear.app/inaturalist/issue/MOB-968/add-edit-icon-to-top-right-of-match-screen), with a [followup ticket to add rollback experience](https://linear.app/inaturalist/issue/MOB-1091/match-screen-edit-flow-should-roll-back-changes-on-back-navigation)

### DIFF
--- a/src/components/ObsEdit/ObsEdit.js
+++ b/src/components/ObsEdit/ObsEdit.js
@@ -36,6 +36,7 @@ const ObsEdit = ( ): Node => {
   const [passesEvidenceTest, setPassesEvidenceTest] = useState( false );
   const [resetScreen, setResetScreen] = useState( false );
   const [needLocation, setNeedLocation] = useState(
+    // take a look 👀
     shouldFetchObservationLocation( currentObservation ),
   );
   const isFocused = useIsFocused( );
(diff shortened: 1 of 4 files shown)


src/components/ObsEdit/ObsEdit.js
```diff
@@ -38,6 +38,7 @@ const ObsEdit = ( ): Node => {
   const [needLocation, setNeedLocation] = useState(
     shouldFetchObservationLocation( currentObservation ),
   );
+  const [hasUserEdited, setHasUserEdited] = useState( false );
```
jtklein: Just looking at the code, I don't think this is necessary. Can we not just use [unsavedChanges](https://github.com/inaturalist/iNaturalistReactNative/blob/cf5ab3abe25ac457b5addb5a8884f6dc9509fc7e/src/stores/createObservationFlowSlice.js#L26-L27) here in places where you introduced a check on hasUserEdited?
So, I think all of the changes here and in EvidenceSectionContainer can be reverted, and in ObsEditHeader we if() on unsavedChanges instead of hasUserEdited, and have the same result.

PR #3321, opened 2026-01-13
implement experiment for running offline suggestions in parallel to online

FLGMwt:
After codebase evolution discussions around #3257, I realized it made more sense to implement this intimately with the online suggestion call rather than trying to entangle the experiment with the coupling complexity that may be addressed in MOB-1081.

Change 1:

`useSuggestions` impl would _like_ to treat useOff/useOn responses homogeneously. but in the downstream `filterSuggestions`, we were normalizing the missing `common_ancestor?.combined_score` which is explicitly redacted from the API version of the response (it's present in offline and we rely on it downstream).

I have moved this normalization from the common flow from `filter` to immediately after the api response. 

Change 2:

I wanted `useOnline` to be able to imperatively "fire and forget" an offline sugg request. We don't need the response for UI, just for logging.

The offline sugg request was wrapped in a component-local callback. I refactored this callback, with no changes to its massaging logic, to a file-local pure function. It had a few setStates or references to prop handler/callbacks. I moved the responsibility of those to the component-local effects or event handlers that cared about them. There should be no functional change.

Change 3 (doing the ticket!):

The "experiment" to fire and forget an offline sugg request is now basically inlined _immediately_ after the API call. including this as part of the query function avoids having to manage any type of "when am I refetching" heuristics trying to guess what our logic is doing or what RQ is doing. Our output is _only_ logging for this case so there's no need to involve the rendering cycle or hooks.

### DIFF
--- a/src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
+++ b/src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
@@ -1,5 +1,6 @@
 import { RealmContext } from "providers/contexts";
 import {
+  useCallback,
   useEffect,
   useState,
 } from "react";
@@ -115,7 +116,7 @@ const useOfflineSuggestions = (
     onFetchError, onFetched, latitude, longitude, tryOfflineSuggestions,
   } = options;
 
-  const refetchOfflineSuggestions = async () => {
+  const fetchOfflineSuggestions = useCallback( async () => {
     try {
       const suggestions = await predictOffline( {
         latitude,
@@ -127,40 +128,22 @@ const useOfflineSuggestions = (
       onFetched( { isOnline: false } );
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch ( predictOfflineError: any ) {
-      // TODO: throw error in a way that doesnt potentially bubble up
       onFetchError( { isOnline: false } );
       setError( predictOfflineError );
     }
-  };
+  }, [latitude, longitude, onFetchError, onFetched, photoUri, realm] );
 
   useEffect( ( ) => {
-    const fetchOfflineSuggestions = async () => {
-      if ( photoUri && tryOfflineSuggestions ) {
-        try {
-          const suggestions = await predictOffline( {
-            latitude,
-            longitude,
-            photoUri,
-            realm,
-          } );
-          setOfflineSuggestions( suggestions );
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        } catch ( predictOfflineError: any ) {
-          // For some reason if you throw here, it doesn't actually buble up. Is
-        // an effect callback run in a promise?
-          onFetchError( { isOnline: false } );
-          setError( predictOfflineError );
-        }
-      }
-    };
-    fetchOfflineSuggestions();
-  }, [photoUri, tryOfflineSuggestions, setError, onFetchError, latitude, longitude, realm] );
+    if ( tryOfflineSuggestions ) {
+      fetchOfflineSuggestions();
+    }
+  }, [fetchOfflineSuggestions, tryOfflineSuggestions] );
 
   if ( error ) throw error;
 
   return {
     offlineSuggestions,
-    refetchOfflineSuggestions,
+    refetchOfflineSuggestions: fetchOfflineSuggestions,
   };
 };
 
--- a/src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
+++ b/src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
@@ -119,15 +119,17 @@ const useOnlineSuggestions = (
 
       // experiment to compare online & offline suggestion results
       // imperatively fire-and-forgetted in online query to circumvent existing stateful mutex logic
-      executeOnRandomPercentile( () => {
+      executeOnRandomPercentile( async () => {
         if ( scoreImageParams ) {
-          predictOffline( {
-            latitude: scoreImageParams.lat,
-            longitude: scoreImageParams.lng,
-            photoUri: scoreImageParams.image.uri,
-            realm,
-          } );
-          // TODO: log results
+          // TODO: log "background" experiment responses & errors
+          try {
+            await predictOffline( {
+              latitude: scoreImageParams.lat,
+              longitude: scoreImageParams.lng,
+              photoUri: scoreImageParams.image.uri,
+              realm,
+            } );
+          } catch ( _error ) { /* empty */ }
         }
       }, SIMULTANEOUS_ONLINE_OFFLINE_SUGGESTION_EXPERIMENT_INTEGER_PERCENTAGE );
       // end experiment


src/sharedHooks/useSuggestions/filterSuggestions.ts
```diff
@@ -78,22 +78,9 @@ const filterSuggestions = (
...
-      // general context: https://github.com/inaturalist/iNaturalistReactNative/blob/505980d3359876a0af383f2ffcc481921f0eb778/src/components/Match/calculateConfidence.ts#L10-L12
-      // online suggs have `score` but _redact_ `combined_score` for commonAncestor https://github.com/inaturalist/iNaturalistAPI/blob/main/lib/controllers/v1/computervision_controller.js#L389
-      // the offline suggs have `combined_score` but don't have `score`
-      // the codebase tends assumes `combined_score` for whenever that matters
-      // the following catches when we're in a "fake" onlineSugg and shims "score" in
-      // `in` operator used for OnlineSuggestion type refinement
-      combined_score: !commonAncestor.combined_score && "score" in commonAncestor
-        ? commonAncestor.score
-        : commonAncestor.combined_score,
-    };
```
FLGMwt, line 95: We were previously having to do what amounts to runtime type inspection here. I have removed this need by moving this logic to normalize these properties immediately after the online response. By the time we get to filterSuggestions here, we can treat the input as homogenous

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
```diff
@@ -15,6 +15,79 @@ const logger = log.extend( "useOfflineSuggestions" );
...
+interface OfflineRequestOptions {
+  latitude?: number;
+  longitude?: number;
+  photoUri: string;
+  realm: Realm;
+}
+
+export const predictOffline = async ( {
+  latitude, longitude, photoUri, realm,
+}: OfflineRequestOptions ) => {
```
FLGMwt, line 28: This is the extraction of the previous component-inline handler for making an offline sugg request. it's now a "pure" function* outside of the react lifecycle so that it can be called imperatively.

The only changes to the function block were to extract the component-bound scope variables. Those concerns (like setState) have been moved to the call locations of `predictOffline` (where in some cases they were already being redundantly called).

*: yes, I know they're not pure because they're making global scope logging calls and realm/api side effects, but pure in the sense of external to react : P

tests/unit/sharedHooks/useSuggestions/isolateHumans.test.js
FLGMwt, line 1: just a bonus unit test for fun. Was gonna do some more tests for `useSuggestions/` but I think it's a bit of a beast that could be broken up further before making that investment

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -42,11 +51,36 @@ interface UseOnlineSuggestionsResponse {
...
 
+const normalizeApiResponseForCommonAncestor
+  = ( apiSuggestions: OnlineSuggestionsApiResponse ): OnlineSuggestionsQueryResponse => {
+    // TODO MOB-1081: maybe we can catch this shim earlier?
+    // general context: https://github.com/inaturalist/iNaturalistReactNative/blob/505980d3359876a0af383f2ffcc481921f0eb778/src/components/Match/calculateConfidence.ts#L10-L12
+    // online suggs have `score` but _redact_ `combined_score` for commonAncestor https://github.com/inaturalist/iNaturalistAPI/blob/main/lib/controllers/v1/computervision_controller.js#L389
+    // the offline suggs have `combined_score` but don't have `score`
+    // the codebase tends assumes `combined_score` for whenever that matters
+    // the following catches when we're in a "fake" onlineSugg and shims "score" in
+    const normalizedCommonAncestor = apiSuggestions.common_ancestor
```
jtklein: Can we use a different name here and in function. In the CV code we often normalize the score in the statistical sense, i.e. get it back in the range of 0-1. Just to avoid confusion with this kind of normalization there.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -42,11 +51,36 @@ interface UseOnlineSuggestionsResponse {
   refetch: () => void;
 }
 
+interface OnlineSuggestionsApiResponse {
+  results: UseSuggestionsOnlineSuggestion[];
+  common_ancestor?: Omit<UseSuggestionsOnlineSuggestion, "combined_score">;
```
jtklein, line 56: I have never seen `Omit` what does it do? It is new to the codebase so I'd prefer to see some explanation or message about it in Slack after a merge as to how it is useful and if we should use this more in general and if there are other places in the code base we should refactor to this syntax.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -82,7 +116,27 @@ const useOnlineSuggestions = (
         ...scoreImageParams,
         ...( !currentUser && { locale } ),
       };
-      return scoreImage( params, optsWithAuth ) as Promise<OnlineSuggestionsQueryResponse>;
+      const suggestionsResponse
+        = await scoreImage( params, optsWithAuth ) as OnlineSuggestionsApiResponse;
+      const normalizedOnlineSuggestionsResponse
```
jtklein: See comment about name above

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -82,7 +116,27 @@ const useOnlineSuggestions = (
...
+        = normalizeApiResponseForCommonAncestor( suggestionsResponse );
+
+      // experiment to compare online & offline suggestion results
+      // imperatively fire-and-forgetted in online query to circumvent existing stateful mutex logic
+      executeOnRandomPercentile( () => {
+        if ( scoreImageParams && scoreImageParams.image ) {
```
jtklein: This raises my eyebrow. If we do not have a `scoreImageParams.image.uri` value also `scoreImage` above will surely not work. I assume you added the check here to satisfy TS not to complain about uri being undefined.
`scoreImage` might not crash the app if we do not give it a uri but for sure the request will be meaningless and hopefully will error out on us. In any regard we probably should have the check for uri being undefined higher up in this block.
But maybe that should be it's own ticket.

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
```diff
@@ -42,97 +115,47 @@ const useOfflineSuggestions = (
...
+        latitude,
+        longitude,
+        photoUri,
+        realm,
+      } );
+      setOfflineSuggestions( suggestions );
+      onFetched( { isOnline: false } );
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch ( predictOfflineError: any ) {
       onFetchError( { isOnline: false } );
```
jtklein: `onFetchError` is called twice here

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
FLGMwt: noted! I think especially since is this _related_ to CV results, it makes sense to avoid "normalize" here.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
FLGMwt: Aye sounds good! will add the following to our running list of these & fyi in slack

[`Omit<Type, "key">`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys), a built-in TS utility type, lets you derive a new type based on an existing type but with some keys omitted:

for example:

```
interface CityState { city: string; state: string }

// the following two type definitions are equivalent:
interface City { city: string }
type City = Omit<CityState, "state">
```

In this case, it's used as a convenience (and clarity) alternative to defining a _whole other type_ for this _one_ special case of a suggestion that has the same type as most other instances of suggestions but _very importantly_ does not have `combined_score`. As opposed to duplicating the entire type tree or unnecessarily complex unioning.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
FLGMwt: Oh yeah, that seems to have been carried over from a previous type check that the newer types help confirm is unnecessary. Fixed and fixed a couple other unnecessary checks.

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
FLGMwt: fixed. Might be clearer from delta to say: fixed in a commit here https://github.com/inaturalist/iNaturalistReactNative/pull/3321/commits/6cc78b1bec3f5ff2d0114576de49485397865ec5 and then refactored for the effect issue in latest commit https://github.com/inaturalist/iNaturalistReactNative/pull/3321/commits/b49875ba7e1c43862f3679a48b0ceac1ebab4c8c

FLGMwt:
@jtklein 

> Second, if I run the app and force an offline fetch, e.g. for example if I change `inatjs.computervision.score_image` to `inatjs.computervision.score_images` in computerVision.ts, the UI does not update with an offline suggestion anymore.

Thank you! That was a good testing suggestion over my trying to mock out some of the flags that drive this happening or not & it uncovered some bad effect expectations in the refactor.

W/ latest commit I fixed this, deduping and refining the distinction between the imperative and reactive cases (refetch handler vs useEffect "when deps tell me to"). I confirmed those work as expected with the "make api fail" test and maintains other existing behavior.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
jtklein: Thanks

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
jtklein: OK, nice. Thanks for the explanation.

src/sharedHooks/useSuggestions/useOnlineSuggestions.ts
```diff
@@ -116,31 +116,32 @@ const useOnlineSuggestions = (
...
-            longitude: scoreImageParams.lng,
-            photoUri: scoreImageParams.image.uri,
-            realm,
-          } );
-          // TODO: log results
+      executeOnRandomPercentile( async () => {
+        if ( scoreImageParams ) {
+          // TODO: log "background" experiment responses & errors
+          try {
+            await predictOffline( {
```
jtklein: I will test now but that means we will await offline and then start online after offline is finished when an experiment is executed. I think we should try to run these in parallel.

src/sharedHooks/useSuggestions/filterSuggestions.ts
jtklein: Great idea!

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
jtklein: I do like this a lot. Not sure why but I like the look of it.

tests/unit/sharedHooks/useSuggestions/isolateHumans.test.js
jtklein: I agree, better to break up first. I marked this [ticket](https://linear.app/inaturalist/issue/MOB-1081/stub-remove-tight-coupling-between-usesuggestions-implementation-and) blocked by this here. We could even tackle this one first before we make tests.

PR #3323, opened 2026-01-13
add zustand slice for feature flags, shared hook, and Debug UI for overriding for test

FLGMwt:
Adds new app-level configuration for feature flags, using ExploreV2 as an exemplary first flag (and will likely add #3321's "feature" once merged).

application code should _only_ need to use `useFeatureFlag` which respects the overrides managed by the Developer / Debug UI.

The benefits as implemented here are:
- easy flag addition / deletion / reads
- ability to toggle forked behavior during runtime

The flags are currently just informed by source-static defaults and the debug-user-initiated local overrides (zustand-persisted), but it's well setup to incorporate whatever type of remote config we may decide on down the line.

https://github.com/user-attachments/assets/e8ef164e-6487-4cdc-9641-6de1772fec93

welp, the video _is_ supposed to have audio, maybe GH just doesn't like my voice : P seems to work if you right-click "Open in new tab"

### DIFF
--- a/src/stores/createFeatureFlagSlice.ts
+++ b/src/stores/createFeatureFlagSlice.ts
@@ -1,16 +1,35 @@
 import type { StateCreator } from "zustand";
 
+// This state slice supports application code's consumption of feature flags, namely:
+// const isMyFeatureEnabled = useFeatureFlag ( FeatureFlag.MyFeatureFlagEnabled );
+
+// The "status" / "enabledness" of whether a feature flag'd feature is "on" or "off" in the live app
+// is driven by the hard-coded entry here in `initialFeatureFlagConfig`. That is,
+// when we want to "turn on" `MyFeature`, we will set its `initialFeatureFlagConfig` to true.
+
+// Once a feature has been enabled, deployed, and verified, the feature flag should be removed
+// along with the newly-deprecated / "old" code.
+
+// To add a new feature flag, add entries below, following the example of `MyFeatureFlagEnabled`
+
+// This slice _also_ supports dynamically enabling and disabling a flag for testing,
+// overriding its default "live" status. This is done through the "Debug" / "Developer" screen.
+// These are not persisted so will be reset to their defaults on app start.
+
 // TODO: move this to more a more global type definition scope
 export enum FeatureFlag {
   // flags should use positive language ending with `Enabled`
-  ExploreV2Enabled = "exploreV2Enabled"
+  // MyFeatureFlagEnabled = "myFeatureFlagEnabled",
+  ExploreV2Enabled = "exploreV2Enabled",
 }
 
 const initialFeatureFlagConfig: Record<FeatureFlag, boolean> = {
+  // [FeatureFlag.MyFeatureFlagEnabled]: false,
   [FeatureFlag.ExploreV2Enabled]: false,
 };
 
 const initialFeatureFlagDebugOverrides: Record<FeatureFlag, boolean | null> = {
+  // [FeatureFlag.MyFeatureFlagEnabled]: null,
   [FeatureFlag.ExploreV2Enabled]: null,
 };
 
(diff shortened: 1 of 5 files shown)


src/components/Developer/DeveloperSharedComponents.tsx
```diff
@@ -0,0 +1,36 @@
+import classnames from "classnames";
```
jtklein, line 1: I have not checked this file as I understand it is just a copy from the Debug screen.

src/stores/createFeatureFlagSlice.ts
```diff
@@ -0,0 +1,39 @@
+import type { StateCreator } from "zustand";
+
+// TODO: move this to more a more global type definition scope
+export enum FeatureFlag {
+  // flags should use positive language ending with `Enabled`
+  ExploreV2Enabled = "exploreV2Enabled"
+}
+
+const initialFeatureFlagConfig: Record<FeatureFlag, boolean> = {
```
jtklein, line 26: I think this file could benefit if you add some explanations as comments here as well. Like how in general this slice is supposed to behave.

abbeycampbell:
This is so great! Thanks for adding this.

I know you covered it briefly during eng meeting, but I couldn't wrap my head around it on the fly: could you explain again why we need to have two toggles here instead of one? Like if a flag is configured with a default value of `false`, can't we just toggle it to be `true`? Why do we need the null state?

I also think it could be nice to document (maybe in the readme, or maybe elsewhere if we don't want it so public facing?) the steps needed to use this feature. It's pretty self explanatory but it's nice when you're in the middle of building a feature to have a checklist ie:
- add feature to `FeatureFlag` enum in this file
- configure a default value
- usage, etc

src/stores/useStore.js
```diff
@@ -77,6 +79,7 @@ const useStore = create( persist(
       // Vestigial un-namespaced values in the layout slice
       isAdvancedUser: state.isAdvancedUser,
       obsDetailsTab: state.obsDetailsTab,
+      featureFlagDebugOverrides: state.featureFlagDebugOverrides,
```
jtklein: I don't think we'd want to persist the overrides. I would expect to open the app and have the "intended" experience first. So that every time I open the app I see the initial slice without overrides. And always have to opt in to test a new feature.

If you strongly favor having these overrides persisted we should probably make a new ticket and postpone adding them to persisted state to get team buy-in.

A corollary problem for adding them to the persisted state would be that then we'd want the key to be in the `layout` key of the layoutSlice. The reason we set it up that way was that in Zustand all slices are merged into a single big state and names can collide, so we'd at least wanted to make sure that we do not end up with having identical names in different slices to persist.

FLGMwt: Sounds good, I agree about the buy-in and being more conservative until our usage patterns are more clear.

On the slice namespace, I'll add something to the backlog. We should be able to do improve this at the slice config level so we don't have to manually namespace and worry about cross-slice collision. That also would be a good opportunity to figure out typing if only _just_ to get type info on `store.????`. Our slices themselves are so nicely typed after all : )

src/stores/createFeatureFlagSlice.ts
FLGMwt: Aye, good thought. I'll add to the readme as well. Also made me realize it'd be good to "announce" this to staff. I'll do that short writeup and add to the ticket, but I'll wait to share that until we actively test something FF'd.

src/components/Developer/DeveloperSharedComponents.tsx
FLGMwt: Yup, just copied over from Debug screen to share among Debug screen files.

FLGMwt:
> Approved if you remove the overrides from persisted state I would say.

^removed persisted state and addressed other documentation suggestions, thanks!

PR #3324, opened 2026-01-14
additional user context for feedback

FLGMwt:
When logging our feedback form, we now are adding additional user context attached directly to the feedback itself, tacked on as just extra lines on the string. For example (everything `-----` and after is new):

```Lorem ipsum dolor sit amet, consectetur adipiscing elit.

Duis in gravida augue, at scelerisque mi. Etiam eros urna, faucibus in rutrum non, scelerisque eu tellus. Ut faucibus convallis purus, id varius turpis malesuada ac.
------------
Feedback Context:
Signed In: yes
Username: myINatUsername
Mode: default
Uploaded Observations: 10-99
Unuploaded Device-Local Observations: 1-9
Identifications: 10-99
```

MOB-142 (not closed with this PR, need some additional grafana changes)

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -20,17 +20,13 @@ import Observation from "realmModels/Observation";
 import User from "realmModels/User";
 import { valueToBreakpoint } from "sharedHelpers/breakpoint";
 import { log } from "sharedHelpers/logger";
-import { useCurrentUser, useTranslation } from "sharedHooks";
-import useStore, { zustandStorage } from "stores/useStore";
+import { useCurrentUser, useLayoutPrefs, useTranslation } from "sharedHooks";
+import { zustandStorage } from "stores/useStore";
 import colors from "styles/tailwindColors";
 
 import MenuItem from "./MenuItem";
 
 const { useRealm } = RealmContext;
-
-function isDefaultMode( ) {
-  return useStore.getState( ).layout.isDefaultMode === true;
-}
 
 interface BaseMenuOption {
   label: string;
@@ -74,6 +70,7 @@ const Menu = ( ) => {
 
   const { isConnected } = useNetInfo( );
 
+  const layoutPrefs = useLayoutPrefs();
   const [modalState, setModalState] = useState<MenuModalState | null>( null );
 
   const menuItems: Record<string, MenuOption> = {
@@ -157,12 +154,12 @@ const Menu = ( ) => {
     navigation.goBack( );
   };
 
-  const onSubmitFeedback = useCallback( ( text: string ) => {
+  const onSubmitFeedback = useCallback( ( feedbackText: string ) => {
     if ( !isConnected ) {
       showOfflineAlert( t );
       return false;
     }
-    const localOnlyObsCount = Observation.filterUnsyncedObservations( realm ).length;
+    const locallySavedOnlyObservations = Observation.filterUnsyncedObservations( realm ).length;
     const getCountBreakpoint = ( count: number ) => valueToBreakpoint( count, [
       [0, "0"],
       [1, "1-9"],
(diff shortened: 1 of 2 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -160,14 +161,52 @@ const Menu = ( ) => {
...
+        return "100-999";
+      }
+      if ( count >= 10 ) {
+        return "10-99";
+      }
+      if ( count >= 1 ) {
+        return "1-9";
+      }
+      return "0";
+    };
```
abbeycampbell: suggestion, non-blocking: I could see extracting this into a helper that lives elsewhere being pretty useful in the future, esp as we add more logs/track user behavior more closely

abbeycampbell (approved):

src/components/Menu/Menu.tsx
```diff
@@ -160,14 +161,52 @@ const Menu = ( ) => {
...
+        : "no"}`,
+      `Username: ${currentUser
+        ? currentUser.login
+        : "loggedout"
+      }`,
+      `Mode: ${isDefaultMode( )
+        ? "default"
+        : "advanced"
+      }`,
+      `Uploaded Observations: ${currentUser
```
jtklein: I think here we should fall back to something else, a string of `NA` maybe. Open for other suggestions, a 0 will indicate to whoever analyzes this that this user had no observations, while maybe there was just something wrong with the code executing.

src/components/Menu/Menu.tsx
```diff
@@ -160,14 +161,52 @@ const Menu = ( ) => {
...
+        ? getCountBreakpoint( currentUser.observations_count || 0 )
+        : "loggedout"
+      }`,
+      `Unuploaded Device-Local Observations: ${getCountBreakpoint( localOnlyObsCount )}`,
+      `Identifications: ${currentUser
+        ? getCountBreakpoint( currentUser.identifications_count || 0 )
```
jtklein: Similarly, also to not fallback to 0.

jtklein (approved):
This looks mostly good to me. Only two fallbacks I think we should change otherwise we might send a confusing data point.

I thought we might also want the `isAllAddObsOptionsMode` because I thought the part in the ticket about ai camera vs obs wheel means that. But better to make it as in the comments of the ticket. We can always add more.

src/components/Menu/Menu.tsx
FLGMwt: This is a good suggestion! I was tempted to make a generic helper originally but deferred to the hardcoded `if` chain for the one off case. You make a good point about this being a potentially common case for upcoming analytics work so I just added that. 

Bonus: I was able to validate the use case both here _and_ to replace our existing analogous `if` chain for media query breakpoints https://github.com/inaturalist/iNaturalistReactNative/pull/3324/changes/9cab1a4346fe16775e1826735b6cab18f3623596#diff-e03ce79ae7c933c8dc8619bba3a391fc0314f9268dc0c8c1095a88536f0b5f2aL12-L28 (the two media query breakpoint implementations will only exist for that single commit so that common test input can be run against them to demonstrate parity).

I gotta get my algo chops back tho, it took me twice as long as the original PR 😆 Not a grievance at all, it was good exercise!

src/sharedHelpers/breakpoint.ts
```diff
@@ -1,6 +1,62 @@
...
+// const label = valueToBreakpoint(
+//   150,
+//   [
+//     [0, "0"],
+//     [1, "1-9"],
+//     [10, "10-99"],
+//     [100, "100+"],
+//   ],
+// );
+// label === "100+"
```
FLGMwt, line 19: man, it'd be nice if jsdoc supported multi-line text and I didn't have to point you to the impl source file 😒

anyway, I made the choice here to use tuples as input and remap them in the impl because I felt that consumers can make much better sense of that than parallel collections that don't visually line up in source:

```
valueToBreakpoint(
  150,
  [0, 1, 10, 100],
  ["0", "1-9", "10-99", "100+],
);
```
^that seems much harder to grok.

tests/unit/helpers/breakpoint.test.js
```diff
@@ -0,0 +1,55 @@
...
+    ] )(
+      "should return appropriate media query label for screenWidth",
+      ( screenWidth, expected ) => {
+        const result = getBreakpoint( screenWidth );
+
+        expect( result ).toBe( expected );
+      },
+    );
+  } );
+} );
```
FLGMwt, line 55: A previous commit in this PR demonstrates the new implementation's parity with the previous implementation using the same test cases:

https://github.com/inaturalist/iNaturalistReactNative/pull/3324/changes/9cab1a4346fe16775e1826735b6cab18f3623596#diff-40b595aa751cee2acd190eec12235fbad96272ec009895e508adec46e1dbde39R31

src/components/Menu/Menu.tsx
abbeycampbell: haha all I meant was just extracting this function out of `Menu` for use with future logs or analytics calls, (if we ever decide to do that kind of thing). I think your approach here is fine though and it's nice that we can re-use it for media queries, though I do wonder if it actually makes the code a bit hard to follow

FLGMwt: Oh hah. oopsies. I think I was figuring that I didn't want to assume these specific breakpoints / messages for likely, but hypothetical cases, which led me to making the shared helper generic. Yeah, it's overkill : p I'm thinking I'll still keep it though to account for your consideration for similar future cases. and I feel okay about it since it's pretty well scoped, documented, and tested.

PR #3326, opened 2026-01-15
Mob 1047 add sorting to my observations

abbeycampbell:
Partially addresses [MOB-1047](https://linear.app/inaturalist/issue/MOB-1047/add-sorting-to-my-observations)

- Adds sorting UI for species and observations that we should just be able to drop into explore when we're ready to use it there.
- Adds sorting for the species/taxa tab; the sort button on the observations tab is commented out for now because this PR is big enough and that work is a bit more fiddly, since we rely on locally cached realm data there. When that PR comes, it will likely also contain some reorganization of `MyObservationsContainer`.

Notes:
A lot of the diff here is adding the new icon, adding translations, and updating snapshots. Still, sorry for the daunting size.

The API doesn't currently support fetching a list of species counts sorted by anything other than count, so alphabetical/taxonomy sorting will need to be added later. I've used switch/case statements in the functions that return API params and handle local sorting, so that we can just drop in the logic that will handle these cases when that time comes. 

Logged in and online users should fetch and see a fresh list on sort, logged out users will have locally sorted taxa, and offline users will see an alert prompting them to try again later.

this PR also adds a new icon, so once it's merged we'll all need to re-build the native layer.

### DIFF
--- a/src/components/SharedComponents/FlashList/CustomFlashList.tsx
+++ b/src/components/SharedComponents/FlashList/CustomFlashList.tsx
@@ -122,27 +122,22 @@ const CustomFlashList = props => {
 
   // To be called when new data is received
   // This needs to be exposed so it can be called from parent component
-  React.useImperativeHandle( ref, ( ) => ( {
-    notifyDataFetched: itemsCount => {
-      if ( fetchInProgress.current ) {
-        flashListTracker.endDataFetch( itemsCount );
-        fetchInProgress.current = false;
-      } else {
-        flashListTracker.beginDataFetch( );
-        flashListTracker.endDataFetch( itemsCount );
-      }
-    },
-    scrollToOffset: params => {
-      if ( internalRef.current ) {
-        internalRef.current.scrollToOffset( params );
-      }
-    },
-    scrollToIndex: params => {
-      if ( internalRef.current ) {
-        internalRef.current.scrollToIndex( params );
-      }
-    },
-  } ) );
+  React.useImperativeHandle( ref, () => {
+    const flashListMethods = internalRef.current || {};
+
+    return {
+      ...flashListMethods,
+      notifyDataFetched: itemsCount => {
+        if ( fetchInProgress.current ) {
+          flashListTracker.endDataFetch( itemsCount );
+          fetchInProgress.current = false;
+        } else {
+          flashListTracker.beginDataFetch();
+          flashListTracker.endDataFetch( itemsCount );
+        }
+      },
+    };
+  } );
 
   const viewabilityConfig = {
     ...defaultViewabilityConfig,


src/components/SharedComponents/Sheets/SortSheet.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import { useTranslation } from "sharedHooks";
+import type { SortItemType } from "types/sorting";
+
+import RadioButtonSheet from "./RadioButtonSheet";
+
+const getObservationSortOptions = ( t: ( key: string ) => string ) => ( {
```
jtklein: There is a type for that from the library: `import type { TFunction } from "i18next";`

src/components/SharedComponents/Sheets/RadioButtonSheet.tsx
```diff
@@ -66,6 +68,7 @@ const RadioButtonSheet = ( {
       insideModal={insideModal}
       onPressClose={onPressClose}
       testID={testID}
+      scrollEnabled={false}
```
jtklein, line 71: This potentially also affects some other tickets in the backlog.
After merging this PR we should remember to check some existing tickets if they are still reproducible.

src/components/SharedComponents/Sheets/SortSheet.tsx
```diff
@@ -0,0 +1,77 @@
...
+} as const );
+
+interface Props {
+  itemType: SortItemType;
+  selectedValue: string;
+  onConfirm: ( sortBy: string ) => void;
+  onPressClose: () => void;
+}
+
+const SortSheet = ( {
```
jtklein: I myself would have probably not opted to add this wrapper component as an abstraction. I find the switch of observation | species here confusing. I think I would have just added the RadioButtonSheet directly to MyObs. Also `selectedValue` and `sortBy` are not string but can be typed more stringent without this wrapper, as they are `ObservationSortOptionId` or `SpeciesSortOptionId`.

src/components/SharedComponents/Sheets/SortSheet.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import { useTranslation } from "sharedHooks";
+import type { SortItemType } from "types/sorting";
+
+import RadioButtonSheet from "./RadioButtonSheet";
+
+const getObservationSortOptions = ( t: ( key: string ) => string ) => ( {
+  created_at_desc: {
```
jtklein: I would like to see a more unforgiving contract here.  See, [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/Explore/Modals/FilterModal.tsx#L162-L163) for example. With an enum we can then make sure we only pass in appropriate values as `selectedValue` and use as `sortBy`.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -317,33 +350,54 @@ const MyObservationsSimple = ( {
...
+            <SortButton
+              onPress={() => setShowSortSheet( true )}
+              accessibilityLabel={t( "Change-species-sort-order" )}
+            />
+          </>
         )}
         { ( activeTab === TAXA_TAB && taxa.length === 0 ) && renderOfflineNotice( )}
       </ViewWrapper>
+      {showSortSheet && (
+        <SortSheet
```
jtklein: I believe having two sheets here one for each tab would make this file more readable. Instead having to switch over observations vs species everywhere.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -317,33 +350,54 @@ const MyObservationsSimple = ( {
...
+            />
+            <SortButton
+              onPress={() => setShowSortSheet( true )}
+              accessibilityLabel={t( "Change-species-sort-order" )}
+            />
+          </>
         )}
         { ( activeTab === TAXA_TAB && taxa.length === 0 ) && renderOfflineNotice( )}
       </ViewWrapper>
+      {showSortSheet && (
```
jtklein: Since this is the second sheet on this screen I think we should make them mutually exclusive. So, at least in the future if we want to have them open at the same time it needs to be a decided effort to do so.
So, instead of one boolean for each sheet, I think we should use an enum to either show NONE or the login sheet or the sort sheet (two if you decide to split them up). Compare the Explore Filter Modal [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/Explore/Modals/FilterModal.tsx#L140-L141)

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -246,6 +262,23 @@ const MyObservationsSimple = ( {
     return null;
   };
 
+  function showOfflineAlert( t: ( _: string ) => string ) {
```
jtklein: TFunction, see below

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -246,6 +262,23 @@ const MyObservationsSimple = ( {
...
+    Alert.alert( t( "You-are-offline" ), t( "Please-try-again-when-you-are-online" ) );
+  }
+
+  const handleSortConfirm = ( optionId: string ) => {
+    if ( currentUser && !isConnected ) {
```
jtklein, line 273: Why is this only for logged in users?
So, looking at the ticket what we want is a warning that the user is offline (because we need the server to retrieve the correct first page of sorted items). But I am unsure why logged-in and logged-out users differ here.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -246,6 +262,23 @@ const MyObservationsSimple = ( {
...
+      showOfflineAlert( t );
+      return;
+    }
+    if ( activeItemType === "observations" ) {
```
jtklein: Again, not so keen on these switches.

src/types/sorting.ts
```diff
@@ -0,0 +1,33 @@
...
+  direction?: SortDirection;
+}
+
+export interface SpeciesSort {
+  by:
+    | "count";
+  direction?: SortDirection;
+}
+
+export interface SpeciesCount {
```
jtklein, line 22: Also created [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/MyObservations/SimpleTaxonGridItem.tsx#L19-L20) and [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/MyObservations/MyObservationsSimple.tsx#L44-L45)

src/components/SharedComponents/FlashList/CustomFlashList.tsx
```diff
@@ -120,29 +122,27 @@ const CustomFlashList = props => {
...
-        }
-      },
-      scrollToOffset: params => {
-        if ( ref && typeof ref !== "function" && ref.current ) {
-          ref.current.scrollToOffset( params );
-        }
-      },
-    };
-  } );
+  React.useImperativeHandle( ref, ( ) => ( {
```
jtklein: I was a bit confused about those changes here, thinking wrongly that removing the `...originalRef` meant we now do no longer expose all other methods of a FlashList ref not mentioned here. But I think the intent of this block was to only expose those three methods and nothing else. But I am wondering why we do that at all now and not just spread all of the FlashList refs together with `notifyDataFetched` up so that we don't have to care about special cases for other FlashList methods.

src/components/SharedComponents/FlashList/CustomFlashList.tsx
```diff
@@ -120,29 +122,27 @@ const CustomFlashList = props => {
...
+        flashListTracker.beginDataFetch( );
+        flashListTracker.endDataFetch( itemsCount );
+      }
+    },
+    scrollToOffset: params => {
+      if ( internalRef.current ) {
+        internalRef.current.scrollToOffset( params );
+      }
+    },
+    scrollToIndex: params => {
```
jtklein: Why did you add this here? I don't see it used anywhere in the new code.

src/components/SharedComponents/FlashList/CustomFlashList.tsx
jtklein: Anyways, not really sure why this refactor was included in this PR here.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -53,6 +57,7 @@ export interface Props {
   isFetchingNextPage: boolean;
   layout: "list" | "grid";
   listRef?: React.RefObject<FlashListRef<RealmObservation> | null>;
+  taxaListRef?: React.RefObject<FlashListRef<SpeciesCount> | null>;
```
jtklein, line 56: Do you want to import SpeciesCount, since SpeciesSortOptionId already is?

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -7,13 +7,17 @@ import { fetchSpeciesCounts } from "api/observations";
 import { RealmContext } from "providers/contexts";
 import React, {
   useCallback,
-  useEffect, useRef,
+  useEffect, useMemo, useRef,
```
jtklein: Nit: newlines here

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -53,6 +57,7 @@ export interface Props {
   isFetchingNextPage: boolean;
   layout: "list" | "grid";
   listRef?: React.RefObject<FlashListRef<RealmObservation> | null>;
```
jtklein, line 55: So, because in CustomFlashList - which both of these props here are handed down into - we replace the ref using an imperative handle the types of these two refs here are not actually FlashListRef, correct?
Just want to avoid that someone working in this file here is confident in using a method that is not actually exposed on this ref here.

src/types/sorting.ts
```diff
@@ -0,0 +1,33 @@
...
+  count: number;
+  taxon: RealmTaxon;
+}
+
+export type ObservationSortOptionId =
```
jtklein: I am obviously influenced by my own work I did on Explore filters but I do think that enums here are more understandable across a big codebase with multiple developers. So, whoever needs to work on this sorting in the future can easily import the one value they need from here. In Explore filters I set up the enum [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/providers/ExploreContext.tsx#L48-L49) and make sure that only these values can be set to the context state [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/providers/ExploreContext.tsx#L268-L269) then, as mentioned in another comment, [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/Explore/Modals/FilterModal.tsx#L163-L164) I make sure that only the values with the enum contract are selectable, and finally map it to the api by only having allowed states [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/Explore/helpers/mapParamsToAPI.js#L38-L39).

src/sharedHelpers/sortingHelpers.ts
```diff
@@ -0,0 +1,87 @@
+import type {
+  ObservationSort, ObservationSortOptionId, SpeciesCount, SpeciesSort, SpeciesSortOptionId,
+} from "types/sorting";
+
+export const OBSERVATION_SORT_MAP: Record<
```
jtklein: This is not imported anywhere.

src/sharedHelpers/sortingHelpers.ts
```diff
@@ -0,0 +1,87 @@
...
+    by: "observed_at",
+    direction: "desc",
+  },
+  observed_at_asc: {
+    by: "observed_at",
+    direction: "asc",
+  },
+};
+
+export const SPECIES_SORT_MAP: Record<SpeciesSortOptionId, SpeciesSort> = {
```
jtklein: This is not imported anywhere.

src/sharedHelpers/sortingHelpers.ts
```diff
@@ -0,0 +1,87 @@
...
+  count_desc: { by: "count", direction: "desc" },
+  count_asc: { by: "count", direction: "asc" },
+};
+
+/**
+ * Maps selected species sort option to API params
+ * Returns null if the sort should be done client-side
+ * @param sortOptionId: the selected sort option
+ */
+export function mapSpeciesSortToAPIParams(
```
jtklein, line 20: As a general comment, this all looks really well encapsulated, and the code works and does what is supposed to do. And maybe I am sensing a striving for mastery in the art of abstraction. As I was reviewing this PR I was comparing it side-by-side with my previous work, and I do think there is some beauty in simplicity looking back at my work (obviously heavily biased here). Especially this block [here](https://github.com/inaturalist/iNaturalistReactNative/blob/409f0edacefc676590187e4f50088bd73c20253f/src/components/Explore/helpers/mapParamsToAPI.js#L35-L54) I can look back at and it takes me a mere few seconds to understand what's going on here because the logic is pretty straight-forward.
I understand that it is not entirely applicable here to do the same because you have an additional case of null you want to handle. Handle it in a way that null signals not to use the API which I did not have. So, I am not asking for a refactor here but just stating that it took me a while to get behind the different levels of abstractions here.

src/sharedHelpers/sortingHelpers.ts
```diff
@@ -0,0 +1,87 @@
...
+  sortOptionId: SpeciesSortOptionId,
+): { order_by?: string; order?: string } | null {
```
jtklein: The return type of `order` is `SortDirection`. `order_by` has no type in `sorting.ts` but maybe we should add one and use here.

src/types/sorting.ts
```diff
@@ -0,0 +1,33 @@
+import type { RealmTaxon } from "realmModels/types";
+
+export type SortItemType = "observations" | "taxa";
+
+export type SortDirection = "asc" | "desc";
```
jtklein, line 5: Not really imported anywhere, but I think it should in sortingHelpers

src/sharedHelpers/sortingHelpers.ts
jtklein: Also the two keys are never undefined.

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -58,6 +58,7 @@ const MyObservationsContainer = ( ): React.FC => {
   const realm = useRealm( );
   const navigation = useNavigation( );
   const listRef = useRef<FlashListRef<RealmObservation>>( null );
+  const taxaListRef = useRef<FlashListRef<SpeciesCount>>( null );
```
jtklein, line 70: Same comment as in MyObsSimple about FlashListRef

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -252,9 +256,11 @@ const MyObservationsContainer = ( ): React.FC => {
   const numOfUserSpecies = zustandStorage.getItem( "numOfUserSpecies" );
   const [activeTab, setActiveTab] = useState( OBSERVATIONS_TAB );
 
-  let numTotalTaxaLocal: number | undefined;
-  const localObservedSpeciesCount: SpeciesCount[] = [];
-  if ( !currentUser ) {
+  const { leafTaxonIds, numTotalTaxaLocal } = useMemo( () => {
```
jtklein: My editor shows me that the inferred type for `leafTaxonIds` is const leafTaxonIds: never[] | number[] but I don't see how it can be never[].

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -271,18 +277,32 @@ const MyObservationsContainer = ( ): React.FC => {
...
-        localObservedSpeciesCount.push( { count: obs.length, taxon: obs[0].taxon } );
-      } );
+    return {
+      leafTaxonIds,
+      numTotalTaxaLocal: leafTaxonIds.length,
+    };
+  }, [currentUser, realm] );
+
+  const localObservedSpeciesCount = useMemo( () => {
+    if ( currentUser || activeTab !== TAXA_TAB || !leafTaxonIds.length ) {
```
jtklein, line 300: Could be  `!numTotalTaxaLocal` instead of `!leafTaxonIds.length`

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -340,6 +361,40 @@ const MyObservationsContainer = ( ): React.FC => {
...
+      ? remoteObservedTaxaCounts
+      : localObservedSpeciesCount;
+
+    // For logged-in users: we get data sorted from the API
+    if ( currentUser && isConnected ) {
+      return unsortedTaxa || [];
+    }
+
+    // For logged-out users: apply client-side sorting to local data
+    return sortSpeciesCounts( unsortedTaxa || [], speciesSortOptionId );
```
jtklein, line 387: Ah, I get it now, that's why we need an offline alert only when logged in.

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -340,6 +361,40 @@ const MyObservationsContainer = ( ): React.FC => {
...
+  useEffect( () => {
+    if ( activeTab === TAXA_TAB ) {
+      // Scroll to top
+      if ( taxaListRef.current ) {
+        taxaListRef.current.scrollToOffset( { offset: 0, animated: true } );
+      }
+
+      // Refetch data with new sort params for logged-in users
+      if ( currentUser ) {
+        refetchTaxa( );
```
jtklein: I think, this is redundant because when speciesSortOptionId changes the useQuery key also changes and hence taxa are refetched anyway.
Or why is this here? If you want to react to some kind of user action, like to the setState of speciesSortOptionId or to the tab being changed we should call refetch directly and not through a useEffect.

src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -340,6 +361,40 @@ const MyObservationsContainer = ( ): React.FC => {
...
+    return sortSpeciesCounts( unsortedTaxa || [], speciesSortOptionId );
+  }, [
+    currentUser,
+    isConnected,
+    remoteObservedTaxaCounts,
+    localObservedSpeciesCount,
+    speciesSortOptionId,
+  ] );
+
+  useEffect( () => {
```
jtklein: Can we try to remove this useEffect? In general I would like to have less in the codebase and not more.

src/components/MyObservations/MyObservationsContainer.tsx
abbeycampbell: I guess typescript was creating a union type here because I say that `leafTaxonIds` is an empty array in the early return, so it makes an inference that it's an array with no elements. 
Opted to type the return value to be explicitly an array of numbers.

src/components/MyObservations/MyObservationsContainer.tsx
abbeycampbell: Yeah that might be a little clearer, but I think I'm going to keep the current implementation because `localObservedSpeciesCount` depends on `leafTaxonIds` anyway. Changing it would mean an extra dependency in the array

src/components/MyObservations/MyObservationsContainer.tsx
abbeycampbell: yep! thank you for this callout. you're right about the refetch being unnecessary and it's probably better to handle the scroll where we do everything else on sort selection, we're already passing the ref down anyway

src/components/MyObservations/MyObservationsContainer.tsx
abbeycampbell: you're totally correct; I think this logic was a remnant of trying different approaches that I either forgot to remove/assumed was still necessary

src/components/MyObservations/MyObservationsSimple.tsx
abbeycampbell: I'm just going to remove this logic for now, since I didn't actually implement observations sorting and we don't even need to handle this case yet. I'll revisit when we actually tackle that.

src/components/SharedComponents/FlashList/CustomFlashList.tsx
abbeycampbell: I included these changes because calling `scrollToOffset` in `MyObservationsSimple` was causing a maximum callstack error on sort, but I like your idea to spread in all the methods and add `notifyDataFetched` on top

src/components/SharedComponents/Sheets/RadioButtonSheet.tsx
abbeycampbell: Anything else off the top of your head besides https://linear.app/inaturalist/issue/MOB-1106/scrolling-on-a-bottom-sheet-with-a-list-should-not-close-the-bottom?

I looked around but didn't see anything obvious

src/sharedHelpers/sortingHelpers.ts
abbeycampbell: Heard and I do agree that your approach is simpler to reason about. 
Since I have two approaches here that both use the SORT_MAP (API request and client-side sorting), and eventually we'll have more cases to handle as well, I wanted to avoid duplicating the logic between the two. I'm going to leave this as it is for now, but I definitely don't want to be over-engineering anything. So please don't hesitate to call it out if you feel like it would be a big improvement to do it differently.

src/components/SharedComponents/Sheets/SortSheet.tsx
abbeycampbell: agree this is a better approach

src/components/MyObservations/MyObservationsSimple.tsx
abbeycampbell: will resolve this when we actually add the sort observations logic

src/components/MyObservations/MyObservationsSimple.tsx
abbeycampbell: I think you got this, but I'll explain for posterity: I decided that since a logged out user has all their data saved locally anyways, we can sort that and show what's available. maybe it's a little crazy to be going out of the way to provide features like this to folks who are logged out, but my thought was that eventually, we could use this same logic for logged-in users who are offline/have a bad connection: sort and show what they have available locally, but with the addition of some kind of toast message or call out that they may not be seeing a fully sorted list of all their observations/species.

src/components/SharedComponents/Sheets/RadioButtonSheet.tsx
jtklein: Yes, I think that's what I remembered.

PR #3318, opened 2026-01-09
MOB-721 Consolidate Obs details components

sepeterson:
Closes [MOB-721](https://linear.app/inaturalist/issue/MOB-721/consolidate-duplicate-code-between-obsdetailsdefaultmode-and)

Mostly just moving stuff around but some files with functional changes are:

- CommunitySection.js and the showStatus prop thru ActivityItem components
- Attribution.tsx to account for slight visual differences between default and advanced mode
- A couple instances where optional boolean props were typed as such

PR #3327, opened 2026-01-16
Make the two ObsDetails container a bit more similar

jtklein:
Looking at https://github.com/inaturalist/iNaturalistReactNative/pull/3318 I re-realized that also the two container share a bunch of code.
Here is one small change to make them more similar so that we can focus on what to de-duplicate at one point.


FLGMwt (approved):
LGTM

PR #3328, opened 2026-01-16
remove instances of forwardRef

abbeycampbell:
closes [MOB-972](https://linear.app/inaturalist/issue/MOB-972/remove-usage-of-forwardref)

### DIFF
--- a/src/components/ObservationsFlashList/ObservationsFlashList.js
+++ b/src/components/ObservationsFlashList/ObservationsFlashList.js
@@ -52,6 +52,9 @@ type Props = {
   onEndReached: Function,
   onLayout?: Function,
   onScroll?: Function,
+  // this ref is being forwarded to the underlying CustomFlashList and used as an imperative handle
+  // so the parent can control behavior like scrolling; it's typed as Function because there's not
+  // a good way to capture this otherwise with Flow.
   ref?: Function,
   renderHeader?: Function,
   showNoResults?: boolean,


src/components/ObservationsFlashList/ObservationsFlashList.js
```diff
@@ -53,13 +52,14 @@ type Props = {
   onEndReached: Function,
   onLayout?: Function,
   onScroll?: Function,
+  ref?: Function,
```
FLGMwt, line 58: This particular one was harder to pull the thread on grokking downstream. If (and only if) you happened to get an understanding of the code flow here while you made the change, would you mind adding a comment here giving context for what this refs role is? It just stood out to me as odd as a `Function` and I'm not confident on its use, e.g., https://github.com/inaturalist/iNaturalistReactNative/blob/b13170c53412d19cdbcc2da2c944256d796ad2f3/src/components/SharedComponents/FlashList/CustomFlashList.tsx#L123-L126

I understand though that this change doesn't necessarily involve that so no worries if not!

abbeycampbell: the ref is just being forwarded through `ObservationsFlashList` to the underlying `CustomFlashList` and used as an imperative handle so the parent can control behavior like scrolling; I typed it as `Function` because there's not really a good way to capture this otherwise with Flow. 

Did you want a comment pertaining to the change here in `ObservationsFlashList`, or the logic in `CustomFlashList`? I actually made a change to that file in my sorting PR (added an internalRef that points to the actual FlashList instance), so I could add an explanation in that PR if you like?

PR #3331, opened 2026-01-21
support log apis extra property

FLGMwt:
Closes MOB-730

EDIT: I pulled in the additional enhancement changes from #3338 into this branch, and updating the description accordingly:

The approach is detailed in the following introduced comment in `src/api/log/enhanceLoggerWithExtra.ts`: 

```
tl;dr: to support `extra`, we "wrap" certain input to react-native-logs and then "unwrap" it
when handling it in our iNat-specific log handler

iNat's /log REST API supports an `extra` field for sending structured data
in addition to the log message.
The logging API exposed by react-native-logs is entirely based on rest params,
making it difficult for our log transport to identify "what" is intented to be treated as `extra`
since a function signature can't have any parameter _after_ a rest parameter.

We don't want to compromise our use of rn-logs for other transports (console, fs, future others)
BUT we also want to make it so that our "special" case for the iNat transport is
1) user-discoverable 2) safe (or as safe as we can get)

The following is probably overkill, but it makes me feel better about hiding the "magic"
case for `extra` from developer code.

The following exposes a "enhancer" function which takes a rn-logs logger (from createLogger)
that looks like:

{
  debug: (...params) => void,
  info: (...params) => void,
  ... and so on
  extend: (namespace) => {
    debug: (...params) => void,
    info: (...params) => void,
    ... and so on
  }
  ...other logger helpers
}

and returns that logger with the additional `[level]WithExtra` functions:

{
  debug: (...params) => void,
  debugWithExtra: (...params) => void,
  info: (...params) => void,
  infoWithExtra: (...params) => void,
  ... and so on
  extend: (namespace) => {
    debug: (...params) => void,
    debugWithExtra: (...params) => void,
    info: (...params) => void,
    infoWithExtra: (...params) => void,
    ... and so on
  }
  ...other logger helpers
}

The `[log]WithExtra`s are passthrus to the default `[log]`s but with some runtime validation
to make sure the last param is acceptable for iNat's `extra` field. They also wrap that argument
in an object with an internal / "magic" property name. This is to allow the transport to more
safely interpret a log param as being intented as an extra since the transport handles generic
input. That is:
log.infoWithExtra(1, 2, 3, { userId: 0 })
becomes
[1, 2, 3, { [magicExtraKey]: { userId: 0 }}]
The trasport looks at its args, sees if the last one is a `{ [magicKey]: {} }` and if so,
separates that value from the rest of the log args as if it was indeed passed separately. The log
API would receive: { message, "1 2 3", extra: { userId: 0 } }

Other transports, like the console, will treat this wrapped value as any other log argument.
```

### DIFF
--- a/react-native-logs.config.ts
+++ b/react-native-logs.config.ts
@@ -1,4 +1,4 @@
-import iNatLogstashTransport from "api/log";
+import iNatLogstashTransport, { enhanceLoggerWithExtra } from "api/log";
 import RNFS from "react-native-fs";
 import {
   consoleTransport,
@@ -24,7 +24,7 @@ const sharedConfig = {
   },
 };
 
-const log = logger.createLogger( {
+const baseLog = logger.createLogger( {
   ...sharedConfig,
   transport: [
     consoleTransport,
@@ -32,6 +32,13 @@ const log = logger.createLogger( {
     iNatLogstashTransport,
   ],
 } );
+
+// given the general react-native-logs logger with debug(), warn(), etx,
+// add wrappers for debugWithExtra(), etc which provides more intentional
+// log interfaces for the iNatLogstash API `extra` proprty. `iNatLogstashTransport`
+// handles this extra data specially while other transports treat it as any other normal
+// log param.
+const log = enhanceLoggerWithExtra( baseLog );
 
 const logWithoutRemote = logger.createLogger( {
   ...sharedConfig,
--- a/src/api/log/index.ts
+++ b/src/api/log/index.ts
@@ -7,6 +7,8 @@ import type { transportFunctionType } from "react-native-logs";
 import { getInstallID } from "sharedHelpers/installData";
 import { isObject, isObjectWithPrimitiveValues } from "sharedHelpers/runtimeTypeUtil";
 
+import { extraSentinelKey } from "./enhanceLoggerWithExtra";
+
 const API_HOST: string
     = Config.API_URL || process.env.API_URL || "https://api.inaturalist.org/v2";
 
@@ -29,9 +31,9 @@ function isError( value: unknown ): value is { message?: string; stack?: string 
   return false;
 }
 
-// if we have anything that looks like:
-// [someObj, 'asdfasdf', 3, { extra: { id: 1, ... } }]
-// where the _last_ rest param is an obj w/ exaclty one `extra` property w/ primitive fields,
+// If we have anything that looks like:
+// [someObj, 'asdfasdf', 3, { [extraSentinelKey]: { id: 1, ... } }]
+// where the _last_ rest param is an obj w/ exactly this sentinel property w/ primitive fields,
 // we infer that last item as intended for the special `extra` API field
 // we return that separately and strip it from the "normal" rest params for later handling
 function extractExtra( rawMsg: unknown ) {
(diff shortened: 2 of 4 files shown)


src/api/log.ts
```diff
@@ -17,23 +18,75 @@ const api = create( {
...
+// at least answers: does it look enough like an Error for logging purposes?
+function isError( value: unknown ): value is { message?: string; stack?: string } {
+  if ( value instanceof Error ) {
+    return true;
+  }
+  if ( isObject( value ) && "stack" in value && "message" in value ) {
+    return true;
+  }
   return false;
 }
```
FLGMwt, line 24: These changes weren't stylistic, they help narrow types which resolved some type errors.

src/api/log/index.ts
```diff
@@ -17,23 +18,75 @@ const api = create( {
...
+    extra: extraCandidate,
+  };
+}
+
+// our transport has no options but this needs to be explicitly `object` for generic typing
+type iNatLogstashTransportOptions = object;
+
 // Custom transport for posting to iNat API logging
-const iNatLogstashTransport: transportFunctionType = async props => {
+const iNatLogstashTransport: transportFunctionType<iNatLogstashTransportOptions> = async props => {
```
FLGMwt, line 76: This one _requires_ a generic type argument which drives the `transportConfig` at the `createLogger` level. We don't have any config for this transport and `object` is the best way to represent `{}` (explicitly not `Record` in this case)

src/api/log.ts
```diff
@@ -47,28 +100,33 @@ const iNatLogstashTransport: transportFunctionType = async props => {
...
+      // eslint-disable-next-line prefer-destructuring
+      message = last.message;
       errorType = last.constructor?.name;
       backtrace = last.stack;
     } else {
-      message = props.rawMsg.join( " " );
+      message = messageParams.join( " " );
     }
   } else {
-    message = props.rawMsg;
```
FLGMwt, line 66: No functional changes here, just switching from using `props.rawMsg` to the resolved local var. Also one switch to `// eslint-disable-next-line prefer-destructuring` for consistency

react-native-logs.config.ts
FLGMwt, line 1: Sorry, this got marked as a new file. No functional changes here just:
- switch to .ts
- added a `TS TODO` comment
- removed local `transport` var and passed it inline to make `log` and `logWithoutRemote` defined exactly the same way.

abbeycampbell:
is there a reason why this is marked draft? Are there more changes coming?

src/api/log/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    return true;
+  }
+  if ( isObject( value ) && "stack" in value && "message" in value ) {
+    return true;
+  }
+  return false;
+}
+
+// If we have anything that looks like:
+// [someObj, 'asdfasdf', 3, { [extraSentinelKey]: { id: 1, ... } }]
```
jtklein, line 35: We will not accept any random object as message, so the comment might be misleading.
If you want an object as first item it has to be an "error" as defined by isError, or you'd need to use JSON.stringify(yourObject) before passing in here. Otherwise on the server we see a "[object Object]" log.

src/api/log/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  // if message is an Error or is an array ending in an error, extract
+  // error_type and backtrace
+  let message;
+  let backtrace;
+  let errorType;
+  if ( typeof ( messageParams ) === "string" ) {
+    message = messageParams;
+  } else if ( isError( messageParams ) ) {
+    // eslint-disable-next-line prefer-destructuring
+    message = messageParams;
```
jtklein, line 109: I don't understand this diff yet. How can we switch from destructuring to reassigning message here?

jtklein, line 109: See comment in the deleted file.
Does this not need to be `message = messageParams.message`?

jtklein (approved):
Looks pretty good. Seems solid but because it is not user facing I did more of a squinty-eye review on this one.
Anyhow, except one line of which I am unsure everything else is good to go, and urgently needed.

PR #3333, opened 2026-01-22
MOB-1089 fix id with comment

sepeterson:
In this PR:

- Proper onPress for comment with ID
- Add prop for BottomSheet that truly is only called when it's dismissed via close or background tap
- Discard ID only when closing sheet from user input
- Added some tests around this bug
- Sorry for the big diff in SuggestIDSheet. Shouldn't be any major changes hidden in there, just fixing a single parent error

### DIFF
--- a/src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
+++ b/src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
@@ -55,7 +55,7 @@ interface Identification {
 interface IdentState {
   comment: string | null;
   commentIsOptional: boolean;
-  identBodySheetShown: boolean;
+  showIdentBodySheet: boolean;
   newIdentification: Identification | null;
   showPotentialDisagreementSheet: boolean;
   showSuggestIdSheet: boolean;
@@ -76,7 +76,7 @@ type IdentAction =
 const initialIdentState: IdentState = {
   comment: null,
   commentIsOptional: false,
-  identBodySheetShown: false,
+  showIdentBodySheet: false,
   newIdentification: null,
   showPotentialDisagreementSheet: false,
   showSuggestIdSheet: false,
@@ -122,19 +122,19 @@ export const identReducer = ( state: IdentState, action: IdentAction ): IdentSta
     case SHOW_EDIT_IDENT_BODY_SHEET:
       return {
         ...state,
-        identBodySheetShown: true,
+        showIdentBodySheet: true,
       };
     case HIDE_EDIT_IDENT_BODY_SHEET:
       return {
         ...state,
-        identBodySheetShown: false,
+        showIdentBodySheet: false,
       };
     case SUBMIT_IDENTIFICATION:
       return {
         ...state,
         showPotentialDisagreementSheet: false,
         showSuggestIdSheet: false,
-        identBodySheetShown: false,
+        showIdentBodySheet: false,
         newIdentification: null,
         identTaxon: null,
       };
@@ -195,7 +195,7 @@ const IdentificationSheets: React.FC<Props> = ( {
   const {
     comment,
     commentIsOptional,
-    identBodySheetShown,
+    showIdentBodySheet,
     identTaxon,
     newIdentification,
     showPotentialDisagreementSheet,
@@ -423,7 +423,7 @@ const IdentificationSheets: React.FC<Props> = ( {
         <AgreeWithIDSheet
           onAgree={onAgree}
           editIdentBody={editIdentBody}
-          hidden={identBodySheetShown}
+          hidden={showIdentBodySheet}
           onPressClose={closeAgreeWithIdSheet}
           identification={agreeIdentification}
         />
@@ -440,7 +440,7 @@ const IdentificationSheets: React.FC<Props> = ( {
         />
       )}
       {/* Sheet for adding comment w/ ID */}
-      {identBodySheetShown && (
+      {showIdentBodySheet && (
         <TextInputSheet
           buttonText={t( "CONFIRM" )}
           onPressClose={onCloseIdentBodySheet}
@@ -453,7 +453,7 @@ const IdentificationSheets: React.FC<Props> = ( {
       {showSuggestIdSheet && (
         <SuggestIDSheet
           editIdentBody={editIdentBody}
-          hidden={identBodySheetShown}
+          hidden={showIdentBodySheet}
           onSuggestId={onSuggestId}
           identification={newIdentification}
           onPressClose={hideSuggestedIdSheet}
--- a/src/components/SharedComponents/Sheets/BottomSheetV2.tsx
+++ b/src/components/SharedComponents/Sheets/BottomSheetV2.tsx
@@ -7,19 +7,22 @@ import classnames from "classnames";
 import { BottomSheetStandardBackdrop, Heading4, INatIconButton } from "components/SharedComponents";
 import { View } from "components/styledComponents";
 import React, { useCallback, useEffect, useRef } from "react";
-import { Dimensions } from "react-native";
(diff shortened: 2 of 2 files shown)


src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -33,15 +33,16 @@ interface Props {
...
-  // closes
+  // OnPressClose *does* get called whenever bottom sheet is dismissed
   onPressClose?: ( ) => void;
   snapPoints?: string[];
   insideModal?: boolean;
   keyboardShouldPersistTaps?: string;
   testID?: string;
   containerClass?: string;
   scrollEnabled?: boolean;
+  // Use this prop for side effects of tapping backdrop or close icon
```
jtklein: See general comment, why not dismiss via swipe here?

src/components/ObsDetailsSharedComponents/Sheets/SuggestIDSheet.tsx
```diff
@@ -16,86 +15,84 @@ interface Props {
...
   };
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
-  onSuggestId:Function;
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
-  editIdentBody: Function;
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
-  onPressClose: Function;
+  onSuggestId: ( ) => void;
+  editIdentBody: () => void;
+  onBackgroundOrCloseIconPress?: ( ) => void;
```
jtklein: Is not undefined in any of the instances this sheet is used.

src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -33,15 +33,16 @@ interface Props {
   hideCloseButton?: boolean;
   headerText?: string;
   onLayout?: ( event: object ) => void;
-  // Callback when the user presses the close button or backdrop, not whenever the sheet
```
jtklein, line 36: This comment might have been misleading because onPress was also used in onDismiss, so every time the modal dismissed.

src/components/SharedComponents/Sheets/BottomSheet.tsx
```diff
@@ -66,6 +68,11 @@ const StandardBottomSheet = ( {
   const sheetRef = useRef<BottomSheet>( null );
   const insets = useSafeAreaInsets( );
 
+  const handleCloseFromUserInput = useCallback( ( ) => {
+    if ( onBackgroundOrCloseIconPress ) onBackgroundOrCloseIconPress( );
+    if ( onPressClose ) onPressClose( );
```
jtklein: `onPressClose` is now called twice per dismissal. Once here and then immediately after through `onDismiss`.

tests/unit/components/ObsDetails/Sheets/SuggestIDSheet.test.js
```diff
@@ -0,0 +1,85 @@
+import { fireEvent, screen } from "@testing-library/react-native";
```
jtklein, line 1: This test looks good.
You are doing byTestID and byText on the same component several times. Probably not necessary but I also kinda like it because it makes sure we have the testID and can find it in e2e tests if we ever need.

tests/unit/components/ObsDetailsDefaultMode/IdentificationSheets.test.js
```diff
@@ -0,0 +1,199 @@
...
+      await waitFor( () => {
+        expect( fetchTaxonAndSave ).toHaveBeenCalledWith(
+          mockTaxon.id,
+          mockRealm,
+        );
+      } );
+    } );
+  } );
+
+  describe( "identReducer", () => {
```
jtklein, line 139: Is there a better way to test the reducer without exporting it solely for testing purposes?

src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
```diff
@@ -446,14 +448,14 @@ const IdentificationSheets: React.FC<Props> = ( {
           headerText={addCommentHeaderText}
           textInputStyle={textInputStyle}
           initialInput={newIdentification?.body}
-          confirm={onChangeIdentBody}
+          confirm={onSuggestIdWithComment}
```
jtklein: This sheet should behave as in the advanced mode here. Pressing CONFIRM to open the SuggestIDSheet back up with the comment in grey.

sepeterson:
> "Add prop for BottomSheet that truly is only called when it's dismissed via close or background tap" Could you expand on the significance of this? I am not sure why we need that. Also, how is this different than when the user swipes the bottom sheet down and dismisses it that way. This also is a user action to dismiss the sheet, why differ it from pressing the close button and backdrop? Would we not need to update all usages of bottom sheet then with the new prop? If I am not mistaken, what I think you wanted to acchieve could have been implemented without renaming onPressClose but instead just removing onDismiss.

The problem was onPressClose was getting called when one of the sheets closed to open the next one, which discarded the in-flight identification. The trouble with removing onDismiss is we don't know if other usages of BottomSheet rely on the previous behavior so creating a new prop is less risky. I also didn't mean to exclude swiping down to close -- I just wasn't aware you could do that 😅 

Fortunately I was able to remove the BottomSheet changes with my second attempt

> The flow in the advanced mode is now correct. Pressing "CONFIRM" on the add-comment-to-id sheet opens the suggest ID sheet back up with the comment body in the grey box above. In the default mode the flow is now broken, pressing "CONFIRM" sends the ID away immediately and I do see an Error there.

Both should behave the same now

> I am a bit confused, I would have asked to split up these two concerns, the refactor and the actual bugfix into separate PRs, but I don't actually see any changes in the advanced mode other than the sheet dismissal changes. So, I am not even sure how the bug got fixed exactly.

I am very sorry but I made this even worse. As I implemented "minimal" fixes it created new bugs until I was making major changes to the component. Happy to jump on a call and explain some of the changes if it'd be helpful!

tests/unit/components/ObsDetailsDefaultMode/IdentificationSheets.test.js
sepeterson: I think the most correct thing would be to just test the behavior of the component. If you'd prefer not to export the reducer, I'm okay with removing these tests.

jtklein (changes requested):
Yes, absolutely loving the refactor of removing sheets and re-using IdentificationSheets here. Everything seems okay to me on the refactor part.
All code changes look solid to me.
Also this fixes the bug with ID with comment, however, it does introduce new bug:
If you X out of a SUGGEST ID sheet and press “SUGGEST ID” once again, then choose any taxon on the next screen the sheet will not open back up again. Tested in Advanced mode but shouldn’t make a difference anymore.

Could you please also do this one minor change to rename identBodySheetShown to showIdentBodySheet, and sort it into the keys.

src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
```diff
@@ -70,12 +70,13 @@ type IdentAction =
...
   | { type: "SHOW_POTENTIAL_DISAGREEMENT_SHEET" }
-  | { type: "SUBMIT_IDENTIFICATION" };
+  | { type: "SUBMIT_IDENTIFICATION" }
+  | { type: "HIDE_SUGGESTED_ID_SHEET" };
 
 const initialIdentState: IdentState = {
   comment: null,
   commentIsOptional: false,
-  identBodySheetShown: false,
+  showIdentBodySheet: false,
```
jtklein, line 79: Thanks for that!

PR #3334, opened 2026-01-23
MOB-968 Match screen edit icon

sepeterson:
closes [MOB-968](https://linear.app/inaturalist/issue/MOB-968/add-edit-icon-to-top-right-of-match-screen)

- Add edit icon to match screen
- When coming from Match, assign topSuggestion taxon to zustand observation

### DIFF
--- a/src/components/ObsEdit/ObsEdit.js
+++ b/src/components/ObsEdit/ObsEdit.js
@@ -36,7 +36,6 @@ const ObsEdit = ( ): Node => {
   const [passesEvidenceTest, setPassesEvidenceTest] = useState( false );
   const [resetScreen, setResetScreen] = useState( false );
   const [needLocation, setNeedLocation] = useState(
-    // take a look 👀
     shouldFetchObservationLocation( currentObservation ),
   );
   const isFocused = useIsFocused( );


src/sharedHooks/useNavigateToObsEdit.ts
```diff
@@ -1,17 +1,34 @@
...
-  function navigateToObsEdit( localObservation: RealmObservation ) {
+  function navigateToObsEdit(
+    localObservation: RealmObservation,
+    lastScreen?: string,
+    taxon?: ApiTaxon | RealmTaxon,
+  ) {
     prepareObsEdit( localObservation );
-    navigation.navigate( "ObsEdit" );
+    if ( taxon ) {
+      updateObservationKeys( {
```
jtklein, line 21: I do like this approach here. Set a taxon if we have one.
Looking at your PR in isolation, I would have requested one minor change here so that owners_identification_from_vision represents if the user chose a cv suggestion or not, so is only true when a user selects the top match, or any of the items in additional match.
I think that:
It needs to be set to false whenever the user makes a decided choice against the cv suggestions.
For the case of a user pressing on "LEARN MORE ABOUT THIS SPECIES" and choosing a non cv alternative over a cv one we do not take care here because in the current flow that saves the  observation directly to the Me tab.
However, in the case of the user selecting one of the "Do you know what group this is in?" we'd need to set this to false.
In other words, using `taxonToSave` means `owners_identification_from_vision: true` and `iconicTaxon` means `owners_identification_from_vision: false`.

However, I can see that in `handleSaveOrDiscardPress` we are doing the same as here. So, I do not request a change here but we should make a follow-up investigation if that is correct. I believe it is not.

sepeterson: Good catch! Created a ticket for this: https://linear.app/inaturalist/issue/MOB-1167/correctly-track-owners-identification-from-vision-through-match-screen

PR #3336, opened 2026-01-26
Update react-native to 0.81

jtklein:
Closes MOB-1119

Compiled and tested in Debug and Release mode on iOS and Android.

There was a compile time error in Android for react-native-audio-recorder-player. I decided to make a patch because this package has a complicated migration to v4 and no v3 version actually seems to fix this error.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -36,7 +36,7 @@ PODS:
     - React-Core
     - React-RCTFabric
     - ReactCommon/turbomodule/core
-  - FBLazyVector (0.80.2)
+  - FBLazyVector (0.81.5)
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -21,7 +21,6 @@
     },
     "types": ["nativewind/types"],
   },
-  "exclude": [
-    "flow-typed"
-  ]
+  "include": ["**/*.ts", "**/*.tsx"],
+  "exclude": ["flow-typed", "**/node_modules", "**/Pods"]
```
abbeycampbell, line 25: just a question: this will silence ts errors in .js files, right?

PR #3342, opened 2026-01-28
Update react-native-camera-roll version and replace our fork with a patch

jtklein:
Closes MOB-1140

An update of this package is potentially required for updating Seek to react-native 0.81.
Since our fork of this package is only slightly deviating from the npm package, indeed only adding functionality there, I'd like to replace this fork with a patch, so that we have one less fork to maintain.


FLGMwt (approved):
👍 nice, I like the idea to switch to a patch for a reasonably small change.

PR #3346, opened 2026-01-30
Update react-native-sensitive-info to 5.6.2

jtklein:
Closes MOB-949
Closes MOB-828

I think this is ready now. Left inline comments for why I made some changes.
Compiled and tested in Debug and Release mode on Android and iOS.
On iOS, I tested with fresh install but also with updating a prior install with a build made from this PR.

### DIFF
--- a/src/components/LoginSignUp/AuthenticationService.ts
+++ b/src/components/LoginSignUp/AuthenticationService.ts
@@ -70,7 +70,18 @@ const clearAuthCache = ( ): void => {
 };
 
 async function getSensitiveItem( key: string, options = {} ) {
-  const exists = await RNSInfo.hasItem( key, options );
+  let exists;
+  try {
+    exists = await RNSInfo.hasItem( key, options );
+  } catch ( e ) {
+    if ( isSensitiveInfoError( e ) ) {
+      const hasItemError = e as SensitiveInfoError;
+      localLogger.info(
+        `RNSInfo.hasItem error for ${key}: ${hasItemError.message}`,
+      );
+    }
+    throw e;
+  }
   if ( !exists ) {
     return null;
   }


src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -69,42 +70,51 @@ const clearAuthCache = ( ): void => {
...
+  }
+
   try {
     return await RNSInfo.getItem( key, options );
   } catch ( e ) {
-    if ( isDebugMode() ) {
-      localLogger.info( `RNSInfo.getItem not available for ${key}, sleeping` );
-    }
-    const getItemError = e as Error;
-    if ( getItemError.message.match( /Protected data not available yet/ ) ) {
```
jtklein, line 79: We were hitting this error very frequently and our first attempt in fixing it was to sleep for a second and try again. However, the error message was a bit misleading as this was happening when the phone was locked. See discussion here: https://linear.app/inaturalist/issue/MOB-868/add-more-logging-around-getsensitiveinfo
This has been fixed in the library, so I just remove our sleep and retry here.

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -69,42 +70,51 @@ const clearAuthCache = ( ): void => {
...
     }
-    throw getItemError;
+    throw e;
   }
 }
 
 async function setSensitiveItem( key: string, value: string, options = {} ) {
+  const actualOptions = {
+    ...options,
+    accessControl: "none" as const,
```
jtklein, line 113: The library switched to use biometrics (Touch ID on iPhones) by default so we disable it here.

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -619,7 +622,7 @@ const registerUser = async ( user: { password: string } ) => {
 };
 
 const isCurrentUser = async ( username: string ): Promise<boolean> => {
-  const currentUsername = await getUsername( );
```
jtklein, line 622: `getUsername` was not actually imported anywhere else so I refactored to a local get here.

__mocks__/react-native-sensitive-info.ts
```diff
@@ -37,6 +37,16 @@ class RNSInfo {
     clearAuthCache();
   } );
 
+  static hasItem = jest.fn( async ( k, o ) => {
```
jtklein, line 40: Note to myself: Double check that the library update did not introduce mocks by itself so we could delete this file instead of adding our own mock here.

jtklein: Checked now, does not seem it has mocks looking at this file here: https://github.com/mCodex/react-native-sensitive-info/compare/v6.0.0-alpha9...v5.6.2?diff=unified&w#diff-8881961dc913eabb59bf9bbd8d06147865e4db2399f71578f1a54131a99a0195R167

patches/react-native-sensitive-info+6.0.0-alpha.9.patch
FLGMwt, line 1: yoooooooo! the only thing better than deleted code is a deleted fork/patch 🎉

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -69,42 +70,51 @@ const clearAuthCache = ( ): void => {
 };
 
 async function getSensitiveItem( key: string, options = {} ) {
+  const exists = await RNSInfo.hasItem( key, options );
```
FLGMwt: does the `hasItem` check not need the same error handling as `getItem`?

src/components/LoginSignUp/AuthenticationService.ts
FLGMwt: Confirming my understanding, would it be correct to say:

A. the package update changed its behavior
B. `accessControl: "none"` gives us our previous behavior
C. the behavior of rn-sensitive-info is unchanged in our app with this PR?

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -285,7 +285,7 @@ const getJWT = async (
   allowAnonymousJWT = false,
   logContext: string | null = null,
 ): Promise<string | null> => {
-  let jwtToken: string | undefined = await getSensitiveItem( "jwtToken" );
+  let jwtToken: string | null | undefined = await getSensitiveItem( "jwtToken" );
```
FLGMwt, line 299: nit: did the package update influence this or is this a factor of our internally inconsistent types? non-urgent, I just want to know if I should add this to my running list of `undef`/`null` nits we have

jtklein: The return signature of `RNSInfo.getItem` changed with the package update from `Promise<string>` to `Promise<string | null>`, hence also our wrapper `getSensitiveItem` changed. So, at this line we have either string or now possible also null as type. `undefined` stays because potentially further down below we receive undefined from the api in `jwtToken = response.data?.api_token`.

src/components/LoginSignUp/AuthenticationService.ts
jtklein: Correct.
The package update changed the default behaviour of calling setItem to use biometrics.
By adding "none" we stay in our previous behaviour of using rn-sensitive-info.

src/components/LoginSignUp/AuthenticationService.ts
jtklein: Good catch. Have added one now.

PR #3353, opened 2026-02-03
disable pan down gestures for PickerSheet

abbeycampbell:
closes [MOB-1106](https://linear.app/inaturalist/issue/MOB-1106/scrolling-on-a-bottom-sheet-with-a-list-should-not-close-the-bottom)

Pickers and bottom sheets don't play nice together.

https://github.com/user-attachments/assets/ce18a6e0-a0a3-45b5-aa5c-a20aa982d885

You can see there is still a bit of a bounce when scrolling down, but you can no longer dismiss the sheet this way--this is how the UI acts for the Explore Filter picker sheets already, so at least the experience is now consistent.

You can also see there is an issue with the scroll animation being duplicated.
I created a separate issue for that here: https://linear.app/inaturalist/issue/MOB-1165/duplicative-scroll-when-selecting-a-language

### DIFF
--- a/src/components/SharedComponents/Sheets/BottomSheet.tsx
+++ b/src/components/SharedComponents/Sheets/BottomSheet.tsx
@@ -42,6 +42,7 @@ interface Props {
   containerClass?: string;
   scrollEnabled?: boolean;
   enablePanDownToClose?: boolean;
+  enableContentPanningGesture?: boolean;
 }
 
 const StandardBottomSheet = ( {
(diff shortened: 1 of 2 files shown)


jtklein (approved):
Looks good! But I think we can get rid of the "bounce" by also using `enableContentPanningGesture={false}` but to be consistent with other places we use a picker in a sheet is good. Maybe change in all places?

Like the fact that we have a different ticket for a different concern now.

PR #3356, opened 2026-02-05
show offline/retry state for notifications when request takes too long

abbeycampbell:
closes [MOB-187](https://linear.app/inaturalist/issue/MOB-187/notifications-should-time-out-when-theres-low-connectivity)

Used a 5 second timeout which felt about right to me and was on the outer edge of being testable with the 'Very Bad Network' preset; sometimes notifications loaded and sometimes I got the offline state.

Opted to put this logic in the `useInfiniteNotificationsScrollHook` so that it was more similar to how we handle suggestions, but I could also see an argument for doing it in the component, since it's kind of presentational--essentially we're saying "you've seen this spinner for long enough, maybe try again later?" Doing it in the hook lets us actually cancel the query though. 

As an aside, it seems like we were eventually showing a `"something went wrong"` if the request takes long enough. In `useAuthenticatedInfiniteQuery`, we have some retry logic that will will eventually make `isError` true. In any case, this is a better experience I think.

### DIFF
--- a/src/sharedHooks/useInfiniteNotificationsScroll.ts
+++ b/src/sharedHooks/useInfiniteNotificationsScroll.ts
@@ -102,7 +102,14 @@ const useInfiniteNotificationsScroll = (
     [notificationParams],
   );
 
-  const infQueryResult = useAuthenticatedInfiniteQuery(
+  const {
+    data,
+    isFetching,
+    isInitialLoading,
+    isError,
+    refetch,
+    fetchNextPage,
+  } = useAuthenticatedInfiniteQuery(
     queryKey,
     async ( { pageParam }: { pageParam: number }, optsWithAuth: ApiOpts ) => {
       const params = { ...BASE_PARAMS, ...notificationParams };
@@ -155,19 +162,19 @@ const useInfiniteNotificationsScroll = (
   // We want to timeout and show an offline/retry state if this request takes too long
   useEffect( () => {
     // Reset if we get data
-    if ( infQueryResult.data !== undefined && !infQueryResult.isFetching ) {
+    if ( data !== undefined && !isFetching ) {
       setLoadingTimedOut( false );
       return undefined;
     }
 
     // Don't set timer if not loading
-    if ( !infQueryResult.isFetching ) {
+    if ( !isFetching ) {
       return undefined;
     }
 
     // Set a timeout and cancel the query if we hit the limit
     const timer = setTimeout( () => {
-      if ( infQueryResult.data === undefined && infQueryResult.isFetching ) {
+      if ( data === undefined && isFetching ) {
         queryClient.cancelQueries( { queryKey } );
         setLoadingTimedOut( true );
       }
@@ -175,26 +182,26 @@ const useInfiniteNotificationsScroll = (
 
     // eslint-disable-next-line consistent-return
     return () => clearTimeout( timer );
-  }, [infQueryResult.data, infQueryResult.isFetching, queryKey, queryClient] );
+  }, [data, isFetching, queryKey, queryClient] );
 
   // Reset when user manually retries
   useEffect( () => {
-    if ( infQueryResult.isFetching ) {
+    if ( isFetching ) {
       setLoadingTimedOut( false );
     }
-  }, [infQueryResult.isFetching] );
+  }, [isFetching] );
 
   return {
-    refetch: infQueryResult.refetch,
-    isError: infQueryResult.isError,
-    isFetching: infQueryResult.isFetching,
-    isInitialLoading: infQueryResult.isInitialLoading,
+    refetch,
+    isError,
+    isFetching,
+    isInitialLoading,
     loadingTimedOut,
     // Disable fetchNextPage if signed out
     fetchNextPage: currentUser
-      ? infQueryResult.fetchNextPage
+      ? fetchNextPage
       : ( ) => undefined,
-    notifications: flatten( infQueryResult?.data?.pages ),
+    notifications: flatten( data?.pages ),
   };
 };
 
--- a/src/sharedHooks/useInfiniteNotificationsScroll.ts
+++ b/src/sharedHooks/useInfiniteNotificationsScroll.ts
@@ -1,3 +1,4 @@
+import { useQueryClient } from "@tanstack/react-query";
 import { fetchObservationUpdates, fetchRemoteObservations } from "api/observations";
 import type {
   ApiNotification,
@@ -7,11 +8,14 @@ import type {
 } from "api/types";
 import flatten from "lodash/flatten";
 import { RealmContext } from "providers/contexts";
+import { useEffect, useMemo, useState } from "react";
 import type Realm from "realm";
 import Observation from "realmModels/Observation";
 import { useAuthenticatedInfiniteQuery, useCurrentUser } from "sharedHooks";
 
 const { useRealm } = RealmContext;
+
+const LOADING_TIMEOUT = 5000;
 
 // Extends API response with data we need in this app
 export interface Notification extends ApiNotification {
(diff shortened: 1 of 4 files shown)


src/sharedHooks/useInfiniteNotificationsScroll.ts
```diff
@@ -142,11 +152,44 @@ const useInfiniteNotificationsScroll = (
...
+    const timer = setTimeout( () => {
+      if ( infQueryResult.data === undefined && infQueryResult.isFetching ) {
+        queryClient.cancelQueries( { queryKey } );
+        setLoadingTimedOut( true );
+      }
+    }, LOADING_TIMEOUT );
+
+    // eslint-disable-next-line consistent-return
+    return () => clearTimeout( timer );
+  }, [infQueryResult.data, infQueryResult.isFetching, queryKey, queryClient] );
```
jtklein: https://tanstack.com/query/v5/docs/eslint/no-unstable-deps
The object returned from useInfiniteQuery and by extension of our wrapper useAuthenticatedInfiniteQuery is not referentially stable and should not be put directly into the dependency array here.
So, please refactor above to destructure out and pass them in here and into the return object of this hook as well.

src/sharedHooks/useInfiniteNotificationsScroll.ts
```diff
@@ -142,11 +152,44 @@ const useInfiniteNotificationsScroll = (
     },
   );
 
+  // We want to timeout and show an offline/retry state if this request takes too long
+  useEffect( () => {
```
jtklein, line 163: I think it is good to follow the existing codebase (useOnlineSuggestions) here, but do we actually need to handle react-query internals here? We have another example we could follow here?
[const timeoutID = setTimeout( ( ) => {
        abortController.abort( );
      }, MS_BEFORE_UPLOAD_TIMES_OUT );
      await uploadObservation( observation, realm, { signal: abortController.signal } );](https://github.com/inaturalist/iNaturalistReactNative/blob/46d2ab7e4b8c8677441d3dba89b690a2c82e27ef/src/components/MyObservations/hooks/useUploadObservations.ts#L112-L116)
Every endpoint we talk to through inaturalistjs, like in `fetchObservationUpdates` accepts a signal option [here](https://github.com/inaturalist/inaturalistjs/blob/c739c2c164b04515fa3ef20e4aeceed65a6afa37/lib/inaturalist_api.js#L121-L122). If we abort the request after 5s we just need to react UI-wise to the error case of `AbortError` (= timed out).
Your approach works but I am wondering if we should decide on one approach over the other to keep the code similar across all inaturalistjs endpoint calls. I can see that a query is different than an upload and have not looked too much into the differences here.
Is the approach here with useEffects better for queries? I would like to not having to add more useEffects to the app.

src/sharedHooks/useInfiniteNotificationsScroll.ts
```diff
@@ -89,8 +94,13 @@ const useInfiniteNotificationsScroll = (
 ): InfiniteNotificationsScrollResponse => {
   const currentUser = useCurrentUser( );
   const realm = useRealm( );
+  const queryClient = useQueryClient();
+  const [loadingTimedOut, setLoadingTimedOut] = useState( false );
 
-  const queryKey = ["useInfiniteNotificationsScroll", JSON.stringify( notificationParams )];
+  const queryKey = useMemo(
```
jtklein, line 100: Have you checked if this is beneficial? I have no data to support my assumption that a useMemo overhead here is worth than a JSON.stringify.

jtklein (changes requested):
Have tested that it does not change any of the existing functionality in terms of actually loading the data and erroring out when the API call has an error. I did not actually test yet the new function of timing out after 5s because I have already one requested change here.

src/sharedHooks/useInfiniteNotificationsScroll.ts
abbeycampbell: ah my bad! I should've been paying closer attention to your PRs the past couple days.
Destructuring also has the added benefit of making everything nicer to read imo

src/sharedHooks/useInfiniteNotificationsScroll.ts
abbeycampbell: I added this because I got this eslint error:
`The 'queryKey' array makes the dependencies of useEffect Hook (at line 175) change on every render. To fix this, wrap the initialization of 'queryKey' in its own useMemo() Hook`

src/sharedHooks/useInfiniteNotificationsScroll.ts
abbeycampbell: I think we could do it this way, tbh I'm not strongly opinionated. Gonna tag in @sepeterson and @FLGMwt for their hot takes

src/sharedHooks/useInfiniteNotificationsScroll.ts
jtklein: Ah, I see. I haven't noticed it is now a dependency below. But looking up the tree `notificationParams` is not referentially stable [here](https://github.com/inaturalist/iNaturalistReactNative/blob/d4b3353cc08d7e51f84ec04ac482256c34cb3b1f/src/components/Notifications/Notifications.tsx#L47-L48) so this useMemo would also change on every render. So, if you keep the useEffect below we need to make the object above stable.

src/sharedHooks/useInfiniteNotificationsScroll.ts
sepeterson: Sorry this isn't terribly helpful but I think I have a slight preference for the abortController if it works but I also wouldn't hold this PR up for it. Maybe something to try in the future

abbeycampbell: I tried out the AbortController approach, and the problem is that it kills requests that with the useEffect approach eventually succeed in the background. Trying this out under "Very Bad Network" conditions with the useEffect implementation eventually shows the user results after a retry or two, but AbortController never does, you might as well have no connection at all. It's possible there's a better way to do it; maybe there's a hybrid approach to take, but for now I think I'll leave as is, barring any strong opinions.

src/sharedHooks/useInfiniteNotificationsScroll.ts
abbeycampbell: oh good catch. I made those values constants instead. 

We could also map them to the `activeTab` values which would let us return only one `NotificationsContainer` I think. But I didn't want to get in the weeds of something kind of unrelated to the ticket ask

src/sharedHooks/useInfiniteNotificationsScroll.ts
jtklein: That is confusing me. Why would we still get responses, I thought the idea is to cancel the query on the 5s timeout? Also, why would we show a UI message that we could not get notifications if they do still appear after the 5s timeout? Is this just an artefact of how we test the app now with the simulated network problems?
I understand that a timeout is better than an "endless" spinner. Would it help to add some logs to your code so that we could track how often we show a timeout message but did actually retrieve the notifications after a while.

jtklein (approved):
The logic changes seem all sound to me now. When comparing this branch to main I do see that using the "Network Link Conditioner" on my real test device:
1.) on the "100% Loss" profile: it never resolves on main, i.e. true endless spinner - this PR fixes that in a way that the offline message appears.
2.) on the "3G" or "Very Bad Network" profile: it loads for a long time but eventually gets a response and shows notifications - this PR shows the offline message after loading for a while, and the responses do appear in the background after a while.
So, while I want to Approve this it does seem to me we should gather real-world data first to decide on the actual timeout.
I think, my preferred approach now would be to change this PR to have a timeout of 10s. But open to other ideas as well.
Then create a follow-up ticket to measure how many false positives we have, i.e. showing the offline message even though we did receive notifications while showing the offline message. And then add an A/B switch of 5 vs 10s timeout and decide how many false positives we can live with.
(Even though I really don't understand why we decide to show the offline message stating something like "We could not get your notifications!" but we keep trying in the background without any UI indication. And completely discarding the response we do receive in the background.)

PR #3358, opened 2026-02-05
Update @tanstack/react-query to 5.90.20

jtklein:
Closes MOB-1103.

Compiled and tested in debug and release mode on Android and iOS.

I am updating this now because I saw that newer version of their eslint plugin has a rule for finding the pattern we fixed in the [PR yesterday](https://github.com/inaturalist/iNaturalistReactNative/pull/3351). And I was actively thinking of making an eslint rule to force destructuring of the useMutation return object. So, at least if we use the return object directly of useMutation we should see an eslint warning now. Does not apply to our wrappers though I assume. https://tanstack.com/query/v5/docs/eslint/no-unstable-deps

Oddly enough the next available patch version breaks one of our tests. But we are able to advance quite a few versions now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -127,7 +127,7 @@
         "@react-native/metro-config": "0.81.5",
         "@react-native/typescript-config": "0.81.5",
         "@stylistic/eslint-plugin": "^3.1.0",
-        "@tanstack/eslint-plugin-query": "5.91.4",
+        "@tanstack/eslint-plugin-query": "^5.91.4",
         "@testing-library/jest-native": "^5.4.3",
         "@testing-library/react-native": "^13.3.3",
         "@types/jest": "^29.5.13",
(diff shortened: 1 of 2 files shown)


FLGMwt:
> Does not apply to our wrappers though I assume. https://tanstack.com/query/v5/docs/eslint/no-unstable-deps

yo, that's actually something I was looking at last month! As a baseline, I was checking to see if we could apply their  [exhaustive deps](https://tanstack.com/query/v5/docs/eslint/exhaustive-deps) / [stable query client](https://tanstack.com/query/v5/docs/eslint/stable-query-client) rules to, for example, our `useAuthenticatedQuery` wrapper.

Outcome: 

- the eslint plugins do _not_ support this from a config options level
- the plugins' implementation and build pipeline do not suggest a convenient (cheap) way for us to fork/patch this to support wrapper function names

PR #3359, opened 2026-02-06
add condition to stop fetching once we've fetched all results

abbeycampbell:
closes [MOB-1168](https://linear.app/inaturalist/issue/MOB-1168/infinite-re-rendering-loading-spinner-in-species-list)

Initially I was thinking of this as a performance-related follow-up to sorting species, but `useInfiniteScroll` is used in several places in the app, so this issue is fixed in these places as well:
- Alternate explore views: `IdentifiersView`, `ObserversView`, and `SpeciesView`, accessed by tapping the dark gray bar at the top
- Project search in Explore
- Project member list

### DIFF
--- a/src/sharedHooks/useInfiniteScroll.js
+++ b/src/sharedHooks/useInfiniteScroll.js
@@ -11,9 +11,9 @@ const useInfiniteScroll = (
   },
 ): Object => {
   const baseParams = {
-    ...newInputParams,
     per_page: 10,
     ttl: -1,
+    ...newInputParams,
   };
 
   const {


jtklein (approved):
I have tested this and it indeed stops the list from ever increasing the page param at the end of all data fetched.
So, I can Approve.
I just got side-tracked by: While the way you accessed the information of `per_page` from `baseParams` is correct given the fact that they are not altered. Looking at some of the components that do use this hook, e.g. [here](https://github.com/inaturalist/iNaturalistReactNative/blob/ed5d6b4c25c3c0fdd3f49b1c74c6ec8cb9dc0523/src/components/Explore/IdentifiersView.js#L30) which uses queryParams that ultimately are defined [here](https://github.com/inaturalist/iNaturalistReactNative/blob/ed5d6b4c25c3c0fdd3f49b1c74c6ec8cb9dc0523/src/components/Explore/ExploreContainer.js#L101) it seems to me that the capability to override `per_page` on a component basis was intended for this one hook, and seems to be broken by spreading `newInputParams` before per_page. But anyway that's for another PR.

Also, we do use the same pattern of: return lastPage.page + 1 in several other places. Do they all need the same fix?
https://github.com/inaturalist/iNaturalistReactNative/blob/ed5d6b4c25c3c0fdd3f49b1c74c6ec8cb9dc0523/src/components/Explore/helpers/exploreParams.ts#L63-L64
https://github.com/inaturalist/iNaturalistReactNative/blob/ed5d6b4c25c3c0fdd3f49b1c74c6ec8cb9dc0523/src/sharedHooks/useInfiniteNotificationsScroll.ts#L138-L139
https://github.com/inaturalist/iNaturalistReactNative/blob/ed5d6b4c25c3c0fdd3f49b1c74c6ec8cb9dc0523/src/sharedHooks/useInfiniteUserScroll.ts#L41-L42

abbeycampbell:
> Also, we do use the same pattern of: return lastPage.page + 1 in several other places. Do they all need the same fix?

I think we're good in `useInfiniteNotificationsScroll`, because we check against the length of `lastPage`-- so maybe we end up making one extra request for an empty page, which isn't a big deal.

`useInfiniteUserScroll` isn't actually broken at the moment, because pagination never happens. The comments [here](https://github.com/inaturalist/iNaturalistReactNative/blob/22d2ca9858719ee276edda4e69df796bd3cb1962/src/sharedHooks/useInfiniteUserScroll.ts#L16) and [here](https://github.com/inaturalist/iNaturalistReactNative/blob/22d2ca9858719ee276edda4e69df796bd3cb1962/src/components/UserProfile/FollowingList.tsx#L76-L77) suggest we're waiting on API support for this?

`exploreParams` seems to handle putting a stop to fetching [here](https://github.com/inaturalist/iNaturalistReactNative/blob/22d2ca9858719ee276edda4e69df796bd3cb1962/src/components/Explore/helpers/exploreParams.ts#L31-L34), so I think we're good on that one too.

PR #3363, opened 2026-02-09
use initialRegion and regionToAnimate for maps

abbeycampbell:
closes [MOB-774](https://linear.app/inaturalist/issue/MOB-774/update-all-usages-of-map-to-not-use-region-prop)

### DIFF
--- a/src/components/Explore/MapView.tsx
+++ b/src/components/Explore/MapView.tsx
@@ -78,7 +78,7 @@ const MapView = ( {
   const { t } = useTranslation( );
   const { state: exploreState, dispatch, defaultExploreLocation } = useExplore( );
   const [showRedoSearchButton, setShowRedoSearchButton] = useState( false );
-  const [regionToAnimate, setRegionToAnimate] = useState<Region | undefined>( undefined );
+  const [regionToAnimate, setRegionToAnimate] = useState<Region | null>( null );
   const isFirstRender = useRef( true );
 
   const mapRef = useRef<RNMapView | null>( null );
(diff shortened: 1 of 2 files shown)


src/components/Explore/MapView.tsx
```diff
@@ -78,6 +78,7 @@ const MapView = ( {
   const { t } = useTranslation( );
   const { state: exploreState, dispatch, defaultExploreLocation } = useExplore( );
   const [showRedoSearchButton, setShowRedoSearchButton] = useState( false );
+  const [regionToAnimate, setRegionToAnimate] = useState<Region | undefined>( undefined );
```
FLGMwt: nit, nonblocker: At least for this file w/ `mapRef`, we use `null` as the explicit "null" / empty state.

It's probably worth a discussion about how we want to treat `null` and `undefined` in general in the codebase, but I do think for state and refs, we tend to do null.

```suggestion
  const [regionToAnimate, setRegionToAnimate] = useState<Region | null>( null );
```

src/components/SharedComponents/Map/Map.tsx
```diff
@@ -377,7 +374,7 @@ const Map = ( {
   useEffect(
     ( ) => {
       if ( Platform.OS === "android" && androidAnimateRegion ) {
-        const curRegion = androidLocalRegion || region;
+        const curRegion = androidLocalRegion;
         const newRegion = {
           ...curRegion, // provides defaults for latitudeDelta and longitudeDelta
```
FLGMwt: nit, non-blocker: since `curRegion` is no longer conditionally defined, maybe we can inline `androidLocalRegion` in the `newRegion` definition?

FLGMwt (approved):
LGTM! 

no expectations on the comments that are unrelated to the changes

PR #3364, opened 2026-02-10
MOB-1176 additional delete diagnosis

FLGMwt:
Ostensibly, some people experiencing the "app slow until I uninstall and re-install" issue have had this problem clearly resolved by logging out and back in, which may help us narrow down the specific issue.

The logout step clears a few items of persisted state:

- [Realm](https://github.com/inaturalist/iNaturalistReactNative/blob/10137c4b83f371d29f9c7f323e61e06e1c4602f9/src/components/LoginSignUp/AuthenticationService.ts#L203-L220)
- [RQ's Cache](https://github.com/inaturalist/iNaturalistReactNative/blob/10137c4b83f371d29f9c7f323e61e06e1c4602f9/src/components/LoginSignUp/AuthenticationService.ts#L224)
- [Logs](https://await%20unlink(%20logfilepath%20);/)
- [Sensitive info](https://github.com/inaturalist/iNaturalistReactNative/blob/10137c4b83f371d29f9c7f323e61e06e1c4602f9/src/components/LoginSignUp/AuthenticationService.ts#L230-L233) (JWT/OAuth)
- [Images](https://github.com/inaturalist/iNaturalistReactNative/blob/10137c4b83f371d29f9c7f323e61e06e1c4602f9/src/components/LoginSignUp/AuthenticationService.ts#L236-L240)
- The MMKV store for [Zustand's persisted slices](https://github.com/inaturalist/iNaturalistReactNative/blob/10137c4b83f371d29f9c7f323e61e06e1c4602f9/src/components/LoginSignUp/AuthenticationService.ts#L243)

We were suspicious about some of these items being the culprit, but identifying the "logout fix" within the app gives us a more clear / explicit way to potentially diagnose them.


src/components/Developer/DiagnosticDeletionScreen.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+const syncWarning = [
+  "Careful! Observations that haven’t been uploaded to",
+  "iNaturalist will be deleted or end up in a broken state.",
+].join( " " );
+
+const operations = {
+  [DIAGNOSTIC_OPERATION.DELETE_REACT_QUERY_CACHE]: {
+    label: "Delete RQ Cache",
+    confirmDescription: "Bloop",
```
jtklein, line 43: Bloop?

src/components/Developer/DiagnosticDeletionScreen.tsx
```diff
@@ -0,0 +1,148 @@
...
+  DELETE_PHOTO_UPLOADS = "DELETE_PHOTO_UPLOADS",
+  DELETE_ROTATED_ORIGINAL_PHOTOS = "DELETE_ROTATED_ORIGINAL_PHOTOS",
+  DELETE_SOUND_UPLOADS = "DELETE_SOUND_UPLOADS",
+  DELETE_ZUSTAND_MMKV = "DELETE_ZUSTAND_MMKV",
+  DELETE_REALM = "DELETE_REALM",
+}
+
+const syncWarning = [
+  "Careful! Observations that haven’t been uploaded to",
+  "iNaturalist will be deleted or end up in a broken state.",
```
jtklein, line 37: Also "You will use data that provides an offline experience of the app."

src/components/Developer/DiagnosticDeletionScreen.tsx
```diff
@@ -0,0 +1,148 @@
...
+  [DIAGNOSTIC_OPERATION.DELETE_ZUSTAND_MMKV]: {
+    label: "Delete Zustand MMKV",
+    confirmDescription: syncWarning,
+    operation: ( ) => {
+      zustandMMKVBackingStorage.clearAll( );
+    },
+  },
+  [DIAGNOSTIC_OPERATION.DELETE_REALM]: {
+    label: "Delete Realm",
+    confirmDescription: syncWarning,
```
jtklein, line 103: In addition to syncWarning: You could likely end up in a broken state of the app and potentially need to login again or re-install the app.

src/components/Developer/DiagnosticDeletionScreen.tsx
```diff
@@ -0,0 +1,148 @@
...
+    operation: ( _queryClient?: QueryClient, realm?: Realm ) => {
+      if ( realm ) {
+        clearRealm( realm );
```
jtklein, line 106: I thought about this some more with the recent testing in relation to other works in AuthenticationService, one problem with only doing a realm delete without a logout, and vice versa only doing a token wipe without a logout, is that we rely on realm to tell the UI wether the user is logged in, but we rely on secret token to tell the API that we the user is logged in.

When pressing this button while logged in, the app crashed for me. So, it can leave the app in a broken state.

So, I am not convinced we get meaningful insights from artificially splitting up the logout deletes. But we can still test it.

jtklein (changes requested):
Could you revert the changes in Log.tsx, UiLibrary.tsx, and AuthenticationService.ts?
Those are separate concerns and not a priority to test at the moment.

If you remove those changes it could be fine to ship this code, I don't think the feature is complete to have users on real devices test it though. You would need to make sure to log before and after information remotely to not rely on the users reporting back to you wether a deletion had an impact in my view.

src/components/Developer/DiagnosticDeletionScreen.tsx
FLGMwt: ^we chatted about this so providing some of that convo here:

@jtklein clarified that the crash involved "Invalid Realm references" where rendered screens didn't like realm dropping out, which I missed because I was frequently restarting the app to test and didn't go through explore, etc. I'm going to take a look at maybe working around that and the UI sync issue as well.

_Separately_, @jtklein might have seen this class of error occur with the logout button in regular use, which is more interesting. I had sometimes seen a blip of a crash / stack trace when logging out but hadn't looked into it and we restart the app anyway so it _technically_ works.

That said, it seems like we have inconsistent logout behavior which _may_ impact some existing issues we've seen with synchronizing Realm & sensitive storage for access tokens. I'm adding this context to those tickets as an avenue of investigation.

Namely: what happens if `clearRealm` fails during logout but we still successfully do the rest of cleanup? or vice versa?

PR #3377, opened 2026-02-12
Mob 1042 type functions that have an eslint warning

jtklein:
Took a bite out of MOB-1042.

This PR includes all components that are tip components, i.e. the function is only used within, or already typed from the tip of the tree, or are already typed from above.


sepeterson (approved):
Lookin good ✅

PR #3379, opened 2026-02-13
MOB-135: Match Container useEffect simplification

sepeterson:
- Memoize topSuggestion and orderedSuggestions
- Update observation directly on place name fetch
- Rm focus listener for effect that’s meant to fire just once

### DIFF
--- a/src/components/Match/MatchContainer.tsx
+++ b/src/components/Match/MatchContainer.tsx
@@ -67,8 +67,7 @@ type Action =
   | { type: "SET_UPLOAD_PARAMS"; scoreImageParams: ImageParamsType }
   | { type: "SET_ONLINE_FETCH_STATUS"; onlineFetchStatus: FETCH_STATUSES }
   | { type: "SET_OFFLINE_FETCH_STATUS"; offlineFetchStatus: FETCH_STATUSES }
-  | { type: "SET_LOCATION"; scoreImageParams: ImageParamsType; shouldUseEvidenceLocation: boolean }
-  | { type: "ORDER_SUGGESTIONS"; orderedSuggestions: ApiSuggestion[] };
+  | { type: "SET_LOCATION"; scoreImageParams: ImageParamsType; shouldUseEvidenceLocation: boolean };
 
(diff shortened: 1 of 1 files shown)


src/components/Match/MatchContainer.tsx
```diff
@@ -300,14 +312,10 @@ const MatchContainer = ( ) => {
       const placeGuess
       = await fetchPlaceName( currentUserLocation?.latitude, currentUserLocation?.longitude );
       if ( placeGuess ) {
-        // Cannot call updateObservationKeys directly from here, since fetchPlaceName might take
-        // a while to return, in the meantime the current copy of the observation might have
-        // changed, so we update the observation from useEffect of currentPlaceGuess, so it will
-        // always have the latest copy of the current observation (see GH issue #584)
-        setCurrentPlaceGuess( placeGuess );
+        updateObservationKeys( { place_guess: placeGuess } );
```
sepeterson, line 314: The previous implementation appears to no longer be necessary. updateObservationKeys will just update the single property in the latest currentObservation in state

src/components/Match/MatchContainer.tsx
abbeycampbell: Let's remove the `ORDER_SUGGESTIONS` action type since we don't need it any more

src/components/Match/MatchContainer.tsx
```diff
@@ -333,71 +341,34 @@ const MatchContainer = ( ) => {
...
-    if ( chosenIndex !== -1 ) {
-      // Set new top suggestion
-      setTopSuggestion( sortedList[chosenIndex] );
-      // We can set the entire list here since we are filtering out the top suggestion in render
-      dispatch( {
-        type: "ORDER_SUGGESTIONS",
-        orderedSuggestions: sortedList,
-      } );
-    }
+    setSelectedSuggestionId( selection.taxon.id );
```
abbeycampbell, line 367: nit: since the `ApiTaxon.id` can be `number | undefined`, can we just define the useState value the same way and initialize it as `undefined` to avoid the type error here?

PR #3382, opened 2026-02-17
MOB-1190: freeze background stacks and clear stacks on tab press

sepeterson:
This is my best reckoning of a fix.

When going between tabs, navigates to the next stack navigator without changing stack at all. Current behavior is to nav to the new tab and place a new screen on top.

When tapping the currently selected tab button, just the stack navigator that you're in is reset. Current behavior is to stay on the current tab and place a new screen on top.

### DIFF
--- a/src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
+++ b/src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
@@ -88,13 +88,15 @@ const CustomTabBarContainer: React.FC<Props> = ( { navigation, state } ) => {
 
   const activeTab = getActiveTab( );
 
-  const getTopScreenOfTab = useCallback( ( tabName: TabName ): string | undefined => {
+  const getTopScreenOfTab = useCallback( ( tabName: TabName ): string | null => {
     const navState = navigation.getState( );
     const tabRoute = navState.routes.find( r => r.name === tabName );
     const tabState = tabRoute?.state;
-    if ( !tabState || !tabState.routes ) return;
-    const topRoute = tabState.routes[tabState.index ?? tabState.routes.length - 1];
-    return topRoute?.name;
+    if ( tabState && tabState.routes ) {
+      const topRoute = tabState.routes[tabState.index ?? tabState.routes.length - 1];
+      return topRoute?.name;
+    }
+    return null;
   }, [navigation] );
 
   const tabs: TabConfig[] = useMemo( ( ) => ( [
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -39,12 +39,15 @@ import { useNetworkActivityDevTools } from "@rozenite/network-activity-plugin";
...
 import { useTanStackQueryDevTools } from "@rozenite/tanstack-query-plugin";
 
+import { enableFreeze } from "react-native-screens";
 import { name as appName } from "./app.json";
 import { log } from "./react-native-logs.config";
 import { getUserAgent } from "./src/api/userAgent";
 
 const logger = log.extend( "index.js" );
 
+enableFreeze( true );
```
sepeterson: This is 95% of the performance fix. If you're curious, more here: https://github.com/software-mansion/react-freeze?tab=readme-ov-file#what-is-this-

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
```diff
@@ -51,6 +53,39 @@ const CustomTabBarContainer: React.FC<Props> = ( { navigation, state } ) => {
...
   };
 
+  const resetStack = useCallback( ( tabName: TabName, screenName: ScreenName ) => {
+    const navState = navigation.getState( );
+    // copy existing stack routes so we don't lose state for tabs we're not resetting
+    const newStacks: Partial<NavigationRoute<ParamListBase, string>>[]
+    = navState.routes.slice();
+    const replaceIndex = newStacks.findIndex( r => r.name === tabName );
+    // add reset stack
+    newStacks.splice( replaceIndex, 1, {
```
sepeterson, line 63: [From react-navigation docs:](https://reactnavigation.org/docs/navigation-actions/#reset)

> Consider the navigator's state object to be internal and subject to change in a minor release. Avoid using properties from the [navigation state](https://reactnavigation.org/docs/navigation-state) state object except index and routes, unless you really need it.

I'm not super pleased to be messing with react-navigation internal state like this. I think theoretically if we're able to do this reset with a useNavigation within the respective stack navigator, it would be cleaner but I haven't found a way to do this.

index.js
sepeterson: I actually think we don't necessarily need this if we have the stack-resetting logic but what I meant was this change alone would fix the performance concern.

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
FLGMwt: I think this _could_ be related to the issue you demonstrated with the "bottom tab rendered but main view blank". Did you determine if that was reproducible with freeze turned off and a hard js reset?

If so, my suspicion is that we might be confusing RNN's tracking of _when_ to freeze something by either the specific configuration of our routing config or this manipulation.

abbeycampbell:
fwiw, I tested this branch on my actual device and didn't manage to trigger the blank screen state. I tried it both with `enableFreeze( true )` and without.

PR #3383, opened 2026-02-17
pass elements to flashlist instead of functions

abbeycampbell:
closes [MOB-717](https://linear.app/inaturalist/issue/MOB-717/flickering-notifications-loading-circle)

The problem here was that our `renderEmptyComponent` function was being redefined when dependencies changed, and thus un/remounted, which made the spinner look glitchy. If we pass a memoized element to the flashlist instead, we just update the tree when a dependency changes, rather than re-mount, so the animation is smooth. Went ahead and made the same change for the footer loading animation.

### DIFF
--- a/src/components/Notifications/NotificationsList.tsx
+++ b/src/components/Notifications/NotificationsList.tsx
@@ -13,6 +13,8 @@ import type { RealmUser } from "realmModels/types";
 import { useTranslation } from "sharedHooks";
 import type { Notification } from "sharedHooks/useInfiniteNotificationsScroll";
 
+const ItemSeparator = ( ) => <View className="border-b border-lightGray" />;
+
 interface Props {
   currentUser: RealmUser | null;
   data: Notification[];
@@ -50,8 +52,6 @@ const NotificationsList = ( {
   const renderItem = useCallback( ( { item }: RenderItemProps ) => (
     <NotificationsListItem notification={item} />
   ), [] );
-
-  const renderItemSeparator = ( ) => <View className="border-b border-lightGray" />;
 
   const footerComponent = useMemo( ( ) => (
     <InfiniteScrollLoadingWheel
@@ -112,7 +112,7 @@ const NotificationsList = ( {
 
   return (
     <CustomFlashList
-      ItemSeparatorComponent={renderItemSeparator}
+      ItemSeparatorComponent={ItemSeparator}
       ListEmptyComponent={emptyComponent}
       ListFooterComponent={footerComponent}
       data={data}


src/components/Notifications/NotificationsList.tsx
```diff
@@ -113,8 +113,8 @@ const NotificationsList = ( {
   return (
     <CustomFlashList
       ItemSeparatorComponent={renderItemSeparator}
-      ListEmptyComponent={renderEmptyComponent}
-      ListFooterComponent={renderFooter}
```
FLGMwt, line 117: ohhhhhhhhhhhhhhhhhhh, @abbeycampbell I think you hit on something very interesting. As you said, I think this might be a productive pattern to address across other instances.

tl;dr: this highlights and fixes a more dubious puzzle than some other `useMemo/useCallback` changes. The `renderFooter` render helper is a secret `React.Component`!

Next comment not commenting on changes of PR, just commenting on existing issue as it presents here, hoping this helps future cases:

FLGMwt: My understanding of this issue, which affirms the PR's fix:

I believe this is different from the common case of passing a `<Button onPress={handlePress} />` where `handlePress` would benefit from being memoized for referential integrity.

`ListFooterComponent` accepts one of two things: `MyCustomFooterComponent` or `<MyCustomFooterComponent />` (an element / instance of a component)

In this case, `ListFooterComponent={renderFooter}`, satisfies the former because `renderFooter`, incidentally, fulfills the contract of a `React.Component`. I believe this use is incidental / unintentional based on: imperative "render"-prefixed variable name, camelCasing, and similarity to the many non-Component "render helpers" in the codebase.

We might say `renderFooter` would be more appropriately called `RenderFooter` (or just `Footer`) as used.

The significant thing is:

Even with `useCallback`, we are defining brand new components during render when deps change to pass to `ListFooterComponent`. When `isConnected` changes, as far as React is concerned, we are switching `<3513423414123 />` out for `<563704568971348 /> without any context for their relationship`. React, in our service, [would very much prefer](https://react.dev/learn/preserving-and-resetting-state#different-components-at-the-same-position-reset-state) we go from `<InfiniteScrollLoadingWheel isConnected={false} />` to `<InfiniteScrollLoadingWheel isConnected={true} />`.

_I believe this case was unintentional, but this fix highlights a situation where defining components in render can cause significant issue._

--------

This PR makes that shift FROM generating a new wrapper component to swap out TO providing an element of the same component that reacts to prop changes.

Before:

`ListFooterComponent={DifferentComponentWhenConnectedChanges}` 

After:

`ListFooterComponent={<InfiniteScrollLoadingWheel isConnected={isConnected} /> }`

------

Note: `ListFooterComponent`, `ListEmptyComponent`, etc are distinctly different from the `renderItem` prop which _does_ behave more like our own "render helpers" and is not necessarily a `React.Component`. Something something performance something virtualization something view recycling.

src/components/Notifications/NotificationsList.tsx
```diff
@@ -53,14 +53,14 @@ const NotificationsList = ( {
 
   const renderItemSeparator = ( ) => <View className="border-b border-lightGray" />;
```
FLGMwt: This follows the same pattern of `footerComponent` and `emptyComponent`. That is, `renderItemSeparator` is treated as an inline component. 

We should either memoize an element value like the others OR since it doesn't have any props, we can extract this as an `ItemSeparator` at the file level and pass it as a component `ItemSeparatorComponent={ItemSeparator}`

FLGMwt (changes requested):
Great catch! yeah, this will definitely inform some other cases.

Just want to make sure `renderItemSeparator` gets captured with this fix, but otherwise gtg

src/components/Notifications/NotificationsList.tsx
abbeycampbell: If I memoize the separator, we are passing the flashlist a specific JSX result instead of something callable, which we need particularly in the case of the separator because it's going to be called multiple times.

So I went with the file level component here, and that's probably the pattern we should adopt with this in those other places too

src/components/Notifications/NotificationsList.tsx
jtklein: So, previously, useCallback provided a referentially stable function that returns a component. Hence types were satisfied. It produces a new element on every render even though the function reference is stable.
With useMemo we get a referentially stable element. Is it stable even through props changing, or only, as long as isFetching, isConnected, and data.length don’t change. I would think only as long as props are unchanged footerComponent is the same object reference.

src/components/Notifications/NotificationsList.tsx
FLGMwt: My mistake! Yeah, this is distinct from the others in that it does not accept an `Element` for that reason https://shopify.github.io/flash-list/docs/usage/#itemseparatorcomponent. The extraction _is_ still significant because `renderItemSeparator` was the `Component` we were using.

Thanks for the update!

PR #3391, opened 2026-02-19
MOB-135: suggestions useEffect simplification

sepeterson:
- Return a function off useNavigateWithTaxonSelected that updates selected taxon rather than relying on an effect dependent on a component state value

### DIFF
--- a/src/components/Suggestions/SuggestionsTaxonSearch.tsx
+++ b/src/components/Suggestions/SuggestionsTaxonSearch.tsx
@@ -15,7 +15,7 @@ const SuggestionsTaxonSearch = ( ) => {
   const { taxa, isLoading, isLocal } = useTaxonSearch( taxonQuery );
   const { t } = useTranslation( );
 
-  const navigateWithTaxonSelected = useNavigateWithTaxonSelected( { vision: true } );
+  const navigateWithTaxonSelected = useNavigateWithTaxonSelected( { vision: false } );
 
   const renderTaxonResult = useCallback( ( { item: taxon, index } ) => (
     <TaxonResult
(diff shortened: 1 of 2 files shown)


src/components/Suggestions/hooks/useNavigateWithTaxonSelected.ts
```diff
@@ -55,20 +48,17 @@ const useNavigateWithTaxonSelected = (
...
   }, [
     currentObservation?.uuid,
     entryScreen,
     lastScreen,
     navigation,
     options?.vision,
-    selectedTaxon,
-    unselectTaxon,
     updateObservationKeys,
     vision,
```
abbeycampbell, line 57: this is a nitpick that has nothing to do with your changes but since `vision` is derived from `options.vision` I think those deps are redundant and we can lose one

src/components/Suggestions/SuggestionsTaxonSearch.tsx
```diff
@@ -12,27 +12,22 @@ import useNavigateWithTaxonSelected from "./hooks/useNavigateWithTaxonSelected";
...
-  const [selectedTaxon, setSelectedTaxon] = useState( null );
   const { taxa, isLoading, isLocal } = useTaxonSearch( taxonQuery );
   const { t } = useTranslation( );
 
-  useNavigateWithTaxonSelected(
-    selectedTaxon,
-    ( ) => setSelectedTaxon( null ),
-    { vision: false },
-  );
+  const navigateWithTaxonSelected = useNavigateWithTaxonSelected( { vision: true } );
```
abbeycampbell: Is it right that this is passing `{ vision: true }`? It was false before, and I think we want it to be false when someone is manually searching, right?

sepeterson: Yes! Fantastic catch

PR #3404, opened 2026-02-26
Make all camera buttons rotate (plus refactors)

jtklein:
Closes MOB-1001

This PR refactors the `Animated.View` instances used to rotate buttons in the two camera screens to a single wrapper component. It also adds this component to all remaining buttons with icons, so that all behave the same. It also adds this wrapper to the icon shown in the shutter button. Another refactor is to make the API of `CameraFlip` more similar to that of `Flash` and `Zoom`.

Landscape view before:
77C8575D-0B71-47FE-93BA-AADEA101973D_1_102_o

After:
C64C13AB-26CB-449A-8112-027235C4F647_1_102_o


src/components/Camera/RotatableIconWrapper.tsx
```diff
@@ -0,0 +1,34 @@
+import type { PropsWithChildren } from "react";
+import React from "react";
+import type { ViewStyle } from "react-native";
+import DeviceInfo from "react-native-device-info";
+import type { AnimatedStyle } from "react-native-reanimated";
+import Animated from "react-native-reanimated";
+
+interface Props extends PropsWithChildren {
+  rotatableAnimatedStyle?: ViewStyle | AnimatedStyle;
```
FLGMwt, line 9: Since we now have an explicit `RotatableIconWrapper`, is there a reason we need to do the prop drilling for `rotatableAnimatedStyle`? It seems like we could just call `useRotation` in this component instead of the camera roots (and another useRotation in `PhotoCarousel` to avoid the `rotation` var we capture in [<StandardCamera />](https://github.com/inaturalist/iNaturalistReactNative/blob/91b38b854bcd325275efb6ef192cf87ce74a97bb/src/components/Camera/StandardCamera/StandardCamera.js#L87).

jtklein: Thought about it, but since it would be adding a listener to device orientation change for each component I think it might be better to have the prop once.

PR #3405, opened 2026-02-26
Update set of uncapitalized common names parts

jtklein:
Closes MOB-790

Following this PR here: https://github.com/inaturalist/inaturalist/pull/4574/changes

Before:
37341613-6CF6-48EA-9ECD-46E5CA5D8FDA_1_102_o
After:
08BF9758-9676-484B-9E2B-66EB2B82931C_1_102_o

### DIFF
--- a/src/sharedHelpers/taxon.ts
+++ b/src/sharedHelpers/taxon.ts
@@ -1,6 +1,7 @@
 import map from "lodash/map";
 import trim from "lodash/trim";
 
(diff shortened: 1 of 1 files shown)


src/sharedHelpers/taxon.ts
```diff
@@ -2,27 +2,29 @@ import map from "lodash/map";
 import trim from "lodash/trim";
 
 const uncapitalized = new Set( [
```
FLGMwt, line 5:
```suggestion
// This collection mirrors web's: https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/ang/models/taxon.js.erb#L187
const uncapitalized = new Set( [
```

PR #3406, opened 2026-02-26
Accuracy circle in location picker should be sized as radius = accuracy

jtklein:
Closes MOB-442.

This PR is changing the calculation of the zoom level of the location picker map, so that the crosshair circle's radius represents the positional accuracy of the observation location.

Before:
1D393BDE-7545-4E66-8776-95D1E0F0D464_4_5005_c

After, a accuracy of 50m correctly displayed on the initial map zoom level.
E2A9CEBC-BEC9-44F1-BAA9-819D91932A28_4_5005_c
90647A08-61CA-43E8-8DE5-39E343EE76B5_4_5005_c

### DIFF
--- a/src/components/SharedComponents/Map/Map.tsx
+++ b/src/components/SharedComponents/Map/Map.tsx
@@ -425,12 +425,6 @@ const Map = ( {
     ? fuzzRegion( unfuzzedMapRegion )
     : unfuzzedMapRegion;
 
-  // In Android, we maintain initialRegion as state localRegion and
-  // pass undefined to parameter initialRegion.
-  const mapInitialRegion = Platform.OS === "android"
-    ? undefined
-    : initialRegion;
-
   const renderDebugZoomLevel = ( ) => {
     if ( isDebug ) {
       return (
@@ -528,7 +522,7 @@ const Map = ( {
       <MapView
         cameraZoomRange={cameraZoomRange}
         className={className}
-        initialRegion={mapInitialRegion}
+        initialRegion={initialRegion}
         loadingEnabled
         loadingIndicatorColor={colors.inatGreen}
         mapType={currentMapType}


src/components/LocationPicker/DisplayLatLng.tsx
```diff
@@ -1,22 +1,22 @@
...
 import { View } from "components/styledComponents";
-import type { Node } from "react";
 import React from "react";
 import { useTranslation } from "sharedHooks";
 import { getShadow } from "styles/global";
 
 const DROP_SHADOW = getShadow( );
 
-type Props = {
-  region: Object,
```
jtklein, line 15: Migrated this to TS because I wanted to make this region prop more clear.

src/components/SharedComponents/Map/Map.tsx
```diff
@@ -82,6 +82,7 @@ interface Props {
   withPressableObsTiles?: boolean;
   zoomEnabled?: boolean;
   zoomTapEnabled?: boolean;
+  onMapLayout?: ( event: { nativeEvent: { layout: { width: number; height: number } } } ) => void;
```
jtklein, line 85: This is to signal the dimensions of the map to higher up.

src/components/SharedComponents/Map/helpers/mapHelpers.ts
```diff
@@ -49,16 +49,29 @@ export function obscurationCellForLatLng( lat: number, lng: number ) {
   };
 }
 
-// Adapted from iNat Android LocationChooserActivity.java computeOffset function
-const EARTH_RADIUS = 6371000; // Earth radius in meters
-export function metersToLatitudeDelta( meters: number, latitude: number ): number {
```
jtklein, line 54: From what I can tell this is actually the calculation to turn longitude delta into meters. So, I switched to use a different calculation. Latitude delta to meters is much less dependent on the latitude, and mostly pretty close to 111km per °. But since the earth is an oblate spheroid there is still a bit of influence from the actual latitude. I did use a bit more complicated calculation (than the 111km per °) because we are an organization that cares about a certain degree of precision in our data.

src/components/LocationPicker/LocationPickerContainer.js
```diff
@@ -218,17 +228,24 @@ const LocationPickerContainer = ( ): Node => {
...
+        place_guess: locationName,
+      };
+      updateObservationKeys( keysToUpdate );
+    }
     navigation.goBack( );
   };
 
+  const onMapLayout = event => {
+    const { height, width } = event.nativeEvent.layout;
+    setRadiusToMapHeight( CROSSHAIRRADIUS / height );
```
jtklein, line 245: We need this ratio because when the map is moved or initialize we need to calculate the radius of the circle as accuracy.

src/components/LocationPicker/LocationPickerContainer.js
```diff
@@ -218,17 +228,24 @@ const LocationPickerContainer = ( ): Node => {
...
+    setMapDimensionsRatio( width / height );
```
jtklein, line 246: We need this ratio because apparently we can not set the latDelta and longDelta with two different values.

src/components/LocationPicker/LocationPicker.js
```diff
@@ -51,9 +52,23 @@ const LocationPicker = ( {
   region,
   selectPlaceResult,
   updateLocationName,
+  onMapLayout,
 }: Props ): Node => {
   const { t } = useTranslation( );
 
+  let regionToDisplay;
```
jtklein, line 59: When loading the location picker we set the initialRegion first, only when the map has been moved we do see a change in region. (Potentially we can see a change in region also when the useEffect in LocationPickerContainer triggers. I have not changed the code there but I do think we will not see this being triggered as much with these changes.)

abbeycampbell (approved):

PR #3407, opened 2026-02-26
MOB-1182: disable firebase for unauthenticated and opted-out users

sepeterson:
closes [MOB-1182](https://linear.app/inaturalist/issue/MOB-1182/disable-firebase-for-unauthenticated-users-and-users-with-prefers-no)

In this PR:

- Default to be opted out of tracking. See ticket for details on this decision
- On user fetch, set tracking options dependent on `prefers_no_tracking`

### DIFF
--- a/src/components/LoginSignUp/AuthenticationService.ts
+++ b/src/components/LoginSignUp/AuthenticationService.ts
@@ -27,7 +27,7 @@ import { getInstallID } from "sharedHelpers/installData";
 import { log, logFilePath, logWithoutRemote } from "sharedHelpers/logger";
 import removeAllFilesFromDirectory from "sharedHelpers/removeAllFilesFromDirectory";
 import safeRealmWrite from "sharedHelpers/safeRealmWrite";
-import { setFirebaseDataCollection } from "sharedHelpers/tracking";
+import { setFirebaseDataCollectionEnabled } from "sharedHelpers/tracking";
 import { unlink } from "sharedHelpers/util";
 import { isDebugMode } from "sharedHooks/useDebugMode";
 import zustandMMKVBackingStorage from "stores/zustandMMKVBackingStorage";
@@ -249,7 +249,7 @@ const signOut = async (
   options.queryClient?.getQueryCache( ).clear( );
 
   // Disable firebase data collection on signout
-  setFirebaseDataCollection( false );
+  setFirebaseDataCollectionEnabled( false );
 
   // switch the app back to the system locale when a user signs out
   const systemLocale = getInatLocaleFromSystemLocale( );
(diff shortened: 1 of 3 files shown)


src/sharedHooks/useUserMe.ts
```diff
@@ -40,6 +45,13 @@ const useUserMe = ( options: UseUserMeOptions ) => {
         realm.create( User, remoteUser, UpdateMode.Modified );
       }, "modifying current user via remote fetch in useUserMe" );
     }
+    if ( remoteUser ) {
+      const enableCollection = !remoteUser.prefers_no_tracking;
+      setAnalyticsCollectionEnabled( getAnalytics(), enableCollection );
+      // This looks unusual but is actually the preferred solution
+      // https://github.com/invertase/react-native-firebase/blob/81a0f1910955a0295b6b308d5c08c17af0384b04/packages/perf/lib/index.d.ts#L459
+      getPerformance().dataCollectionEnabled = enableCollection;
```
FLGMwt: Yeah, that's odd, seems like the docs are stale. Seems like analytics also has multiple ways to toggle? idk, if it works, it works!

sepeterson: 🤷

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -247,6 +248,9 @@ const signOut = async (
   // to the React Query context (maybe it could...)
   options.queryClient?.getQueryCache( ).clear( );
 
+  // Disable firebase data collection on signout
+  setFirebaseDataCollection( false );
+
```
FLGMwt, line 253: 🤙

src/sharedHelpers/tracking.ts
```diff
@@ -19,6 +21,17 @@ export const logFirebaseEvent = (
   }
 };
 
+export const setFirebaseDataCollection = ( enabled: boolean ) => {
```
FLGMwt: Can we name this `setFirebaseDataCollectionEnabled`? This would normally be a small nit, but I think it's more important here because the user option flag is inverted from what one might expect (`true` drives `disabled`).

ios/Podfile
```diff
@@ -47,6 +47,9 @@ use_frameworks! linkage: :static
 # rubocop:disable Style/GlobalVars
 # Required for react-native-firebase: https://rnfirebase.io/#configure-firebase-with-ios-credentials-react-native--077
 $RNFirebaseAsStaticFramework = true
+# Disable ad ID support: https://rnfirebase.io/analytics/usage#disable-ad-id-usage-on-ios
+# Conext: https://linear.app/inaturalist/issue/MOB-1182/disable-firebase-for-unauthenticated-users-and-users-with-prefers-no#comment-8c9d3cc2
+$RNFirebaseAnalyticsWithoutAdIdSupport = true
```
FLGMwt, line 52: thanks so much for adding this!

sepeterson: 😎

PR #3408, opened 2026-02-26
add and configure react-navigation rozenite plugin

FLGMwt:
Idk why this didn't make my pass adding rozenite in originally, but w/ the navigation work we're looking at, it might be helpful! Seems to work find out of the box.

The redux-devtool-style action / state log is pretty intuitive and the "time travel" seems to work well for simple jumps (I haven't tried it for nested explore stuff)

image

### DIFF
--- a/src/navigation/OfflineNavigationGuard.tsx
+++ b/src/navigation/OfflineNavigationGuard.tsx
@@ -2,6 +2,7 @@ import {
   useNetInfo,
 } from "@react-native-community/netinfo";
 import { NavigationContainer } from "@react-navigation/native";
+import { useReactNavigationDevTools } from "@rozenite/react-navigation-plugin";
 import type { PropsWithChildren } from "react";
 import React, { useRef } from "react";
 import { Alert } from "react-native";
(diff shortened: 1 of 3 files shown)


src/navigation/OfflineNavigationGuard.tsx
```diff
@@ -15,6 +16,8 @@ const OfflineNavigationGuard = ( { children }: PropsWithChildren ) => {
   const { isConnected } = useNetInfo( );
   const { t } = useTranslation( );
 
+  useReactNavigationDevTools( { ref: navigationRef } );
```
abbeycampbell, line 19: wouldn't we wanna put this in `render.js`?

FLGMwt: I'm not sure what I would expect we'd want out of this for our jest tests. Am I misunderstanding what file you mean?

abbeycampbell: Sorry, dumb question on my part lol. I couldn't remember ever seeing this file in my nav work so in my mind, our actual `NavigationContainer` wrapping the all the app code lived somewhere else. I did a grep for that, saw `render.js`, and thought yeah that sounds right without actually looking at it. Disregard 🙃

FLGMwt: oh, yeah, I also found it odd to put in here, but it happens to be the outermost place we "care" about nav and therefore where our root `NavigationContainer` is. I figured since we're considering a routing makeover, it'd be something that would probably more clearly placed anyhow. 🙏

PR #3414, opened 2026-03-02
allow gobacks, if possible, during dismiss photo permissions and add …

abbeycampbell:
…evidence flows

Follow-up to https://github.com/inaturalist/iNaturalistReactNative/pull/3403 and https://github.com/inaturalist/iNaturalistReactNative/pull/3409


abbeycampbell:
> Thanks. I have just tested this with the following flow: In advanced mode, start an observation without media, press plus icon, press camera. X in the permission gate brings you back to ObsEdit as it should. On main it was resetting the stack which is not wanted.

Thanks for catching/calling this out!

PR #3416, opened 2026-03-03
Restructure explore full screen experiences

jtklein:
Closes MOB-1018

This PR introduces the two new screens we need for the Explore V2.
The first new screen is just a full react-navigation screen that displays the already built "Explore Filters". For this screen I have duplicated the FiltersModal component (into FiltersModalV2) and removed the parts that open additional search modals on top of Explore Filters. In V2 these search experiences will be combined in a single "Universal Search" screen. If you compare FiltersModal with FiltersModalV2 there are mostly deletions (only?), and no functionality has been changed.
The second screen added is this "Universal Search" screen where I just copied the search UIs that have been shown as modals before. These are not copies but show the existing components because this screen needs extensive make-over before release and I would expect none of these components to be taken forward.
All actual filter setting functionality is stubbed out because we have to re-architect the state keeping at one point.
Both new screens are wrapped in the ExploreProvider so have their own explore context, this is of course non-sensical because we need a shared context with the other screens. But just leaving these new screens without this context means I have to rip out all the state setting we might need ti have context for when redoing it later. So, it is meant to be temporary until we redo the state.

Few actual changes:
Changed a type to re-use what we export from it's source

### DIFF
--- a/src/components/Explore/ExploreSearchContainer.tsx
+++ b/src/components/Explore/ExploreSearchContainer.tsx
@@ -1,3 +1,4 @@
+import type { RouteProp } from "@react-navigation/native";
 import { useNavigation, useRoute } from "@react-navigation/native";
 import type { ApiPlace, ApiProject, ApiTaxon } from "api/types";
 import { View } from "components/styledComponents";
@@ -15,7 +16,8 @@ import ExploreUserSearch from "./SearchScreens/ExploreUserSearch";
 
 const ExploreSearchContainerWithContext = () => {
   const navigation = useNavigation();
-  const { params } = useRoute( );
+  const { params } = useRoute<RouteProp<Record<string, { initialSearchMode?: string }>, string>>( );
+
   const {
     hasPermissions,
     renderPermissionsGate,


src/components/Explore/ExploreSearchContainer.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { useLocationPermission } from "sharedHooks";
+
+import ExploreLocationSearch from "./SearchScreens/ExploreLocationSearch";
+import ExploreProjectSearch from "./SearchScreens/ExploreProjectSearch";
+import ExploreTaxonSearch from "./SearchScreens/ExploreTaxonSearch";
+import ExploreUserSearch from "./SearchScreens/ExploreUserSearch";
+
+const ExploreSearchContainerWithContext = () => {
+  const navigation = useNavigation();
+  const { params } = useRoute( );
```
sepeterson: Could be typed like

```
const { params } = useRoute<RouteProp<Record<string, { initialSearchMode?: string }>, string>>();
```

src/components/Explore/Modals/FilterModalV2.tsx
```diff
@@ -0,0 +1,1331 @@
...
+        return t( "DATE-OBSERVED-OLDEST" );
+      case SORT_BY.MOST_FAVED:
+        return t( "MOST-FAVED" );
+      case SORT_BY.DATE_UPLOADED_NEWEST:
+      default:
+        return t( "DATE-UPLOADED-NEWEST" );
+    }
+  };
+
+  const sortByValues = {
```
sepeterson, line 149: What do you think about moving these big consts to a separate file?

src/components/Explore/Modals/FilterModalV2.tsx
```diff
@@ -0,0 +1,1331 @@
...
+              if ( taxonName === "unknown" ) {
+                if ( ( iconicTaxonNames || [] ).indexOf( taxonName ) >= 0 ) {
+                  updateTaxon( null );
+                } else {
+                  filterByIconicTaxonUnknown();
+                }
+              } else if ( taxon?.name?.toLowerCase() === taxonName ) {
+                updateTaxon( null );
+              } else {
+                const selectedTaxon = realm
```
sepeterson, line 738: I know this was just copy/pasted but in general, how do we feel about making a goal of ExploreV2 to decouple it from Realm?

src/components/Explore/Modals/FilterModalV2.tsx
jtklein: I did not touch this yet because there will be quite some rearrangement when we extract the sorting logic which will need to be shared across different screens. https://linear.app/inaturalist/issue/MOB-1047/add-sorting-to-my-observations

src/components/Explore/Modals/FilterModalV2.tsx
jtklein: Not quite sure what you mean with decouple from realm. Do you mean only always show data retrieved from the API directly without using cached resources? Being able to see data you have on your device like taxa you once searched was important for this app in the past, but I don't know where we stand on this now. Anyway, what we do want is to re-do the state architecture at one point. Feel free to add your thoughts to this ticket here: https://linear.app/inaturalist/issue/MOB-134/placeholder-redo-explore-architecturally

src/components/Explore/ExploreSearchContainer.tsx
jtklein: Have added this. From what I have built in the past I remember a central file where navigation and routes and params are typed.

src/components/Explore/Modals/FilterModalV2.tsx
sepeterson: Yes I was thinking explore could rely on just API data in a way that a user's personal observations might not be able to. Additionally, since we likely need to move away from realm in the long term, this could be a good opportunity to try alternative solutions for persisting data long term if that's something we need to do

sepeterson (approved):
LGTM

PR #3417, opened 2026-03-04
MOB-1190 reset stacks on nav away

sepeterson:


src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
```diff
@@ -53,51 +53,64 @@ const CustomTabBarContainer: React.FC<Props> = ( { navigation, state } ) => {
...
-  const activeTab = getActiveTab( );
+    // If pressing the currently active tab, reset its stack
+    if ( targetTabName === activeTabName ) {
+      const idx = newStacks.findIndex( r => r.name === targetTabName );
+      newStacks.splice( idx, 1, {
+        name: targetTabName,
+        state: { index: 0, routes: [{ name: targetScreenName }] },
+      } );
+      needsReset = true;
+    }
```
FLGMwt, line 92: I'm sorry, I'm only now realizing the implications you mentioned on MOB-1190 for this behavior. This seems a bit jarring to me; in most apps, I think I'm used to no action on pressing an active tab. aka: the button for the active tab might as well be disabled . If we _do_ want this, is it maybe worth a modal warning that your "session" is being reset? 

happy to defer that Product question since this is implementing as described.

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
```diff
@@ -53,51 +53,64 @@ const CustomTabBarContainer: React.FC<Props> = ( { navigation, state } ) => {
...
-          },
-        ],
-      } ),
-    );
-  }, [navigation] );
+      = navState.routes.slice();
+    let needsReset = false;
+
+    // Reset Menu and Notifications stacks when navigating away from them
+    const resetOnLeave: [TabName, ScreenName][] = [
```
FLGMwt: naming nit:

`resetOnLeave` sounds active, like an event handler whereas this is more of a config object. Can we try another name for this? I think this also fills the same "config" role of the `getActiveTab` mapping helper. I think both of those might be clearer as file-level vars and maybe also defined the same way (either as a helper function or config object with a default).

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
sepeterson: We covered this in planning but there is precedent for this. The home tab on Amazon resets that stack when tapped

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
sepeterson: Good spot of a couple opportunities to move stuff out of a component! I am not sure I totally captured your vision so feel free to lmk

src/navigation/BottomTabNavigator/CustomTabBarContainer.tsx
FLGMwt: Thanks for clarifying!

PR #3425, opened 2026-03-10
Mob 1146 update react native reanimated to v4 latest

abbeycampbell:
closes [MOB-1146](https://linear.app/inaturalist/issue/MOB-1146/update-react-native-reanimated-to-v4-latest)

tested on debug mode for android and ios, release mode for ios. couldn't get android release mode to work properly but I might have been doing something wrong. This feels like the least important of all of them but would appreciate a check from someone else!

### DIFF
--- a/tests/unit/components/Camera/__snapshots__/PhotoCarousel.test.js.snap
+++ b/tests/unit/components/Camera/__snapshots__/PhotoCarousel.test.js.snap
@@ -70,6 +70,7 @@ exports[`PhotoCarousel renders correctly 1`] = `
             }
           }
           jestInlineStyle={{}}
+          nativeID="0"
           style={
             [
               {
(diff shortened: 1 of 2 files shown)


babel.config.js
```diff
@@ -26,13 +26,8 @@ module.exports = {
         uploaders: "./src/uploaders",
       },
     }],
-    // Reanimated plugin has to be listed last https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation/
-    [
-      "react-native-reanimated/plugin",
-      {
-        processNestedWorklets: true,
```
jtklein, line 33: This is no longer needed?

abbeycampbell: it wasn't listed in the [docs](https://docs.swmansion.com/react-native-worklets/docs/worklets-babel-plugin/plugin-options) as an option for the new plugin

PR #3429, opened 2026-03-12
MOB-1186: logged persisted storage

sepeterson:


src/components/StartupService.tsx
```diff
@@ -95,6 +97,26 @@ const StartupService = ( ) => {
...
 
+      const logStorageMetrics = async ( ) => {
+        try {
+          const realmBytes = realm?.path
+            ? ( await RNFS.stat( realm.path ).catch( () => ( { size: 0 } ) ) ).size
+            : 0;
+          const logFileBytes = logFilePath
+            ? ( await RNFS.stat( logFilePath ).catch( () => ( { size: 0 } ) ) ).size
+            : 0;
+          logger.infoWithExtra( "storage_metrics", {
```
sepeterson, line 108: The metrics should all be super lightweight to retrieve which is why I'm alright sending them at startup

PR #3430, opened 2026-03-12
MOB-1186: log storage metrics on startup

sepeterson:
In this PR:

- On startup, log some persisted data storage
  - I did not include photo storage as that, as near as I could tell, would require relatively expensive reading of the file system

### DIFF
--- a/src/components/StartupService.tsx
+++ b/src/components/StartupService.tsx
@@ -6,20 +6,19 @@ import { RealmContext } from "providers/contexts";
 import { useEffect } from "react";
 import { LogBox } from "react-native";
 import DeviceInfo from "react-native-device-info";
-import RNFS from "react-native-fs";
 import Orientation from "react-native-orientation-locker";
 import Realm from "realm";
 import clearCaches from "sharedHelpers/clearCaches";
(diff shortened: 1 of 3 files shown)


src/components/StartupService.tsx
```diff
@@ -95,6 +97,25 @@ const StartupService = ( ) => {
...
       );
 
+      const logStorageMetrics = async ( ) => {
+        try {
+          const realmBytes = realm?.path
+            ? ( await RNFS.stat( realm.path ).catch( () => ( { size: 0 } ) ) ).size
+            : 0;
+          const logFileBytes = logFilePath
+            ? ( await RNFS.stat( logFilePath ).catch( () => ( { size: 0 } ) ) ).size
+            : 0;
```
FLGMwt: Wouldn't we just want these to throw and get caught at `storage_metrics collection failed`?

sepeterson: The rationale was these could throw on first install if for example the log file didn't exist yet and sending to the catch block means the log doesn't get sent. There probably isn't much to learn about persisted storage from the first session but I know the realm db has a surprising amount in it on a fresh install

PR #3432, opened 2026-03-12
Mob 725 reset password form field obscured by keyboard

abbeycampbell:
closes [MOB-725](https://linear.app/inaturalist/issue/MOB-725/reset-password-form-field-obscured-by-keyboard)

Followed the example set by the sign up screen, which is also a single input and submit button which moves to the top of the screen when the keyboard is active. It's a little abrupt IMO but at least the behavior is consistent now, and it lets us get rid of some measuring/scrolling logic.

https://github.com/user-attachments/assets/36dc5859-8e49-4ca4-96e7-c80256268c08

### DIFF
--- a/src/components/LoginSignUp/LoginSignUpWrapper.tsx
+++ b/src/components/LoginSignUp/LoginSignUpWrapper.tsx
@@ -3,7 +3,7 @@ import {
   ImageBackground, ScrollView, View,
 } from "components/styledComponents";
 import type { PropsWithChildren } from "react";
-import React, { useEffect, useRef } from "react";
+import React, { useCallback, useEffect, useRef } from "react";
 import type {
   ImageSourcePropType,
   ImageStyle,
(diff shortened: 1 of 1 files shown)


src/components/LoginSignUp/LoginSignUpWrapper.tsx
```diff
@@ -34,42 +36,37 @@ const LoginSignupWrapper = ( {
   children,
   imageStyle,
 }: Props ) => {
-  const scrollViewRef = useRef( null );
+  const scrollViewRef = useRef<ScrollViewRef>( null );
```
jtklein, line 39: Can we not just use ScrollView as the type here?

abbeycampbell: I tried and got ```Property 'scrollTo' does not exist on type '{}'.```, I think because our `ScrollView` is actually a styled wrapper and doesn't provide type access to `scrollTo`

PR #3433, opened 2026-03-13
MOB-992 start using rolling daily logs instead of a single append-only

FLGMwt:
Phew, this was meant to be a quick dopamine ticket. The _actual_ change is just tweaking the rn-logs config from `filename: "inaturalist-rn-log.txt"` => `filename: "inaturalist-rn-log.{date-today}.txt"` and rn-logs automatically creates a new file every day. One line!

The trickyness was that we want to actually _consume_ the log files somehow while keeping support for existing log files (which will be orphaned, but still potentially valuable for issues with existing installs or updated installs).

I did a lot of refactoring with this to hopefully make some other parts of the Debug screen easier to work with. I tried to make those commits as progressive as possible. I also took some liberties with the layout of the screen since it's a kind of "back of house" UI, but lemme know if there's concerns or feedback on that, especially the "legacy description" copy.

The old log file behavior should be unchanged: we still let you preview the old log and share/email it.

For the new logs, when we share them, we're grabbing the last 20 "days"* of logs, concatenating them into a temp file and immediately sharing that. I figured that was better than making a zip and we shouldn't have issues anymore with the files themselves getting massive. Previewing the logs is similar but we're just collecting them in memory for the UI. *: not necessarily 20 calendar days but the last 20 days where there's been any activity in the app, resulting in a new log file.

Screenshot 2026-03-13 at 10 58 24 AM

Note: we only render the "legacy" section here if we notice that there _is_ the old log file around. I removed the share/email actions from the log preview screen since that's duplicated in the debug menu.

### DIFF
--- a/src/components/Developer/logManagementHelpers.ts
+++ b/src/components/Developer/logManagementHelpers.ts
@@ -22,14 +22,21 @@ const getSortedDailyLogFileInfo = async ( n: number ) => {
     .map( ( { name, size, path } ) => {
       // assumes log files will be prefix.yyyy-mm-dd.txt
       const dateString = name.split( "." )[1];
+      const date = Date.parse( dateString );
+      if ( isNaN( date ) ) {
+        // just log to console, this would only happen if a file is intentionally stuffed in here
+        console.warn( `Unable to parse date from rolling logfile: ${path}` );
+        return null;
+      }
       // RN-logs doesn't zero-pad months/days, so we have to parse date in order to sort
       return {
         name,
         path,
         size,
-        date: Date.parse( dateString ),
+        date,
       };
     } )
+    .filter( a => !!a )
     // flipped for descending
     .sort( ( a, b ) => b.date - a.date );
 
@@ -129,7 +136,7 @@ const concatenateLogsForSharing = async () => {
     // we want to start with the oldest and _add_ newer ones as we go
     .reverse();
 
-  for ( const { path } of mostRecentLogs.reverse() ) {
+  for ( const { path } of mostRecentLogs ) {
     // eslint-disable-next-line no-await-in-loop
     const chunkContents = await RNFS.readFile( path );
     // eslint-disable-next-line no-await-in-loop
@@ -160,7 +167,7 @@ interface LogPreview {
   length: number;
 }
 
-async function getRecentLogContents() {
+async function getRecentLogContentPreview() {
   // we don't need to be precise here, we can jam 10 together and limit it later
   const recentLogsPaths = ( await getSortedDailyLogFileInfo( 10 ) )
     .map( foo => foo.path )
@@ -184,7 +191,7 @@ export function useLogPreview( { legacy }: { legacy: boolean } ): LogPreview | n
     const getLogPreview = async () => {
       const logContents = legacy
         ? await getLegacyLogContents()
-        : await getRecentLogContents();
+        : await getRecentLogContentPreview();
 
       const lines = logContents.split( "\n" );
       const trimmedContent = lines


src/components/Developer/Log.tsx
```diff
@@ -11,38 +11,22 @@ import {
...
-      const lines = logContents.split( "\n" );
-      const trimmedContent = lines
-        .slice( lines.length - 1000, lines.length )
-        .join( "\n" );
-      setContent( { text: trimmedContent, length: lines.length } );
-    } );
-  }, [] );
-
-  if ( !content ) {
+  const logPreview = useLogPreview( { legacy: !!params?.isLegacyLogs } );
```
jtklein, line 21: `!!` only to make TS happy? Or should we actually handle something like the param being `undefined`?

src/components/Developer/logManagementHelpers.ts
```diff
@@ -8,12 +9,37 @@ import {
...
+    .filter( ( { name } ) => name.startsWith( logFileNamePrefix ) )
+    .map( ( { name, size, path } ) => {
+      // assumes log files will be prefix.yyyy-mm-dd.txt
+      const dateString = name.split( "." )[1];
+      // RN-logs doesn't zero-pad months/days, so we have to parse date in order to sort
+      return {
+        name,
+        path,
+        size,
+        date: Date.parse( dateString ),
```
jtklein: Can this become undefined, and then the sort below erroring out somehow?

src/components/Developer/logManagementHelpers.ts
```diff
@@ -99,3 +125,48 @@ export async function shareLegacyLogFile( ) {
...
 }
+
+interface LogPreview {
+  text: string;
+  length: number;
+}
+
+async function getRecentLogContents() {
+  // we don't need to be precise here, we can jam 10 together and limit it later
+  const recentLogsPaths = ( await getSortedDailyLogFileInfo( 10 ) )
```
jtklein, line 172: Why 10? I thought we make 20?

src/components/Developer/logManagementHelpers.ts
```diff
@@ -118,6 +118,35 @@ export async function getLegacyLogfileExists() {
...
+
+const concatenateLogsForSharing = async () => {
+  // this will overwrite / clear an existing temp one if it exists
+  await RNFS.writeFile( temporaryLogForSharingPath, "" );
+
+  const mostRecentLogs = ( await getSortedDailyLogFileInfo( 20 ) )
+    // we want to start with the oldest and _add_ newer ones as we go
+    .reverse();
+
+  for ( const { path } of mostRecentLogs.reverse() ) {
```
jtklein: One reverse too many here. When sharing the combined log files I can see the days in reversed (but the the log lines not). So, end of day 17th next line is beginning of day 16th.

When looking at the logs on the Log screen it is fine.

jtklein: Also, why not re-use getRecentLogContents here?

FLGMwt: The difference here is that we're reading a single file into memory and appending before moving on to the next file. The log preview screen necessarily loads everything into memory it for the display which is also why I went for fewer "recent" files for that one, especially since we trim to 1000 lines anyway.

Since we're writing this on the fly the _moment_ someone presses a email/share I wanted to be extra sure we wouldn't lock or crash due to holding or writing too much at a time.

src/components/Developer/logManagementHelpers.ts
FLGMwt: See comment above: https://github.com/inaturalist/iNaturalistReactNative/pull/3433#discussion_r2948207051

Since we're needing to load it all to display for preview, I'm being more conservative w/ this one since we limit it to 1000 lines anyway.

The shared file is pulling the last 20.

FLGMwt: actually though, for both those reasons, I'll fix the function naming since it's only ever used for the preview

src/components/Developer/Log.tsx
FLGMwt: This isn't a type assertion, it's two `!` NOT operators that do just that to coerce falsy values to a boolean: if `params?.isLegacyLogs` is `undefined`, `!!params?.isLegacyLogs` evaluates to `false`.

FLGMwt: I forget I have this all the time because I never see it, but relevant ; D (there's a C# `??` on the other side)
image

jtklein (approved):
LGTM

PR #3441, opened 2026-03-16
Add @ mention autocomplete

abbeycampbell:
closes [MOB-94](https://linear.app/inaturalist/issue/MOB-94/mentions-autocomplete) as part of creative day 2/2026 (so this was done without real design input, just as a callout)

Adds @ mention autocomplete to comments, comments w/ IDs, editing comments, and notes on observations, but not for the feedback text input. 

https://github.com/user-attachments/assets/de24e273-403b-4f63-aaf3-c36784dee625

After this gets merged I'll do a follow-up to use the new hook in explore user search and fix a little UI bug there.

### DIFF
--- a/src/components/UserList/UserListItem.tsx
+++ b/src/components/UserList/UserListItem.tsx
@@ -9,13 +9,22 @@ import React from "react";
 import User from "realmModels/User";
 import { useTranslation } from "sharedHooks";
 
+type IconVariant = "small" | "medium" | "large";
+
 interface Props {
   item: object;
   countText: string;
   onPress?: ( ) => void;
   accessibilityLabel?: string;
   pressable?: boolean;
+  iconVariant?: IconVariant;
 }
+
+const ICON_VARIANT_SIZE: Record<IconVariant, number> = {
+  small: 22,
+  medium: 62,
+  large: 134,
+};
 
 const UserListItem = ( {
   item,
@@ -23,9 +32,13 @@ const UserListItem = ( {
   onPress,
   accessibilityLabel: accessibilityLabelProp,
   pressable = true,
+  iconVariant,
 }: Props ) => {
   const { t } = useTranslation( );
   const user = item?.user;
+  const fallbackIconSize = iconVariant
+    ? ICON_VARIANT_SIZE[iconVariant]
+    : 40;
 
   const UserListItemContainer = pressable
     ? ( { children } ) => (
@@ -54,11 +67,18 @@ const UserListItem = ( {
     <UserListItemContainer>
       <View className="flex-row items-center w-5/6">
         {user?.icon_url
-          ? <UserIcon uri={User.uri( user )} medium />
+          ? (
+            <UserIcon
+              uri={User.uri( user )}
+              small={iconVariant === "small"}
+              medium={iconVariant === "medium"}
+              large={iconVariant === "large"}
+            />
+          )
           : (
             <INatIcon
               name="person"
-              size={62}
+              size={fallbackIconSize}
             />
           )}
         <View className="ml-3 shrink">
--- a/src/components/UserList/UserList.tsx
+++ b/src/components/UserList/UserList.tsx
@@ -62,6 +62,7 @@ const UserList = ( {
           }
         }}
         accessibilityLabel={accessibilityLabel}
+        iconVariant="medium"
       />
     );
   };


src/components/SharedComponents/Sheets/TextInputSheet.js
```diff
@@ -24,6 +25,7 @@ type Props = {
   buttonText: string,
   confirm: Function,
   description?: string,
+  enableMentions?: boolean,
```
FLGMwt: can we change this to `mentionsEnabled`? the `{thing}Enabled` convention is pretty customary for functionality prop flags, [trimmed example](https://github.com/inaturalist/iNaturalistReactNative/blob/4cf8248beef2c1791f2ecf1653297a67d8dc27a5/src/components/SharedComponents/Map/Map.tsx#L522-L536):

```jsx
loadingEnabled
rotateEnabled={false}
scrollEnabled={scrollEnabled}
zoomEnabled={zoomEnabled}
zoomTapEnabled={zoomTapEnabled}
```

A declarative name for a flag may help clarify it from a similar imperative / name, for example: `function enableMentions() { mentionsEnabled = true; }`

src/components/SharedComponents/Sheets/TextInputSheet.js
```diff
@@ -116,22 +121,38 @@ const TextInputSheet = ( {
...
+                maxLength={maxLength}
+                maxFontSizeMultiplier={2}
+                multiline
+                onChangeText={setInput}
+                placeholder={placeholder}
+                style={[inputStyle, textInputStyle]}
+                testID="TextInputSheet.notes"
+                value={input ?? ""}
+              />
+            )}
```
FLGMwt: Since these are passed identical props, it might be clearer if we decide on the component outside of jsx, something like `const Input = enableMentions ? MentionTextInput : BottomSheetTextInput;`. We have precedent for this: https://github.com/inaturalist/iNaturalistReactNative/blob/4cf8248beef2c1791f2ecf1653297a67d8dc27a5/src/components/SharedComponents/Sheets/BottomSheet.tsx#L103-L105

We have props spreading disabled, otherwise I'd have recommended that ; D

src/components/SharedComponents/Sheets/TextInputSheet.js
```diff
@@ -144,12 +165,7 @@ const TextInputSheet = ( {
               <CharLimit current={input?.length} limit={maxLength} t={t} />
             ) }
             <Pressable
-              onPress={() => {
-                textInputRef?.current?.clear();
-                if ( Platform.OS === "android" ) {
-                  setInput( "" );
-                }
-              }}
```
FLGMwt, line 152: ooh, neat, do we not need this handling anymore?

src/components/SharedComponents/MentionTextInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+    if ( query === "" ) {
+      setDebouncedQuery( "" );
+      return;
+    }
+
+    debounceRef.current = setTimeout( ( ) => {
+      setDebouncedQuery( query );
+      debounceRef.current = null;
+    }, DEBOUNCE_MS );
+  }, [onChangeText] );
```
FLGMwt, line 84: Maybe a good opportunity to add a shared hook. I see that we do this for debouncing in our search bar as well, but I think both could use a simpler effect-based implementation without the ref: https://github.com/uidotdev/usehooks/blob/main/index.js#L239-L253. Then (I think?) we could just do:

```js
const query = getMentionQuery(props.value);
const debouncedQuery = useDebounce(query, ms);
// ...
onChangeText={props.handleChangeText}
```

src/components/SharedComponents/MentionTextInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+    ( item: { login?: string } ) => {
+      const username = item?.login;
+      if ( !username ) return;
+      const lastAt = getLastAtIndex( value );
+      if ( lastAt < 0 ) return;
+      const newValue = `${value.slice( 0, lastAt )}@${username} `;
+      onChangeText( newValue );
+    },
+    [value, onChangeText],
+  );
```
FLGMwt, line 96: It seem odd to me that we should have any "unsuccessful" user selection here. We have all this information before we render the user list right? If so, I think we should filter them based on this criteria up front so that we don't hit this case of selecting a user having no effect.

src/sharedHooks/useUserSearch.ts
```diff
@@ -0,0 +1,35 @@
+import { fetchSearchResults } from "api/search";
+import type { ApiOpts, ApiUser } from "api/types";
+import { useAuthenticatedQuery } from "sharedHooks";
+
+const USER_SEARCH_FIELDS = "user.id,user.login,user.icon_url,user.observations_count";
```
sepeterson, line 5: Looks like we might not need observations_count

abbeycampbell: Yeah not for this. but I left it in because as a follow up I wanna have the explore user search use this hook too and for that, we show you the user's obs count in the list.

src/components/SharedComponents/MentionTextInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+      return [
+        style,
+        { height: Math.max( flattenedStyles.height - MENTION_LIST_MAX_HEIGHT, MIN_INPUT_HEIGHT ) },
+      ];
+    }
+    return style;
+  }, [style, showList] );
+
+  const emptyListContent = useMemo( ( ) => {
+    if ( !showList ) return null;
```
FLGMwt: I don't think `return null` ever happens, right? Since we conditionally render the parent of this with `!showList`.

In any case, this var should not be memoized since it's just a container for markup. That is, it's no different than an inline ternary.

src/sharedHooks/useUserSearch.ts
```diff
@@ -0,0 +1,35 @@
...
+        sources: "users",
+        fields: USER_SEARCH_FIELDS,
+      },
+      optsWithAuth,
+    ),
+    { enabled: shouldFetch },
+  );
+
+  return {
+    users: users ?? [],
```
FLGMwt, line 29: does this handle a case separate from the defaulting of `users` above?

src/components/SharedComponents/MentionTextInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import type { TextInputProps } from "react-native";
+import { ScrollView, StyleSheet } from "react-native";
+import { useTranslation } from "sharedHooks";
+import useUserSearch from "sharedHooks/useUserSearch";
+
+const MENTION_MATCH = /\B@([A-Za-z][\w-]*)$/;
```
FLGMwt, line 23: Would you mind adding a comment describing the pattern and the flow of this component? My regex is rusty and I think I need a bit of help understanding the actual lifecycle of input => match dropdown here, considering wordbreaks and editing in the middle of an input

src/components/UserList/UserListItem.tsx
```diff
@@ -70,9 +70,11 @@ const UserListItem = ( {
...
             </Body1>
           )}
-          <List2 className="mt-1" maxFontSizeMultiplier={1.5}>
-            {countText}
-          </List2>
+          {!!countText && (
+            <List2 className="mt-1" maxFontSizeMultiplier={1.5}>
+              {countText}
+            </List2>
+          )}
```
FLGMwt, line 97: nice! good spot

src/components/SharedComponents/MentionTextInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+      {showList && (
+        <View
+          className="bg-white border border-lightGray rounded-lg overflow-hidden max-h-[200px]"
+        >
+          <ScrollView keyboardShouldPersistTaps="handled">
+            {users.length === 0
+              ? emptyListContent
+              : users.map( ( item, index ) => (
+                <View key={item?.id}>
+                  {index > 0 && <View className="border-b border-lightGray" />}
```
FLGMwt, line 150: similar to the `handleSelectUser` comment, why would `item` be nullable here? do we need to filter `users` beforehand?

Also: afaict, users is coming straight from the api, is `index === 0` a special case?

src/sharedHooks/useUserSearch.ts
FLGMwt: In that case, I'd lean towards having different fields for each search and either adding an option to `useUserSearch` or exporting two different / case-specific hooks (that wrap the shared "guts" search hook)

src/components/SharedComponents/Sheets/TextInputSheet.js
FLGMwt: We chatted about this: I missed the different `InputComponent` prop! So you're right, we can't do that exactly.

As we mentioned, for now, at least moving the "different" prop to the top or bottom helps highlight that for now while we discuss this type of thing with the team.

abbeycampbell: We talked about this in person, but putting my reply here for posterity. the props aren't quite identical. `MentionTextInput` is actually a wrapper and takes `BottomSheetTextInput` as a prop. I _could_ pull all the shared props to a const and spread them in if we're fine with an eslint-disable, and I think that might be more clear because it would make the `InputComponent` prop very obvious. But I think for now we decided to leave this be, other than moving the `InputComponent` prop to the top.

src/components/SharedComponents/Sheets/TextInputSheet.js
abbeycampbell: Before this was an uncontrolled input, and we needed to imperatively tell the native text field to clear, but android handles native events differently which necessitated the android path. Now we have a controlled input, and react state fully owns the displayed text so clearing the input works the same on both platforms.

src/components/SharedComponents/MentionTextInput.tsx
abbeycampbell: I like this idea. Made a follow-up ticket to implement: https://linear.app/inaturalist/issue/MOB-1253/add-a-hook-that-handles-debounces

src/components/SharedComponents/MentionTextInput.tsx
abbeycampbell: good point, I'm just gonna remove these unnecessary defensive checks

src/components/SharedComponents/MentionTextInput.tsx
abbeycampbell: whoops true

src/sharedHooks/useUserSearch.ts
abbeycampbell: Yes, it handles null cases and makes sure we always return an array; after the first default (`users = []`), `users` can still be null. It's mainly there to help readability in my mind "this will be an array" but I can remove it if you want, ln 29 does handle null and undefined

src/components/SharedComponents/MentionTextInput.tsx
abbeycampbell: lol I certainly cannot read regex 🙃 so yes I will happily add this
fwiw I lifted the matchstring from web: https://github.com/inaturalist/inaturalist/blob/2484c983cbc2e79244b44bba99cbeb1a000cc397/app/assets/javascripts/application.js.erb#L1479
With a minor adjustment to disallow the six extra characters between `Z` and `a` since our usernames may only start with a letter.

src/components/SharedComponents/MentionTextInput.tsx
abbeycampbell: no we do not, good call out. 
The `index > 0` check is just to make sure we don't render an item separator border before the first user in the list.

src/sharedHooks/useUserSearch.ts
abbeycampbell: I kinda feel like that adds a bit of overhead for what is just one extra field. 
I made the choice to not surface observation count in the user list because that's how web does it, however we already have a precedent in the app for showing obs counts here so maybe we would rather be consistent--I'm about to ask abhas about this and what he thinks about this overall. If that's the case, this hook will have the same fields for both cases. If not, I'll revisit this.

src/components/SharedComponents/MentionTextInput.tsx
FLGMwt: omg I totally missed that this was an independent View and not a conditional on the item render 🤦 I'm 0/2 for literacy

src/sharedHooks/useUserSearch.ts
FLGMwt: SGTM!

I was originally hesitant because I know that some fields, if requested, can have a significant impact on what the server actually _does_ and may materially change response time. I clarified with @.pleary though that `user.observations_count` isn't one of those fields.

src/components/SharedComponents/MentionTextInput.tsx
FLGMwt: SGTM!

src/sharedHooks/useUserSearch.ts
FLGMwt: I guess I'm not quite seeing why we want to default at all. UQ's `data` is `undef` when not loaded and `User[]` when loaded, right? That in mind, I don't think it'd ever be `null` either.

src/components/SharedComponents/MentionTextInput.tsx
FLGMwt: Thank you!

FLGMwt (approved):
Thanks for the followups!

I just have the question about the non-loaded state and defaulting for `users`, but shouldn't be blocking

src/sharedHooks/useUserSearch.ts
abbeycampbell: `fetchSearchResults` can [return null](https://github.com/inaturalist/iNaturalistReactNative/blob/ab4c1c87b0bb430d9fe8b0a045e9ab76119571b0/src/api/search.ts#L57)

PR #3447, opened 2026-03-18
Move clearing of cv photos cache to app startup

jtklein:
Closes MOB-1245

As documented in the deliverable of MOB-977 there is a possibility for a user in Default mode to never trigger the cache cleanup for photos resized to send to the `score_image` API and therefore to accumulate an unchecked cache.

Here I am moving the cleanup logic to app startup (where we already have other cache clean up running). This is a fix in Default mode, but it is a difference for Advanced mode users as well because there we were clearing the cache every time the user opened Suggestions before creating a new cached image of the one they are currently looking at. So, in other words in Advanced mode there was usually only ever as much photos in this cache as the latest observation the user looked at had.

I am also adding a small duration measurement to the log.

### DIFF
--- a/src/components/StartupService.tsx
+++ b/src/components/StartupService.tsx
@@ -128,7 +128,7 @@ const StartupService = ( ) => {
         }
         // Run startup tasks when app launches
         if ( realm?.path ) {
-          await clearCaches( isDebugMode( ), realm );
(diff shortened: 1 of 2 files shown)


FLGMwt (approved):
great thought to add the timing log, LGTM

PR #3450, opened 2026-03-18
Update react-native-logs jest mock

jtklein:
This increases the mock we use for logs in jest tests to include the new functions of ...WithExtra.

Reason why this is currently failing: https://github.com/inaturalist/iNaturalistReactNative/pull/3449


FLGMwt (approved):
oh, sorry I missed this! It must not have been exercised this way until now, thank you 🙏

PR #3451, opened 2026-03-19
Migrate PhotoLibrary to TS

jtklein:
Migrated this file to TS as good as I could. There is heavy reliance on the zustand store here which is not typed yet.

### DIFF
--- a/src/components/PhotoImporter/PhotoLibrary.tsx
+++ b/src/components/PhotoImporter/PhotoLibrary.tsx
@@ -109,16 +109,21 @@ const PhotoLibrary = ( ) => {
         throw new Error( "No fileName in pick photo response" );
       }
       const destPath = `${path}/${fileName}`;
-      let sourcePath = `${RNFS.TemporaryDirectoryPath}/${fileName}`;
-      // Get image from uri on android. TemporaryDirectoryPath results in an ANR
-      if ( Platform.OS === "android" ) {
-        if ( !uri ) {
-          throw new Error( "No URI in pick photo response" );
-        }
-        sourcePath = uri;
-      }
+      const getSourcePath = Platform.select( {
+        ios: ( ) => `${RNFS.TemporaryDirectoryPath}/${fileName}`,
+        // Get image from uri on android. TemporaryDirectoryPath results in an ANR.
+        android: ( ) => {
+          if ( !uri ) {
+            throw new Error( "No URI in pick photo response" );
+          }
+          return uri;
+        },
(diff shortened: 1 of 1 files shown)


src/components/PhotoImporter/PhotoLibrary.tsx
```diff
@@ -174,7 +201,7 @@ const PhotoLibrary = ( ): Node => {
       return;
     }
 
-    const importedPhotoUris = selectedImages.map( x => x.uri );
+    const importedPhotoUris = selectedImages.map( x => x.image.uri );
```
jtklein, line 209: This is probably a functional change but I have seen no effect of it. The type of selectedImages helped to find that we were not accessing the uri correctly here.

src/components/PhotoImporter/PhotoLibrary.tsx
```diff
@@ -84,17 +98,26 @@ const PhotoLibrary = ( ): Node => {
...
       // Get image from uri on android. TemporaryDirectoryPath results in an ANR
-      const sourcePath = Platform.OS === "android"
-        ? image.uri
-        : `${RNFS.TemporaryDirectoryPath}/${fileName}`;
+      if ( Platform.OS === "android" ) {
+        if ( !uri ) {
+          throw new Error( "No URI in pick photo response" );
+        }
+        sourcePath = uri;
+      }
```
FLGMwt: Can we break this out into explicitly platform-specific flows? like a `getSourcePath = Platform.select()`

src/components/PhotoImporter/PhotoLibrary.tsx
FLGMwt: That seems a little spooky. Do you have an idea of where it _should_ have presented? like are we handling around this bug somewhere else?

jtklein: I am not sure but after looking into this for 15min now it seems to me we are not actually using `photoLibraryUris` values from the store anywhere downstreams of this file.
In the two code paths where we are setting something to photoLibraryUris: When adding photos to an existing observation we seem to rely on `evidenceToAdd`. When importing for a new obs we seem to only use `groupedPhotos`. The only thing we use it for is to also do an updateObsExifFromPhotos for an existing observation, but probably we have not seen any buggy behaviour because we push users UI-wise to make sure every observation has a date and a location when first saved.

That being said, we are using the .length of the array in the StandardCamera and for that a [undefined, undefined] is just as well as the real values.

PR #3454, opened 2026-03-20
Add some TS to our use of react-navigation

jtklein:
Closely following the documentation here: https://reactnavigation.org/docs/typescript

Started with the basic typing of param lists and screen props, as outlined in this documentation. Centralized to a type file.

Added two examples of how we would use these in a screen that uses useRoute.

Typed all instances of navigator creation.

### DIFF
--- a/src/navigation/types.ts
+++ b/src/navigation/types.ts
@@ -29,10 +29,12 @@ export type OnboardingStackParamList = {
   Onboarding: undefined;
 };
 
+// Tab-only routes (not from SharedStackScreens). Intersected with SharedStackParamList
+// so TabStackParamList matches TabStackNavigator + SharedStackScreens.
 // Note from the documentation:
 // The type containing the mapping must be a type alias. It cannot be an interface.
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
-export type TabStackParamList = {
+export type BaseTabStackParamList = {
   Menu: undefined;
   ObsList: undefined;
   RootExplore: undefined;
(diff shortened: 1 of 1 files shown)


src/navigation/types.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export type LoginStackScreenProps<T extends keyof LoginStackParamList> =
+  CompositeScreenProps<
+    NativeStackScreenProps<LoginStackParamList, T>,
+    RootStackScreenProps
+  >;
+
+// https://reactnavigation.org/docs/typescript/?config=dynamic#specifying-default-types-for-usenavigation-link-ref-etc
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
```
jtklein, line 161: I wanted this snippet to be identical to the documentation so I disabled whatever these warnings are. Have not even looked into those.

src/components/LoginSignUp/LoginForm.tsx
```diff
@@ -31,21 +31,11 @@ interface Props {
...
-interface ParamList {
-  LoginFormParams: LoginFormParams;
-}
-
 const LoginForm = ( {
   scrollViewRef,
 }: Props ) => {
   const navigation = useNavigation( );
-  const { params } = useRoute<RouteProp<ParamList, "LoginFormParams">>( );
+  const { params } = useRoute<LoginStackScreenProps<"Login">["route"]>();
```
FLGMwt, line 38: This is excellent! very exited to chip away at more redsquiggles and one-off `ParamList` stubs because this hadn't been centralized before.

Not for this PR, but what would you think about making our own (pseudo types) `useRoute<TStack: keyof Stacks, TScreen: keyof Stacks[TStack]['route']>`?

src/navigation/types.ts
```diff
@@ -0,0 +1,159 @@
...
+  FollowingList: undefined;
+  Debug: undefined;
+  UILibrary: undefined;
+  UiLibraryItem: undefined;
+  Log: { isLegacyLogs: boolean };
+  Settings: undefined;
+  About: undefined;
+  Donate: undefined;
+  Help: undefined;
+};
```
sepeterson, line 63: Do you think we should intersect this with the shared stack params? Like `& SharedStackParamList`

PR #3455, opened 2026-03-20
Mob 1184 startup time time to interactive

abbeycampbell:
closes [MOB-1184](https://linear.app/inaturalist/issue/MOB-1184/stub-startup-time-time-to-interactive)

- Adds [react-native-performance](https://github.com/oblador/react-native-performance?tab=readme-ov-file) on Ryan's recommendation so we can capture native initialization, JS bundle load times, and the time it takes for content to appear (true tti) easily and without writing any timing code.
- Adds a coordinator (`startupPerformanceTracker` that watches for `StartupService`and `useSyncObservations` to finish their work and the target screen to appear before firing the log.

```
{
  "extraSentinelKey": {
    "screenInteractiveMs": 4735, // the metric we care most about, when can the user actually interact?
    "contentAppearedMs": 3200, // real tti, the user sees content
    "jsBundleLoadedMs": 838, // js loading time
    "nativeLaunchMs": 533, // native launch time
    "startType": "cold", // only recording cold starts for now but maybe someday we'll want warm ones too
    "targetScreen": "OnboardingCarousel",
    "loggedIn": false
  }
}
```

### DIFF
--- a/tests/jest.setup.js
+++ b/tests/jest.setup.js
@@ -20,6 +20,9 @@ global.ReanimatedDataMock = {
   now: () => 0,
 };
 
+global.requestIdleCallback = callback => setTimeout( callback, 0 );
+global.cancelIdleCallback = id => clearTimeout( id );
+
 jest.mock( "react-native-volume-manager", () => ( {
   VolumeManager: {
     getVolume: jest.fn( () => Promise.resolve( 0.5 ) ),


src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+);
+
+const isReady = () => startupServiceDone && syncDone !== false && screenMeta !== null;
+
+const tryEmit = () => {
+  if ( emitted || !isReady() || !screenMeta ) { return; }
+  emitted = true;
+
+  try {
+    const nativeLaunchStart = getMark( "nativeLaunchStart" );
```
jtklein: How did you arrive at those strings? Are they documented somewhere?

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+      screenInteractiveMs,
+      contentAppearedMs,
+      jsBundleLoadedMs,
+      nativeLaunchMs,
+      startType: "cold",
+      targetScreen: screenMeta.targetScreen,
+      loggedIn: screenMeta.loggedIn,
+    } );
+  } catch ( e ) {
+    logger.info( "startup_tti collection failed", e );
```
jtklein: We should use logger.error here.

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+      : "NA";
+
+    logger.infoWithExtra( "startup_tti", {
+      screenInteractiveMs,
+      contentAppearedMs,
+      jsBundleLoadedMs,
+      nativeLaunchMs,
+      startType: "cold",
+      targetScreen: screenMeta.targetScreen,
+      loggedIn: screenMeta.loggedIn,
```
jtklein: Are you considering having to wait for auth status as time spent before being interactive?

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+    const screenInteractiveMs = originMs !== undefined
+      ? Math.round( performance.now() - originMs )
+      : "NA";
+
+    logger.infoWithExtra( "startup_tti", {
+      screenInteractiveMs,
+      contentAppearedMs,
+      jsBundleLoadedMs,
+      nativeLaunchMs,
+      startType: "cold",
```
jtklein: We are not actually doing anything here to filter out warm starts it seems to me. We would need to add code to the native side of the app to filter out warm starts from being measured, right? So, essentially with this setup we don't know if we are measuring a warm start or a cold start. Or your assumption is that in a warm start this will not get called because the state is still there?

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+    // (ms since performance.timeOrigin ≈ native launch)
+    const screenInteractiveMs = originMs !== undefined
+      ? Math.round( performance.now() - originMs )
+      : "NA";
+
+    logger.infoWithExtra( "startup_tti", {
+      screenInteractiveMs,
+      contentAppearedMs,
+      jsBundleLoadedMs,
+      nativeLaunchMs,
```
jtklein: I was seeing entries of < 10ms for nativeLaunchMs during testing, Is that possible? Or maybe it is because of development mode? Either way, do we have to wait for this somehow: https://github.com/oblador/react-native-performance/blob/66d91c03d1a6dc9d2b0c4f697440c9c21340064c/packages/react-native-performance/README.md?plain=1#L147

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+// null = no sync expected (e.g. new user), false = pending, true = done
+let syncDone: boolean | null = null;
+let screenMeta: ScreenMeta | null = null;
+let emitted = false;
+
+const getMark = ( name: string ): PerformanceEntry | undefined => (
+  performance.getEntriesByName( name, "react-native-mark" )[0]
+);
+
+const isReady = () => startupServiceDone && syncDone !== false && screenMeta !== null;
```
jtklein: Waiting for syncDone: is this an assumption that the sync process is somehow keeping the app from being interactive? I don't think we should assume that. I think we should rather trigger the tti log as soon as the app is interactive.
For that end: Can we not just add a useFocusEffect to the first screen that should render and trigger the emit in a callback as soon as the JS thread is idle? Previously like with InteractionManager in newer RNN versions with requestIdleCallback?

abbeycampbell: This is a really good point; I think I got really in the weeds with this after our discussion on the ticket where we agreed we'd consider tti for our purposes 'when heavy startup work finishes and the app is actually responsive', but using requestIdleCallback seems like a good way doing this

src/sharedHelpers/startupPerformanceTracker.ts
abbeycampbell: I'm not sure about this, I was seeing more like ~500ms when doing a dev build to my device. I'll look into it

src/sharedHelpers/startupPerformanceTracker.ts
abbeycampbell: yeah, this value is based on the assumption that this wouldn't re-fire on warm starts because `emitted = true`. But your right that adding a check on the native side is a better approach.

src/sharedHelpers/startupPerformanceTracker.ts
abbeycampbell: no, this is just data I thought it might be helpful to know, since someone can have a target screen of `MyObservations` and be logged in or out. This is just our usual auth status realm read

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+  targetScreen: "MyObservations" | "OnboardingCarousel";
+  loggedIn: boolean;
+}
+
+// Module-level state — the module itself is the singleton
+let startupServiceDone = false;
+// null = no sync expected (e.g. new user), false = pending, true = done
+let syncDone: boolean | null = null;
+let screenMeta: ScreenMeta | null = null;
+let emitted = false;
```
FLGMwt: As far as I know, we don't do this type of "stateful module" / singleton in the codebase and I don't think it's common in the ecosystem. We have the precedent of `[MapPerformanceTracker](https://github.com/inaturalist/iNaturalistReactNative/blob/cc39a844e8375964ecb932fd61dc6c060a50da46/src/sharedHelpers/mapPerformanceTracker.ts#L1)` using `class` for a very similar purpose which makes the encapsulation boundaraies of private state / public methods easier to read.  Can we switch to follow that pattern?

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+    syncDone = true;
+  }
+  tryEmit();
+};
+
+// Called from landing screens to register screen metadata.
+// For MyObservations: always called after expectSync() in the same useFocusEffect,
+// so syncDone is already false when tryEmit runs here and isReady() returns false.
+// For OnboardingCarousel: no sync is expected, so syncDone stays null and isReady()
+// is satisfied as soon as the startup service completes.
```
FLGMwt: This seems overly complicated. It's fishy that we have particular sequence requirements of our callers based on our internal state and implementation. Couldn't we have separate startup & carousel methods exposed for? I think we're doing _more_ work to expose a shared api but making them care about each other.

src/sharedHelpers/startupPerformanceTracker.ts
```diff
@@ -0,0 +1,112 @@
...
+let emitted = false;
+
+const getMark = ( name: string ): PerformanceEntry | undefined => (
+  performance.getEntriesByName( name, "react-native-mark" )[0]
+);
+
+const isReady = () => startupServiceDone && syncDone !== false && screenMeta !== null;
+
+const tryEmit = () => {
+  if ( emitted || !isReady() || !screenMeta ) { return; }
```
FLGMwt: wait, when wouldn't we have `screenMeta`? isn't that what drives calling tryEmit?

index.js
```diff
@@ -43,6 +44,8 @@ import { name as appName } from "./app.json";
 import { log } from "./react-native-logs.config";
 import { getUserAgent } from "./src/api/userAgent";
 
+beginColdStart();
```
FLGMwt: Do we need this? The module is initializing the variables on cold launch, I think this only ever gets called _immediately_ after those are initialized.

src/components/Onboarding/OnboardingCarousel.js
```diff
@@ -186,6 +187,13 @@ const OnboardingCarousel = ( ) => {
     );
   }
 
+  // Reached on iOS once images are loaded, and immediately on Android.
+  // The tracker's emitted guard makes repeated render calls safe.
+  markScreenReached( {
+    targetScreen: "OnboardingCarousel",
+    loggedIn: false,
+  } );
```
FLGMwt: this should not be called on every render

FLGMwt (commented):
My main concern is the coupling between the caller and the implementation around when we _expect_ you to call something like `expectSync`. I think maybe the sync-y workflow would fit more into a "return stopTimerAndLog" similar to the firebase trace: https://github.com/inaturalist/iNaturalistReactNative/blob/cc39a844e8375964ecb932fd61dc6c060a50da46/src/stores/createFirebaseTraceSlice.ts#L80

I don't know where y'all landed on splitting the work for this ticket and a followup, but I'd like to at least clear up the above

src/sharedHelpers/startupPerformanceTracker.ts
abbeycampbell: I can look into this in a follow-up, it seems not as straightforward as I thought and I'm not all that familiar with Obj-C stuff

src/sharedHelpers/startupPerformanceTracker.ts
abbeycampbell: I think now that we're using `requestIdleCallback`, we don't need to worry about this, because our JS thread is already idle and we should be able to read all the marks

index.js
abbeycampbell: You're correct! We don't need this, it was a vestige left over from my first pass

src/components/Onboarding/OnboardingCarousel.js
abbeycampbell: yeah, my idea here was that because of the emitted flag, any subsequent calls would be no-ops and we could avoid having a useEffect dependent on `imagesLoaded`, which seems to be an odd android thing I'm sure we'll have to deal with soon anyway. But we can do it that way instead.

PR #3456, opened 2026-03-20
MOB-1224: refactor project fetching for ProjectList

sepeterson:
In this PR:

- Update ProjectListContainer to accept either an observationUuid or userId and use it to fetch projects rather than passing them entirely in route params
- Update navigations to ProjectList

### DIFF
--- a/src/components/ProjectList/ProjectListContainer.tsx
+++ b/src/components/ProjectList/ProjectListContainer.tsx
@@ -3,6 +3,7 @@ import { useNavigation, useRoute } from "@react-navigation/native";
 import type { ApiProject } from "api/types";
 import { fetchUserProjects } from "api/users";
 import {
+  ActivityIndicator,
   Body1,
   ViewWrapper,
 } from "components/SharedComponents";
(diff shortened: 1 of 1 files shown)


src/components/ProjectList/ProjectListContainer.tsx
```diff
@@ -1,20 +1,71 @@
...
   const { t } = useTranslation( );
 
+  const { remoteObservation } = useRemoteObservation(
+    observationUuid,
+    !!observationUuid,
+  );
+
+  const traditionalProjects = remoteObservation?.project_observations?.map(
+    p => p.project,
+  ) || [];
```
jtklein, line 42: Is there a possibility for some early renders that show the ListEmptyComponent because the data is initialized as a [] now instead of undefined. Or in other words should we show a loading state while the respective query isLoading/isFetching instead of the empty array UI fallback?
This is phrased as a question because I haven't really seen this now while testing but just from looking at the code it would seem to be necessary to me. Probably it helps that we just come from a screen where we did the exact same query, and hence still have some cache.

src/components/UserProfile/UserProfile.js
```diff
@@ -199,7 +171,7 @@ const UserProfile = ( ): Node => {
             text={t( "VIEW-PROJECTS" )}
             onPress={( ) => navigation.navigate( "ProjectList", {
               userId,
-              headerOptions: projectsHeaderOptions,
+              userLogin: user.login,
```
jtklein, line 174: Ah, yes, that also works. That's better than having to query on the project list for the user again.

PR #3467, opened 2026-03-24
track simultaneous suggestions in firebase

FLGMwt:
MOB-981

As discussed with the team, this really flexes the pain points of Firebase and why it's _probably_ not well-suited for this task. Tell you what though, I learned a _ton_ (even if the learning wasn't necessarily pleasant).

Main thing here is that the capabilities of event parameters and structured data are really poorly documented. I was originally optimistic about structured data after seeing a smattering of examples using _at least_ arrays of primitives, but was thrown off again seeing arrays of _objects_ and yet again after seeing neither of those seem to work. I finally discovered that structured data _is_ supported but only for the `items` property of some of the built-in ecommerce events. So what I settled with:

Since we're likely never going to use `purchase` and `view_item_list` (ostensibly for a product page) _and_ we can remap events at the Firebase level and even more explicitly at the BigQuery level, I opted to use those for the structured data support with the expectation that we wouldn't consume those events as-is. Similarly, the item-level `category` built-in properties capture taxonId and score for the suggestion results.

This is janky for sure, so I'll pose the question: does everyone prefer this? Or is it _too_ janky? The alternative would be flattening the result structure which puts more mapping work on the consumer. That is, we'd have to do something like:

```json
{
  // ...
  taxonId0: number,
  score0: number,
  taxonId1: number,
  score1: number,
  taxonId2: number,
  score2: number,
  // ... up until 10 or FB's property limit considering other props
}
```

### DIFF
--- a/src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
+++ b/src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
@@ -1,0 +1,75 @@
+import { logFirebaseEvent } from "sharedHelpers/tracking";
+import { v4 as uuidv4 } from "uuid";
+
+import type { OfflineSuggestionsResponse } from "./useOfflineSuggestions";
+import type { OnlineSuggestionsQueryResponse } from "./useOnlineSuggestions";
+
+// GA has very limited support for structured data. Only certain built-in events have support for
+// non-primitive data structures. We're piggybacking on those built-ins for this reporting.
+// These are then mapped to more appropriately named suggestion events using "event modifications"
+// in the Firebase Events Config.
+// https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#purchase
+const offlineEventName = "purchase";
+// https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#view_item_list
+const onlineEventName = "view_item_list";
+// similarly, GA, at least through Firebase, doesn't seem to respect custom properties on `items`
+// so we're using the generic "item_category" properties which will also be remapped.
+const taxonIdPropertyName = "item_category";
+const taxonScorePropertyName = "item_category2";
+
+const logSuggestionAnalytics = (
+  optimisticObservationUuid: string,
+  offlineSuggestions: OfflineSuggestionsResponse,
+  onlineSuggestions: OnlineSuggestionsQueryResponse,
+) => {
+  const transactionId = uuidv4();
+
+  logFirebaseEvent( offlineEventName, {
+    transactionId,
+    optimisticObservationUuid,
+    prediction_source: "offline",
+    commonAncestorTaxonId: offlineSuggestions.commonAncestor?.taxon.id ?? "NA",
+    commonAncestorCombinedScore: offlineSuggestions.commonAncestor?.combined_score ?? "NA",
+    items: offlineSuggestions.results
+      .slice( 0, 10 )
+      .map( suggestion => ( {
+        item_id: String( suggestion.taxon.id ),
+        [taxonIdPropertyName]: String( suggestion.taxon.id ),
+        [taxonScorePropertyName]: String( suggestion.combined_score ),
+      } ) ),
+  } );
+
+  logFirebaseEvent( onlineEventName, {
+    transactionId,
+    optimisticObservationUuid,
+    prediction_source: "online",
+    commonAncestorTaxonId: onlineSuggestions.common_ancestor?.taxon.id ?? "NA",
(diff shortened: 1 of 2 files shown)


src/sharedHelpers/tracking.ts
```diff
@@ -7,7 +7,7 @@ import { log } from "sharedHelpers/logger";
 
 const logger = log.extend( "tracking.ts" );
 
-type FirebaseParameters = Record<string, string | number | string[]>
+type FirebaseParameters = Record<string, unknown>
```
jtklein, line 10: I think we should stay more strict with the types here.
Top level keys of param need to have primitive values. Then we can extend to allow items with an Item[] that we also type according to the restrictions there.

src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const logSuggestionAnalytics = (
+  optimisticObservationUuid: string,
+  offlineSuggestions: OfflineSuggestionsResponse,
+  onlineSuggestions: OnlineSuggestionsQueryResponse,
+) => {
+  const transactionId = uuidv4();
+
+  logFirebaseEvent( offlineEventName, {
+    transactionId,
```
jtklein, line 28: Do ew need this transactionId?

src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
```diff
@@ -0,0 +1,75 @@
+import { logFirebaseEvent } from "sharedHelpers/tracking";
+import { v4 as uuidv4 } from "uuid";
+
+import type { OfflineSuggestionsResponse } from "./useOfflineSuggestions";
+import type { OnlineSuggestionsQueryResponse } from "./useOnlineSuggestions";
+
+// GA has very limited support for structured data. Only certain built-in events have support for
+// non-primitive data structures. We're piggybacking on those built-ins for this reporting.
+// These are then mapped to more appropriately named suggestion events using "event modifications"
+// in the Firebase Events Config.
```
jtklein, line 10: I have not seen this in the console now yet. I do see the events coming in as purchase and view_item_list, but I assume this is something we can do after the merge as well.

src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
```diff
@@ -0,0 +1,75 @@
...
+const taxonScorePropertyName = "item_category2";
+
+const logSuggestionAnalytics = (
+  optimisticObservationUuid: string,
+  offlineSuggestions: OfflineSuggestionsResponse,
+  onlineSuggestions: OnlineSuggestionsQueryResponse,
+) => {
+  const transactionId = uuidv4();
+
+  logFirebaseEvent( offlineEventName, {
```
jtklein, line 27: Just curious, why did you decide to split offline and online into two different events? The actual shape of the data is the same in both cases. I think I would have recorded them under the same event with the prediction_source as offline or online.

src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
FLGMwt: chatted offline: needs to happen as a follow-up

src/sharedHelpers/tracking.ts
FLGMwt: I'm still working through _what_ Firebase actually accepts / respects here which unfortunately in this case has lag time because it's not always apparent in just debug view 😒

Once the parameters settle, I'll come back to this and ratchet the types down.

src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
FLGMwt: chatted offline: again with the lag time, I'm _hoping_ to see after we have a few days of events if this can eliminate the need for separate event types for online/offline. txnId is _supposed_ to prevent event squashing, but doesn't appear to in debug view. Since debug view has a number of unclear and undocumented differences from actual FB output, I want to see if this makes an impact.

src/sharedHooks/useSuggestions/suggestionComparisonExperiment.ts
FLGMwt: chatted offline: mentioned above w/ txnId, with the same event name, FB squashes logs that are close together and sequential so online and offline were getting merged. that is: w/ the same event name, `items` would include both online and offline suggestions. using separate event names prevents that.

PR #3471, opened 2026-03-25
Switch to react-native-exify

jtklein:
Closes MOB-191, also closes MOB-192

There are actually not a lot of changes here. The most basic realisation after working on this for several hours was that even though there is code in the native side of the react-native-exif-reader to take the timezone offset of the datetime into account in the end before sending it over to the app from the library we still only see the "wall clock" part of the date time. So, e.g. a photo from Australia with an exif DateTimeOriginal of "2026:01:05 18:37:19" and an OffsetTimeOriginal of "+11:00" was still returned to the app as "2026-01-05T18:37:19.000". We just take this string and after re formatting only send it to the server to let it figure out which timezone this is in based on the coordinates. The "timezone_offset" from the exif-reader we never ingested anywhere.

I think this also closes the ticket MOB-192, at least on my Android device I am able to import a photo and make an observation that ends up in the correct time zone. But I'd appreciate extra testing with importing photos taken in different time zones than yours.

### DIFF
--- a/src/sharedHelpers/parseExif.ts
+++ b/src/sharedHelpers/parseExif.ts
@@ -63,7 +63,7 @@ const formatExifDateAsString = ( tags: ExifTags ): string | null => {
 
 // Parse the EXIF of all photos - fill out details (lat/lng/date) from all of these,
 // in case the first photo is missing EXIF
-const readExifFromMultiplePhotos = async ( photoUris: string[] ): Promise<object> => {
+const readExifFromMultiplePhotos = async ( photoUris: string[] ) => {
   const unifiedExif: {
     latitude?: number;
     longitude?: number;


src/components/SharedComponents/DateDisplay.tsx
```diff
@@ -56,6 +57,9 @@ const DateDisplay = ( {
     || taxonGeoprivacy === "private";
 
   const formattedDate = useMemo( () => {
+    if ( isDebug ) {
+      return dateString;
```
jtklein, line 61: I found it helpful while developing this to look at the unformatted datetime string here. Does not need to be pulled in though.

src/sharedHelpers/parseExif.ts
```diff
@@ -0,0 +1,131 @@
+import type { ExifTags } from "@lodev09/react-native-exify";
```
jtklein, line 1: Sorry about this being a new file. I did a git mv first without any code changes, but somehow this still shows me this as a new file here in the Github PR UI. It's probably easiest to copy the previous parseExif.js into a new file in this branch and compare it with this new version.

src/sharedHelpers/parseExif.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const logger = log.extend( "parseExif.ts" );
+
+class UsePhotoExifDateFormatError extends Error {}
+// https://wbinnssmith.com/blog/subclassing-error-in-modern-javascript/
+Object.defineProperty( UsePhotoExifDateFormatError.prototype, "name", {
+  value: "UsePhotoExifDateFormatError",
+} );
+
+const formatExifDateAsString = ( tags: ExifTags ): string | null => {
```
jtklein, line 15: Have combined all previous datetime helper functions into one.

src/sharedHelpers/parseExif.ts
```diff
@@ -0,0 +1,131 @@
...
+// in case the first photo is missing EXIF
+const readExifFromMultiplePhotos = async ( photoUris: string[] ): Promise<object> => {
+  const unifiedExif: {
+    latitude?: number;
+    longitude?: number;
+    observed_on_string?: string | null;
+    positional_accuracy?: number;
+  } = {};
+
+  // TODO: when uri starts with content do we need to check if we have required permission
```
jtklein, line 74: Not having these permissions means that for "content://" uri's personal exif metadata like GPS coords get's redacted before we get the file here. This is essentially this ticket here: https://linear.app/inaturalist/issue/MOB-471/android-geotagged-location-not-imported
So, I opted not to tackle it here but in follow-up PR because it might spiral into some UI work as well, like possibly having to ask for permissions first.

src/sharedHelpers/parseExif.ts
```diff
@@ -0,0 +1,131 @@
...
+    : uri ) );
+  const responses = await Promise.allSettled( normalizedUris.map( uri => Exify.read( uri ) ) );
+
+  // If any of the EXIF reads were rejected, log the reasons, but do continue
+  const rejectedReasons = responses
+    .filter( r => r.status === "rejected" )
+    .map( r => r.reason );
+  if ( rejectedReasons.length > 0 ) {
+    rejectedReasons.forEach(
+      reason => logger.error( "Failed to read EXIF data from a photo:", reason ),
```
jtklein, line 88: Added a server log for failures.

src/sharedHelpers/parseExif.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  // this returns a string, in the same format as photos which fall back to the
+  // photo timestamp instead of exif data
+  return formatISONoTimezone( zonedDate );
+};
+
+// Parse the EXIF of all photos - fill out details (lat/lng/date) from all of these,
+// in case the first photo is missing EXIF
+const readExifFromMultiplePhotos = async ( photoUris: string[] ): Promise<object> => {
```
sepeterson: I would remove the `: Promise<object>` here and allow the return type to be inferred

src/sharedHelpers/parseExif.ts
```diff
@@ -0,0 +1,131 @@
...
+  if ( !match ) return null;
+
+  const year = Number( match[1] );
+  const month = Number( match[2] );
+  const day = Number( match[3] );
+  const hour = Number( match[4] );
+  const minute = Number( match[5] );
+  const second = Number( match[6] );
+
+  const utcMs = Date.UTC( year, month - 1, day, hour, minute, second );
```
sepeterson: lol that month starts at 0 but day starts at 1

sepeterson (approved):
Looks good to me! I was also able to make observations from android with the correct time zone.

I had a couple of the obs I made on this branch fail to upload but I can't say for certain it has to do with these changes. Something to keep an eye out for.

src/sharedHelpers/parseExif.ts
jtklein: Well, JS. The third param is actually date and not day, I did misname this here.

PR #3472, opened 2026-03-25
MOB-1091 roll back ObsEdit changes on nav back to Match screen

sepeterson:
In this PR:

When a user navigates Match -> ObsEdit, save a snapshot of currentObservation in zustand and make copies of any photos on the obs in the file system. When  a user taps the back button, they have an option to discard whatever changes they made and be returned to the Match screen with the saved values restored to currentObservation

I wanted to store the snapshot in component state but there are so many paths away from and back to ObsEdit that preserving component state would require a huge refactor.


src/components/ObsEdit/ObsEditHeader.js
```diff
@@ -86,14 +100,11 @@ const ObsEditHeader = ( {
...
-      // When coming from the match screen, we don't have a version of the match to roll back to
-      // so if there are changes, they need to restart
-      // In the future, we'll support a rollback https://linear.app/inaturalist/issue/MOB-1091/match-screen-edit-flow-should-roll-back-changes-on-back-navigation
+    } else if ( canRollbackToMatch ) {
       if ( unsavedChanges ) {
-        setDiscardObservationSheetVisible( true );
+        setDiscardChangesSheetVisible( true );
       } else {
-        navigation.goBack( );
+        navigation.dispatch( StackActions.popTo( "Match" ) );
```
jtklein, line 109: Why this instead of goBack?

src/components/ObsEdit/ObsEditHeader.js
```diff
@@ -44,12 +47,23 @@ const ObsEditHeader = ( {
...
-  const discardChanges = useCallback( ( ) => {
+  const discardChanges = useCallback( async ( ) => {
     setDiscardChangesSheetVisible( false );
-    exitObservationFlow( {
-      navigate: ( ) => navigateToObsDetails( navigation, currentObservation?.uuid ),
-    } );
-  }, [currentObservation?.uuid, exitObservationFlow, navigation] );
+    if ( canRollbackToMatch ) {
+      await rollback( );
+      navigation.dispatch( StackActions.popTo( "Match" ) );
```
jtklein, line 55: Why this instead of goBack?

src/navigation/types.ts
```diff
@@ -11,7 +11,7 @@ import type { NativeStackScreenProps } from "@react-navigation/native-stack";
 // The type containing the mapping must be a type alias. It cannot be an interface.
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
 export type SharedStackParamList = {
-  ObsEdit: undefined;
+  ObsEdit: { lastScreen?: string } | undefined;
```
jtklein: Technically not any string but a defined set of screen names. I am not asking to change this because it would unnecessary increase the scope of this PR I think, especially because some of the screens that set/get this param are not in TS yet.
So, we should probably make a ticket to have the goal of 1) type this param correctly, and a prerequisite of subtickets to type the screens that set/get this param.

src/sharedHooks/index.ts
```diff
@@ -21,6 +21,7 @@ export { default as useLocalObservation } from "./useLocalObservation";
 export { default as useLocalObservations } from "./useLocalObservations";
 export { default as useLocationPermission } from "./useLocationPermission";
 export { default as useNavigateToObsEdit } from "./useNavigateToObsEdit";
+export { default as useObsEditRollback } from "./useObsEditRollback";
```
jtklein: Maybe we should move away from using barrel import files. I am fine if you do a direct import instead of via an index.ts.

src/sharedHelpers/rollbackPhotos.ts
```diff
@@ -0,0 +1,109 @@
...
+    const dirExists = await RNFS.exists( rollbackPhotosPath );
+    if ( dirExists ) {
+      await unlink( rollbackPhotosPath );
+    }
+  } catch ( e ) {
+    logger.error( `clearRollbackPhotos error: ${e}` );
+  }
+};
+
+export const backupObservationPhotos = async (
```
jtklein, line 25: Why do we need to do this backup and restore?

src/sharedHelpers/rollbackPhotos.ts
```diff
@@ -0,0 +1,109 @@
...
+  observations: RealmObservationPojo[],
+  currentObservationIndex: number,
+): Promise<BackupMapping[]> => {
+  try {
+    const observation = observations[currentObservationIndex];
```
jtklein: If all we do with these two params is to get the observation, we can only use the obs as one param?

jtklein (changes requested):
I haven't finished the review but I already left some questions/comments before I recorded the video I shared in Slack.

So, I'll request changes now so that the inline comments become canon.

src/components/ObsEdit/ObsEditHeader.js
sepeterson: If you add evidence or go to suggestions to change the taxon, your nav state ends up as like Match | ObsEdit | Suggestions | ObsEdit so `goBack` takes you back to suggestions. Not ideal but fixing the underlying issue is its own can of worms

src/components/ObsEdit/ObsEditHeader.js
sepeterson: Same as above

src/sharedHelpers/rollbackPhotos.ts
sepeterson: Good point! I'll change this

src/sharedHelpers/rollbackPhotos.ts
sepeterson: It was more complicated to change how the delete methods work than to simply keep existing logic and make a copy to revert to. Addressed why we can't roll back using camera roll data in the slack thread

src/components/ObsEdit/ObsEditHeader.js
```diff
@@ -86,14 +100,11 @@ const ObsEditHeader = ( {
...
-      // so if there are changes, they need to restart
-      // In the future, we'll support a rollback https://linear.app/inaturalist/issue/MOB-1091/match-screen-edit-flow-should-roll-back-changes-on-back-navigation
+    } else if ( canRollbackToMatch ) {
       if ( unsavedChanges ) {
-        setDiscardObservationSheetVisible( true );
+        setDiscardChangesSheetVisible( true );
       } else {
-        navigation.goBack( );
+        navigation.dispatch( StackActions.popTo( "Match" ) );
       }
```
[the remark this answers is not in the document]
sepeterson: This is really an edge case but I think I agree. Implementing

sepeterson:
@jtklein Thanks for the review! I think I've addressed all your comments. I'm not sure on two of the copilot comments & will hopefully be able to taker a closer look at them later today.

src/sharedHelpers/rollbackPhotos.ts
```diff
@@ -0,0 +1,107 @@
...
+  backupPath: string;
+}
+
+const getLocalPhotoPath = ( localFilePath: string ): string | null => {
+  const pieces = localFilePath?.split( "photoUploads/" );
+  if ( !pieces || pieces.length <= 1 ) return null;
+  return `${photoUploadPath}/${pieces[1]}`;
+};
+
+export const clearRollbackPhotos = async ( ): Promise<void> => {
```
jtklein: This logic lies in clearCache for the other folders.

src/sharedHelpers/rollbackPhotos.ts
```diff
@@ -0,0 +1,107 @@
...
+    if ( photosWithLocalPath.length === 0 ) return [];
+
+    await RNFS.mkdir( rollbackPhotosPath );
+
+    const results = await Promise.all(
+      photosWithLocalPath.map( async ( op ): Promise<BackupMapping | null> => {
+        const { localFilePath } = op.photo;
+        if ( !localFilePath ) return null;
+        const originalPath = getLocalPhotoPath( localFilePath );
+        const filename = localFilePath.split( "/" ).pop( );
```
jtklein, line 45: It might have been good to make a prerequisite PR to change how we save backup photos to always have the photo uuid as filename.

src/sharedHelpers/rollbackPhotos.ts
```diff
@@ -0,0 +1,107 @@
...
+          return null;
+        }
+
+        await RNFS.copyFile( originalPath, backupPath );
+        return { originalPath, backupPath };
+      } ),
+    );
+
+    const mappings = results.filter(
+      ( m ): m is BackupMapping => m !== null,
```
jtklein: What does this part do "m is BackupMapping"? At this point m is `BackupMapping | null`?

src/sharedHelpers/rollbackPhotos.ts
```diff
@@ -0,0 +1,107 @@
...
+        const filename = localFilePath.split( "/" ).pop( );
+        if ( !originalPath || !filename ) return null;
+
+        const backupPath = `${rollbackPhotosPath}/${filename}`;
+        const sourceExists = await RNFS.exists( originalPath );
+        if ( !sourceExists ) {
+          logger.warn(
+            `backupObservationPhotos: source not found: ${originalPath}`,
+          );
+          return null;
```
jtklein, line 54: Do we need to error the entire operation out at this point? If we fail to get a backup will the user be able to delete the original and expect to be able to get it back?

src/components/ObsEdit/ObsEditHeader.js
jtklein: But that's working around an existing bug then. Accepting a suggestion should already nav back to the existing ObsEdit in the stack. If we have a ticket for this bug do we need to comment to adapt this line of code as well? In general it seems to be no problem to use popTo.

src/stores/createObservationFlowSlice.js
```diff
@@ -205,6 +210,43 @@ const createObservationFlowSlice = ( set, get ) => ( {
   setAICameraSuggestion: suggestion => set( {
     aICameraSuggestion: suggestion,
   } ),
+  restoreRollbackSnapshot: ( ) => set( state => {
+    const snapshot = state.rollbackSnapshot;
+    if ( !snapshot ) return {};
+    return {
+      observations: snapshot.observations,
+      currentObservation: snapshot.observations[snapshot.currentObservationIndex],
```
jtklein, line 229: Your changes should only work for Match to ObsEdit I wonder if it is easier on us if we make this also apparent in the API here, so instead of observations, currentObservation, and currentObservationIndex, only always keep one observation in the snapshot.

src/sharedHooks/useObsEditRollback.ts
```diff
@@ -0,0 +1,59 @@
...
+  );
+  const backupMappings = useStore( state => state.backupMappings );
+  const setBackupMappings = useStore( state => state.setBackupMappings );
+
+  const hasBackedUp = useRef( false );
+
+  useEffect( ( ) => {
+    if ( hasBackedUp.current || canRollbackToMatch || params?.lastScreen !== "Match" ) return;
+
+    hasBackedUp.current = true;
```
jtklein: Does it make sense to put this after snapshot and photo backup?

src/sharedHooks/useObsEditRollback.ts
```diff
@@ -0,0 +1,59 @@
...
+  const observations = useStore( state => state.observations );
+  const currentObservationIndex = useStore(
+    state => state.currentObservationIndex,
+  );
+  const backupMappings = useStore( state => state.backupMappings );
+  const setBackupMappings = useStore( state => state.setBackupMappings );
+
+  const hasBackedUp = useRef( false );
+
+  useEffect( ( ) => {
```
jtklein: Have you considered doing this not in a useEffect but instead create the snapshot when the user presses the button to get to ObsEdit from Match?

jtklein (changes requested):
Finally managed to get through this. From my testing on a real iPhone (Android doe not navigate to ObsEdit from Match but already in main):

What works well:
- Data rollback: Changing the taxon, location, date, captive/cultivated, obscuration, notes in ObsEdit and pressing back correctly rolls back to the original state on Match.
- Suggestions roundtrip: Match -> ObsEdit -> Suggestions -> choose different taxon -> ObsEdit -> Back correctly rolls back to original Match state.
- Multiple sequential cycles: Doing edit-rollback-edit-rollback multiple times seems correct.

Have tested in Default and Advanced mode for all different obs create start points with Match as first screen after obs creation. Most of the times it works as expected, however, I found two bugs that need to be fixed:
Critical crash: When the Match screen hits the "The AI is not confident" state and I select an iconic taxon "suggestion" I replicatably see a crash in ObsEditHeader on the next screen.
UX bug: Now that we no longer have to Discard the entire observation, when the user enters ObsEdit and makes no changes we should just navigate back to the Match screen without the Discard sheet. Currently in this PR: I think because `unsavedChanges` is always `true` when entering ObsEdit from Match because `updateObservationKeys` unconditionally sets it. This causes a "DISCARD CHANGES" prompt even when the user made no changes.

PS.: Comment about multiple observations ins snapshot: I am not able to get to the Match screen for multiple obs creation or from editing existing observations, or am I missing something here?

src/components/ObsEdit/ObsEditHeader.js
sepeterson: There's an existing ticket for the bug. Added a comment linking to this discussion https://linear.app/inaturalist/issue/MOB-958/regression-in-navigation-behaviour-obsedit

src/sharedHelpers/rollbackPhotos.ts
sepeterson: In theory, this is a type predicate that makes it so typescript treats the return value as `BackupMapping` instead of `BackupMapping | null` if the function `m !== null` returns true

In practice, it is unnecessary because ts can infer that without it so I'll remove it

src/stores/createObservationFlowSlice.js
sepeterson: I considered this but decided to capture everything in the snapshot because otherwise we would need to do some logic to splice the one observation back in. This should be very simple, but would definitely require _some_ logic that considers the other places this data is used.

By copying everything, we can naively return the entire state to a place it was on a previous screen.

src/sharedHelpers/rollbackPhotos.ts
sepeterson: Here's a ticket https://linear.app/inaturalist/issue/MOB-1305/use-photo-uuid-as-filename-for-local-photo-storage

src/sharedHooks/useObsEditRollback.ts
sepeterson: Yes, I don't like that it requires setting up the snapshot and then consuming it in a different file. That said it's probably worth excising a useEffect is we can. I'll give it a shot

src/sharedHooks/useObsEditRollback.ts
sepeterson: Nullified by removing the useEffect but I think you'd want it first to prevent the logic from being executed twice

src/sharedHooks/useNavigateToObsEdit.ts
```diff
@@ -23,6 +26,11 @@ function useNavigateToObsEdit() {
         taxon,
       } );
     }
+    if ( lastScreen === "Match" ) {
```
jtklein, line 29: Thanks for doing this, I think that's much cleaner now.

src/stores/createObservationFlowSlice.js
```diff
@@ -62,7 +76,7 @@ const updateObservationKeysWithState = ( keysAndValues, state ) => {
   const updatedObservations = observations;
   const updatedObservation = {
     ...observationToJSON( currentObservation ),
-    ...keysAndValues,
+    ...keysAndValuesToJSON( keysAndValues ),
```
sepeterson, line 79: Fairly major change here. The crash you observed Johannes was because assigning the iconic taxa from Match was storing a RealmTaxon directly in zustand, which we make make a point of not doing for RealmObservations via `observationToJSON`

For whatever reason, this hadn't caused issues in the past but the rollback logic triggered it. Honestly, I do not fully understand the mechanism of the crash but it seems that not storing Realm objects in zustand state fixes it.

src/sharedHelpers/rollbackPhotos.ts
sepeterson: Sorry for the delay here -- this has been difficult to test. 

I think we should keep this as is. I don't see a strong benefit to erroring the entire operation (blocking the user from getting to ObsEdit) and in my testing, even if we fail to back up one of the photos, we can still recover others.

sepeterson:
@jtklein Okey dokey should be good to go again

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
```diff
@@ -122,13 +125,17 @@ const useOfflineSuggestions = (
   } = options;
 
   const fetchOfflineSuggestions = useCallback( async () => {
+    if ( !photoUri ) return;
```
jtklein, line 128: This is related to the crash fix merged in earlier this week or last?

src/stores/createObservationFlowSlice.js
jtklein: Great you figured out what was crashing. Is it related to any of the "Accessing invalidate or deleted realm object" we were seeing.

jtklein (approved):
Thank you so much for all the changes done here.

I have not seen any more crashes, errors or unexpected behaviours with this.

src/stores/createObservationFlowSlice.js
sepeterson: Maybe -- the exact error was something like "isValid() is not a function" which seemed to be coming from Realm internals in some capacity

src/sharedHooks/useSuggestions/useOfflineSuggestions.ts
sepeterson: I'm actually not totally sure. iirc this was caused by an intermediate loading state where photoUri was temporarily `null`

PR #3474, opened 2026-03-25
update react-native from 82.1 to 83.4 (includes react@19.2)

FLGMwt:
we finally landed on something new just in time for it being relevant to something we're working on! I'm in the trenches of syncObs logic and I think that area could benefit from `useEffectEvent`, even if it's just for me moving things around during investigation. I'm not thinking about introducing it to the codebase without some more discussion / guidance, just doing this rn upgrade for react 19.2 to be able to experiment with it.

----

This feels unfair after @jtklein's monumental work sheparding RN New Architecture & React Compiler and most of the other RN upgrades (shoutout @sepeterson also for the most recent one!). 

This just happens to be the _very first_ RN release [without breaking changes](https://react-native-community.github.io/upgrade-helper/?from=0.82.1&to=0.83.4) so yeah, it feels like cheating.

I followed upgrade path here: https://react-native-community.github.io/upgrade-helper/?from=0.82.1&to=0.83.4 app seemed fine without the non-package.json changes, but I went ahead with them for consistency.

----

why do that takes 2 minutes to do a few times a year when you could take ten minutes to write a script for it that will inevitably be invalidated the next time you have to do it: https://gist.github.com/FLGMwt/902916867d7dd3926345e63b0d2a5bbc


patches/react-native+0.83.4.patch
```diff
@@ -0,0 +1,44 @@
...
+--- a/node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js
++++ b/node_modules/react-native/Libraries/Renderer/implementations/ReactNativeRenderer-dev.js
+@@ -2433,8 +2433,15 @@ __DEV__ &&
+           (isDeeplyEqual = !1));
+       for (var _key in next)
+         if (_key in prev) {
+-          var key = prev[_key];
+-          var nextValue = next[_key];
++          // https://github.com/realm/realm-js/issues/7086
++          // https://github.com/facebook/react/issues/35126
```
FLGMwt, line 34: great find, thanks @jtklein!

This is an interesting consideration from the issue reporter / patch source: https://github.com/realm/realm-js/issues/7086#issuecomment-4065625218

> Longer-term, the safer pattern is to avoid passing live Realm objects directly into UI props / lists / state, and instead map them to small plain JS view models at the Realm -> UI boundary.

^that's certainly an option for fixing the instances of this crash we've been looking at _generally_ for the app (separate from this renderer bug). That is: map all Realm objects to pojos when passing as props.

FLGMwt:
last blocker was restoring flow, now completed, ready to go @jtklein!

learning for me: RN update introduced two new flow errors, I went down a rabbit hole of updating flow to the pinned RN version which currently reports ~200 errors in our codebase that would require FlowFixMes. And then I saw we have a laundry list of RN file excludes in .flowconfig for probably exactly this reason. 🙃

patches/react-native+0.83.4.patch
jtklein: This is interesting. I didn't really look past the analysis of where in the code this comes from and how to patch it.
Definitely, a better solution to fix it closer to our source without patching react-native.
Let me know if you want to apply the fix of "map realm to pojo" in this PR or if we should make a new issue for this.

package.json
```diff
@@ -175,7 +175,7 @@
     "@types/jsrsasign": "^10.5.15",
     "@types/lodash": "^4.17.23",
     "@types/markdown-it": "^14.1.2",
-    "@types/react": "^19.1.1",
```
jtklein, line 178: The upgrade helper uses `^19.2.0` here.

jtklein (approved):
Have tested in Debug and Release mode with basic functions of the app, logging in, and creating several observations. Apart from one change in text style I could not find any regressions.
So, I think this is good to go.

Here is the regression I am seeing on iOS. In a scientific name the second part seems to be bold now. I think we can merge this PR though and fix it in a new PR.

Before:

Image

Image

After:

Image

Image

package.json
FLGMwt: ah, interesting. I hadn't seen any issues, but I'll switch to that for safety.

patches/react-native+0.83.4.patch
FLGMwt: oh, certainly not for this PR, but I think it'd be a good thing to talk about soon. I'm _pretty_ sure it would obviate all of MOB-139 depending on how callback scoping stragglers are treated.

Separately (and more selfishly), it would also be an opportunity to talk about the "is this from the API? is this from Realm?" ambiguity that i think tooooooo many view-layer components are worried about themselves. That is: if we happened to do work to map realms to pojos, we might as well map _everything_ and give it a discriminating `source: 'api' | 'realm'` property AND/OR normalize snake_case properties and use it as a forcing factor to make components not care about `source`

anyway, that's gonna _have_ to be it for me hand-waving today

FLGMwt:
@jtklein holy hell, good eye

I'll spend an hour or so this weekend looking at the font thing (side-eyeing "NO BREAKING CHANGES" RN), but I'm inclined to agree. we've got some headroom before next release and it would be ideal to vet this update in general for a bit and spot-fix the font issue if we need to.

PR #3478, opened 2026-03-27
Mob 1264 announcements design changes

abbeycampbell:
closes [MOB-1264](https://linear.app/inaturalist/issue/MOB-1264/announcements-design-changes)

IMG_1766

### DIFF
--- a/tests/unit/components/MyObservations/Announcements.test.js
+++ b/tests/unit/components/MyObservations/Announcements.test.js
@@ -116,7 +116,7 @@ describe( "Announcements", () => {
 
       await waitFor( () => expect( inaturalistjs.announcements.search ).toHaveBeenCalled() );
 
-      const button = screen.queryByTestId( "announcements-dismiss" );
+      const button = screen.queryByLabelText( "Dismiss announcement" );
       expect( button ).toBeNull();
     } );
   } );
@@ -134,9 +134,8 @@ describe( "Announcements", () => {
 
       await waitFor( () => expect( inaturalistjs.announcements.search ).toHaveBeenCalled() );
 
-      const button = await screen.findByTestId( "announcements-dismiss" );
+      const button = await screen.findByLabelText( "Dismiss announcement" );
       expect( button ).toBeTruthy();
-      expect( button.props.accessibilityLabel ).toBe( "Dismiss announcement" );
     } );
   } );
 


src/components/MyObservations/Announcements.js
```diff
@@ -120,17 +128,29 @@ const Announcements = ( {
...
-        source={{ html: body }}
+        source={{ html: announcementHtml }}
         scrollEnabled={false}
         testID="announcements-webview"
       />
       {dismissible && (
-        <Button
-          className="m-3"
-          text={t( "DISMISS" )}
+        <Pressable
```
jtklein, line 136: We have a custom component for when an icon is supposed to act as a button: INatIconButton.tsx
And we already have a wrapper for when in a transparent circle I think: TransparentCircleButton.tsx

src/components/MyObservations/Announcements.js
```diff
@@ -103,6 +104,13 @@ const Announcements = ( {
     .sort( ( a, b ) => new Date( a.start ) - new Date( b.start ) );
   const topAnnouncement = homeAnnouncements[0];
   const { id, dismissible, body } = topAnnouncement;
+  const announcementHtml = `
```
jtklein, line 107: Is this necessary? This does mean we override the html and body tag if the announcement has one right? I haven't seen any announcement that has them just doing a spot check now but I don't know if we can categorically sure it never will have.

src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -237,6 +237,33 @@ const MyObservationsSimple = ( {
...
+    if ( layout !== "grid" ) {
+      return headerContent;
+    }
+
+    const TARGET_SPACING = 10;
+
+    // our HALF_GUTTER margin value is 7.5, so when we try to cancel it out around announcements we
+    // can get odd rounding behavior that causes 1px margins. Using Math.ceil accounts for this.
+    return (
+      <View
```
jtklein, line 254: I honestly don't remember why announcements are a ListHeaderComponent of the obs list. Probably they are supposed to scroll away with the list, idk I always dismiss them before I scroll I think.
But having to do this acrobatics here smells to me like maybe it should be a sibling to the obs list which only has the padding.

jtklein: I just quickly edited the file to move the Announcements out of the list header into the fragment above the list and it unfortunately does not show the announcement. So, something in the auto-height function of the webview is not working when outside of the list. Maybe I am just missing a flex somewhere.
Anyways, your solution does work but does look a bit to me like a patch for not having the time to place this component properly in the tree in the beginning.

tests/unit/components/MyObservations/Announcements.test.js
```diff
@@ -108,8 +116,8 @@ describe( "Announcements", () => {
 
       await waitFor( () => expect( inaturalistjs.announcements.search ).toHaveBeenCalled() );
 
-      const text = screen.queryByText( "DISMISS" );
-      expect( text ).toBeNull();
+      const button = screen.queryByTestId( "announcements-dismiss" );
```
jtklein: It would be preferred if you use getByLabelText instead of byTestId here.
https://oss.callstack.com/react-native-testing-library/docs/guides/how-to-query#idiomatic-query-predicates

tests/unit/components/MyObservations/Announcements.test.js
```diff
@@ -126,8 +134,9 @@ describe( "Announcements", () => {
 
       await waitFor( () => expect( inaturalistjs.announcements.search ).toHaveBeenCalled() );
 
-      const button = await screen.findByText( "DISMISS" );
+      const button = await screen.findByTestId( "announcements-dismiss" );
```
jtklein: Same as above

src/components/MyObservations/MyObservationsSimple.tsx
jtklein: PS: On Android legacy the announcements are sister to the list on iOS legacy and iNat Next they scroll with the list. So, we are not even consistent with this.

src/components/MyObservations/Announcements.js
abbeycampbell: That's correct. I looked through recent announcements and didn't see any that would cause an issue, and I wasn't sure how else to negate the webview spacing that is extra important now that we want announcements to go all the way to the edge of the screen. I figured I would just call this out to the engagement team, once merged.

src/components/MyObservations/Announcements.js
abbeycampbell: yeah, I opted not to use TransparentCircleButton or INatIconButton because the requirements of the announcement design are pretty different. We want the announcement close button to be quite small (ostensibly so people are less likely to tap it, and so it doesn't obscure the html content which we don't have much control over) but still needs to have an accessible pressable touch target of 44px. I can adapt the api of the TransparentCircleButton to work with the different size requirements and bg color, but the extra pressable area outside the circle, and the fact that this pressable has to play nice with announcements which we don't actually own, made me feel like it's actually preferable in this case for Announcements to determine the shape here.

For me, this brings up the bigger question of when it's appropriate to push back on design. Like, maybe this is a case where I'm adapting to the design when maybe what's actually preferable is to say "no, we have an existing component that's close enough, so it's going to work like this instead"

src/components/MyObservations/MyObservationsSimple.tsx
abbeycampbell: Yes I tried moving it out of the list first but ran into the same issue, and figured even if I fixed that, the change to scrolling behavior was not desired anyway

src/components/MyObservations/Announcements.js
abbeycampbell: I looked into this again and I feel pretty strongly that the amount of adapting we'd need to do in TransparentCircleButton and also likely INatIconButton to accommodate the design here isn't worth it.

jtklein: Thanks for looking into this.

PR #3480, opened 2026-03-27
convert creatObservationFlowSlice to ts

sepeterson:
5 ts errors but I believe they all require functional changes


FLGMwt (approved):
This is awesome and really thorough! I appreciate separating the actions and state since this one's pretty chonky.

Did you have a reason for keeping it draft?

PR #3485, opened 2026-03-31
MOB-90 add some error handling around photo gallery import

sepeterson:
- Add some try/catch blocks that deliver the user back to the app
- Add logging to catch future failures

### DIFF
--- a/src/components/PhotoImporter/PhotoLibrary.tsx
+++ b/src/components/PhotoImporter/PhotoLibrary.tsx
@@ -155,17 +155,27 @@ const PhotoLibrary = ( ) => {
     // According to the native code of the image picker library, it never rejects the promise,
     // just returns a response object with errorCode
     // https://github.com/react-native-image-picker/react-native-image-picker?tab=readme-ov-file#Asset-Object
-    const response = await launchImageLibrary( {
-      selectionLimit: fromAICamera
-        ? FROM_AICAMERA_MAX_PHOTOS_ALLOWED
-        : MAX_PHOTOS_ALLOWED,
-      mediaType: "photo",
-      includeBase64: false,
-      // forceOldAndroidPhotoPicker is necessary because the "new" picker strips key EXIF data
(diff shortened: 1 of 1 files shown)


src/components/PhotoImporter/PhotoLibrary.tsx
```diff
@@ -152,20 +155,35 @@ const PhotoLibrary = ( ) => {
...
-      includeBase64: false,
-      // forceOldAndroidPhotoPicker is necessary because the "new" picker strips key EXIF data
-      forceOldAndroidPhotoPicker: true,
-      chooserTitle: t( "Import-Photos-From" ),
-      presentationStyle: "overFullScreen",
-    } );
+    let response;
+    try {
+      // Adding a try/catch anyway. The error state described in https://linear.app/inaturalist/issue/MOB-90/importing-photo-while-offline-in-android-gets-stuck-on-photogallery
+      // is now fixed but matches the screen state shown if I manually throw an error here
```
FLGMwt: I'm not sure I understand the role of this comment (and maybe the behavior it explains?). Can you describe you intention here and if we maybe need to correct the comment above this one?

FLGMwt (commented):
omg, thank you for calling out semantic diff, that was _super_ clear on the try/catch lift.

no expectations, but did you learn anything about `exitObservationsFlow` in this? seems a lil funny that it's a hook, but nevermind if it didn't play a role in this PR

jtklein (approved):
LGTM

I think it is great that we think about recovery paths and error fallback behaviour.

src/components/PhotoImporter/PhotoLibrary.tsx
sepeterson: I consolidated the two as I think they're both important for full context. If you'd like more explanation sometime, lmk!

FLGMwt: oh oh oh, that makes sense now, thank you!

PR #3486, opened 2026-03-31
Mob 1056 nav bar UI changes

abbeycampbell:
[MOB-1056](https://linear.app/inaturalist/issue/MOB-1056/nav-bar-ui-changes)

`INatIconButton` has 44px touch target guards that don't play nice with the inner 36px icon frame we want to add here, so here I'm using `INatIcon` instead and `NavButton` is our wrapper that maintains the desired touch target around the tab icons and their labels. 

Also switched to using `thumbUri` for the me tab; it's a bit blurrier but this is consistent with our other 32px user icon.

IMG_1820
IMG_1819

### DIFF
--- a/src/navigation/BottomTabNavigator/NavButton.tsx
+++ b/src/navigation/BottomTabNavigator/NavButton.tsx
@@ -1,25 +1,24 @@
-// @flow
 import { Body4, INatIcon, UserIcon } from "components/SharedComponents";
 import { Pressable, View } from "components/styledComponents";
 import NotificationsIconContainer from "navigation/BottomTabNavigator/NotificationsIconContainer";
+import type { PropsWithChildren } from "react";
 import * as React from "react";
 import colors from "styles/tailwindColors";
 
-type IconFrameProps = {
-  children: React.Node,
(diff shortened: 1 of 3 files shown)


src/navigation/BottomTabNavigator/NavButton.tsx
```diff
@@ -28,52 +39,41 @@ const NavButton = ( {
...
+          size={size}
+        />
+      </IconFrame>
     );
   }
 
   return (
     <Pressable
-      className="flex-column items-center justify-end"
+      className="min-h-[44px] min-w-[44px] items-center justify-end"
```
jtklein, line 75: Why are the min values necessary? You could also make the IconFrame 44x44?

abbeycampbell: The designs specify a 36x36 box around each icon, which is what the IconFrame is about. This is to maintain at least a 44x44 touch target on the whole Pressable, which includes the label text for each tab--the NavButtons actually end up being taller than 44px, but we just want to keep enforcing it for accessibility.

PR #3490, opened 2026-04-01
Add rozenite controls plugin with ff toggle

FLGMwt:
another convenience opportunity I uncovered doing the "airplane mode" investigation: I couldn't _actually_ use airplane mode because I still needed to be connected to the debugger _and_ see our attempted http traffic. So I had some janky code in a wrapper for `const { isConnected } = useNetInfo( );` but I couldn't actively config it helpfully.

I remembered that rozenite just added `controls` which is _perfect_ for this sort of thing.

I'm not certain if I'm going to want to keep the `useNetInfo` wrapper (that is, introduce it to the codebase), so I'm using our current ExploreV2 Feature Flag as a test case for this plugin.

https://github.com/user-attachments/assets/46f34c9e-64e4-4ca7-ae96-5c53704e90a4

also: moved rozenite plugin calls to a new file / hook because it was starting to bloat our app root.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,10 +7,7 @@ import "react-native-get-random-values";
 import "react-native-url-polyfill/auto";
 
 import { BottomSheetModalProvider } from "@gorhom/bottom-sheet";
-import {
-  Alert, AppRegistry,
-  LogBox,
-} from "react-native";
+import { Alert, AppRegistry } from "react-native";
 import { getCurrentRoute } from "navigation/navigationUtils";
 import { zustandStorage } from "stores/useStore";
 import zustandMMKVBackingStorage from "stores/zustandMMKVBackingStorage";
@@ -38,8 +35,6 @@ import useRozenite from "sharedHooks/useRozenite";
 import { name as appName } from "./app.json";
 import { log } from "./react-native-logs.config";
 import { getUserAgent } from "./src/api/userAgent";
-
-LogBox.ignoreAllLogs();
 
 const logger = log.extend( "index.js" );
 


package.json
```diff
@@ -165,6 +165,7 @@
     "@react-native/eslint-config": "0.82.1",
     "@react-native/metro-config": "0.82.1",
     "@react-native/typescript-config": "0.82.1",
+    "@rozenite/controls-plugin": "^1.6.0",
```
jtklein: Other plugins are in dependencies.

index.js
```diff
@@ -33,15 +34,13 @@ import { SafeAreaProvider } from "react-native-safe-area-context";
...
-import { useNetworkActivityDevTools } from "@rozenite/network-activity-plugin";
-import { useRequireProfilerDevTools } from "@rozenite/require-profiler-plugin";
-import { useTanStackQueryDevTools } from "@rozenite/tanstack-query-plugin";
-
+import useRozenite from "sharedHooks/useRozenite";
 import { name as appName } from "./app.json";
 import { log } from "./react-native-logs.config";
 import { getUserAgent } from "./src/api/userAgent";
 
+LogBox.ignoreAllLogs();
```
jtklein: We shouldn't do that.

src/sharedHooks/useRozenite.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const sections = useMemo(
+    () => [
+      createSection( {
+        id: "toggles",
+        title: "Toggles",
+        items: [
+          {
+            id: "explore-v2",
+            type: "button",
```
jtklein: Why not use toggle? https://www.rozenite.dev/docs/official-plugins/controls#toggle

jtklein (changes requested):
I think this is a good example of how to use this plugin. Only minor changes.

index.js
FLGMwt: ah, shoot, didn't mean to leave that in.

package.json
FLGMwt: moved, thx

src/sharedHooks/useRozenite.ts
FLGMwt: hey, that's a _great_ point, updated

PR #3492, opened 2026-04-01
MOB-1009 get android detox running locally again

sepeterson:


android/app/build.gradle
```diff
@@ -152,7 +152,11 @@ android {
 
 dependencies {
     // Detox Android setup
-    androidTestImplementation('com.wix:detox:+')
+    androidTestImplementation('com.wix:detox:+') {
+        // accessibility-test-framework pulls in protobuf-lite:3.0.1 which conflicts
+        // with protobuf-javalite:3.25.5 used by Firebase, crashing the app on launch
+        exclude group: 'com.google.protobuf', module: 'protobuf-lite'
+    }
```
FLGMwt, line 159: 💯 omg this was probably the thing I couldn't track down last year, great find!

android/build.gradle
```diff
@@ -36,3 +36,24 @@ allprojects {
...
+        sub.afterEvaluate {
+            if (it.hasProperty('android')) {
+                if (it.android.defaultConfig.minSdkVersion != null
+                    && it.android.defaultConfig.minSdkVersion.apiLevel < rootProject.ext.minSdkVersion) {
+                    it.android.defaultConfig.minSdkVersion rootProject.ext.minSdkVersion
+                }
+            }
+        }
+    }
+}
```
FLGMwt, line 59: nice workaround!

I left a comment in MOB-298 linking to this in case we end up replacing rn-mail and not needing this. 

noting #3016 for android / email relevance

patches/react-native-worklets-core+1.6.3.patch
```diff
@@ -0,0 +1,13 @@
...
+index b6af1c3..67c3bd3 100644
+--- a/node_modules/react-native-worklets-core/android/build.gradle
++++ b/node_modules/react-native-worklets-core/android/build.gradle
+@@ -152,6 +152,8 @@ android {
+       "**/libturbomodulejsijni.so",
+       "**/libreact_nativemodule_core.so",
+       "**/libjscexecutor.so",
++      "**/libhermesvm.so",
++      "**/libhermestooling.so",
+     ]
```
FLGMwt, line 11: copying from internal ticket comment for public posterity:

In that last attempt to get this working, I found other people only experiencing this w/ Detox in New Architecture. It seems like New Arch native modules have a laundry list of jni libs (.co) that they necessarily produce but aren't expected to be exposed to client apps, whats' builds seem to expect to rely on the maven r-n android plugin dependency for the jnis (see .gradle path below). This creates the following native lib load error:

```
Execution failed for task ':react-native-worklets-core:mergeDebugAndroidTestNativeLibs'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.MergeNativeLibsTask$MergeNativeLibsTaskWorkAction
   > 2 files found with path 'lib/arm64-v8a/libhermestooling.so' from inputs:
      - /Users/ryan/git/iNatRN/node_modules/react-native-worklets-core/android/build/intermediates/library_jni/debug/copyDebugJniLibsProjectOnly/jni/arm64-v8a/libhermestooling.so
      - /Users/ryan/.gradle/caches/9.0.0/transforms/81e986bf84b62dfee3f4f82452117344/transformed/react-android-0.82.1-debug/jni/arm64-v8a/libhermestooling.so
```

Fix: this isn't the first instance of this issue, just the latest. It seems like lib authors (and consumers w/ patch-packages like us) have been having to play whackamole with the changing output jnis form rn internals. Libs have corrected it with this [gradle exclude config](https://github.com/margelo/react-native-worklets-core/blob/main/android/build.gradle#L137-L156)

Probably due to the less common use of somethingTest gradle tasks, Detox in particular complains about more of these than a typical user or lib author would hit.

Anyhow, this seems to be resolved by just adding some extra excludes to that list, "**/libhermesvm.so" & "**/libhermestooling.so". The second one was the next whackamole mole that surfaced when the first was fixed. Following that, the build continued and succeeded. [patch-package here](https://github.com/inaturalist/iNaturalistReactNative/compare/detox-march-spike?expand=1#diff-2c9a3bcda3a5035096be3d4c34aea94cbd6db2939cecfe49a0ad787f092eebbb).

PSA: for this flavor of patch-package, running patch-package to generate the patch after a build with the modification gives you a massive patch that includes native build artifacts. Steps I took to get focused patch:

1. Make actual change in node_modules/some-package, in this case adding to the list
2. Run build to validate the change
3. rm -rf node_modules/some-package
4. npm i to restore fresh download of some-package
5. Reapply change from Step 1
6. before building again,patch-package some-package

If we are able to scramble up more instances of this, I'll post them in comments so that people on these libs can get the oss repo issue links surfaced. I think the following at least demonstrate this (and might be fixed upstream):
- https://github.com/mrousavy/react-native-mmkv/issues/510 
- https://github.com/software-mansion/react-native-gesture-handler/issues/3311
- https://github.com/mrousavy/react-native-vision-camera/issues/3540

src/sharedHelpers/installData.e2e-mock
```diff
@@ -0,0 +1,23 @@
+import { MMKV, useMMKVBoolean } from "react-native-mmkv";
+import * as uuid from "uuid";
+
+const MMKV_ID = "install-data";
+const INSTALL_ID = "installID";
+const ONBOARDING_SHOWN = "onboardingShown";
+export const IS_FRESH_INSTALL = "isFreshInstall";
+
+export const store = new MMKV( { id: MMKV_ID } );
```
FLGMwt: can we make `store` a stub and drop the mmkv import?

.detoxrc.js
```diff
@@ -1,6 +1,20 @@
+const { execSync } = require( "child_process" );
 const fs = require( "fs" );
 const { version } = require( "./package.json" );
 
+function getAvailableAVD( ) {
+  try {
+    const avds = execSync( "emulator -list-avds", { encoding: "utf8", timeout: 5000 } )
+      .trim()
+      .split( "\n" )
+      .filter( Boolean );
```
FLGMwt, line 10: after my own heart <3

.detoxrc.js
```diff
@@ -1,6 +1,20 @@
+const { execSync } = require( "child_process" );
 const fs = require( "fs" );
 const { version } = require( "./package.json" );
 
+function getAvailableAVD( ) {
```
FLGMwt, line 5: okay this is killer, awesome idea!

package.json
```diff
@@ -28,7 +28,7 @@
     "translate": "babel-node src/i18n/i18ncli.js build",
     "prepare-fastlane-metadata": "babel-node src/i18n/i18ncli.js prepare-fastlane-metadata",
     "e2e:android": "npm run e2e:build:android && npm run e2e:test:android",
-    "e2e:build:android": "MOCK_MODE=e2e npx detox build --configuration android.release",
+    "e2e:build:android": "npm run add-example-model -- -f=main && MOCK_MODE=e2e npx detox build --configuration android.release",
```
FLGMwt: I'd prefer we remove this change. all of our other scripts have the presumption of "setup complete" as described in the readme so I don't have a good feeling as to why we'd make an exception here.

sepeterson: Ahhh I see this step is already in the CI setup. Great call-out!

src/sharedHelpers/installData.e2e-mock
sepeterson: Yaaa totally

PR #3494, opened 2026-04-02
Defer some StartupService tasks to idle callbacks

jtklein:
Closes MOB-1235

Moves non-critical startup tasks to a new StartupService. Before, all initialization work was running in a single sequential await chain all competing with the initial render for JS thread time.

This PR splits startup work into two components:

StartupService (slimmed down), keeps only tasks that must run immediately: fresh-install check/sign-out, version logging, several config steps.
DeferredStartupService (new), schedules non-critical work via `requestIdleCallback` so it runs during idle periods after the initial render completes. Each task gets its own idle callback for per-task isolation in separate try catch blocks, logging purposes, and to potentially interleave user interactions between tasks.

Those tasks were mostly just moved over with some refactors:
checkForPreviousCrash, logSentinelFiles, logStorageMetrics, clearCache split up into 4 individual cache clears (rotated originals, gallery photos, CV photos, synced media)

Unclear to me where addARCameraFiles fits. I'd suggest we need user data to decide this.

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -13,7 +13,7 @@ import {
 import { isDebugMode } from "sharedHooks/useDebugMode";
 
 import AppStateListener from "./AppStateListener";
-import DeferredStartupService from "./DeferredStartupService";
+import useDeferredStartup from "./hooks/useDeferredStartup";
 import useLinking from "./hooks/useLinking";
 import NetworkService from "./NetworkService";
 import StartupService from "./StartupService";
@@ -73,13 +73,13 @@ const App = ( { children }: Props ): Node => {
 
   useLinking( currentUser );
   useShare( onShare );
+  useDeferredStartup( );
 
   // this children prop is here for the sake of testing with jest
   // normally we would never do this in code
   return (
     <>
       <StartupService />
-      <DeferredStartupService />
       <NetworkService />
       <AppStateListener />
       {children || <RootStackNavigator />}
--- a/src/components/hooks/useDeferredStartup.ts
+++ b/src/components/hooks/useDeferredStartup.ts
@@ -1,17 +1,14 @@
 /**
- * DeferredStartupService
+ * useDeferredStartup
  *
- * A renderless component that schedules non-critical startup work to run
- * during idle periods of the JS event loop, after the initial render and
- * navigation transitions have completed. Hopefully, this improves Time-To-Interactive
- * (TTI) by keeping the JS thread free for layout, painting, and user input
- * during app launch.
+ * Schedules non-critical startup work to run during idle periods of the JS
+ * event loop, after the initial render and navigation transitions have
+ * completed. Hopefully, this improves Time-To-Interactive (TTI) by keeping
+ * the JS thread free for layout, painting, and user input during app launch.
  *
  * Uses requestIdleCallback to schedule each task in its own callback so the
  * runtime can interleave user interactions between them.
- *
  */
-
 import { RealmContext } from "providers/contexts";
 import { useEffect } from "react";
 import {
@@ -76,7 +73,7 @@ const checkForPreviousCrash = async () => {
   }
 };
 
-const DeferredStartupService = ( ) => {
+const useDeferredStartup = ( ) => {
   const realm = useRealm( );
 
   useEffect( ( ) => {
@@ -106,8 +103,6 @@ const DeferredStartupService = ( ) => {
       cancelIdleCallback( id7 );
     };
   }, [realm] );
-
-  return null;
 };
 
-export default DeferredStartupService;
+export default useDeferredStartup;


__mocks__/react-native-fs.ts
```diff
@@ -4,7 +4,7 @@ module.exports = {
   appendFile: jest.fn( ),
   CachesDirectoryPath: "caches/directory/path",
   DocumentDirectoryPath: "document/directory/path",
-  exists: jest.fn( async ( ) => true ),
+  exists: jest.fn( async name => !name.includes( "sentinelFiles" ) ),
```
jtklein, line 7: We were treating every "app start" during testing as having had some sentinel files present from a previous session, i.e. having had an error during a previous camera usage session. Changing to not having sentinelFiles present seems a more sensible default for testing to me.

src/components/hooks/useDeferredStartup.ts
```diff
@@ -0,0 +1,113 @@
...
+      console.log( `${name} completed in ${duration}ms` );
+    }
+  } catch ( error ) {
+    logger.error( `${name} failed after ${Date.now( ) - start}ms`, error );
+  }
+}, timeout
+  ? { timeout }
+  : undefined );
+
+const checkForPreviousCrash = async () => {
```
jtklein, line 63: This function is copied over without changes.

src/components/hooks/useDeferredStartup.ts
```diff
@@ -0,0 +1,113 @@
...
+    const id1 = deferTask( "checkForPreviousCrash", checkForPreviousCrash, 30000 );
+    const id2 = deferTask( "logSentinelFiles", logSentinelFiles, 30000 );
+    const id3 = deferTask( "logStorageMetrics", async () => {
+      const metrics = await getStorageMetrics( realm?.path );
+      logger.infoWithExtra( "storage_metrics", metrics );
+    }, 30000 );
+
+    // Each cache directory gets its own idle callback so that we can still have
+    // user interactions between potentially slow filesystem operations.
+    const id4 = deferTask( "clearRotatedOriginalPhotos", clearRotatedOriginalPhotosDirectory );
```
jtklein, line 91: I decided to not add a timeout to these four because I think it is less of a problem if they don't run on every app start. But I am also assuming that at one point the user opens the app with little enough load and these do run at least from time to time. We don't really have any intel built in now if they do run or not because we only log when the task is too long.

src/components/StartupService.tsx
```diff
@@ -5,27 +5,14 @@ import { getUserAgent } from "api/userAgent";
...
-} from "sharedHooks";
+import { usePerformance } from "sharedHooks";
 import { isDebugMode } from "sharedHooks/useDebugMode";
-import { zustandStorage } from "stores/useStore";
-
-// Ignore warnings about 3rd parties that haven't implemented the new
-// NativeEventEmitter interface methods yet. As of 20230517, this is coming
-// from react-native-share-menu.
-// https://stackoverflow.com/questions/69538962
-LogBox.ignoreLogs( ["new NativeEventEmitter"] );
```
jtklein, line 28: Shouldn't be part of this PR but I was in this file already deleting things. Every other change in this file is moved to the new component.

src/sharedHelpers/sentinelFiles.ts
```diff
@@ -66,9 +66,9 @@ const deleteSentinelFile = async ( sentinelFileName: string ): Promise<void> =>
   }
 };
 
-const findAndLogSentinelFiles = async ( ) => {
```
jtklein, line 69: We are nowhere consuming the return.

src/sharedHelpers/clearCaches.ts
```diff
@@ -82,19 +79,9 @@ const clearSyncedMediaForUpload = async realm => {
   );
 };
 
-const clearCaches = async realm => {
```
jtklein, line 85: Not needed anymore.

src/components/hooks/useDeferredStartup.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+};
+
+const DeferredStartupService = ( ) => {
+  const realm = useRealm( );
+
+  useEffect( ( ) => {
+    // Diagnostic tasks that we need to finish even on a busy thread
+    // should have a timeout to ensure they run eventually.
+    const id1 = deferTask( "checkForPreviousCrash", checkForPreviousCrash, 30000 );
```
FLGMwt, line 82: nit, nonblocking:

the `id1` `id2` thing seems a little wonky. and maybe fragile if we add a new one and forget to cancel it (if we somehow didn't hit `unused-vars`). can we switch this to being a lil more programmatic:

```js
const idleCallbackIds = [
  deferTask( "checkForPreviousCrash", checkForPreviousCrash, 30000 ),
  deferTask( "logSentinelFiles", logSentinelFiles, 30000 ),
  // ...
];
// ...
for (const idleCallbackId of idleCallbackIds) {
  cancelIdleCallback( idleCallbackId );
}

File: src/components/hooks/useDeferredStartup.ts [hunk 8]
```diff
@@ -0,0 +1,113 @@
...
+  try {
+    await run( );
+    const duration = Date.now( ) - start;
+    // These tasks should not be UI blocking, but maybe good to get intel
+    // from users devices if some of these tasks are very long running.
+    if ( duration > 1000 ) {
+      logger.info( `${name} completed in ${duration}ms` );
+    } else {
+      console.log( `${name} completed in ${duration}ms` );
+    }
```
FLGMwt, line 55: I really like this idea, I'll keep it in mind for other timing concerns.

src/components/hooks/useDeferredStartup.ts
FLGMwt: yeah, that makes sense. I really appreciate that `deferTask`'d things have the start/complete/error logs so that we can start getting insight for this.

src/components/App.js
```diff
@@ -78,6 +79,7 @@ const App = ( { children }: Props ): Node => {
   return (
     <>
       <StartupService />
+      <DeferredStartupService />
```
FLGMwt: I know we're mirroring prior art, but while we're doing active dev here, is there a reason why we want this to be a component? if the component only exists to host a `useEffect` call, I think I'd prefer that useEffect to be wrapped in a `useDeferredStartupTasksOnMount()` over `<DeferredStartupService / >`

src/sharedHelpers/sentinelFiles.ts
```diff
@@ -66,9 +66,9 @@ const deleteSentinelFile = async ( sentinelFileName: string ): Promise<void> =>
...
+const logSentinelFiles = async ( ) => {
   const directoryExists = await RNFS.exists( sentinelFilePath );
-  if ( !directoryExists ) { return null; }
+  if ( !directoryExists ) { return; }
   const files = await RNFS.readDir( sentinelFilePath );
 
   files.forEach( async file => {
```
abbeycampbell, line 74: existing issue, but it could mean we miss handling some errors and probably won't log correct duration for `logSentinelFiles`: `forEach` can't handle async callbacks. We could use a for/of loop instead.

abbeycampbell: clarification: this was here before so I don't think fixing it should block this PR. I think we should look into fixing it at some point, but I'm not requesting changes.

Not opinionated about the DeferredStartupService as a component vs hook issue.

src/components/App.js
jtklein: Thanks for the suggestion. Have refactored to hook.

PR #3510, opened 2026-04-07
useNetInfo wrapper for debugging network status changes

FLGMwt:
This comes out of MOB-1214, an investigation into the long-standing long-install performance issue. Some users experience "app gets laggy after having it installed a while", similar to a memory leak issue but it persists across launches. Some users reported that toggling airplane mode on and off "heals" it. We don't have a local repro, but it's consistent enough that we need to address it.

I needed to be able emulate toggling "airplane mode", as far as the app handles it, while trying to repro without code changes or affecting the packager / dev tools / http debugging. I'm also not able  just "turn off" our `/ping` url that netinfo is [configured to use as a `reachabilityUrl`](https://github.com/inaturalist/iNaturalistReactNative/blob/632e7d7b2969fea5b0ce615b03044f18576d81f1/src/components/StartupService.tsx#L35) for this because netinfo reacts differently to device network changes (like airplane mode).

I initially thought I would not PR this test code but there's likely other situations where this could be convenient for development. I also realized that we'll soon-ish need to handle responding to planned server downtime messaging and app behavior which would need an "offline mode" status wrapper anyhow. The feature flag code got a little cleaner in the meantime.

Config implementation: this is stored, read, and set exactly like a Feature Flag except that I thought it was important to namespace it differently since this kind of thing _oughtn't_  follow the semantics of a Feature Flag. That is: it's never going to be "released". Distinction captured in the [flag definitions here](https://github.com/inaturalist/iNaturalistReactNative/compare/netinfo-offline-wrapper?expand=1#diff-1b353904c7b884bc645ca34232a0e3662f76a1064d9f9a70d2aa7dbd3a86393aR1-R21):

```
// FeatureFlag: used to toggle a feature on or off during development,
// allowing us to ship in progress code and turn it on for testing
// it in a certain build of the app.

// Associated Feature is expected to be eventually be turned on by default,
// and then eventually remove the FeatureFlag and "any old" code.

VS

// DevOnlyFlag: similar to FeatureFlag allowing us to toggle app
// behavior in development, but not intended to be exposed to end users.

// Associated flag and code is expected to remain unless it no longer makes sense.
```

Open to feedback and I'm fine if we don't want to pull this in, I'll still have the branch around for continued testing.


FLGMwt:
closing now as no immediate need for generalizing this tooling. I'll hold onto the branch for if/when this comes up again and when we tackle remote config tooling.

PR #3512, opened 2026-04-07
MOB-1196 enable unauthenticated announcements

FLGMwt:
EDIT, correction, no longer splitting the implementation

we now support querying logged out announcements

for logged out announcements that art tagged dismissible, we allow _local_ dismissal. To enable this, we have a local persisted store for dismissed announcements IDs.


src/components/MyObservations/Announcements.tsx
```diff
@@ -35,52 +36,104 @@ const Webshell = makeWebshell(
...
   const { autoheightWebshellProps } = useAutoheight( {
     webshellProps,
   } );
   // eslint-disable-next-line react/jsx-props-no-spreading
   return <Webshell {...autoheightWebshellProps} />;
 };
 
-type Props = {
-  isConnected: boolean,
+interface Announcement {
```
jtklein: This should be ApiAnnouncement and live in api/types

abbeycampbell:
Testing this branch out! I'm unable to see this announcement in a logged out state: https://staging1.inaturalist.org/announcements/2387

Curious if I'm doing something wrong? If you get a chance, could you outline your testing process for this one?

src/stores/useStore.js
```diff
@@ -40,6 +41,7 @@ const useStore = create( persist(
       createRootExploreSlice( ...args ),
       createSyncObservationsSlice( ...args ),
       createUploadObservationsSlice( ...args ),
+      createLoggedOutAnnouncementsSlice( ...args ),
```
jtklein: Long time ago we made the decision to keep all persisted zustand information in the layout slice in the `layout` key specifically. So that the new information is already added by the partialize => layout below.
For this PR I'd say we stick to that convention and merge this new slice into createLayoutSlice.

src/components/MyObservations/Announcements.tsx
```diff
@@ -35,52 +36,104 @@ const Webshell = makeWebshell(
...
+    // local _is_ on the model but we need to figure this out
     locale: currentUser?.locale || "en",
     per_page: 20,
   };
+
   // TODO: if there are more than 20 announcements, should we paginate and get more?
+  const isAuthenticated = !!currentUser;
   const {
     data: announcements,
-    refetch: refetchAnnouncements,
```
jtklein, line 66: Why do we not need this anymore?

jtklein (changes requested):
I was not able to test this yet, because an announcement I created on staging is not showing up for me.
But I did leave a few code level requested changes.

src/components/MyObservations/Announcements.tsx
FLGMwt: reverted this change for consistency. It shouldn't be necessary if we're invalidating queries as well (which triggers a reset), but it's removed from the changeset to defer discussion if it comes up again.

src/stores/useStore.js
FLGMwt: To clarify: anything we persist in zustand should be namespaced in the `layout` slice, regardless of whether it's related to layout or not?

Do you have more context to the decision? I would think it favorable to have any (exceptional) persisted slices explicitly named in the partialize. The separate slice also avoids the extra layer of state merging.

I switched to this convention in https://github.com/inaturalist/iNaturalistReactNative/pull/3512/changes/3bcefe13288af6223b3c507b2c77d31143b0aa95 but lemme know if we're open to changing this in the future

FLGMwt:
^we may have to add extra handling for e2e tests & announcements since we can't do the "cleaning" in a script for the locally "dismissed" announcements. we _know_ we have some announcements we might not normally have live for testing this, but I'll keep an eye on builds

src/components/MyObservations/Announcements.tsx
jtklein: I was actually looking a bit into that after I posted the question here. We should definitely remove the refetch because I think it leads to duplicate fetches (invalidate and refetch). There are also other places in the app we do this for other queries. But we can do it in another PR as well.

FLGMwt: yeah, I'll make a ticket to audit. I don't want to apply it blanketly because while they do happen to be synonymous here, invalidateQueries could have broader effect than a local refecth

src/components/MyObservations/Announcements.tsx
```diff
@@ -48,37 +49,32 @@ const AutoheightWebView
...
 
   const queryFnWithoutDismissedIds = async () => {
     const announcements = await queryFn( { api_token: null } );
     return announcements
       .filter( ( { id } ) => !dismissedAnnouncementIds.includes( id ) );
   };
   const unauthenticatedQuery = useQuery(
-    queryKey,
+    // include dismissedIds because they inform the query response
+    [...queryKey, dismissedAnnouncementIds],
```
jtklein, line 77: How do they inform the query response? Isn't invalidation an information enough? Changing the query key does not result in different data, like for example the dismissed announcements being excluded from fetching.

src/components/MyObservations/Announcements.tsx
jtklein: I meant in other places in the app we do specifically also a invalidate plus refetch right after.

src/stores/useStore.js
jtklein: Probably open but we'd need to find documentation as to the original decision and probably will not have any near term priority.

jtklein: "anything we persist in zustand should be namespaced in the layout slice"
Yes.

", regardless of whether it's related to layout or not?"
There shouldn't be any non-layout things there. "Layout" not only in the sense of how to layout the UI but also what to layout, so certain announcements that we should not show do fit into the layout paradigm there similar as to certain PivotCards that we should not show. Otherwise, data is in realm, other persisted non-layout app state is in installData in mmkv.

src/api/types.d.ts
```diff
@@ -37,6 +37,13 @@ export interface ApiObservationsUpdatesParams extends ApiParams {
 
 // Model types, need to be ordered by reference
 
+interface ApiAnnouncement {
+  id: string;
```
jtklein, line 41: Not a requested change but other interfaces here have optional keys.

src/components/MyObservations/Announcements.tsx
```diff
@@ -1,5 +1,3 @@
-// @flow
```
jtklein, line 1: Thanks for doing the TS migration here.

src/components/MyObservations/Announcements.tsx
```diff
@@ -35,52 +37,101 @@ const Webshell = makeWebshell(
...
+  queryFn: QueryFunction<ApiAnnouncement[]>,
+  isAuthenticated: boolean,
+): AnnouncementQueryResponse => {
+  const dismissedAnnouncementIds
+    = useStore( state => state.layout.dismissedAnnouncementIds );
+
+  const queryFnWithoutDismissedIds = async () => {
+    const announcements = await queryFn( { api_token: null } );
+    return announcements
+      .filter( ( { id } ) => !dismissedAnnouncementIds.includes( id ) );
```
jtklein, line 73: Using dismissedAnnouncementIds in the key and function of the useQuery here does mean we add a new query every time we dismiss an announcement, right? When dismissing multiple announcements while logged out I can see queries accumulate. Have not checked if they actually make all fresh requests when onSucess we invalidate them all at once.

src/api/types.d.ts
FLGMwt: I'd prefer not to type this incorrectly for the sake of convention. We only consume existing announcements and the optional would necessitate unrealized (and misleading) codepaths since we are not POSTing id-less announcements. 

Is api/types meant to be exactly reflective of the API's models? If so, I can make it optional, but I should add `endDate` and `clients`. 

happy to do that in this PR, just lemme know

jtklein (approved):
This works great as far as I have tested:
While logged out, I was able to Dismiss multiple announcements. Was able to see an announcement that can not be dismissed without the x button. When I logged in I was able to see different announcements based on targeting but also based on date as it should be.

Thanks for making all the changes requested already.

src/components/MyObservations/Announcements.tsx
FLGMwt: Glad we chatted about this this morning : ) I understand now that you are referring to the _query instances_ themselves (in addition to the API calls). This makes a ton of sense in relation to #3537, it's a week of excellent coincidences!

We settled this on the call, but just noting for posterity:

Conversation outcome: keeping as is

I decided to include the filtering _in_ the queryFn to keep the announcement queries' code as close as possible which meant the qK needed the ids as an effect-like "dependency array". The alternative would be to filter outside of the query. The tradeoff was doing slightly more work than necessary (extra query instances / api calls) for more readable code.

This is not as concerning as the ballooning polling queries in #3537 because here, for better or worse, we're doing "legitimately different" queries where inputs inform results.

src/stores/useStore.js
FLGMwt: convo moved to internal chat, will update with outcome

PR #3513, opened 2026-04-08
Mob 1177 ios jwt refresh returning 503 during downtime 2

abbeycampbell:
[MOB-1177](https://linear.app/inaturalist/issue/MOB-1177/ios-jwt-refresh-returning-503-during-downtime)

### DIFF
--- a/src/sharedHelpers/logging.js
+++ b/src/sharedHelpers/logging.js
@@ -68,7 +68,7 @@ function reactQueryRetry( failureCount, error, options = {} ) {
   // load. The logger.warn above still fires so these errors remain visible in
   // Grafana, but we skip handleError (which would fire the log transport again)
   // since /v2/log itself returns 503 during downtime.
-  if ( error.status >= 500 || error.response?.status >= 500 ) {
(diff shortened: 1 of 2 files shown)


src/sharedHelpers/logging.js
```diff
@@ -64,6 +64,13 @@ function reactQueryRetry( failureCount, error, options = {} ) {
     `reactQueryRetry, error: ${error.message}, failureCount: ${failureCount}, options:`,
     inspect( options?.queryKey ),
   );
+  // We don't want to retry server-side errors — retrying a downed server amplifies
+  // load. The logger.warn above still fires so these errors remain visible in
+  // Grafana, but we skip handleError (which would fire the log transport again)
+  // since /v2/log itself returns 503 during downtime.
+  if ( error.status >= 500 || error.response?.status >= 500 ) {
+    return false;
```
abbeycampbell: I think this is fine, but we could also only skip retries specifically for 503 only

jtklein (approved):
LGTM
I think this is fine but I was not able to set up a proxy to test if it actually works yet, so maybe @FLGMwt can help with testing this here.

PR #3531, opened 2026-04-13
Replace deprecated react-native-gesture-handler Components

jtklein:
Closes MOB-1262

Following the migration path here: https://docs.swmansion.com/react-native-gesture-handler/docs/guides/upgrading-to-2


abbeycampbell (approved):
lgtm

PR #3532, opened 2026-04-13
UserProfile TS migration

jtklein:
Originally, I wanted to A/B test with using fields param for queries here unil I figured out that we do actually use fields already for both queries.

In general, I started checking TS files from the tips until I was at UserProfile and did a migration of this screen and all it touches (route params to and from other screens).

### DIFF
--- a/src/api/relationships.js
+++ b/src/api/relationships.js
@@ -10,8 +10,8 @@ const PARAMS = {
 
 const fetchRelationships = async ( params: Object = {}, opts: Object = {} ): Promise<?Object> => {
   try {
-    const response = await inatjs.relationships.search( { ...PARAMS, ...params }, opts );
-    return response;
+    const { results } = await inatjs.relationships.search( { ...PARAMS, ...params }, opts );
+    return results;
   } catch ( e ) {
     return handleError( e, { context: { functionName: "fetchRelationships", opts } } );
   }
@@ -19,8 +19,7 @@ const fetchRelationships = async ( params: Object = {}, opts: Object = {} ): Pro
 
 const createRelationships = async ( params: Object = {}, opts: Object = {} ): Promise<?Object> => {
   try {
-    const response = await inatjs.relationships.create( { ...PARAMS, ...params }, opts );
-    return response;
+    return await inatjs.relationships.create( { ...PARAMS, ...params }, opts );
   } catch ( e ) {
     return handleError( e, { context: { functionName: "createRelationships", opts } } );
   }
@@ -28,8 +27,7 @@ const createRelationships = async ( params: Object = {}, opts: Object = {} ): Pr
 
 const updateRelationships = async ( params: Object = {}, opts: Object = {} ): Promise<?Object> => {
   try {
-    const response = await inatjs.relationships.update( { ...PARAMS, ...params }, opts );
-    return response;
+    return await inatjs.relationships.update( { ...PARAMS, ...params }, opts );
   } catch ( e ) {
     return handleError( e, { context: { functionName: "updateRelationships", opts } } );
   }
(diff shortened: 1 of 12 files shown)


src/navigation/types.ts
```diff
@@ -38,20 +39,38 @@ export type BaseTabStackParamList = {
...
+  Explore: {
+    user: ApiUser;
+    worldwide: boolean;
+  };
   ExploreFilters: undefined;
   ExploreSearch: undefined;
   ObsDetails: undefined;
   Notifications: undefined;
-  UserProfile: undefined;
+  UserProfile: { userId: number } & { login: string };
```
jtklein: Not sure about & here

src/api/relationships.js
```diff
@@ -10,43 +10,31 @@ const PARAMS = {
...
   try {
-    const response = await inatjs.relationships.update( { ...PARAMS, ...params }, opts );
-    return response;
+    return await inatjs.relationships.update( { ...PARAMS, ...params }, opts );
   } catch ( e ) {
     return handleError( e, { context: { functionName: "updateRelationships", opts } } );
   }
 };
 
-const deleteRelationships = async ( params: Object = {}, opts: Object = {} ): Promise<?Object> => {
```
jtklein, line 38: This was not used

src/api/relationships.js
```diff
@@ -10,43 +10,31 @@ const PARAMS = {
 
 const fetchRelationships = async ( params: Object = {}, opts: Object = {} ): Promise<?Object> => {
   try {
-    const response = await inatjs.relationships.search( { ...PARAMS, ...params }, opts );
-    return response;
+    const { results } = await inatjs.relationships.search( { ...PARAMS, ...params }, opts );
```
jtklein, line 13: These new returns follow more what we do in other API files.

src/components/UserProfile/UserProfile.tsx
```diff
@@ -124,6 +100,29 @@ const UserProfile = ( ): Node => {
...
   }
 
+  let hasRelationshipWithCurrentUser;
+  if ( relationships?.length > 0 ) {
+    hasRelationshipWithCurrentUser = relationships.find(
+      relationship => relationship.friendUser.id === userId,
+    );
+  }
+
+  const onObservationPressed = ( ) => {
```
jtklein, line 110: Those two don't need to be callbacks as they are only passed down to be Pressable onPress functions. Also if user is null they are never used anywhere, and should not be available because we don't want to navigate for a user without a user.

src/api/types.d.ts
```diff
@@ -154,6 +165,15 @@ export interface ApiObservation extends ApiRecord {
   uuid: string;
 }
 
+export interface ApiRelationship extends ApiRecord {
+  following?: boolean;
+  // For some reason inaturalistjs is changing friend_user to friendUser on the fly.
+  // https://github.com/inaturalist/inaturalistjs/blob/35534d90f67c4e724a5679a53e99a39cc167a8f5/lib/models/relationship.js#L7-L8
```
FLGMwt, line 171: yeah, that is very odd, I don't see `inaturalistjs` doing that for any other property

src/navigation/types.ts
```diff
@@ -38,20 +39,48 @@ export type BaseTabStackParamList = {
...
+  //   userLogin: user.login,
+  // }
+  ProjectList: {
+    observationUuid?: string;
+    userId?: number;
+    userLogin?: string;
+  };
+  FollowersList: {
+    // TODO: don't send the entire user object over here, only an ID or ID+login
+    user: ApiUser;
```
FLGMwt, line 78: good callout

FLGMwt (approved):

PR #3537, opened 2026-04-14
Use refetchInterval instead of new queryKey

jtklein:
Use react-query's polling instead of encoding time into the key. See
https://tanstack.com/query/latest/docs/framework/react/guides/polling
https://tanstack.com/query/latest/docs/framework/react/examples/auto-refetching

Before: NotificationsIconContainer polled by changing the query key every minute (numFetchIntervals in the key). That created new cached queries over time (unbounded growth).
After: query key is stable and uses refetchInterval: 60_000 for polling, so one query gets refetched instead of accumulating.


FLGMwt (approved):
good find and this is a good thing to be aware of!

PR #3538, opened 2026-04-14
Mob 713 project search typing in k causes a crash

abbeycampbell:
closes [MOB-713](https://linear.app/inaturalist/issue/MOB-713/project-search-typing-in-k-causes-a-crash)

Malformed dates for one of the projects returned when searching 'K' was causing this crash. we actually already have a note about this issue at line 417 in `dateAndTime.ts`, and I agree with the note that it would be nice to just be able to expect normalized dates from the API. But in the meantime, we'll handle this more gracefully by treating malformed dates as missing dates, and fallback to displaying the project type instead. 
Here's the project with the bad dates, now showing project type:
626A4D3D-7BAE-4147-B6B6-59A8ADB64391_1_201_a
Also includes some typescript improvements.

### DIFF
--- a/src/components/Projects/helpers/displayDates.ts
+++ b/src/components/Projects/helpers/displayDates.ts
@@ -65,28 +65,44 @@ const formatProjectDate = ( project, t, i18n ) => {
     ?.filter( pref => pref.field === "observed_on" ) );
   const months = getFieldValue( project?.rule_preferences
     ?.filter( pref => pref.field === "month" ) );
+  const formattedStartDate = projectStartDate
+    ? formatProjectsApiDatetimeLong( projectStartDate, i18n, { missing: null } )
+    : null;
+  const formattedEndDate = projectEndDate
+    ? formatProjectsApiDatetimeLong( projectEndDate, i18n, { missing: null } )
+    : null;
+  const formattedObservedOnDate = observedOnDate
+    ? formatProjectsApiDatetimeLong( observedOnDate, i18n, { missing: null } )
+    : null;
 
   if ( projectStartDate && !projectEndDate ) {
-    projectDate = t( "project-start-time-datetime", {
-      datetime: formatProjectsApiDatetimeLong( projectStartDate, i18n ),
-    } );
+    projectDate = formattedStartDate
+      ? t( "project-start-time-datetime", {
+        datetime: formattedStartDate,
+      } )
+      : null;
   }
-  if ( projectStartDate && projectEndDate ) {
+  if ( formattedStartDate && formattedEndDate ) {
     projectDate = t( "date-to-date", {
-      d1: formatProjectsApiDatetimeLong( projectStartDate, i18n ),
-      d2: formatProjectsApiDatetimeLong( projectEndDate, i18n ),
+      d1: formattedStartDate,
+      d2: formattedEndDate,
     } );
   }
-  if ( observedOnDate ) {
-    projectDate = formatProjectsApiDatetimeLong( observedOnDate, i18n );
+  if ( formattedObservedOnDate ) {
+    projectDate = formattedObservedOnDate;
   }
   if ( months ) {
     const monthList = months.split( "," );
-    projectDate = monthList.map( numberOfMonth => monthValues[numberOfMonth].label ).join( ", " );
+    const monthLabels = monthList
+      .map( numberOfMonth => monthValues[numberOfMonth]?.label )
+      .filter( Boolean );
+    projectDate = monthLabels.length > 0
+      ? monthLabels.join( ", " )
+      : null;
   }
   return {
     projectDate,
-    shouldDisplayDateRange: projectStartDate && projectEndDate
+    shouldDisplayDateRange: !!( formattedStartDate && formattedEndDate )
       && project?.project_type !== "traditional",
   };
 };
--- a/src/components/Projects/helpers/displayDates.ts
+++ b/src/components/Projects/helpers/displayDates.ts
@@ -8,7 +8,7 @@ interface ProjectRulePreference {
 }
 
 type ProjectWithDateRules = Omit<ApiProject, "project_type"> & {
-  project_type?: ApiProject["project_type"] | "traditional";
+  project_type?: ApiProject["project_type"];
   rule_preferences?: ProjectRulePreference[] | null;
 };
 
@@ -88,7 +88,7 @@ const formatProjectDate = (
   return {
     projectDate,
     shouldDisplayDateRange: !!( formattedStartDate && formattedEndDate )
-      && project?.project_type !== "traditional",
+      && project?.project_type !== "", // "" means "traditional"
   };
 };
 
--- a/src/components/Projects/helpers/displayDates.ts
+++ b/src/components/Projects/helpers/displayDates.ts
@@ -60,7 +60,7 @@ const formatProjectDate = (
     ? formatProjectsApiDatetimeLong( observedOnDate, i18n, { missing: null } )
     : null;
 
-  if ( projectStartDate && !projectEndDate ) {
+  if ( formattedStartDate && !formattedEndDate ) {
     projectDate = formattedStartDate
       ? t( "project-start-time-datetime", {
         datetime: formattedStartDate,
--- a/src/components/Projects/helpers/displayDates.ts
+++ b/src/components/Projects/helpers/displayDates.ts
@@ -1,61 +1,46 @@
+import type { ApiProject } from "api/types";
+import type { i18n as i18next, TFunction } from "i18next";
 import { formatProjectsApiDatetimeLong } from "sharedHelpers/dateAndTime";
 
-const getFieldValue = item => item?.[0]?.value;
+interface ProjectRulePreference {
+  field?: string;
+  value?: string | null;
+}
+
+type ProjectWithDateRules = Omit<ApiProject, "project_type"> & {
+  project_type?: ApiProject["project_type"] | "traditional";
+  rule_preferences?: ProjectRulePreference[] | null;
+};
+
+interface FormattedProjectDate {
+  projectDate: string | null;
+  shouldDisplayDateRange: boolean;
+}
+
+const getFieldValue = ( item?: ProjectRulePreference[] | null ) => item?.[0]?.value ?? null;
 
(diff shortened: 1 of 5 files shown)


src/sharedHelpers/dateAndTime.ts
```diff
@@ -332,9 +333,16 @@ function formatDateString(
     timeZone = Intl.DateTimeFormat( ).resolvedOptions( ).timeZone;
   }
 
+  const parsedDate = parseISO( isoDateString );
+  if ( !isValid( parsedDate ) ) {
+    return options.missing === undefined
```
jtklein, line 338: I think that is fair. But I am wondering if returning the unaltered isoDateString would also work here in terms of getting user feedback. We don't want the app to crash with an invalid string but we probably do want to have some visibility into a string that is expected by the user to be seen as a date but comes out of this clause as missing date. I wonder if a malformed date string is easier to spot than a "Missing date" where the user knows that there is a date.
My main concern is that isValid might be more stringent than we as iNat are in other places. Maybe on the web we would show some dates that are filtered out by this check.

src/components/Projects/helpers/displayDates.ts
```diff
@@ -65,28 +50,44 @@ const formatProjectDate = ( project, t, i18n ) => {
...
+    ? formatProjectsApiDatetimeLong( projectStartDate, i18n, { missing: null } )
+    : null;
+  const formattedEndDate = projectEndDate
+    ? formatProjectsApiDatetimeLong( projectEndDate, i18n, { missing: null } )
+    : null;
+  const formattedObservedOnDate = observedOnDate
+    ? formatProjectsApiDatetimeLong( observedOnDate, i18n, { missing: null } )
+    : null;
 
   if ( projectStartDate && !projectEndDate ) {
```
jtklein, line 63: Do you want to switch those two to use formatted.. instead?

src/components/Projects/helpers/displayDates.ts
```diff
@@ -65,28 +50,44 @@ const formatProjectDate = ( project, t, i18n ) => {
...
-    projectDate = formatProjectsApiDatetimeLong( observedOnDate, i18n );
+  if ( formattedObservedOnDate ) {
+    projectDate = formattedObservedOnDate;
   }
   if ( months ) {
     const monthList = months.split( "," );
-    projectDate = monthList.map( numberOfMonth => monthValues[numberOfMonth].label ).join( ", " );
+    const monthLabels = monthList
+      .map( numberOfMonth => monthValues[numberOfMonth] )
+      .filter( ( label ): label is string => Boolean( label ) );
```
jtklein, line 83: I don't think I understand what you are guarding against here. Even if a language has no translation we should get the en fallback, or if i18n is not initialized yet we should get the key back which is the same as the string.

jtklein (approved):
Thanks for doing this. Really appreciate the slow increase of type coverage in the codebase as well.

I think this PR works fairly well, only one minor concern cutting across different product dimensions.

src/components/Projects/helpers/displayDates.ts
```diff
@@ -1,61 +1,46 @@
...
-    11: {
-      label: t( "November" ),
-      value: 11,
-    },
-    12: {
-      label: t( "December" ),
-      value: 12,
-    },
+const formatProjectDate = (
+  project: ProjectWithDateRules | null | undefined,
```
jtklein, line 24: I understand that you looked at the call sites of this function as to what is passed in here. Not your code, but I do think that it would be better if we only call this function higher up for a valid project object.

FLGMwt: I don't think it's a responsibility for this PR, but I agree with Johannes. I would like if we could start to favor guarding at the callsite rather than expecting guards downstream. Certainly, this is coming more to light as we add types so thank you!

as is, I do have one question: is the `| undefined` intentional? I think we've typically done `project?:` for optionals/undefs.

src/components/Projects/helpers/displayDates.ts
```diff
@@ -1,61 +1,46 @@
...
 import { formatProjectsApiDatetimeLong } from "sharedHelpers/dateAndTime";
 
-const getFieldValue = item => item?.[0]?.value;
+interface ProjectRulePreference {
+  field?: string;
+  value?: string | null;
+}
+
+type ProjectWithDateRules = Omit<ApiProject, "project_type"> & {
+  project_type?: ApiProject["project_type"] | "traditional";
```
FLGMwt, line 11: oh, I was wondering about this, maybe you have context w/ making this change:

How are we treating `project_type: ""` vs `project_type: "traditional"` in the context of the comment on the ApiProject type: https://github.com/inaturalist/iNaturalistReactNative/blob/869adc74704f582587bce842954232d2fa2149e3/src/api/types.d.ts#L23, namely `// FYI "" means "traditional"`

Am I missing somewhere where we're _setting_ `project_type` to `"traditional"` when `project_type === ""`?

src/components/Projects/helpers/displayDates.ts
```diff
@@ -1,61 +1,46 @@
...
-    12: {
-      label: t( "December" ),
-      value: 12,
-    },
+const formatProjectDate = (
+  project: ProjectWithDateRules | null | undefined,
+  t: TFunction,
+  i18n: i18next,
+): FormattedProjectDate => {
+  const monthValues: Record<string, string> = {
```
FLGMwt, line 28: I'm so thankful of this deduping!

src/components/Projects/helpers/displayDates.ts
```diff
@@ -65,28 +50,44 @@ const formatProjectDate = ( project, t, i18n ) => {
...
+      .map( numberOfMonth => monthValues[numberOfMonth] )
+      .filter( ( label ): label is string => Boolean( label ) );
+    projectDate = monthLabels.length > 0
+      ? monthLabels.join( ", " )
+      : null;
   }
   return {
     projectDate,
-    shouldDisplayDateRange: projectStartDate && projectEndDate
+    shouldDisplayDateRange: !!( formattedStartDate && formattedEndDate )
```
FLGMwt, line 90: I'm not sure the `!!` is doing anything here. `(expression1 && expression2)` resolves strictly to a `boolean` no matter what because of the `&&` and the parens.

Was there a particular runtime or type error that necessitated this?

src/sharedHelpers/dateAndTime.ts
FLGMwt: would it be a fair middle ground to add a line to log "invalid" dates and investigate the occurrence going forward? b/c it seems like users aren't seeing them in the first place because the app crashes. This change might give less complete information but it fixes the crash.

nit: can we alias the import to like `isValidDateString` or something? mainly because it's a 500ln file and there's plennnnnnty of stuff that might be invalid : p

jtklein: Sure, a log also works. Even thoug this function is used for the crashing projects strings, it is not only used for projects. It is also used for other dates that might be not "isValid" but don't crash the app?

abbeycampbell: In my opinion, "missing date" is better UX/less confusing than a malformed date string. But I like the idea of logging them, so we can have better visibility into this and know how prevalent it is.

tbh, I didn't even think to check how the web was handling this, which is something I definitely need to get in the habit of considering. It seems like they use moment, which will return a display string "Invalid date". So we end up displaying "Invalid date - June 6, 2023" for the particular project which triggered this work. So yes Johannes, you're correct that we'd display at least part of the date range, when possible. I can create a follow-up ticket to take this approach in the app as well, if we think that's preferred--maybe it's something we can add to the iOS parity project, since we'll be fleshing out projects as part of that.

I can alias the import as well, but I'm gonna call it isValidDate, because `isValid` checks for the validity of date objects.

src/components/Projects/helpers/displayDates.ts
abbeycampbell: I don't think that's true? 
`&&` returns a value, and we have to coerce it to a boolean.
For example:
```
const test1 = ('str1' && 'str2') // 'str2'
const test2 = ('str1' && null) // null
const test3 = (null && null) //null
```
but if we coerce, we get:
```
const test1 = !!('str1' && 'str2') // true
const test2 = !!('str1' && null) // false
const test3 = !!(null && null) // false
```

src/components/Projects/helpers/displayDates.ts
abbeycampbell: good call out, I was confused about this too tbh. It doesn't look like we check for traditional projects using `""`, so I'm gonna remove the expansion I added here and just stay consistent with that.

I do think that's something we could normalize in the future; it's always nice to have explicit naming.

src/components/Projects/helpers/displayDates.ts
abbeycampbell: this isn't for handling translation fallbacks, it's meant to guard against invalid month values like "13" or "0", or if there's whitespace, undefined values, etc, so we'll skip those instead of crashing.

src/components/Projects/helpers/displayDates.ts
abbeycampbell: I agree that guarding at the call site and not having to fiddle with null/undefined in the helper is definitely better. I can write a follow-up, and maybe we can handle when we're spending a bit more time of projects.

`| undefined` is intentional to preserve argument order, because required parameters can't follow optional parameters

src/sharedHelpers/dateAndTime.ts
jtklein: I do like the "Invalid date - June 6, 2023", as a combination of both concerns. 1) Sth went wrong but 2) we have a date to show you. We can make a follow-up ticket about this. iNat Classic on iOS does not even show the date for projects, so not in iOS parity.

src/components/Projects/helpers/displayDates.ts
jtklein: Thanks for the explanation. I was missing that `months` is user (or server) generated.

jtklein (approved):
I have no further comments/questions, and all my previous ones are resolved.

src/components/Projects/helpers/displayDates.ts
FLGMwt: omg I'm so sorry, I haven't written C# in years but my C# brain manifested here. 

you're totally right!

FLGMwt (approved):
lgtm! thank you for correcting my syntax flub, sorry about that

PR #3541, opened 2026-04-15
Fix: Navigation issue in useExitObservationFlow, migrate to TS

jtklein:
Closes MOB-1316

Sorry for the huge PR but the TS migrations were very helpful in isolating the issue.
In terms of TS I was trying to stick to only type the navigation of screens that useExitObservationFlow is imported in, and the screens that route to those screens (because in the beginning useExitObservationFlow also had a useRoute and params check which I needed to see typed before I removed it).

The actual performance fix is [the last commit](https://github.com/inaturalist/iNaturalistReactNative/pull/3541/commits/cb76191195b58fe790d56c34ec8d43380bf5f028):
This is a fix for a performance issue we have seen where saving multiple observations and editing them leads to a slow app.
What is the problem? As the migration to TS made more clear, this hook is imported in screens in both the NoBottomTabStack and the TabStack.
Before: if you hit this navigation action in a screen in the NoBottomTabStack it works fine, the app navigates to the TabNavigator here and loads ObsList as the first screen. However, if you call it while already being in the TabNavigator it just pushes another ObsList unto the stack and not removing any previous screens (even though they are UI wise essentially unreachable).
After: Every time we hit this action we just reset the entire navigation to only show the Me tab with ObsList loaded. This frees up all accumulated screens in either the NoBottomTabStack or the TabStack (depending on from which one it is called in).

This PR also [removes one of the navigation flows](https://github.com/inaturalist/iNaturalistReactNative/pull/3541/commits/92e41df22cb2944b6f9c66ac0b57386d005db8b8) of the useExitObservationFlow, why?
1.) To remove some level of complexity in our navigation flows.
2.) Migrating all to TypeScript reveals that, this is an edge case for a user being on an ObsDetails screen and then starting a new obs from the AddObsButton. This has nothing to do with a user being on ObsDetails for an observation and starting to edit this same observation. In other words, this navigation flow meant that when a user starts a new obs and saves it, we navigate them back to the ObDetails they were on when starting to create a new obs.
3.) This is actually kind of buggy: If a user is in the Notifications tab looking at an ObsDetails of someone else's observation, then starting a new observation and finishing it, we are resetting the navigation to be in the Me tab looking at the same ObsDetails of someone else, which doesn't make a lot of sense.

Before this PR, for a flow of new install - import 3 photos from gallery - Save all - For each of the three new obs: One by one go into 1) ObsDetails - 2) ObsEdit - 3) Suggestions (select a taxon) - 4) ObsEdit (Save) - repeat for next one:
Screenshot 2026-04-15 at 23 10 35
Memory usage is only accumulating
After: Same flow
Screenshot 2026-04-15 at 22 47 57
Several cleanups hit

Before: Same flow as above but only 2 observations:
Screenshot 2026-04-15 at 21 34 49
Navigation state accumulated screens
After: Same flow as above but only 2 observations:
Screenshot 2026-04-15 at 21 37 02
Navigation reset after last obs Saved (Note that also the NoBottomTab has disappeared because after the import of 3 photos has finished with all being saved, the navigation is already reset once.

### DIFF
--- a/src/components/Camera/StandardCamera/hooks/useBackPress.ts
+++ b/src/components/Camera/StandardCamera/hooks/useBackPress.ts
@@ -1,5 +1,6 @@
 import { useFocusEffect, useNavigation } from "@react-navigation/native";
 import { getCurrentRoute } from "navigation/navigationUtils";
+import type { NoBottomTabStackScreenProps } from "navigation/types";
 import {
   useCallback,
   useState,
@@ -10,7 +11,7 @@ import {
 import useExitObservationFlow from "sharedHooks/useExitObservationFlow";
 
 const useBackPress = ( shouldShowDiscardSheet: boolean ) => {
-  const navigation = useNavigation( );
+  const navigation = useNavigation<NoBottomTabStackScreenProps<"Camera">["navigation"]>( );
   const exitObservationFlow = useExitObservationFlow( );
 
   const [showDiscardSheet, setShowDiscardSheet] = useState( false );
--- a/src/components/Camera/StandardCamera/hooks/useBackPress.ts
+++ b/src/components/Camera/StandardCamera/hooks/useBackPress.ts
@@ -1,5 +1,4 @@
-import { useFocusEffect, useNavigation } from "@react-navigation/native";
-import { getCurrentRoute } from "navigation/navigationUtils";
+import { useFocusEffect, useNavigation, useRoute } from "@react-navigation/native";
 import type { NoBottomTabStackScreenProps } from "navigation/types";
 import {
   useCallback,
@@ -12,6 +11,7 @@ import useExitObservationFlow from "sharedHooks/useExitObservationFlow";
 
 const useBackPress = ( shouldShowDiscardSheet: boolean ) => {
   const navigation = useNavigation<NoBottomTabStackScreenProps<"Camera">["navigation"]>( );
+  const { params } = useRoute<NoBottomTabStackScreenProps<"Camera">["route"]>( );
   const exitObservationFlow = useExitObservationFlow( );
 
   const [showDiscardSheet, setShowDiscardSheet] = useState( false );
@@ -19,17 +19,15 @@ const useBackPress = ( shouldShowDiscardSheet: boolean ) => {
   const handleBackButtonPress = useCallback( ( ) => {
     if ( shouldShowDiscardSheet ) {
       setShowDiscardSheet( true );
+    } else if ( params?.addEvidence ) {
+      navigation.navigate( "ObsEdit" );
     } else {
-      const currentRoute = getCurrentRoute();
-      if ( currentRoute?.params?.addEvidence ) {
-        navigation.navigate( "ObsEdit" );
-      } else {
-        exitObservationFlow( );
-      }
+      exitObservationFlow( );
     }
   }, [
     exitObservationFlow,
     navigation,
+    params?.addEvidence,
     setShowDiscardSheet,
     shouldShowDiscardSheet,
   ] );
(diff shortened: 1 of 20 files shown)


src/components/Camera/StandardCamera/hooks/useBackPress.ts
```diff
@@ -10,25 +10,24 @@ import {
...
   const [showDiscardSheet, setShowDiscardSheet] = useState( false );
 
   const handleBackButtonPress = useCallback( ( ) => {
     if ( shouldShowDiscardSheet ) {
       setShowDiscardSheet( true );
+    } else if ( params?.addEvidence ) {
+      navigation.navigate( "ObsEdit" );
     } else {
-      const currentRoute = getCurrentRoute();
-      if ( currentRoute?.params?.addEvidence ) {
```
jtklein, line 23: The only thing the currentRoute is used here is to get params, so I replaced it with useRoute to the same effect.

src/components/Explore/Modals/ExploreTaxonSearchModal.tsx
```diff
@@ -30,7 +28,6 @@ const ExploreTaxonSearchModal = ( {
       modal={(
         <ExploreTaxonSearch
           closeModal={closeModal}
-          hideInfoButton={hideInfoButton}
```
jtklein, line 33: This prop is nowhere used

src/navigation/types.ts
```diff
@@ -4,20 +4,176 @@
...
+  //   hideSkip: true,
+  // }
+  // From PhotoSharing
+  // { lastScreen: "PhotoSharing" }
+  // From GroupPhotosContainer
+  // {
+  //   entryScreen: "GroupPhotos",
+  //   lastScreen: "GroupPhotos",
+  // }
+  Suggestions: {
```
jtklein, line 163: It's kind of hard to see the inclusivity and exclusivity of different keys in the different routes this way, so I also added comments for each route. Not sure how to use TS as a documentation of all possible routes param configs here.

src/components/PhotoImporter/PhotoLibrary.tsx
```diff
@@ -84,6 +72,7 @@ const PhotoLibrary = ( ) => {
 
   const navBasedOnUserSettings = useCallback( async ( ) => {
     if ( isDefaultMode ) {
+      // TODO: why do we need to define higher navigator here
```
sepeterson, line 75: Did you figure out if these were necessary?

src/components/PhotoImporter/PhotoLibrary.tsx
```diff
@@ -195,6 +184,7 @@ const PhotoLibrary = ( ) => {
         navToObsEdit();
 
         // Determine if we need to go back to ObsList or ObsDetails screen
+      // TODO: the typing of params seems correct to me, so we might never hit this case here?
```
sepeterson: Is this still a TODO?

sepeterson (approved):
Nice find! A couple TODOs I wasn't sure were still needed but if you'd like to keep them, that's fine with me

Also pulled locally and ran through a few scenarios that worked fine ✅

src/components/PhotoImporter/PhotoLibrary.tsx
jtklein: I meant more in follow-up work to not bloat this PR too much.

src/components/PhotoImporter/PhotoLibrary.tsx
jtklein: This one I can remove actually. I am seeing "ObsDetails" in params.previousScreen.name so, rather something is off with the type.

PR #3544, opened 2026-04-16
prevent duplicate photo submissions by disabling checkmark while processing

abbeycampbell:
closes [MOB-1084](https://linear.app/inaturalist/issue/MOB-1084/multicapture-camera-multiple-copies-of-photos-can-be-saved)

### DIFF
--- a/src/components/Camera/AICamera/AICameraButtons.tsx
+++ b/src/components/Camera/AICamera/AICameraButtons.tsx
@@ -81,6 +81,7 @@ const AICameraButtons = ( {
         handleZoomButtonPress={handleZoomButtonPress}
         disabled={!modelLoaded || takingPhoto}
         disabledPhotoLibrary={takingPhoto}
+        confirmDisabled={false}
         flipCamera={flipCamera}
         hasFlash={hasFlash}
         hasPhotoLibraryButton
--- a/src/components/Camera/AICamera/AICameraButtons.tsx
+++ b/src/components/Camera/AICamera/AICameraButtons.tsx
@@ -81,7 +81,10 @@ const AICameraButtons = ( {
         handleZoomButtonPress={handleZoomButtonPress}
         disabled={!modelLoaded || takingPhoto}
         disabledPhotoLibrary={takingPhoto}
-        confirmDisabled={false}
+        // TODO: once we re-visit tablet views, we'll want to ensure users cannot spam the submit
+        // button while taking photos. see:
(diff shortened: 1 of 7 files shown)


src/components/Camera/AICamera/AICameraButtons.tsx
```diff
@@ -81,6 +81,7 @@ const AICameraButtons = ( {
         handleZoomButtonPress={handleZoomButtonPress}
         disabled={!modelLoaded || takingPhoto}
         disabledPhotoLibrary={takingPhoto}
+        confirmDisabled={false}
```
jtklein, line 84: I don't think we should actively update TabletButtons if we are not testing them. You could leave this new code as a comment linking to the ticket and say that once tablet views are being used again we need to test if this works.

src/components/Camera/StandardCamera/CameraOptionsButtons.js
```diff
@@ -70,6 +72,7 @@ const CameraOptionsButtons = ( {
       takePhoto={takePhoto}
       handleClose={handleClose}
       disabled={disabled}
+      confirmDisabled={confirmDisabled}
```
jtklein, line 75: Same as above

jtklein (approved):
Thanks for looking into this. I tested and after I did no longer get duplicate photos saved to my gallery, so the fix is working.

In general we also have a PressableWithDebounce component that can help in similar cases. But I do think that in your specific case there is an argument that it is possible for the process you are handling here might take longer than the 400ms debounce in this wrapper, so I am not requesting you to change it here.

src/components/Camera/CameraContainer.tsx
```diff
@@ -298,6 +308,7 @@ const CameraContainer = ( ) => {
         device={device}
         flipCamera={flipCamera}
         handleCheckmarkPress={handleNavigation}
+        confirmPhotosInProgress={confirmPhotosInProgress}
```
FLGMwt, line 311: is there a reason we can't pass `confirmPhotosInProgressRef.current` and avoid defining `confirmPhotosInProgress`?

jtklein: Because thou shall not read or write ref.current during render https://react.dev/reference/react/useRef#caveats

FLGMwt: ah! I'm a fool!

in that case, it may seem entirely in contrast to what I've been jabbering about, but I wonder if we might ought to have an effect _specifically_ listening for changes to the ref and updating the props-passed state value 

I'll have a think on this this weekend

PR #3550, opened 2026-04-20
Update MatchContainer to import loads function directly

jtklein:
This way we don't need to require the entire lodash package.


FLGMwt (approved):
LGTM

what do you think about using either a lint for this or a babel transform? seems to fit the flavor of "we want this but we don't have to think about it"

lint: https://github.com/idok/eslint-plugin-lodash/blob/HEAD/docs/rules/import-scope.md
transform: https://www.npmjs.com/package/babel-plugin-transform-imports

PR #3551, opened 2026-04-20
Replace a useEffect with a useFocusEffect

jtklein:
useNavigation replaced with useFocusEffect: focus handling uses the navigation hook meant for screen focus.


FLGMwt (approved):
nice, good call

PR #3552, opened 2026-04-20
add lint rule for lodash imports

FLGMwt:
hah @jtklein I thought this was a lot more pervasive than.... two remaining instances ; D only one instance after your #3550 
image

it doesn't have an autofix which is fine since we shouldn't have any more instances after this


jtklein (approved):
Thanks! I think this is a great idea.

PR #3553, opened 2026-04-21
Switch to @native-html/render to support markdown again

jtklein:
Closes MOB-1197

The previously used library is no longer maintained and has a migration notice to this new package: https://github.com/meliorence/react-native-render-html/issues/674

Just swapping out the library meant we already get back to supporting all markdown elements I checked for (didn't check for all in the list of supported html tags).
Added a snapshot test for the part that takes user text and generates html.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -2,6 +2,7 @@ module.exports = {
   presets: ["module:@react-native/babel-preset"],
   plugins: [
     "babel-plugin-react-compiler", // must run first!
+    "@babel/plugin-transform-export-namespace-from",
     "react-native-worklets-core/plugin",
     "transform-inline-environment-variables",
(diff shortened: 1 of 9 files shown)


babel.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   presets: ["module:@react-native/babel-preset"],
   plugins: [
     "babel-plugin-react-compiler", // must run first!
+    "@babel/plugin-transform-export-namespace-from",
```
FLGMwt, line 5: Is this needed for @native-html/render?

jtklein: Yes, Without the babel plugin the app does not start because of an error with the new package.

FLGMwt (approved):

PR #3556, opened 2026-04-21
Use zustand for debug mode state

FLGMwt:
We were using our mmkv store _directly_ instead of using the zustand store. This meant that react wasn't aware of updates to it. This adds that key to the persisted layout slice and updates all usages to a zustand-aware hook with two changes:
- the hook no longer returns the "toggle" function. This is _only_ used in one place and shouldn't be "accidentally" importable.
- the hook file no longer exports an imperative "get value". Again, this is only used in one place so is now defined in that one place (AuthenticationService). It still uses the store w/ an imperative `useStore.getState()` (specifically to _not_ call functions on mmkv)

We do have a TODO "don't do this" against this exact thing, I think we just hadn't noticed the impact.
```ts
// TODO do *not* export this. This allows any consumer to overwrite *any* part
// of state, circumventing any getter/setter logic we have in the stores. If
// you need to modify state, you should be doing so through a store.
export const zustandStorage = {
```

This instance has marginal user impact but I noticed it looking at obs counts / upload workflow which is bugged partially because we do this thing. I'll get to those next and we _might_ be able to remove the `export`.

A few remaining "violators" I'll fix:
- "numOfUserObservations" / "numOfUserSpecies": is set and read from `layout`, should _definitely_ be fixed, causes actual bugs
- `useStoredLayout`: I think this just shouldn't exist and can all be moved to `layout`
- "mapType": isolated but impact UI & easily ported to `layout`

Skipping
- "lastDeletedSyncTime": isolated, I'll leave this alone
- "LAST_CRASH_DATA": isolated, I'll leave this alone
- "availableLocales": isolated, I'll leave this alone

### DIFF
--- a/src/components/About.tsx
+++ b/src/components/About.tsx
@@ -13,6 +13,7 @@ import { t } from "i18next";
 import React, { useState } from "react";
 import { getBuildNumber, getVersion } from "react-native-device-info";
 import { useDebugMode } from "sharedHooks";
+import useStore from "stores/useStore";
 
 const aboutID = "about";
 const teamID = "team";
@@ -23,7 +24,8 @@ const About = ( ) => {
   const [count, setCount] = useState( 0 );
   const appVersion = getVersion();
   const buildVersion = getBuildNumber();
-  const { isDebug, toggleDebug } = useDebugMode( );
+  const { isDebug } = useDebugMode( );
+  const toggleDebug = useStore( state => state.layout.toggleDebugMode );
 
   const onTermsPressed = async () => {
     const url = "https://www.inaturalist.org/pages/terms";
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -123,7 +123,7 @@ const AICamera = ( {
   const changeDebugFormat = ( ) => {
     setDebugFormatIndex( prev => ( prev + 1 ) % device.formats.length );
   };
-  const debugFormat = isDebugMode()
+  const debugFormat = isDebug
     ? device.formats[debugFormatIndex]
     : undefined;
 
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -23,7 +23,6 @@ import {
   usePerformance,
   useTranslation,
 } from "sharedHooks";
-import { isDebugMode } from "sharedHooks/useDebugMode";
 import useStore from "stores/useStore";
 import colors from "styles/tailwindColors";
 
@@ -146,9 +145,11 @@ const AICamera = ( {
   const { loadTime } = usePerformance( {
     isLoading: camera?.current !== null,
   } );
-  if ( isDebugMode( ) && loadTime ) {
-    logger.info( loadTime );
-  }
+  useEffect( () => {
+    if ( isDebug && loadTime ) {
+      logger.info( loadTime );
+    }
+  }, [isDebug, loadTime] );
 
   const resetCameraOnFocus = useCallback( ( ) => {
     setResult( null );
--- a/src/components/LoginSignUp/AuthenticationService.ts
+++ b/src/components/LoginSignUp/AuthenticationService.ts
@@ -30,8 +30,12 @@ import removeAllFilesFromDirectory from "sharedHelpers/removeAllFilesFromDirecto
 import safeRealmWrite from "sharedHelpers/safeRealmWrite";
 import { setFirebaseDataCollectionEnabled } from "sharedHelpers/tracking";
 import { unlink } from "sharedHelpers/util";
-import { isDebugMode } from "sharedHooks/useDebugMode";
+import useStore from "stores/useStore";
 import zustandMMKVBackingStorage from "stores/zustandMMKVBackingStorage";
+
+function isDebugModeSync( ): boolean {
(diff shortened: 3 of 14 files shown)


src/components/Camera/AICamera/AICamera.js
```diff
@@ -146,9 +145,11 @@ const AICamera = ( {
...
     isLoading: camera?.current !== null,
   } );
-  if ( isDebugMode( ) && loadTime ) {
-    logger.info( loadTime );
-  }
+  useEffect( () => {
+    if ( isDebug && loadTime ) {
+      logger.info( loadTime );
+    }
+  }, [isDebug, loadTime] );
```
FLGMwt, line 152: This exact delta is pretty much most of the PR. I took the one extra-functional step of wrapping this in an effect to at least prevent the log on _every_ render but I stopped short of refactoring usePerformance

src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -30,9 +30,13 @@ import removeAllFilesFromDirectory from "sharedHelpers/removeAllFilesFromDirecto
 import safeRealmWrite from "sharedHelpers/safeRealmWrite";
 import { setFirebaseDataCollectionEnabled } from "sharedHelpers/tracking";
 import { unlink } from "sharedHelpers/util";
-import { isDebugMode } from "sharedHooks/useDebugMode";
+import useStore from "stores/useStore";
 import zustandMMKVBackingStorage from "stores/zustandMMKVBackingStorage";
 
+function isDebugModeSync( ): boolean {
+  return useStore.getState().layout.debugModeEnabled === true;
+}
```
FLGMwt, line 38: this is the only place we really "need" to imperatively get this so defining it in here

src/sharedHooks/useDebugMode.ts
```diff
@@ -1,33 +1,10 @@
...
-    setDebug( !isDebug );
-  }, [isDebug] );
+import useStore from "stores/useStore";
 
+const useDebugMode = ( ): { isDebug: boolean } => {
+  const isDebug = useStore( state => state.layout.debugModeEnabled );
   return {
     isDebug,
-    toggleDebug,
   };
```
FLGMwt, line 7: We're no longer returning a toggle function and instead pulling it from layout the one place we need it.

I was on the fence about removing the wrapping object or just switching consumers to `useStore` for the flag, but decided to keep the interface for minimal disruption.

src/stores/createLayoutSlice.ts
```diff
@@ -110,6 +110,14 @@ const createLayoutSlice = set => ( {
         dismissedAnnouncementIds: [],
       },
     } ) ),
+    // State to control debug mode
+    debugModeEnabled: false,
+    toggleDebugMode: () => set( state => ( {
+      layout: {
+        ...state.layout,
+        debugModeEnabled: !state.layout.debugModeEnabled,
+      },
```
FLGMwt, line 119: here we're adding the flag to the layout slice which is persisted.

this _does_ mean that the previous `"debugMode"` key will be orphaned for existing installs, but I think that's fine, not worth migrating for an "internal" flag

jtklein: There are actually quite a lot of users that have debug mode enabled, see Kibana logs for usePerformance logs.
But I do agree that we make no promise to keep this internal flag consistent across production releases.

src/components/Camera/AICamera/AICamera.js
jtklein: I think I would have preferred to refactor usePerformance to be imperative before doing this here, to avoid adding useEffects. But if we have a ticket and commitment for the refactor of usePerformance in the near future, I am okay with adding this here.

src/components/Menu/Menu.tsx
```diff
@@ -22,8 +22,10 @@ import User from "realmModels/User";
 import { valueToBreakpoint } from "sharedHelpers/breakpoint";
 import { log } from "sharedHelpers/logger";
 import getStorageMetrics from "sharedHelpers/storageMetrics";
-import { useCurrentUser, useLayoutPrefs, useTranslation } from "sharedHooks";
-import { zustandStorage } from "stores/useStore";
+import {
+  useCurrentUser, useDebugMode,
```
jtklein, line 26: Weird format

src/components/About.tsx
```diff
@@ -23,7 +24,8 @@ const About = ( ) => {
   const [count, setCount] = useState( 0 );
   const appVersion = getVersion();
   const buildVersion = getBuildNumber();
-  const { isDebug, toggleDebug } = useDebugMode( );
+  const { isDebug } = useDebugMode( );
+  const toggleDebug = useStore( state => state.layout.toggleDebugMode );
```
jtklein, line 28: For me it could have stayed in useDebugMode but also fine.

jtklein (approved):
LGTM.
Do you want to make tickets for the rest?

src/components/Camera/AICamera/AICamera.js
FLGMwt: I was unclear here, I didn't introduce a need for a `useEffect`. we _ought_ to have had a `useEffect` before this change. MOB-1369 created to track this tho.

FLGMwt:
> LGTM. Do you want to make tickets for the rest?

@jtklein I made a ticket for `usePerformance`, was there another ticket I missed?

PR #3557, opened 2026-04-21
Fix list component remounting in ProjectMembers.tsx

jtklein:
Closes MOB-1202

ProjectMembers now passes memoized elements (useMemo) to UserList for ListEmptyComponent and ListFooterComponent instead of useCallback render functions, so FlashList doesn’t remount those subtrees when deps change (same approach as MOB-717 / #3383).

Also added types for CustomFlashList.


FLGMwt (approved):
LGTM thanks for the types!

PR #3559, opened 2026-04-21
Fix list component remounting in Projects.tsx

jtklein:
Closes [MOB-1204](https://linear.app/inaturalist/issue/MOB-1204)

Projects now passes memoized elements (useMemo) to ProjectList for ListEmptyComponent and ListFooterComponent instead of useCallback render functions (same approach as [MOB-717](https://linear.app/inaturalist/issue/MOB-717) / https://github.com/inaturalist/iNaturalistReactNative/pull/3383).
Also refactored out ItemSeparator in ProjectsList following the same PR as an example.

### DIFF
--- a/src/components/ProjectList/ProjectList.tsx
+++ b/src/components/ProjectList/ProjectList.tsx
@@ -11,6 +11,8 @@ import {
 
 import ProjectListItem from "./ProjectListItem";
 
+const ItemSeparator = () => <View className="border-b border-lightGray" />;
+
 interface Props {
   projects: ApiProject[];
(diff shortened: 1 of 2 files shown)


src/components/ProjectList/ProjectList.tsx
```diff
@@ -11,10 +11,12 @@ import {
 
 import ProjectListItem from "./ProjectListItem";
 
+const ItemSeparator = () => <View className="border-b border-lightGray" />;
```
sepeterson, line 14: Excellent

sepeterson (approved):
Looks great

PR #3560, opened 2026-04-22
Toggle advanced mode for users that have 100 obs while logging in

jtklein:
Closes MOB-1309

We are already fetching the remote user after authentication, so all I did there was to refactor the return type to also return the number of already uploaded observations (`observations_count` from API). We save the info to realm as well but we do need the count immediately from the caller.

I did not take into consideration a way to manipulate the actual threshold at runtime, as we are not really setup for that yet. So, currently a hard-coded const of 100 obs.

Since we already have the number of obs setting the boolean zustand that is "Advanced mode" is already done before we navigate away from the login screen, so there should be no awkward switch when on the Me tab. (Or do we need to "await" setIsDefaultMode?)

Some things to note for the logic part: if the app is in default or advanced mode before pressing "Log in" shouldn't matter for this bit because we want to enforce isDefaultMode to false when >100 obs and do not change it when <100 obs.

One open question I had: Why is the remote user typed so defensively with some fields being undefined?

PS: Sorry, couldn't help myself and added some navigation types. Bad example because they should be in a separate PR.

### DIFF
--- a/src/components/LoginSignUp/AuthenticationService.ts
+++ b/src/components/LoginSignUp/AuthenticationService.ts
@@ -542,7 +542,9 @@ async function verifyCredentials(
   return afterVerifyCredentials( tokenResponse, apiClient );
 }
 
-export interface AuthenticateUserResult { success: boolean }
+export type AuthenticateUserResult =
+| { success: true; observationsCount?: number }
+| { success: false };
 
 async function afterAuthenticateUser(
   userDetails: UserDetails | null,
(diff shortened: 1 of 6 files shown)


FLGMwt:
changes LGTM as implementing the thing in the ticket : )

Reviewing this clarified a product concern I have. Is it fair to withhold an approve until I clear that up? It's spooky to me to change app behavior this dramatically without notice.

FLGMwt (approved):

PR #3561, opened 2026-04-22
fix UiLibrary fast refresh

FLGMwt:
this happens to be a very similar concern to the `<FlashList ItemSeparatorComponent={???} />` thing. I think it manages specifically to break Fast Refresh because it _happens_ to cross a module boundary at the same time we're treating the components as render helpers: `return MyComponent()` is very different than `return <MyComponent />`. React can't help you with the former _if_ the render helper definition lives outside the render tree.

idk how much we use this but this should make it nicer. just saw the kueda comment and made the fix while I was here.


jtklein (approved):
LGTM

PR #3564, opened 2026-04-22
When a user has 100 obs uploaded show Advanced Mode banner

jtklein:
Closes MOB-1335, also closes MOB-1355

For logged-in users still in default mode with ≥100 uploaded observations, show a dismissible banner on My Observations below server announcements and above the observation list. Dismiss is remembered until reinstall or logout. Tap switches to Advanced Mode immediately (no loading state required).

Also, consolidated list header content (announcements + banner + missing-basics row) via SimpleHeader.
BannerDismissButton extracted from Announcements and reused on the new banner.
AdvancedModeBanner is purely presentational; SimpleHeader computes eligibility and passes numTotalObservations from the container plus onDismiss / onSwitchToAdvancedMode handlers.


src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -238,12 +237,14 @@ const MyObservationsSimple = ( {
...
 
   const observationsHeader = ( ) => {
-    const headerContent = obsMissingBasicsExist
-      ? <SimpleErrorHeader isConnected={isConnected} />
-      : <Announcements isConnected={isConnected} />;
-
     if ( layout !== "grid" ) {
-      return headerContent;
+      return (
+        <SimpleHeader
```
jtklein, line 242: I realized we have Announcements in both versions of the headerContent and the "variable" part is if we need to show the error message. After this PR, we also would have the banner in both versions. So, I think a header component that decides if we need to show an error is better for now.

src/components/MyObservations/LoginBanner.tsx
```diff
@@ -29,38 +31,16 @@ const LoginBanner = ( {
...
-        <View
-          className="absolute self-center top-0"
-        >
+        <View className="absolute self-center top-0">
           <View className="pt-[20px] px-[20px]" pointerEvents="box-none">
-            <Pressable
-              className="absolute top-0 right-0 justify-center items-center h-[44px] w-[44px] z-20"
-              accessibilityRole="button"
-              onPress={() => dismissLoginBanner()}
+            <BannerDismissButton
```
jtklein, line 39: One requirement for the ticket was to make the dismiss button of the new banner identical to the one used in announcements, so I also re-used it here which is technically a different ticket. But I think it makes more sense to combine the two in one PR.

src/components/MyObservations/SimpleHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <>
+      <Announcements isConnected={isConnected} />
+      {shouldShowAdvancedModeBanner && (
+        <AdvancedModeBanner
+          onDismiss={( ) => setAdvancedModeBannerDismissed( )}
+          onSwitchToAdvancedMode={( ) => setIsDefaultMode( false )}
+        />
+      )}
+      { obsMissingBasicsExist && (
+        <View className="flex-row items-center px-[32px] py-[20px]">
```
jtklein, line 57: This part is based on the previous SimpleErrorHeader

src/components/MyObservations/Announcements.tsx
```diff
@@ -185,24 +186,11 @@ const Announcements = ( {
...
-        <Pressable
-          accessibilityRole="button"
+        <BannerDismissButton
           accessibilityLabel={t( "Dismiss-announcement" )}
-          className="absolute top-0 right-0 h-[44px] w-[44px] items-center
-          justify-center z-10"
           onPress={dismiss}
           testID="announcements-dismiss"
-        >
-          <View
```
jtklein, line 196: This was the basis for BannerDismissButton

src/components/MyObservations/BannerDismissButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+  onPress: ( ) => void;
+  testID: string;
+}
+
+const BannerDismissButton = ( {
+  accessibilityLabel,
+  onPress,
+  testID,
+}: Props ) => (
+  <Pressable
```
jtklein, line 17: This was taken as is from Announcements, `active:opacity-75` is an addition here to provide tap feedback

src/components/MyObservations/SimpleHeader.tsx
```diff
@@ -0,0 +1,76 @@
...
+    state => state.layout.setAdvancedModeBannerDismissed,
+  );
+  const setIsDefaultMode = useStore( state => state.layout.setIsDefaultMode );
+
+  const currentUser = useCurrentUser();
+
+  const shouldShowAdvancedModeBanner = !!currentUser
+      && isDefaultMode
+      && numTotalObservations
+      && numTotalObservations >= ADVANCED_MODE_ONBOARDING_MIN_OBSERVATIONS
```
sepeterson, line 44: The copy says "more than 100 observations" not >= 100 😉

sepeterson (approved):
lookin good

src/components/MyObservations/AdvancedModeBanner.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onDismiss,
+  onSwitchToAdvancedMode,
+}: Props ) => {
+  const { t } = useTranslation( );
+
+  return (
+    <View className="relative px-[27px] py-[20px]">
+      <BannerDismissButton
+        accessibilityLabel={t( "Close" )}
+        onPress={( ) => onDismiss( )}
```
sepeterson: I know sometimes one can wrap a prop with an arrow function to make sure a child component doesn't pass arguments you don't expect but is that necessary here? Could we just do `onPress={onDismiss}`

PR #3565, opened 2026-04-22
Sample workflow: Slack notify pattern sandbox

simonsanchez:
Adds a low-risk `workflow_dispatch` sandbox (`.github/workflows/sample_echo.yml`) that exercises the same Slack notify pattern used in `test.yml` against a trivial echo job. Useful for iterating on the Slack setup without touching the real CI.

- `force-fail` input lets you simulate a failed upstream job.
- Slack post uses the official `slackapi/slack-github-action` pinned by SHA (v3.0.1).
- `SLACK_WEBHOOK_URL` is set at workflow-level `env` so the `if:` guard correctly skips the post when the secret is unset (e.g. on forks).

Draft — opening for visibility / early feedback.

### DIFF
--- a/.github/workflows/sample_echo.yml
+++ b/.github/workflows/sample_echo.yml
@@ -16,6 +16,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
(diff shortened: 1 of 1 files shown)


FLGMwt:
Thanks, @simonsanchez, was fun to jam on this! I'm closing the branch for now but we'll leverage this and the guidance you shared when we get back to tackling this.

PR #3566, opened 2026-04-22
MOB-1317: replace delete mode overlay modal with wrapping pressable

sepeterson:
Replace the modal-based "delete photo mode" with a <Pressable> that wraps the camera UI and conditional blocks taps on the children via `pointerEvents`

the SemanticDiff is pretty helpful here


src/components/Camera/StandardCamera/PhotoCarousel.tsx
```diff
@@ -112,9 +114,9 @@ const PhotoCarousel = ( {
 
   const showDeletePhotoMode = useCallback( ( ) => {
     if ( deletePhotoFromObservation ) {
-      setDeletePhotoMode( mode => !mode );
+      setDeletePhotoMode( !deletePhotoMode );
```
jtklein: I think this is a minor regression: Previously we didn't need `deletePhotoMode` in the dependencies which causes extra callback work.

jtklein (approved):
Works like a charm. I completely forgot we had this complicated measuring setup here.

src/components/Camera/StandardCamera/PhotoCarousel.tsx
sepeterson: Good catch -- in practice this function only sets to true, so I'll just use the literal `true`

PR #3573, opened 2026-04-27
retrigger notification query on effect rather than new query

FLGMwt:
closes MOB-1382

Highlighting the distinction between:
- API call
- Query: instance of a React Query returned from `useQuery`, one instance per `queryKey`. maintains its own request state and cache entry. triggering a query's queryFn usually results in an API call

before: every new "viewed at" timestamp update would trigger a new Query (and API call)
after: every new "viewed at" timestamp update will trigger the same Query's `refetch` (and API call)

here's the only place that's set w/ a `new Date()`
https://github.com/inaturalist/iNaturalistReactNative/blob/26d89434d0799afa3cb509868e31799fb9ce61e3/src/components/ObsDetailsSharedComponents/hooks/useMarkViewedMutation.js#L45-L47

### DIFF
--- a/src/components/Notifications/NotificationsTab.tsx
+++ b/src/components/Notifications/NotificationsTab.tsx
@@ -23,9 +23,6 @@ const NotificationsTab = ( { id, text }: TabComponentProps ) => {
     [
       "NotificationsTab",
       "notificationsCount",
-      // We want to check for notifications when the user views an
-      // observation, because that might make the indicator go away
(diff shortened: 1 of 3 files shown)


jtklein (approved):
Have tested, and can confirm that this does indeed create less query instances when looking at several notifications in one session.

FLGMwt:
ty for the merge!

PR #3574, opened 2026-04-27
Update react hooks eslint for 19.2

FLGMwt:
I was thinking that useEffectEvent _may_ be appropriate for the useQuery fixes. I wasn't planning on introducing it yet, but it uncovered that our version of react-hooks linting wasn't "effect event"-aware. Specifically, `exhaustive-deps` should _not_ fire when "effect events" are omitted from the dep array. In fact, it should _stop_ you from using one in the dep array: https://react.dev/reference/react/useEffectEvent#effect-event-in-deps.

Anyway, I'm bumping this so that we can explore that in the future.

Additionally,  new 7.x plugin is more closely aligned with React Compiler and it uncovered some _great_ spots for the stuff we were already fixing in one-offs. I added the `:recommended` and explicitly disabled those with a comment. It'd be nice to set those to warn a bit at a time to start knocking them down.

as of now, the error count is:
```
react-hooks/static-components: 33
react-hooks/set-state-in-effect: 31
react-hooks/refs: 26
react-hooks/preserve-manual-memoization: 11
react-hooks/purity: 3
react-hooks/immutability: 3
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -113,14 +113,14 @@ module.exports = {
 
     // @FLGMWT 20260427: the following were introduced with `react-hooks/recommended` and represent
     // cases that are very likely to cause bugs and _have definitely_ been the cause of a number of
-    // bugs fixed in the past few months. Ideally, we'll progressively turn these into warnings
-    // and then errors as we fix them : ) noting here in case it ever gets lost in Linear.
-    "react-hooks/static-components": 0,
-    "react-hooks/set-state-in-effect": 0,
-    "react-hooks/refs": 0,
-    "react-hooks/preserve-manual-memoization": 0,
-    "react-hooks/purity": 0,
-    "react-hooks/immutability": 0,
+    // bugs fixed in the past few months. Turning these on as warnings and we can eventually
+    // switch to errors as we fix them : ) noting here in case it ever gets lost in Linear.
+    "react-hooks/static-components": 1,
+    "react-hooks/set-state-in-effect": 1,
+    "react-hooks/refs": 1,
+    "react-hooks/preserve-manual-memoization": 1,
+    "react-hooks/purity": 1,
+    "react-hooks/immutability": 1,
 
     "react-native/no-inline-styles": "error",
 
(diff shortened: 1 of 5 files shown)


src/components/FullPageWebView/FullPageWebView.tsx
```diff
@@ -206,7 +206,6 @@ const FullPageWebView = ( ) => {
       }
       // TODO: I am not sure how to make the react compiler happy here, so I disabled it
       // for this hook
-      // eslint-disable-next-line react-hooks/react-compiler
```
FLGMwt, line 209: there's no longer a `react-compiler` rule for "heyyyyyyyy, really, I mean it, don't do this". I think that's fine, but if we'd like to keep this behavior, I can find a generally configurable "disable disable" rule

.eslintrc.js
```diff
@@ -105,9 +106,21 @@ module.exports = {
...
+    // @FLGMWT 20260427: the following were introduced with `react-hooks/recommended` and represent
+    // cases that are very likely to cause bugs and _have definitely_ been the cause of a number of
+    // bugs fixed in the past few months. Ideally, we'll progressively turn these into warnings
+    // and then errors as we fix them : ) noting here in case it ever gets lost in Linear.
+    "react-hooks/static-components": 0,
+    "react-hooks/set-state-in-effect": 0,
+    "react-hooks/refs": 0,
+    "react-hooks/preserve-manual-memoization": 0,
+    "react-hooks/purity": 0,
+    "react-hooks/immutability": 0,
```
FLGMwt: https://react.dev/reference/eslint-plugin-react-hooks/lints/static-components
https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect
https://react.dev/reference/eslint-plugin-react-hooks/lints/refs
https://react.dev/reference/eslint-plugin-react-hooks/lints/preserve-manual-memoization
https://react.dev/reference/eslint-plugin-react-hooks/lints/purity
https://react.dev/reference/eslint-plugin-react-hooks/lints/immutability

abbeycampbell (approved):
Thanks for doing this. I'm down with this plan. I think it'll help us avoid common pitfalls going forward; we can handle it incrementally and the error count list isn't too scary

PR #3576, opened 2026-04-27
Mob 1239 obs sheet not dismissing

abbeycampbell:
closes [MOB-1239](https://linear.app/inaturalist/issue/MOB-1239/obs-sheet-not-dismissing)

Big diff is mostly snapshot changes, see comments. The actual fix is in the first commit: switched from unmounting the sheet when hidden to explicitly opening/closing it through the bottom sheet API-- the sheet could linger or reappear after navigation because cleanup didn’t run reliably when we were returning null.

I also verified that we didn't regress on the issue amanda was fixing when she first added the 'return null' paths, where dismissing a sheet by tapping on the background meant you couldn't reopen it.

I'm finding this bottom sheet stuff pretty confusing so I tried to do a little consolidation/deduping of the logic.

This PR also includes some type error fixes, which necessitated splitting the render path. I actually think this makes things easier to reason about, but I think we could take it even further and have two separate components or just not use the `insideModal` path at all? I think we're only really using that prop in explore filters, which are about to become their own screen, if I'm not mistaken.

### DIFF
--- a/tests/unit/components/AddObsBottomSheet/AddObsButton.test.js
+++ b/tests/unit/components/AddObsBottomSheet/AddObsButton.test.js
@@ -30,7 +30,9 @@ describe( "AddObsButton", () => {
 
   it( "renders correctly", () => {
     renderComponent( <AddObsButton /> );
-    // Snapshot test
+    // Snapshot includes AddObs BottomSheet subtree due @gorhom/bottom-sheet test mock
+    // rendering children unconditionally. Hidden/open behavior is asserted in
+    // integration tests.
     expect( screen ).toMatchSnapshot();
   } );
   it( "does not render tooltip in default state", () => {
--- a/src/components/SharedComponents/Sheets/BottomSheet.tsx
+++ b/src/components/SharedComponents/Sheets/BottomSheet.tsx
@@ -71,6 +71,7 @@ const StandardBottomSheet = ( {
 
   const { t } = useTranslation( );
   const sheetRef = useRef<SheetHandle>( null );
+  const skipNextOnPressCloseRef = useRef( false );
   const insets = useSafeAreaInsets( );
 
   // The optional `sheet` arg lets the unmount cleanup pass a captured handle;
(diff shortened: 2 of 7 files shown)


tests/unit/components/AddObsBottomSheet/AddObsButton.test.js
```diff
@@ -30,7 +30,9 @@ describe( "AddObsButton", () => {
 
   it( "renders correctly", () => {
     renderComponent( <AddObsButton /> );
-    // Snapshot test
+    // Snapshot includes AddObs BottomSheet subtree due @gorhom/bottom-sheet test mock
+    // rendering children unconditionally. Hidden/open behavior is asserted in
+    // integration tests.
```
abbeycampbell, line 35: Alternatively, we could just remove this snapshot test, as well as the CustomNavBar one.

jtklein (changes requested):
The code changes seem okay to me.
1. If we will remove insideModal when Explore V2 is done then we can probably leave it as is now, and don't need to make separate components?

Two requested changes:
1. We should also make these changes in BottomSheetV2.
2. One regression I am seeing is that for the Add and ID modal, what used to happen was press add ID -> modal opens -> press add comment -> second modal opens -> accept comment -> first (add ID) modal opens back up. Now on accept comment nothing happens.

abbeycampbell:
> 2. One regression I am seeing is that for the Add and ID modal, what used to happen was press add ID -> modal opens -> press add comment -> second modal opens -> accept comment -> first (add ID) modal opens back up. Now on accept comment nothing happens.

I am only seeing this happen on the Agree with ID flow, which seems to already be an existing bug in the Internal build:
- press Agree
- tap Add a comment 
- add a comment and confirm
- modal closes/nothing happens
- if you tap the agree button agin, the original modal will open and allow you to post an ID, but not with a comment.

It seems to work correctly for the Suggest an ID with comment flow, both in the Internal build and in this branch--unless I'm missing something?

Regardless, I fixed the bottom sheet closes/nothing happens issue. But your comment uncovered a new bug/functionality issue, which is that we lead people to believe they can add a comment to an agreeing ID, but currently can't. I think it's a simple fix, but one that should be made separately from this. So I'll make a ticket for it

jtklein (approved):
You are right. I was confusing the Suggest and ID and Agree with ID sheets to be identical components (with different title) but they are not. The bug is pre-existing to this PR.
Thanks for fixing the nothing happens issue.

PR #3577, opened 2026-04-28
MOB-1321: approximate location for android

sepeterson:
The intent of this PR is to support approximate (coarse) location permissions for android. The old approach in useWatchPosition doesn't work for this case but I wanted to retain a single hook that sourced this information. 
Main changes are: 
- Instead of "subscribing" to a user's location in useWatchPosition, we retry a few times. I think this preserves the intent of the hook while simplifying the logic
- For coarse permissions in useWatchPosition, just fetch the location once since it already has a ceiling on accuracy
- Add a special condition for permissionResultFromMultiple to accept android allowing coarse OR fine permissions

### DIFF
--- a/src/components/SharedComponents/PermissionGateContainer.tsx
+++ b/src/components/SharedComponents/PermissionGateContainer.tsx
@@ -69,7 +69,14 @@ export const WRITE_MEDIA_PERMISSIONS = Platform.OS === "ios"
 
 export const LOCATION_PERMISSIONS = Platform.OS === "ios"
   ? [PERMISSIONS.IOS.LOCATION_WHEN_IN_USE]
-  : [PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION];
+  : [
+    PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION,
+    PERMISSIONS.ANDROID.ACCESS_COARSE_LOCATION,
+  ];
+
+export interface MultiResult {
+  [permission: string]: PermissionStatus;
+}
 
 interface Props extends PropsWithChildren {
   blockedPrompt?: string;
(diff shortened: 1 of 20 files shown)


src/components/SharedComponents/PermissionGateContainer.tsx
```diff
@@ -91,16 +98,32 @@ interface Props extends PropsWithChildren {
...
     throw new Error(
       "permissionResultFromMultiple received something other than an object. "
       + "Make sure you're using it with checkMultiple and not check",
     );
   }
+  // On Android 12+, the user may grant only approximate (coarse) location,
+  // leaving fine location denied. Use OR logic: if ANY location permission is
+  // granted, the overall result is GRANTED.
+  const coarseKey = PERMISSIONS.ANDROID.ACCESS_COARSE_LOCATION;
+  if ( coarseKey in multiResults ) {
```
sepeterson, line 112: I want to acknowledge that this is a little goofy but that I prefer it to the alternatives of setting up a new function or adding some sort of optional boolean. This function is used in two pretty different contexts and the updates to the calling sites would get messy

src/sharedHooks/useWatchPosition.ts
```diff
@@ -20,113 +18,84 @@ export interface UserLocation {
...
-  // Listen for focus. We only want to fetch location when this screen has focus.
+    return ( ) => { cancelled = true; };
+  }, [shouldFetchLocation] );
+
+  // Cancel any in-flight fetch when we lose focus
   useEffect( ( ) => {
-    const unsubscribe = navigation.addListener( "focus", ( ) => {
-      setHasFocus( true );
-    } );
+    const unsubscribe = navigation.addListener( "blur", cancel );
```
abbeycampbell, line 91: maybe I'm missing something, but we dropped our focus listener. should we be re-triggering a fetch for location on focus?

abbeycampbell (commented):
I like the approach of using a specified number of retries. Just one question from me

PR #3578, opened 2026-04-28
capture if user was defaulted into adv mode and log that for feedback

FLGMwt:


sepeterson:
I'm not totally sure about this change. I'd prefer to extract this information after the fact if we need it at all

src/components/LoginSignUp/LoginForm.tsx
```diff
@@ -126,6 +130,7 @@ const LoginForm = ( {
       // If a user that just logged in already has more than the threshold number of observations,
       // we assume they are an advanced user and switch them to advanced mode.
       setIsDefaultMode( false );
+      setUserWasDefaultedToAdvancedMode();
```
abbeycampbell, line 133: we can't know that a user was defaulted to advanced mode here, right? They could have already been in advanced mode while logged out, in which case, if they leave feedback, this flag would be misleading.

src/components/Menu/Menu.tsx
```diff
@@ -240,6 +241,7 @@ const Menu = ( ) => {
         username: "loggedout",
         identifications: "loggedout",
         remoteObservations: "loggedout",
+        userWasDefaultedToAdvancedMode,
```
abbeycampbell, line 244: This is in the wrong place, right? Since this flag is set on login, this value needs to be in the `currentUser === true` branch

abbeycampbell (changes requested):
I'm personally not opposed to adding this info to feedback logs, since it's cheap to do and will mean the info is more immediately accessible to Abhas. But I do want to make sure we really think through all the possible flows and make sure we don't send misleading info

src/components/LoginSignUp/LoginForm.tsx
jtklein: Maybe the "switch" in my comment above is misleading here. This path get's hit for a user that is in default mode as well as for a user that is already in advanced mode. The set is switching the user to advanced mode from default mode, but has no effect if the user is already in advanced mode.

src/stores/createLayoutSlice.ts
```diff
@@ -118,6 +118,14 @@ const createLayoutSlice = set => ( {
         debugModeEnabled: !state.layout.debugModeEnabled,
       },
     } ) ),
+    // Diagnostic flags
+    userWasDefaultedToAdvancedMode: null,
```
jtklein, line 122: This is not really a layout relevant state.

jtklein (commented):
I am not sure why we would want to log this proactively.

FLGMwt:
Closing, I had poor assumptions, apologies.

PR #3582, opened 2026-04-29
useQuery's refetch doesn't respect enabled

FLGMwt:
my bad, I switched refetch to use an effect but I wasn't considering that that still needs to consider `!!currentUser`. `refetch` doesn't take `enabled` into consideration


src/components/Notifications/NotificationsTab.tsx
abbeycampbell, line 52: do we need to guard this one too?

abbeycampbell (approved):

src/components/Notifications/NotificationsTab.tsx
FLGMwt: great catch! thank you, will update

PR #3585, opened 2026-05-01
MOB-1329 ExploreV2 state and initial obs fetching

sepeterson:
In this PR:
- Navigator for ExploreV2
- Context/Provider for ExploreV2
- Query building for ExploreV2
- Stub screens for ExploreObservations, UniversalSearch, and AdvancedSearch

I also have a funky lil modal for debugging in the branch `mob-1329-debug`

https://github.com/user-attachments/assets/817abdd2-7f60-48b6-95ac-b1f99e7b83e8

### DIFF
--- a/src/components/Explore/ExploreV2/ExploreV2Container.tsx
+++ b/src/components/Explore/ExploreV2/ExploreV2Container.tsx
@@ -14,7 +14,6 @@ const ExploreV2WithProvider = ( ) => {
   const { state, dispatch } = useExploreV2( );
   const {
     hasPermissions,
-    renderPermissionsGate,
   } = useLocationPermission( );
   const previousHasPermissions = useRef<boolean | undefined>( undefined );
 
@@ -53,10 +52,7 @@ const ExploreV2WithProvider = ( ) => {
   }, [hasPermissions] );
 
   return (
-    <>
-      <ExploreStackNavigator />
-      {renderPermissionsGate( undefined )}
-    </>
+    <ExploreStackNavigator />
   );
 };
 
--- a/src/components/Explore/ExploreV2/ExploreV2Container.tsx
+++ b/src/components/Explore/ExploreV2/ExploreV2Container.tsx
@@ -19,11 +19,7 @@ const ExploreV2WithProvider = ( ) => {
   const previousHasPermissions = useRef<boolean | undefined>( undefined );
 
   const onPermissionsGained = useEffectEvent( async ( ) => {
-    // State not empty. Do nothing
-    if (
-      state.placeMode !== EXPLORE_V2_PLACE_MODE.NEARBY
-      || state.lat !== undefined
-    ) return;
+    if ( state.placeMode !== EXPLORE_V2_PLACE_MODE.UNINITIALIZED ) return;
 
     const next = await defaultExploreV2Location( );
     if ( next.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY ) {
@@ -51,7 +47,7 @@ const ExploreV2WithProvider = ( ) => {
   useEffect( ( ) => {
     if (
       hasPermissions === false
-      && state.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY
+      && state.placeMode === EXPLORE_V2_PLACE_MODE.UNINITIALIZED
     ) {
       dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
     }
--- a/src/components/Explore/ExploreV2/ExploreV2Container.tsx
+++ b/src/components/Explore/ExploreV2/ExploreV2Container.tsx
@@ -7,7 +7,7 @@ import {
   ExploreV2Provider,
   useExploreV2,
 } from "providers/ExploreV2Context";
-import React, { useEffect, useRef } from "react";
+import React, { useEffect, useEffectEvent, useRef } from "react";
 import useLocationPermission from "sharedHooks/useLocationPermission";
 
 const ExploreV2WithProvider = ( ) => {
@@ -18,42 +18,33 @@ const ExploreV2WithProvider = ( ) => {
   } = useLocationPermission( );
   const previousHasPermissions = useRef<boolean | undefined>( undefined );
 
-  // Resolve initial location once we know permission state. NEARBY with no
-  // coords means we haven't fetched the user's coarse location yet.
-  useEffect( ( ) => {
-    let cancelled = false;
-    async function resolveLocation( ) {
-      if (
-        hasPermissions
-        && !previousHasPermissions.current
-        && state.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY
-        && state.lat === undefined
-      ) {
-        const next = await defaultExploreV2Location( );
-        if ( cancelled ) return;
-        if (
-          next.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY
-          && next.lat !== undefined
-          && next.lng !== undefined
-          && next.radius !== undefined
-        ) {
-          dispatch( {
-            type: EXPLORE_V2_ACTION.SET_LOCATION_NEARBY,
-            lat: next.lat,
-            lng: next.lng,
-            radius: next.radius,
-          } );
-        } else {
-          dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
-        }
-      }
-      previousHasPermissions.current = hasPermissions;
+  const onPermissionsGained = useEffectEvent( async ( ) => {
+    // State not empty. No op
+    if (
+      state.placeMode !== EXPLORE_V2_PLACE_MODE.NEARBY
+      || state.lat !== undefined
+    ) return;
+
+    const next = await defaultExploreV2Location( );
+    if ( next.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY ) {
+      dispatch( {
+        type: EXPLORE_V2_ACTION.SET_LOCATION_NEARBY,
+        lat: next.lat,
+        lng: next.lng,
+        radius: next.radius,
+      } );
+    } else {
+      dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
     }
-    resolveLocation( );
-    return ( ) => {
-      cancelled = true;
-    };
-  }, [hasPermissions, state.placeMode, state.lat, dispatch] );
+  } );
+
+  // handle granting location permissions on Explore
+  useEffect( ( ) => {
+    if ( hasPermissions && !previousHasPermissions.current ) {
+      onPermissionsGained( );
+    }
+    previousHasPermissions.current = hasPermissions;
+  }, [hasPermissions] );
 
   // Per the ticket: default to "Worldwide" when location is denied (or blocked).
   // Once permission state is known to be false, fall back from NEARBY.
--- a/src/components/Explore/ExploreV2/ExploreV2Container.tsx
+++ b/src/components/Explore/ExploreV2/ExploreV2Container.tsx
@@ -1,0 +1,83 @@
+import ExploreStackNavigator
+  from "navigation/StackNavigators/ExploreStackNavigator";
+import {
+  defaultExploreV2Location,
+  EXPLORE_V2_ACTION,
+  EXPLORE_V2_PLACE_MODE,
+  ExploreV2Provider,
+  useExploreV2,
+} from "providers/ExploreV2Context";
+import React, { useEffect, useRef } from "react";
+import useLocationPermission from "sharedHooks/useLocationPermission";
+
+const ExploreV2WithProvider = ( ) => {
+  const { state, dispatch } = useExploreV2( );
+  const {
+    hasPermissions,
+    renderPermissionsGate,
+  } = useLocationPermission( );
+  const previousHasPermissions = useRef<boolean | undefined>( undefined );
+
+  // Resolve initial location once we know permission state. NEARBY with no
+  // coords means we haven't fetched the user's coarse location yet.
+  useEffect( ( ) => {
+    let cancelled = false;
+    async function resolveLocation( ) {
+      if (
+        hasPermissions
+        && !previousHasPermissions.current
+        && state.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY
+        && state.lat === undefined
+      ) {
+        const next = await defaultExploreV2Location( );
+        if ( cancelled ) return;
+        if (
+          next.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY
+          && next.lat !== undefined
+          && next.lng !== undefined
+          && next.radius !== undefined
+        ) {
+          dispatch( {
+            type: EXPLORE_V2_ACTION.SET_LOCATION_NEARBY,
+            lat: next.lat,
+            lng: next.lng,
+            radius: next.radius,
+          } );
+        } else {
+          dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
+        }
+      }
+      previousHasPermissions.current = hasPermissions;
+    }
+    resolveLocation( );
+    return ( ) => {
+      cancelled = true;
+    };
+  }, [hasPermissions, state.placeMode, state.lat, dispatch] );
+
+  // Per the ticket: default to "Worldwide" when location is denied (or blocked).
+  // Once permission state is known to be false, fall back from NEARBY.
+  useEffect( ( ) => {
+    if (
+      hasPermissions === false
+      && state.placeMode === EXPLORE_V2_PLACE_MODE.NEARBY
+    ) {
+      dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
+    }
+  }, [hasPermissions, state.placeMode, dispatch] );
+
+  return (
+    <>
+      <ExploreStackNavigator />
+      {renderPermissionsGate( undefined )}
+    </>
+  );
+};
+
+const ExploreV2Container = ( ) => (
+  <ExploreV2Provider>
+    <ExploreV2WithProvider />
+  </ExploreV2Provider>
+);
+
+export default ExploreV2Container;
--- a/src/components/Explore/ExploreV2/ExploreV2Container.tsx
+++ b/src/components/Explore/ExploreV2/ExploreV2Container.tsx
@@ -34,24 +34,23 @@ const ExploreV2WithProvider = ( ) => {
     }
   } );
 
-  // handle granting location permissions on Explore
+  // default to "Worldwide" when location is denied
+  // hasPermissions === false always means permission has been denied or blocked
+  const onPermissionsDenied = useEffectEvent( ( ) => {
+    if ( state.placeMode === EXPLORE_V2_PLACE_MODE.UNINITIALIZED ) {
+      dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
+    }
+  } );
+
+  // handle location permission changes on Explore
   useEffect( ( ) => {
-    if ( hasPermissions && !previousHasPermissions.current ) {
+    if ( hasPermissions === true && previousHasPermissions.current !== true ) {
       onPermissionsGained( );
+    } else if ( hasPermissions === false && previousHasPermissions.current !== false ) {
+      onPermissionsDenied( );
     }
     previousHasPermissions.current = hasPermissions;
   }, [hasPermissions] );
-
-  // default to "Worldwide" when location is denied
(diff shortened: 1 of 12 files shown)


src/components/Explore/ExploreV2/ExploreV2Container.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const ExploreV2WithProvider = ( ) => {
+  const { state, dispatch } = useExploreV2( );
+  const {
+    hasPermissions,
+    renderPermissionsGate,
+  } = useLocationPermission( );
+  const previousHasPermissions = useRef<boolean | undefined>( undefined );
+
+  const onPermissionsGained = useEffectEvent( async ( ) => {
```
sepeterson, line 20: Gave `useEffectEvent` a whirl. lmk if this usage looks okay

jtklein (changes requested):
Two questions before diving into the actual changes here:
1. Why not use the screen components that I already made for the new V2 designs in #3416 and only lift the provider to a new component surrounding the three new screens? ExploreFiltersContainer = your AdvancedSearch; ExploreSearchContainer = your UniversalSearch; ExploreV2 probably some kind of combination of your ExploreStackNavigator and ExploreObservations that need to be entangled at one point. If I am missing the point as to why we need new screens we need to at least clean up the ones introduced for ExploreV2 already.
2. Can we avoid the new navigator and use a Stack.Group instead: https://reactnavigation.org/docs/nesting-navigators#best-practices-when-nesting?

sepeterson:
> 1. Why not use the screen components that I already made for the new V2 designs in #3416 and only lift the provider to a new component surrounding the three new screens? ExploreFiltersContainer = your AdvancedSearch; ExploreSearchContainer = your UniversalSearch; ExploreV2 probably some kind of combination of your ExploreStackNavigator and ExploreObservations that need to be entangled at one point. If I am missing the point as to why we need new screens we need to at least clean up the ones introduced for ExploreV2 already.

My thought was to implement the state/navigation separately from existing UI for this ticket and then the tickets for the screens could draw from your existing work when they get taken up. The main reason for the new screens is just to use the same namespace (calling the screens Universal Search and Advanced Search) as the tickets. I think I would prefer to leave the stub screens as is to keep the diff cleaner but also think it's good to have your work around for future reference. What's the right level of cleaned-up for this PR?

> 2. Can we avoid the new navigator and use a Stack.Group instead: https://reactnavigation.org/docs/nesting-navigators#best-practices-when-nesting?

Unfortunately Stack.Group can't be used like that. If you use the [screenLayout prop](https://reactnavigation.org/docs/group/#screen-layout) to wrap those screens with ExploreV2Provider, it would mount a separate provider around each screen.

That said, I was thinking it might be worth it to implement the search screens as fullscreen sheets instead of RN screens because that would *also* enable us to avoid the nested navigators. I think the main questions are: would that require design changes? Would it be able to gracefully handle navigating into ExploreV2 directly into a search screen?

src/components/Explore/ExploreV2/screens/ExploreObservations.tsx
```diff
@@ -0,0 +1,55 @@
...
+          onEndReached={fetchNextPage}
+          showNoResults={!canFetch || totalResults === 0}
+          testID="ExploreV2ObservationsList"
+        />
+      </View>
+    </ViewWrapper>
+  );
+};
+
+export default ExploreObservations;
```
jtklein: Not final but name not correct, in the end this will not only show observations but also taxa and user like V1 does now.

src/components/Explore/ExploreV2/screens/ExploreResults.tsx
```diff
@@ -0,0 +1,55 @@
+import { useNetInfo } from "@react-native-community/netinfo";
```
jtklein, line 1: Haven't looked at this file as I assume everything here is preliminary.

src/components/Explore/ExploreV2/buildQueryParams.ts
```diff
@@ -0,0 +1,81 @@
+import type { ExploreV2State } from "providers/ExploreV2Context";
```
jtklein, line 1: Haven't looked at this file as I assume everything here is preliminary.

src/providers/ExploreV2Context.tsx
```diff
@@ -0,0 +1,210 @@
+import * as React from "react";
```
jtklein, line 1: Haven't looked at this file as I assume everything here is preliminary.

tests/unit/components/Explore/ExploreV2/buildQueryParams.test.js
```diff
@@ -0,0 +1,167 @@
+import buildExploreV2QueryParams
```
jtklein, line 1: Haven't looked at this file as I assume everything here is preliminary.

tests/unit/providers/ExploreV2Context.test.js
```diff
@@ -0,0 +1,226 @@
+import {
```
jtklein, line 1: Haven't looked at this file as I assume everything here is preliminary.

src/components/Explore/ExploreV2/ExploreV2Container.tsx
jtklein: I am not familiar with this yet but @FLGMwt has looked into it.

src/components/Explore/ExploreV2/ExploreV2Container.tsx
```diff
@@ -0,0 +1,65 @@
+import ExploreStackNavigator
```
jtklein, line 1: Haven't looked at this file in too much details, as I assume everything here is preliminary. Do I understand correctly though that your approach is to split the current ExploreContainer up into one part that is a container for the Navigator (this file) and one container for the Explore results UI (yet untouched)?

src/components/Explore/ExploreV2/ExploreV2Container.tsx
```diff
@@ -0,0 +1,65 @@
...
+  // hasPermissions === false always means permission has been denied or blocked
+  const onPermissionsDenied = useEffectEvent( ( ) => {
+    if ( state.location.placeMode === EXPLORE_V2_PLACE_MODE.UNINITIALIZED ) {
+      dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
+    }
+  } );
+
+  // handle location permission changes on Explore
+  useEffect( ( ) => {
+    if ( hasPermissions === true && previousHasPermissions.current !== true ) {
```
jtklein: ExploreObservations has no permission gate yet, have you tested if this works here?

src/navigation/StackNavigators/ExploreStackNavigator.tsx
```diff
@@ -0,0 +1,36 @@
...
+import ExploreObservations
+  from "components/Explore/ExploreV2/screens/ExploreObservations";
+import UniversalSearch
+  from "components/Explore/ExploreV2/screens/UniversalSearch";
+import { hideHeader } from "navigation/navigationOptions";
+import type { ExploreStackParamList } from "navigation/types";
+import React from "react";
+
+// When navigating out of this stack, useNavigation should be typed like:
+// useNavigation<ExploreStackScreenProps<"ExploreObservations">["navigation"]>( );
```
jtklein: Why? Would using e.g. "UniversalSearch" instead of "ExploreObservations" not work?

src/navigation/StackNavigators/TabStackNavigator.tsx
```diff
@@ -197,7 +203,9 @@ const TabStackNavigator = ( { route }: BottomTabProps ) => {
         />
         <Stack.Screen
           name={SCREEN_NAME_ROOT_EXPLORE}
-          component={RootExploreContainer}
+          component={exploreV2Enabled
+            ? ExploreV2Container
```
jtklein, line 207: Is removing the RootExplore - Explore separation part of the Explore Re-Design project? If not do we also need to use ExploreV2Container instead of ExploreContainer below?

src/navigation/types.ts
```diff
@@ -382,6 +391,14 @@ export type TabStackScreenProps<T extends keyof TabStackParamList> =
     BottomTabProps
   >;
 
+// ExploreStackNavigator is nested inside RootExplore. This composite type
+// acknowledges ExploreV2 screens access to outer-stack routes
+export type ExploreStackScreenProps<T extends keyof ExploreStackParamList> =
+  CompositeScreenProps<
+    NativeStackScreenProps<ExploreStackParamList, T>,
+    TabStackScreenProps<"RootExplore">
```
jtklein, line 399: I am not sure I understand all implications of why keeping this here not generic is better. I think it does make sense because the new navigator will always be inside the RootExplore screen. But this means that we should do the same in some of the other navigators, like for example LoginStackScreenProps could be with RootStackScreenProps<"LoginStackNavigator">? Or because the actual component of the Stack.Screen with the name LoginStackNavigator is the LoginStackNavigator it is different?

jtklein (approved):
Thanks for you r answers above. I think I am just confused as to what the purpose of this single PR is. Given the fact that the entire Explore Re-design should mostly be cosmetic (not changing any query logic) the logic migration here seems intermediate to me. If I understand right that subsequent tickets will do more of the migration then for me personally it makes more sense to review the diff when this migration is complete. To that effect I added a ticket to do that review later. So, for this PR I think we should just merge it and review most of the changes after the project is done instead.

I'd prefer a smaller PR that is actually reviewable. Personally, I think refactoring the state of the context (where does `subject` come from?) could have been done after the new UI is fully done, if at all. If you think that refactoring the state internals is fundamental then it could have been a separate PR that applies to ExploreV1 and V2 as is now at the same time.

src/components/Explore/ExploreV2/ExploreV2Container.tsx
FLGMwt: If I understand correctly that this bit of logic is _only_ meant to respond to changes to `hasPermissions` and _not_ meant to otherwise respond to changes to state.location.placeMode, then I think this is the _exact_ intended use case of `useEffectEvent`. 

_in fact_, in this, case, it makes it so that we can remove the `previousHasPermissions` tracking ref. Consider: if the effect is triggered when `hasPermissions` changes we can know for certain that it's not equal to the previous value.

That said, I've been hesitant to introduce useEffectEvent because it's new and we know we have effect bugs. Maybe it makes sense to introduce it here in _new_ code but maybe with some extra comments? I'll probably still hold off for existing files until we feel more confident in it.

Thoughts @jtklein @sepeterson ? 

Also, maybe worth a 15m "here's when to use it" chat?

src/components/Explore/ExploreV2/ExploreV2Container.tsx
```diff
@@ -0,0 +1,65 @@
...
+} from "providers/ExploreV2Context";
+import React, { useEffect, useEffectEvent, useRef } from "react";
+import useLocationPermission from "sharedHooks/useLocationPermission";
+
+const ExploreV2WithProvider = ( ) => {
+  const { state, dispatch } = useExploreV2( );
+  const {
+    hasPermissions,
+  } = useLocationPermission( );
+  const previousHasPermissions = useRef<boolean | undefined>( undefined );
```
FLGMwt: Mentioned below, in `useEffectEvent` comment, I think we can remove this.

src/providers/ExploreV2Context.tsx
```diff
@@ -0,0 +1,210 @@
...
+  | { placeMode: EXPLORE_V2_PLACE_MODE.WORLDWIDE }
+  | {
+    placeMode: EXPLORE_V2_PLACE_MODE.NEARBY;
+    lat: number;
+    lng: number;
+    radius: number;
+  };
+
+export async function defaultExploreV2Location( ): Promise<DefaultExploreV2Location> {
+  const location = await fetchCoarseUserLocation( );
```
FLGMwt, line 163: Not related to this, but it highlights that we do this full perm/location fetch loop on demand whenever we use location in the app. I wonder if this is something that we should consider moving to an app-wide store? And subscribe to perm updates in one place.

src/navigation/StackNavigators/TabStackNavigator.tsx
FLGMwt: Did I understand you correctly @sepeterson that the plan was to focus on _just_ the Explore tab for now and that these other v2 ports would come later?

src/providers/ExploreV2Context.tsx
```diff
@@ -0,0 +1,210 @@
...
+export const ExploreV2Provider = ( { children }: ExploreV2ProviderProps ) => {
+  const [state, dispatch] = React.useReducer( exploreV2Reducer, initialExploreV2State );
+
+  const value = React.useMemo(
+    () => ( { state, dispatch } ),
+    [state],
+  );
+
+  return (
+    <ExploreV2Context.Provider value={value}>
```
FLGMwt: trivial: we can omit the `.Provider` as of react 19 https://react.dev/reference/react/createContext#provider

src/providers/ExploreV2Context.tsx
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// more options to be potentially added in MOB-1333
+export enum EXPLORE_V2_SORT {
+  DATE_UPLOADED_NEWEST = "DATE_UPLOADED_NEWEST",
+  DATE_UPLOADED_OLDEST = "DATE_UPLOADED_OLDEST",
+  DATE_OBSERVED_NEWEST = "DATE_OBSERVED_NEWEST",
+  DATE_OBSERVED_OLDEST = "DATE_OBSERVED_OLDEST",
+  MOST_FAVED = "MOST_FAVED"
+}
```
FLGMwt, line 32: We talked about this being an alternative but not expected or necessary for this PR:

Since this will be ideally consumed by TS-only files, we can consider a stringy alternative to enums. No functional difference but they can be easier to type against and potentially more ergonomic to use. e.g., `dispatch( { type: 'typeSafeActionType', ...`. The result is the same w/ the same type safety but can be easier to read and no import is required.

src/providers/ExploreV2Context.tsx
```diff
@@ -0,0 +1,210 @@
...
+      return { ...state, subject: null };
+    case EXPLORE_V2_ACTION.SET_LOCATION_NEARBY:
+      return {
+        ...state,
+        location: {
+          placeMode: EXPLORE_V2_PLACE_MODE.NEARBY,
+          lat: action.lat,
+          lng: action.lng,
+          radius: action.radius,
+        },
```
FLGMwt, line 124: I don't know how complex this reducer is going to get, but I can already see where it could make sense to physically separate concerns such as "location" and "subject" w/ the slice/combine reducer pattern: https://redux.js.org/api/combinereducers

I wouldn't expect to do that here at this stage, just want to keep it in mind as an option as this gets closer to the full implementation.

src/components/Explore/ExploreV2/screens/ExploreObservations.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from "components/SharedComponents";
+import { View } from "components/styledComponents";
+import { EXPLORE_V2_PLACE_MODE, useExploreV2 } from "providers/ExploreV2Context";
+import React, { useMemo } from "react";
+
+const ExploreObservations = ( ) => {
+  const { state } = useExploreV2( );
+  const { isConnected } = useNetInfo( );
+
+  const queryParams = useMemo( () => buildExploreV2QueryParams( state ), [state] );
```
FLGMwt: This should not be memoized, the query object building should be trivial

src/providers/ExploreV2Context.tsx
```diff
@@ -0,0 +1,210 @@
...
+  subject: null,
+  location: { placeMode: EXPLORE_V2_PLACE_MODE.UNINITIALIZED },
+  sortBy: EXPLORE_V2_SORT.DATE_UPLOADED_NEWEST,
+  filters: {},
+};
+
+export function exploreV2Reducer(
+  state: ExploreV2State,
+  action: ExploreV2Action,
+): ExploreV2State {
```
FLGMwt, line 110: I really like the context / reducer approach! I think this is a great way to capture the correlated actions & state.

The types make it super clean as well. I think it's a great idea to keep the user API at the raw `dispatch`as opposed to adding helpers / wrappers such as `clearSubject`, at least for now while this is still under development.

FLGMwt (approved):
I walked through this with Seth on Friday, thanks @sepeterson !

I'm clear on and approve the context / query building as written and as it serves these screens. I'm expecting we'll iterate on it and the navigation, but I am good w/ the approach.

I'm still unclear though on the overall routing / history / navigation expectations, maybe that's something that we need to clear up? I would expect that a "new" instance of the ExploreV2Context would be route params-informed, is that correct? Or is it only ever going to get the initializing default state and then `setSubject`'d?

FLGMwt:
Oh, also, I think we should pull in mob-1329-debug! Once this is in, could you PR that behind a new Feature Flag? I think it'd be helpful to have some testing controls while working on this before it's "fully real"

src/components/Explore/ExploreV2/ExploreV2Container.tsx
sepeterson: Good call on `previousHasPermissions` I will push that removal and add a comment marking it as a new pattern.

Let's definitely find time to discuss, maybe in standup as makes sense

src/components/Explore/ExploreV2/screens/ExploreObservations.tsx
sepeterson: Great point, I'm renaming to ExploreResults

src/components/Explore/ExploreV2/ExploreV2Container.tsx
sepeterson: Yes, this container will handle updates pertinent to the whole navigator and ExploreResults will handle the "container" role for that screen. Perhaps all 3 screens will necessitate splitting into Container/Screen patterns

src/components/Explore/ExploreV2/ExploreV2Container.tsx
sepeterson: Yes it also fires on first render to set the initial state to Nearby/Worldwide so I was able to test both branches

src/navigation/StackNavigators/ExploreStackNavigator.tsx
sepeterson: Yes, you would replace it with whatever `keyof ExploreStackParamList` you were on so it expects the correct params

src/navigation/StackNavigators/TabStackNavigator.tsx
sepeterson: Exactly. I'd like to support entry from other screens with https://linear.app/inaturalist/issue/MOB-1337/explore-entry-points-from-other-screens

src/navigation/types.ts
sepeterson: Yes I believe the difference here is this is our only navigator inside a Screen. So we *must* use TabStackScreenProps which requires a key.

LoginStackScreenProps could indeed be:

```
export type LoginStackScreenProps<T extends keyof LoginStackParamList> =                                    
    CompositeScreenProps<
      NativeStackScreenProps<LoginStackParamList, T>,                                                         
      NativeStackScreenProps<RootStackParamList, "LoginStackNavigator">
    >;   
```

I think this makes the exact parent more specific (instead of a union of all higher routes) but that isn't used in practice

sepeterson:
> I'm still unclear though on the overall routing / history / navigation expectations, maybe that's something that we need to clear up? I would expect that a "new" instance of the ExploreV2Context would be route params-informed, is that correct? Or is it only ever going to get the initializing default state and then `setSubject`'d?

On the Explore Tab, it will always initialize to some default state and then be updated via `setSubject`. We'll need to support some route param initialization when users are able to navigate from other routes

src/providers/ExploreV2Context.tsx
sepeterson: *Definitely* interested in this in the future!

src/providers/ExploreV2Context.tsx
sepeterson: Cool! Good callout, will push.

src/providers/ExploreV2Context.tsx
sepeterson: Ya good point! Absolutely going to be necessary once we get to filters imo

src/components/Explore/ExploreV2/screens/ExploreObservations.tsx
sepeterson: Thanks -- will push!

sepeterson:
> Oh, also, I think we should pull in [mob-1329](https://linear.app/inaturalist/issue/MOB-1329)-debug! Once this is in, could you PR that behind a new Feature Flag? I think it'd be helpful to have some testing controls while working on this before it's "fully real"

Sounds good! I'll look into it

PR #3588, opened 2026-05-04
Fix list component remounting in GroupPhotos

jtklein:
Closes MOB-1206

Also, migrating GroupPhotos and non-typed children to TS.

### DIFF
--- a/src/components/PhotoImporter/GroupPhotoImage.tsx
+++ b/src/components/PhotoImporter/GroupPhotoImage.tsx
@@ -5,11 +5,11 @@ import { Pressable } from "components/styledComponents";
 import type { Node } from "react";
 import React from "react";
 
-type Props = {
-  item: Object,
-  selectedObservations: Object[],
-  selectObservationPhotos: Function,
-  style?: Object
+interface Props {
+  item: object;
+  selectedObservations: object[];
+  selectObservationPhotos: Function;
+  style?: object;
 }
 
 const GroupPhotoImage = ( {
--- a/src/components/PhotoImporter/GroupPhotoImage.tsx
+++ b/src/components/PhotoImporter/GroupPhotoImage.tsx
@@ -2,10 +2,17 @@ import ObsImagePreview from "components/ObservationsFlashList/ObsImagePreview";
 import { Pressable } from "components/styledComponents";
 import React from "react";
 
+interface Item {
+  photos: {
+    image: {
+      uri: string;
+    };
+  }[];
+}
 interface Props {
-  item: object;
-  selectedObservations: object[];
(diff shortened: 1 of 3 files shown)


src/components/PhotoImporter/GroupPhotos.tsx
```diff
@@ -13,24 +12,43 @@ import {
...
+  photos: {
+    image: {
+      uri: string;
+    };
+  }[];
+}
+
+type GroupPhotosListItem = Item | { empty: true };
+
+function isEmptyGridItem( item: GroupPhotosListItem ): item is { empty: true } {
```
sepeterson, line 38: Beautifully typed

src/components/PhotoImporter/GroupPhotos.tsx
```diff
@@ -124,13 +137,12 @@ const GroupPhotos = ( {
 
   const extraData = {
     selectedObservations,
-    gridItemWidth,
```
sepeterson, line 127: Is this intentional? No explanation needed if so just wanted to be sure

sepeterson (approved):
Ready to go assuming the gridItemWidth change was intentional

src/components/PhotoImporter/GroupPhotos.tsx
jtklein: Yes. Was not actually exported from the hook above and so forever `undefined`.

src/components/PhotoImporter/GroupPhotos.tsx
FLGMwt: hard +1

src/components/PhotoImporter/GroupPhotoImage.tsx
```diff
@@ -1,23 +1,28 @@
...
-  style?: Object
+interface Item {
+  photos: {
+    image: {
+      uri: string;
+    };
+  }[];
+}
+interface Props {
+  item: Item;
```
FLGMwt, line 14: very minor question, no change requested, I'm just curious about naming preferences here: `<GroupPhotoImage item={Item} />`

`item` totally makes sense as a prop name. For the type name, I might have gone for `item: GroupPhotoImageItem`. The type isn't exported but it _would_ show up in an editor as a consumer fills in the prop.

FLGMwt (commented):
LGTM, thanks for the types!

src/components/PhotoImporter/GroupPhotoImage.tsx
jtklein: I didn't think too hard about it because I assume from top we prop in something like an entire observation here and at a later point we have to redo the type here when connecting the screen to these children here.

PR #3591, opened 2026-05-05
Restrict queries to only stats, no data fetched

jtklein:
Closes MOB-1359

### DIFF
--- a/src/components/ProjectDetails/ProjectDetailsContainer.js
+++ b/src/components/ProjectDetails/ProjectDetailsContainer.js
@@ -71,8 +71,6 @@ const ProjectDetailsContainer = ( ): Node => {
     }, optsWithAuth ),
   );
 
-  // TODO: enable fields if it makes sense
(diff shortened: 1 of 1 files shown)


src/components/ProjectDetails/ProjectDetailsContainer.js
```diff
@@ -71,17 +71,14 @@ const ProjectDetailsContainer = ( ): Node => {
     }, optsWithAuth ),
   );
 
-  // TODO: enable fields if it makes sense
-  // https://linear.app/inaturalist/issue/MOB-1359/enable-fields-for-searchobservations-in-projectdetailscontainer
   const { data: projectStats } = useAuthenticatedQuery(
     ["searchObservations", id],
```
sepeterson: Mostly a nit but I wonder if we should use a more specific key for this. Seems possible another query in the future would use `["searchObservations", id]`

Maybe:

```suggestion
    ["searchObservations", "projectStats", id],
```

sepeterson (approved):
Good to merge after the query key thing

PR #3592, opened 2026-05-05
Omit fields if not given

jtklein:
While looking at query results from useLinking, I have seen that we are handling a response that looks like this: `{ uuid: xy, observation_photos: undefined, observationPhotos: undefined }`. This changes so that an observations without photos does not add those fields when mapped to local schema.

### DIFF
--- a/src/api/observations.js
+++ b/src/api/observations.js
@@ -15,8 +15,16 @@ function mapObsPhotoToLocalSchema( obsPhoto ) {
   return obsPhoto;
 }
 function mapToLocalSchema( observation ) {
-  observation.observationPhotos = observation?.observationPhotos?.map( mapObsPhotoToLocalSchema );
-  observation.observation_photos = observation?.observation_photos?.map( mapObsPhotoToLocalSchema );
+  // eslint-disable-next-line camelcase
+  const { observationPhotos, observation_photos } = observation;
+  if ( observationPhotos != null ) {
+    observation.observationPhotos = observationPhotos.map( mapObsPhotoToLocalSchema );
+  }
+  // eslint-disable-next-line camelcase
+  if ( observation_photos != null ) {
+    // eslint-disable-next-line camelcase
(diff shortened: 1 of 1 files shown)


src/api/observations.js
```diff
@@ -15,8 +15,16 @@ function mapObsPhotoToLocalSchema( obsPhoto ) {
...
 function mapToLocalSchema( observation ) {
-  observation.observationPhotos = observation?.observationPhotos?.map( mapObsPhotoToLocalSchema );
-  observation.observation_photos = observation?.observation_photos?.map( mapObsPhotoToLocalSchema );
+  // eslint-disable-next-line camelcase
+  const { observationPhotos, observation_photos } = observation;
+  if ( observationPhotos != null ) {
+    observation.observationPhotos = observationPhotos.map( mapObsPhotoToLocalSchema );
+  }
+  // eslint-disable-next-line camelcase
+  if ( observation_photos != null ) {
```
FLGMwt, line 24: This is prior art, but I'm confused why we would be handling `observationPhotos` _at all_ here. This function isn't exported and it's only used as a mapping function _directly_ after an api response. Aren't we _only_ ever going to have `observation_photos`?

If I'm missing something and we _would_ be consuming this, I don't think we ever set `observationPhotos` to null so this would be a functional change switching the `?.` falsy check to a null check so we may no longer catch a previous case.

FLGMwt: More clearly: I don't think this line will ever get hit: https://github.com/inaturalist/iNaturalistReactNative/pull/3592/changes#diff-ca2b2d3b0a1a3f2709eb943b66d1d2a07d1d23ac06d21dd4358af3a20b54e372R21

jtklein: I actually don't know, I can only assume Ken-ichi knew why he was adding this snake and camel at the time. Maybe some inaturalisjs thing? But maybe better to change only when certain why camel case is here, I'll close this for now.

PR #3597, opened 2026-05-06
Remove TODO for searchObservations in useLinking

jtklein:
Closes MOB-1358

Also migrate useLinking to TS. One note, since this hook is imported in App.js between the NavigationContainer and the RootStack I don't know how to type navigation in the hook.

### DIFF
--- a/src/components/hooks/useLinking.ts
+++ b/src/components/hooks/useLinking.ts
@@ -38,7 +38,7 @@ const useLinking = ( currentUser?: object ) => {
     }
   }, [navigation, observationId] );
 
-  const checkAllowedHosts = useCallback( ( url: string ) => {
+  const checkAllowedHosts = useCallback( ( url: string | null ) => {
     if ( typeof url !== "string" ) { return; }
     const { host, pathname } = new URL( url );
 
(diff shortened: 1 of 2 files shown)


src/components/hooks/useLinking.ts
```diff
@@ -12,9 +10,9 @@ import { Linking } from "react-native";
 const newAccountConfirmedUrl = "https://www.inaturalist.org/users/sign_in?confirmed=true";
 const existingAccountConfirmedUrl = "https://www.inaturalist.org/home?confirmed=true";
 
-const useLinking = ( currentUser: ?Object ) => {
+const useLinking = ( currentUser?: object ) => {
```
FLGMwt, line 13: maybe not appropriate for a ts-only migration but this does happen to only have one consumer.

We're _only_ using `currentUser` to check `isAuthenticated`. we can make this a flag to get rid of the `object` typing TODO.

Here and in other parts of the codebase where we string `currentUser` for the same purpose (this is very prevalent), it has the additional benefit of being a primitive value. This means that we wouldn't ever have to care about the memoization of it across function / prop boundaries

FLGMwt (approved):

PR #3598, opened 2026-05-07
Fix: sounds getting deleted during upload

jtklein:
Closes MOB-1400

Source of the bug was that when an upload errored out after upload of the sound was successfull but attaching the sound to the observations was not, we marked the observation_sound as synced. This was in contrast to photos were RealmPhoto is responsible for holding state wether the upload has succeeded and RealmObservationPhoto is responsible for holding state wether the attachment has succeeded. So, I refactored the upload process for sounds to be the same as photos which I think also removes some complexity of differing flows. Also migrated sound models to TS closely following the photo models.

Composer2 Summary:
Sound evidence no longer treats `inatjs.sounds.create` like an ObservationSound Realm update. Blob upload is modeled like photos: Sound carries the standalone media upload + server id + _synced_at, while ObservationSound is used for observation_sounds.create (attach) and stays “pending” until that step succeeds. That avoids marking ObservationSound as synced during file upload alone, which could drop sounds from retries when attach hadn’t finished.


src/realmModels/ObservationSound.ts
```diff
@@ -21,14 +29,15 @@ class ObservationSound extends Realm.Object {
     return this._synced_at !== null;
   }
 
-  static async new( observationSound ) {
+  static async new( uri: string ) {
+    const sound = await Sound.new( uri );
```
jtklein, line 33: cf ObservationPhoto

src/realmModels/Sound.ts
```diff
@@ -38,47 +39,47 @@ class Sound extends Realm.Object {
...
+    srcPath: string,
+    options: { basename: string },
+  ) {
     const srcFileName = srcPath.split( "/" ).at( -1 );
     const srcFileExt = srcFileName.split( "." ).at( -1 );
-    let fileName = srcFileName;
-    if ( options.basename ) {
-      fileName = `${options.basename}.${srcFileExt}`;
-    }
+    const fileName = `${options.basename}.${srcFileExt}`;
```
jtklein, line 48: The only call site of this function was not passing this in optionally.

src/realmModels/Sound.ts
```diff
@@ -38,47 +39,47 @@ class Sound extends Realm.Object {
...
+  static getLocalSoundUri( localPathOrUri?: string ) {
     const pieces = localPathOrUri?.split( "soundUploads/" );
     if ( !pieces || pieces.length <= 1 ) return null;
     return `file://${soundUploadPath}/${pieces[1]}`;
   }
 
-  static deleteSoundFromDeviceStorage( path ) {
-    unlink( path );
+  static deleteSoundFromDeviceStorage( path: string ) {
+    const localSound = Sound.getLocalSoundUri( path );
```
jtklein, line 81: cf Photo

src/uploaders/dataTransformation/prepareMediaForUpload.ts
```diff
@@ -8,10 +8,11 @@ function prepareMediaForUpload(
   action: ActionType,
   observationId?: number | null,
 ): object {
-  if ( type === "Photo" || type === "ObservationPhoto" ) {
+  if ( type === "Photo" ) {
```
jtklein, line 11: This is not a functional change, only for clarity see commit description.

src/uploaders/mediaUploader.ts
```diff
@@ -201,8 +203,21 @@ const filterMediaForUpload = ( observation: RealmObservation ): {
     ? observation.observationSounds.filter( item => !item.wasSynced( ) )
     : [];
 
+  // extract the actual sound objects
+  const unsyncedSounds = unsyncedObservationSounds?.map( os => {
```
jtklein, line 207: This is the main part of the bugfix, we handle uploading sounds the same as photos now with a new source of truth wether sth is synced successfully in the RealmSound model instead of RealmObservationSound.

jtklein (commented):
Few explanations:

src/realmModels/ObservationSound.ts
FLGMwt: TIL `cf`, thank you!

src/realmModels/ObservationSound.ts
```diff
@@ -37,36 +46,41 @@ class ObservationSound extends Realm.Object {
...
       ? "mp4"
       : "m4a";
 
     return {
-      uuid: observationSound.uuid,
       file: new FileUpload( {
-        uri: Sound.getLocalSoundUri( observationSound.sound.file_url ),
-        name: `${observationSound.uuid}.${fileExt}`,
+        uri,
+        name: uri?.split( "/" ).pop( ),
```
FLGMwt, line 58: That's nifty, I haven't seen that done before

src/uploaders/mediaUploader.ts
FLGMwt: That makes a lot of sense, good find.

FLGMwt (approved):

PR #3599, opened 2026-05-07
do not fail lint-staged on warnings

FLGMwt:
We intentionally introduced warnings in #3574. I missed that we _were_ preventing commits on warnings, we just happened to be warning free. Team consensus on that issue was that we desired the warnings. Discussion w/ @jtklein this week confirmed that we feel safe _allowing_ warnings to exist and not preventing commits.


jtklein (approved):
Thanks! LGTM

PR #3602, opened 2026-05-08
Android save match crash

sepeterson:
Wait for the exif writes to settle before navigating away


jtklein (approved):
Thanks for fixing this, and adding an error log is a good idea as well.

PR #3603, opened 2026-05-08
MOB-1321 minimal

sepeterson:
Different approach to [this PR ](https://github.com/inaturalist/iNaturalistReactNative/pull/3577)that maintains useWatchPosition functionality while handling the coarse location only path in a wrapper


src/components/SharedComponents/PermissionGateContainer.tsx
```diff
@@ -91,16 +94,37 @@ interface Props extends PropsWithChildren {
...
     throw new Error(
       "permissionResultFromMultiple received something other than an object. "
       + "Make sure you're using it with checkMultiple and not check",
     );
   }
+  // On Android 12+, the user may grant only approximate (coarse) location,
+  // leaving fine location denied. If ANY location permission is granted,
+  // the overall result is GRANTED.
+  const coarseKey = PERMISSIONS.ANDROID.ACCESS_COARSE_LOCATION;
+  if ( coarseKey in multiResults ) {
```
sepeterson, line 112: Copying this comment from the last PR: 

> I want to acknowledge that this is a little goofy but that I prefer it to the alternatives of setting up a new function or adding some sort of optional boolean. This function is used in two pretty different contexts and the updates to the calling sites would get messy

PR #3604, opened 2026-05-08
add worklets patch for extra react native host preventing profiling

FLGMwt:
The Profiling Dev Tools tab is disabled with message "Multiple React Native hosts not supported". This is related to us seeing two duplicate attach options when launching the dev tools through the dev server.

Issue: https://github.com/react-native-community/discussions-and-proposals/discussions/954

Comment w/ patch:https://github.com/react-native-community/discussions-and-proposals/discussions/954#discussioncomment-16707157

As I understand, this change only affects users debugging _against_ worklets itself.

With this, I am able to start and stop a profiling session and get a result.

Pretty much anything "useful", even a couple taps in under 5s results in the dev tools hanging and becoming unresponsive. I discovered that it _eventually_ completes processing, even if the device's connection with dev tools times out. Once it's done, you get the pop up of "dev tools disconnected. Reconnect? Dismiss?" <- you have to dismiss to get to the actual results or you'll lose the profiling session.

EDIT: it took a while for me to trust that it actually eventually worked consistently. I now trust that it does, one just has to be patient.

Would welcome anyone trying this out!

fixes MOB-1407


sepeterson (approved):
Had trouble with this the other day. Much appreciated!

jtklein (approved):
This works on my end as well. (Also Android still working after the PR)

PR #3605, opened 2026-05-09
Basic news screen

jtklein:
This is a PR to prepare for MOB-1312.
Here I was able to salvage a few pieces from a creative day pursuit: A news screen behind a feature flag.
I did add an API call to get posts and display a list as well.


src/api/posts.ts
```diff
@@ -0,0 +1,27 @@
...
+import handleError from "api/error";
+import inatjs from "inaturalistjs";
+import Config from "react-native-config";
+
+const fetchUserPosts = async (
+  params: Record<string, unknown> = {},
+  opts: Record<string, unknown> = {},
+): Promise<Record<string, unknown> | null | ErrorWithResponse | INatApiError> => {
+  try {
+    inatjs.setConfig( { apiURL: "https://api.inaturalist.org/v1" } );
```
jtklein: This is a bit tricky here. When using v2 for this endpoint I get an error that GET is not supported here. So, until checking back with the api team, I think there is just no v2 version of this available yet. I also did not see an endpoint call only options override for apiURL, like POST endpoints for example allow in inaturalistjs.
So, I am switching all of inatjs to v1 and back, which I think as long as this is behind a feature flag is not too bad.
Changing this to a supported v2 endpoint is a blocker for release to me, so for me this can stay like that until we switch over.

sepeterson: Ya that's fine w/ me since it's behind a feature flag. Maybe we can add a tracking ticket?

src/components/News/News.tsx
```diff
@@ -0,0 +1,36 @@
...
+    return (
+      <View className="flex-1 bg-white">
+        <ActivityIndicator size={50} />
+      </View>
+    );
+  }
+
+  return (
+    <View className="flex-1 bg-white">
+      <PostList posts={posts} />
```
sepeterson, line 57: Seeing a ts error on the `posts` prop here

`Type 'Record<string, unknown> | ErrorWithResponse | INatApiError | null | undefined' is not assignable to type 'ApiPost[]'.`

src/components/News/PostListItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+  if ( !item ) {
+    return null;
+  }
+
+  const excerpt = item.body
+    ? stripHtmlForPreview( item.body )
+    : "";
+
+  let bodyNumberOfLines = 3;
+  if ( item.title.length > 70 ) {
```
sepeterson, line 37: `item.title` is possibly undefined according to ts here

src/components/News/PostList.tsx
```diff
@@ -0,0 +1,33 @@
...
+const ItemSeparator = () => <View className="border-b border-lightGray" />;
+
+interface Props {
+  posts: ApiPost[];
+}
+
+const PostList = ( {
+  posts,
+}: Props ) => {
+  const renderPost: ListRenderItem<ApiPost> = ( { item } ) => (
```
sepeterson, line 18: Absolutely a nit but this could be a `useCallback` or even lifted out of the component like ItemSeparator above

sepeterson (approved):
Looks great! Up to you if you'd like to address any of the comments or if they'll be irrelevant with later code

src/api/posts.ts
FLGMwt: This `setConfig` isn't safe as is. We should at least move it to `finally` or probably better, fire off the `for_user` promise then immediately set it back before awaiting the promise.

Since this is a one-off for now, I'd feel more comfortable if we skipped using `inatjs` altogether and constructed the v1 request manually. That way, we can be sure that we're not interfering with the "normal" consumers of inatjs and its global config.

src/components/News/PostList.tsx
FLGMwt: FlashList's `renderItem` _does_ actually behave differently than `ItemSeparatorComponent`. For optimization, even if it gets a valid "React.Component", it's treated as an element-returning function. As opposed to ItemSeparatorComponent which calls React.createElement on Components.

I think callback is still valid here though.

But that's a nit followup to a nit ; D

src/components/News/PostListItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+} from "components/SharedComponents";
+import {
+  View,
+} from "components/styledComponents";
+import React from "react";
+import { formatDifferenceForHumans } from "sharedHelpers/dateAndTime";
+import { useTranslation } from "sharedHooks";
+
+function stripHtmlForPreview( body: string ): string {
+  return body.replace( /<[^>]+>/g, " " ).replace( /\s+/g, " " ).trim( );
```
FLGMwt: Personal blanket request for regexes: could we add a comment describing what this does and a before/after example

FLGMwt (changes requested):
Awesome! Excited to see a new FeatureFlag, not to mention a new iOS parity feature coming into the mix!

Requesting changes on the inatjs api url config.

src/components/News/PostList.tsx
jtklein: The way it is now follows the flashlist documentation. Where it is also recommended ti measure if adding a useCallback is making performance better before adding it. So, I will leave as is now.

PR #3609, opened 2026-05-11
reduce rework in use local observations, improve locking UI thread

FLGMwt:
MyObsContainer and Notifications call useLocalObservations. This hook returns Realm Observations mapped to pojos. It attaches listeners to the realm.objects( "Observation" ) collection to keep this list up to date. Aside: each use of hook maintains its own state / change handlers.

As far as mapping goes, this is a pretty expensive due to Realm objects being "live" managed objects (sort of like js Proxies).  

With the exception of insertion-only, this function was reacting to changes by remapping the entire collection. This becomes an issue because `realm.objects( "Observation" )` grows as the app is used: every time an Obs is loaded from the API, including during infinite scroll, it is added to Realm, increasing the amount of work this function needs to do.

**This change:**

instead of rebuilding and remapping all entities on handleChange, this change uses the previously built list and `newModifications` to only remap what is necessary. Note that additions and deletions are handled implicitly.

Other opportunities:
- consolidating the state / listeners so we're not doing duplicate work in notifs & myobs
- improve and/or defer initial build. we currently lock the UI thread with this so we may need to manually chunk it to
- reduce reliance on a full build at all
- garbage collect `realm.objects( "Observation" )`. This may be an opportunity w/ MyObs search work.

closes MOB-1384 (more tickets to come)

### DIFF
--- a/src/sharedHooks/useLocalObservations.js
+++ b/src/sharedHooks/useLocalObservations.js
@@ -75,12 +75,14 @@ const useLocalObservations = ( ): Object => {
           // Mode change requires full remap
           mappedObservations = Array.from( filteredObservations )
             .filter( obs => obs.isValid() )
-            .map( obs => mapObservation( obs, isDefaultMode ) );
+            .map( mapObservation );
         } else {
-          const modifiedUuids = newModifications
-            .map( index => localObservations[index] )
-            .filter( obs => obs?.isValid() )
-            .map( obs => obs.uuid );
+          const modifiedUuids = new Set(
+            newModifications
+              .map( index => localObservations[index] )
+              .filter( obs => obs?.isValid() )
+              .map( obs => obs.uuid ),
+          );
 
           const previousObsByUuid = Object.fromEntries(
             prevListRef.current.list.map( obs => [obs.uuid, obs] ),
@@ -89,7 +91,7 @@ const useLocalObservations = ( ): Object => {
           mappedObservations = Array.from( filteredObservations )
             .filter( obs => obs.isValid() )
             .map( obs => {
-              if ( modifiedUuids.includes( obs.uuid ) ) {
+              if ( modifiedUuids.has( obs.uuid ) ) {
                 return mapObservation( obs, currentIsDefaultMode );
               }
               return previousObsByUuid[obs.uuid] ?? mapObservation( obs, currentIsDefaultMode );
--- a/src/sharedHooks/useLocalObservations.js
+++ b/src/sharedHooks/useLocalObservations.js
@@ -7,6 +7,7 @@ import {
 } from "react";
 import Observation from "realmModels/Observation";
 import useStore from "stores/useStore";
+import { v4 as uuidv4 } from "uuid";
 
 function isDefaultMode( ) {
   return useStore.getState( ).layout.isDefaultMode === true;
@@ -18,12 +19,6 @@ const deletionFilters
   = "_deleted_at == nil OR _pending_deletion == false OR _pending_deletion == nil";
 
 const sortedFilters = [["needs_sync", true], ["_created_at", true]];
-
-const mapObservations = ( observations, isDefault ) => ( isDefault
-  ? observations
-    .map( observation => Observation.mapObservationForMyObsDefaultMode( observation ) )
-  : observations
-    .map( observation => Observation.mapObservationForMyObsAdvancedMode( observation ) ) );
 
 const useLocalObservations = ( ): Object => {
   const setNumUnuploadedObservations = useStore( state => state.setNumUnuploadedObservations );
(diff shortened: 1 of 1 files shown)


src/sharedHooks/useLocalObservations.js
```diff
@@ -71,23 +65,37 @@ const useLocalObservations = ( ): Object => {
...
+
+          const previousObsByUuid = Object.fromEntries(
+            prevListRef.current.list.map( obs => [obs.uuid, obs] ),
+          );
+
+          mappedObservations = Array.from( filteredObservations )
+            .filter( obs => obs.isValid() )
+            .map( obs => {
+              if ( modifiedUuids.has( obs.uuid ) ) {
+                return mapObservation( obs, currentIsDefaultMode );
```
jtklein: mapObservation no longer needs currentIsDefaultMode

src/sharedHooks/useLocalObservations.js
```diff
@@ -19,12 +19,6 @@ const deletionFilters
...
 
-const mapObservations = ( observations, isDefault ) => ( isDefault
-  ? observations
-    .map( observation => Observation.mapObservationForMyObsDefaultMode( observation ) )
-  : observations
-    .map( observation => Observation.mapObservationForMyObsAdvancedMode( observation ) ) );
-
 const useLocalObservations = ( ): Object => {
   const setNumUnuploadedObservations = useStore( state => state.setNumUnuploadedObservations );
   // Use refs to maintain state without triggering re-renders of hook consumers
```
jtklein: This comment seem irrelevant

src/sharedHooks/useLocalObservations.js
```diff
@@ -71,23 +65,37 @@ const useLocalObservations = ( ): Object => {
...
-            .filter( o => o.isValid() );
+        const mapObservation = currentIsDefaultMode
+          ? Observation.mapObservationForMyObsDefaultMode
+          : Observation.mapObservationForMyObsAdvancedMode;
 
-          const newMappedObservations = mapObservations( newObservations, currentIsDefaultMode );
+        let mappedObservations;
 
-          mappedObservations = [...prevListRef.current.list, ...newMappedObservations];
+        if ( currentIsDefaultMode !== prevListRef.current.isDefaultMode ) {
```
jtklein: When I am logged in as a default user, and press the "Switch to Advanced Mode" link in the new banner a breakpoint in here is not immediately stopping. So if you say the mode change "requires" a full remap, we need to add `isDefaultMode` as a dependency to the hook. As symptom I can see that after switching to Advanced Mode every obs show "No location" and "Missing date" in obs list view.
The remap does eventually get triggered by scrolling in the list, and then I do see a sudden change in UI at this point which is undesired.
Advanced mode to default doesn't seem problematic because pressing the Me tab away from settings resets the nav stack and remounts this hook I suppose.

src/sharedHooks/useLocalObservations.js
```diff
@@ -71,23 +65,37 @@ const useLocalObservations = ( ): Object => {
...
+            .filter( obs => obs.isValid() )
+            .map( mapObservation );
         } else {
-          // Full rebuild needed (modifications, deletions, or mode change)
-          const validObservations = Array.from( filteredObservations ).filter( o => o.isValid() );
-
-          mappedObservations = mapObservations( validObservations, currentIsDefaultMode );
+          const modifiedUuids = new Set(
+            newModifications
+              .map( index => localObservations[index] )
```
jtklein, line 78: Doc comment to Collection.addListener states: "@note
deletions and oldModifications report the indices in the collection before the change happened, while insertions and newModifications report the indices into the new version of the collection." The naming of things is a bit confusing here maybe. I would expect you have to get the newModifications from `collection` here and not from `localObservations`. But I can see above that we already do the filteredObservations from localObservations.
I guess my question is, if a handleChange get's triggered, the `localObservation` is already updated with the deletions removed, insertions added, and modifications pointing to the new indices.

src/sharedHooks/useLocalObservations.js
```diff
@@ -71,23 +65,37 @@ const useLocalObservations = ( ): Object => {
...
+          mappedObservations = Array.from( filteredObservations )
+            .filter( obs => obs.isValid() )
+            .map( mapObservation );
         } else {
-          // Full rebuild needed (modifications, deletions, or mode change)
-          const validObservations = Array.from( filteredObservations ).filter( o => o.isValid() );
-
-          mappedObservations = mapObservations( validObservations, currentIsDefaultMode );
+          const modifiedUuids = new Set(
+            newModifications
```
jtklein, line 77: When I put a breakpoint here, I think I am still seeing way to many invocations of this. Sometimes tens to hundreds of newModifications several in a row. (I would not expect any modifications server-side from staging that much).
I would feel much more comfortable if we have 1) an integration test, and 2) instrumentation around this crucial hook. Also, to see if this PR is a step in the right direction.
That being said, I do not see a lot of stops at the same breakpoint that have only insertions and so would have fallen into the deleted if clause from before. So, all in all just looking at the code I do think we will be doing less work after because we have to map smaller arrays.
Anyway, I also checked main just now, and I also see a lot of invocations there with hundreds of modifications.

src/sharedHooks/useLocalObservations.js
```diff
@@ -71,23 +65,35 @@ const useLocalObservations = ( ): Object => {
...
-          const newMappedObservations = mapObservations( newObservations, currentIsDefaultMode );
+        let mappedObservations;
 
-          mappedObservations = [...prevListRef.current.list, ...newMappedObservations];
+        if ( currentIsDefaultMode !== prevListRef.current.isDefaultMode ) {
+          // Mode change requires full remap
+          mappedObservations = Array.from( filteredObservations )
+            .filter( obs => obs.isValid() )
+            .map( obs => mapObservation( obs, isDefaultMode ) );
         } else {
```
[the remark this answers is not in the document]
sepeterson: Looks like there's two other usages of mapObservation where you can get rid of the 2nd arg

jtklein (changes requested):
I have a suspicion that overall I concur with your believe that we will be seeing performance benefits from this. Looking at the code there is potential for this flow to cause less work than before keeping exiting functionality as far as I can see. However, I do not see any data confirming this claim. Could there be user for which by far the most filteredObservations will trigger a mapObservation, and they also have the additional work of creating the Set and lookup table? I don't know, I don't see any data here. I am seeing a lot of breakpoints triggering with many more insertions + modifications than I would have expected from my user behaviour while testing.
Overall, this could have benefited from a standalone TS migration PR first.

There is one visible UI issue which is a requested change: Obs list is not updating immediately when switching to Advanced mode now, see inline comment.

I do feel a bit iffy about pushing this out as an experiment in production without having an integration test that tells me that there aren't too many invocations happening now, and some instrumentation of are we actually making things better here for most users.

sepeterson (approved):
Code looks great and I was able to do some light smoke testing locally

src/sharedHooks/useLocalObservations.js
FLGMwt: incomplete refactor, thanks!

src/sharedHooks/useLocalObservations.js
FLGMwt: removed

src/sharedHooks/useLocalObservations.js
FLGMwt: Good catch, I should have toggled back in testing. Switched to use zustand and add to dep array. Pushed, but I still need to test this.

src/sharedHooks/useLocalObservations.js
sepeterson: >      * @example
>      * wines.addListener((collection, changes) => {
>      *  // collection === wines
>      *  console.log(`${changes.insertions.length} insertions`);
>      *  console.log(`${changes.oldModifications.length} oldModifications`);
>      *  console.log(`${changes.newModifications.length} newModifications`);
>      *  console.log(`${changes.deletions.length} deletions`);
>      *  console.log(`new size of collection: ${collection.length}`);
>      * });

`collection === localObservations ` so they can be used interchangeably but maybe using `collection` is more readable

src/sharedHooks/useLocalObservations.js
FLGMwt: modifications here doesn't necessarily mean anything has changed server-side, just that the Obs in realm has been written to. For example, `markViewedLocally` writes to a RealmObs which triggers an API call which fetches latest remote and rewrites to a RealmObs.

I should have been more explicit that I was _not_ intending to address that here. Since this is crucial, to your point, I'm trying to not touch external behavior and just improve this mapping for the vast majority of unnecessary cases.

I expect there's a lot more triggers / modifications of this handleChange _and_ more individual "unexpected" remaps happening than just the obs that triggered the collection onChange.

To that point:
> Anyway, I also checked main just now, and I also see a lot of invocations there with hundreds of modifications.

_just speaking to what Realm is handing us_, did you notice any concerning differences between the number of handleChange calls or number of items in modifications?

src/sharedHooks/useLocalObservations.js
FLGMwt: I agree, I should be extra certain about this, I'll spend some time this evening and follow up.

FLGMwt: Thank you both for pointing to the jsdoc! With that and reviewing the documentation again, I feel confident in saying:

- `collection` and `localObservations` in _all_ cases are equivalent and not subject to closure staleness in this instance because of their "live" Realm behavior. In this change, I was following the previous dev's preference to just use one reference for clarity. This ought to be explicit and documented.
  - **Outcome**: I will rename `collection` to `_collection`. Given both references are equivalent, we will use the reference available to both containing scopes. I will add a comment clarifying this.
- `newModifications` has the correct indexes for `collection`/`localObs` based on their contemporaneously-relevant values and `collection`/`localObs` are both contemporaneously correct. We would only want to use `oldModifications` if we were referencing a context _external to realm_ where the the index of the item in the _pre-change copy_ had significance. Such as if we stored a manual cache of non-realm values that changed outside of realm. We're using this data to _inform_ a local cache of non-realm values but we're not depending on it for consuming the indexes.

FLGMwt:
I shared some proper test results in iNat chat, but posting some here for public posterity. After this change, this user would face 3 fewer long periods of UI-blocking unresponsiveness. The exact timing ("how many ms saved?")is not reliable because of development mode / profiling, but the impact is demonstrable.

Test: 
- be a user w/ 10k obs
- load ~1k obs into Realm by scrolling down on myobs a bunch
- close the app
- launch the app
- start perf trace
- navigate to myobs
- wait 5s
- stop perf trace

Before:
Four calls to handleChange lasting ~700ms
Screenshot of rn performance dev tools highlighting 4 function calls lasting ~700ms

After:
One call to handleChange lasting ~700ms, three lasting 30-50ms
Screenshot of rn performance dev tools highlighting 1 function call lasting ~700ms, three lasting 30-50ms

PR #3611, opened 2026-05-12
don't use problematic useLocalObservations in Notifications

FLGMwt:
closes MOB-1420

same concern as #3609. The observation Realm=>pojo mapping that `useLocalObservations` handles is hook-local so both consumers were doing their own expensive mapping.

For users with a non-trivial amount of locally loaded Observations, the app freezes multiple times during upload. The problem gets worse over time as the Realm obs collection grows. 

If a user ever navigates to the Notifications tab, for the reset of the app session, the freezes are twice as long because we are attaching a second mapping listener.

The solution here is to just not use it. We just need to know this is a new user w/ <10 obs to see if we need to prompt them w/ an onboarding modal. We don't need the pojo list, we can just query realm directly.

### DIFF
--- a/src/components/Notifications/Notifications.tsx
+++ b/src/components/Notifications/Notifications.tsx
@@ -1,11 +1,10 @@
 import { NotificationOnboarding } from "components/OnboardingModal/PivotCards";
 import { Tabs } from "components/SharedComponents";
 import { View } from "components/styledComponents";
+import { RealmContext } from "providers/contexts";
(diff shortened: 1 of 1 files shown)


jtklein (approved):
Have tested and the onboarding modal is still showing for a new user with only 2 observations in default mode.
This is probably also a good reminder to not import hooks if we only need a fraction of the functionality of the hook.

PR #3613, opened 2026-05-12
MOB-1396: disable suggest id buttons while pending

sepeterson:
Use the loading state from `useAuthenticatedMutation` to show a loading spinner on buttons until the suggestion is successful.

Implemented according to this in the descripton "We should only allow a user to submit an ID once, and then show a loading wheel on the button to indicate it has been tapped and the ID is being submitted."

But now I'm wondering if we shouldn't just do a debounce thing. Lmk what you think!

  <details>                                                                                                            
  <summary>videos showing new behavior</summary>
                                                                                                                       
https://github.com/user-attachments/assets/f797839e-5222-40f1-aca4-88b53c5c0652

https://github.com/user-attachments/assets/dbaef0ba-c707-4864-8600-8c9ca71da709 

  </details>

### DIFF
--- a/src/components/ObsDetailsSharedComponents/Sheets/SuggestIDSheet.tsx
+++ b/src/components/ObsDetailsSharedComponents/Sheets/SuggestIDSheet.tsx
@@ -15,6 +15,7 @@ interface Props {
     body?: string;
     taxon: { id: number };
   };
+  loading?: boolean;
   onSuggestId: ( ) => void;
   editIdentBody: () => void;
   onPressClose?: () => void;
@@ -23,6 +24,7 @@ interface Props {
 const SuggestIDSheet = ( {
   hidden,
   identification,
+  loading,
   onSuggestId,
   editIdentBody,
   onPressClose,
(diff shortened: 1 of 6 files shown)


src/components/ObsDetailsSharedComponents/Sheets/SuggestIDSheet.tsx
```diff
@@ -66,6 +68,7 @@ const SuggestIDSheet = ( {
                 editIdentBody( );
               }}
               className="mx-2 flex-1"
+              disabled={loading}
```
jtklein, line 71: Also a good idea to disable this while ID is loading.

jtklein (approved):
Our Button component also has a debounce built in already for rapid succession taps. But since we start an API mutation round trip here, the button debounce is too short to be meaningful. Adding a loading state is the correct solution I believe, and you don't need to add another debounce add sheet level.

Great PR, good you also changed the Comment buttons to be disabled while an ID is loading.

PR #3616, opened 2026-05-12
MOB-1321 minimal

sepeterson:
Different approach to https://github.com/inaturalist/iNaturalistReactNative/pull/3577 that maintains useWatchPosition functionality while handling the coarse location only path in a wrapper

### DIFF
--- a/src/components/Match/MatchContainer.tsx
+++ b/src/components/Match/MatchContainer.tsx
@@ -27,9 +27,10 @@ import fetchPlaceName from "sharedHelpers/fetchPlaceName";
 import saveObservation from "sharedHelpers/saveObservation";
 import shouldFetchObservationLocation from "sharedHelpers/shouldFetchObservationLocation";
 import {
-  useExitObservationFlow, useLocationPermission, useSuggestions, useWatchPosition,
+  useExitObservationFlow, useLocationPermission, useSuggestions,
 } from "sharedHooks";
 import useDebugMode from "sharedHooks/useDebugMode";
+import useObservationLocation from "sharedHooks/useObservationLocation";
 import {
   internalUseSuggestionsInitialSuggestions,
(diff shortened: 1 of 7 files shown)


src/components/Match/MatchContainer.tsx
```diff
@@ -337,8 +341,6 @@ const MatchContainer = ( ) => {
     refetchSuggestions,
     scoreImageParams,
     scrollToTop,
-    currentUserLocation?.latitude,
-    currentUserLocation?.longitude,
```
FLGMwt, line 341: nice

src/sharedHooks/useObservationLocation.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    return ( ) => { cancelledRef.current = true; };
+  }, [shouldFetchLocation] );
+
+  useEffect( ( ) => {
+    const unsubscribe = navigation.addListener( "blur", ( ) => {
+      cancelledRef.current = true;
+      setIsFetchingCoarse( false );
+      setCoarseLocation( null );
+      setIsCoarseOnly( null );
```
FLGMwt, line 49: I grok the lifecycle and role of each of these, but it does take some thinkering.

Not for this case, but maybe one complexity higher, I would probably vote for a reducer with discrete lifecycle states.

FLGMwt (approved):
code review LGTM. goooood call on the cancel ref and thanks for the tests!

Anything I should focus on testing on device?

sepeterson:
> Anything I should focus on testing on device?

The main flow that gave me trouble was Fresh Install -> AI Camera -> Match Screen -> Add Location to observation -> grant approximate permission only

src/sharedHooks/useObservationLocation.ts
sepeterson: Ya totally agree!

PR #3622, opened 2026-05-14
add ExploreV2 debug sheet

sepeterson:
DEBUG ONLY so don't spend your precious time on it

As suggested here: https://github.com/inaturalist/iNaturalistReactNative/pull/3585#issuecomment-4422706923

I didn't put it behind a dedicated feature flag since it'll be obsolete after we get UniversalSearch implemented.

I also am fine if we'd rather not merge it. It's so self-contained that it's simple enough to just pull it in for testing


jtklein (approved):
We already have a debug button in the V1, so if and when we want to port it over the two buttons might conflict, but we can deal with it then.
Do you want to put the pink button behind the isDebug flag, like we have the V1 button?

sepeterson:
> Do you want to put the pink button behind the isDebug flag, like we have the V1 button?

Ya good idea!

PR #3624, opened 2026-05-14
Infinite explore ts conversion

sepeterson:
Helpful ts conversion supporting ExploreV2

### DIFF
--- a/src/components/Explore/hooks/useInfiniteExploreScroll.ts
+++ b/src/components/Explore/hooks/useInfiniteExploreScroll.ts
@@ -1,3 +1,4 @@
+import type { QueryStatus } from "@tanstack/react-query";
 import { useQueryClient } from "@tanstack/react-query";
 import { searchObservations } from "api/observations";
 import type {
@@ -31,7 +32,7 @@ interface UseInfiniteExploreScrollReturn {
   isFetchingNextPage: boolean;
   handlePullToRefresh: ( ) => Promise<void>;
   observations: ApiObservation[];
-  status: "pending" | "error" | "success";
+  status: QueryStatus;
   totalBounds: ApiTotalBounds | undefined;
   totalResults: number | null;
 }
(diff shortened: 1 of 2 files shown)


src/components/Explore/hooks/useInfiniteExploreScroll.ts
jtklein, line 93: I wonder why we are doing this here. We are not using removeQueries in any other infinite scroll list.

jtklein, line 80: Would be nice to have src/sharedHooks/useAuthenticatedInfiniteQuery.ts better typed before typing this. (Not requesting it here though)

src/components/Explore/hooks/useInfiniteExploreScroll.ts
```diff
@@ -64,13 +94,15 @@ const useInfiniteExploreScroll = ( { params: newInputParams, enabled }: Object )
     await refetch( );
   };
 
-  let observations = flatten( data?.pages?.map( r => r.results ) ) || [];
-  let totalResults = data?.pages?.[0].total_results;
+  const pages = data?.pages as ApiObservationsSearchResponse[] | undefined;
+
+  let observations: ApiObservation[] = flatten( pages?.map( r => r.results ) ) || [];
+  let totalResults: number | null | undefined = pages?.[0]?.total_results;
```
jtklein, line 100: All these optional chains, and possibilities of totalResults being null or undefined are just artefacts of our incomplete typing of using useQuery right?
In principle, we either have data or we don't.
If we don't we have to wait until we have or we have an error and need to show an error state to the user.
If we have data we have pages.
If pages is an empty array we have queried successfully but we have no data for it on the server.
I don't want to change functionality in this PR but at tone point we need to reconcile what we are doing here with the truth of possible returns from useQuery.

jtklein (approved):
Thanks for doing this. There is some uncertainty in what is the final stage of this because we haven't really typed useInfiniteQuery yet. We can already do a better job there because updating tanstack query has provided us with better typing to use since we first made this wrapper. But no need to increase scope here.

src/components/Explore/hooks/useInfiniteExploreScroll.ts
sepeterson: Ya I started on that in a branch after I pushed this -- may finish it next week.

sepeterson: Appears to be requirement of the paginated fetching. In useAuthenticatedInfiniteQuery we use a prebuilt `useInfiniteQuery` hook that presumably handles all that for us

src/components/Explore/hooks/useInfiniteExploreScroll.ts
sepeterson: Yes I'm on board with that. We could clean up a bunch of this with the the useAuthenticatedInfiniteQuery ts conversions

PR #3625, opened 2026-05-14
Reuse FollowersList as layout for news/journal screen

jtklein:
Closes MOB-1425

Everything behind a feature flag.
The new screen can be routed to from 1) user profile, 2) project details, and 3) menu (for "News").


src/i18n/l10n/en.ftl
```diff
@@ -150,6 +150,8 @@ attribution-cc-by-sa = some rights reserved (CC BY-SA)
 August = August
 # Returns user to login screen
 BACK-TO-LOGIN = BACK TO LOGIN
+Blog = Blog
+BLOG = BLOG
```
sepeterson, line 154: Is there reason to have a separate key for the all-caps version?

src/components/ProjectDetails/ProjectDetails.js
```diff
@@ -93,6 +95,13 @@ const ProjectDetails = ( {
     [navigation, project],
   );
 
+  const onJournalPostsPressed = ( ) => {
```
sepeterson, line 98: Okay *this* time I'm pretty sure this could be wrapped with useCallback since the other onPress functions are 😄

sepeterson (approved):
Lookin good!

src/i18n/l10n/en.ftl
jtklein: Yes, so that translators can translate the two texts differently, if required.

PR #3626, opened 2026-05-14
Only send user id and login via route params

jtklein:
We do not need anything else from `user` on the other side.

Removes two TODOs.

### DIFF
--- a/src/navigation/types.ts
+++ b/src/navigation/types.ts
@@ -241,12 +241,12 @@ export type BaseTabStackParamList = {
     userLogin?: string;
   };
   FollowersList: {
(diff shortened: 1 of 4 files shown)


src/navigation/types.ts
```diff
@@ -241,12 +241,12 @@ export type BaseTabStackParamList = {
     userLogin?: string;
   };
   FollowersList: {
-    // TODO: don't send the entire user object over here, only an ID or ID+login
```
FLGMwt, line 244: TODONE!

FLGMwt (approved):
LGTM!

I've still got the internal struggle of whether a deleted LoC or a deleted comment is better, but it's gotta be that a deleted _TODO_ comment is the best.

PR #3628, opened 2026-05-15
Fallback UI for hidden content

jtklein:
The easy part of MOB-203.
Closes MOB-1429.

Make sure we query the API for the `hidden` field for photos, sounds, IDs, comments.
Display a fallback UI in case any of these is supposed to be hidden.

In contrast to web or classic apps we only show a generic explanation that this content is hidden, and we do not show details of the ModeratorAction.

Will create a follow-up ticket to include the eye icon like classic does. [Edit] Have also added the wanted icon, so closes MOB-1428 as well.

Started to migrate to TS everything I touch but there are just too many children here that would increase the scope too much. I did leave two files in that should be straight-forward.


src/components/ObsDetailsSharedComponents/ActivityTab/FloatingButtons.tsx
```diff
@@ -15,17 +13,17 @@ const DROP_SHADOW = getShadow( {
   shadowOpacity: 0.2,
 } );
 
-type Props = {
-  navToSuggestions: Function,
-  showAddCommentSheet: Function,
-  openAddCommentSheet: Function
+interface Props {
+  navToSuggestions: ( ) => void;
+  showAddCommentSheet: boolean;
```
FLGMwt, line 18: Oh, looks like this accidental "Function" typing cross a few layers. Nice find.

src/sharedHooks/useLocalObservation.ts
```diff
@@ -63,23 +57,6 @@ const useLocalObservation = ( uuid: string ): UseLocalObservation => {
     }, "marking viewed locally in ObsDetailsContainer" );
   };
 
-  Object.defineProperties( observation, {
```
FLGMwt, line 66: Great removal! It rules this out as a sneaky side-effect culprit in mapping / passing around Realm obs.

FLGMwt (approved):
LGTM, very happy to remove those `defineProperty` calls on the realm `observation`.

FLGMwt:
> Started to migrate to TS everything I touch but there are just too many children here that would increase the scope too much. I did leave two files in that should be straight-forward.

Followup from our conversation about this since we both hit this boundary at the same time: MOB-1438 is slotted for this / next week to add a global / app-level `AppObservation` type representing this pojo separate from the existing Realm-related types or existing API types.

PR #3629, opened 2026-05-17
Observation flash list render defined component

FLGMwt:
This is an additional case of a class of issue discovered in #3383: `<FlashList HeaderComponent={() => <View />} />` or its equivalent of defining a `*Component` prop in render is Bad.

In this case, I believe it's causing `Announcements`, which loads dynamic html into a WebView on mount, to unnecessarily mount/remount (not just rerender).

UPDATE: 

1) After testing, that doesn't seem to be the case. This is still a desired change following the pattern of #3383.
2) Related we _are_ getting two calls to announcements whenever we render it, introduced when we added unauth'd announcements. This is due to an issue with how we check "is user authenticated". MOB-1431 to track this

----

Looking at this class of issue again here, I'm realizing now that we should probably _never_ pass a `useCallback`'d prop to one of FlashList's `*Component` props. At best, with no props, it's identical to passing a useMemo()'d Element. W/ props, it's strictly worse because we mount/unmount instead of just re-rendering. I think we can also _test_ removing the memos as well. They probably aren't having the impact we think they're having and they complicate the code.

^not suggesting that immediately but mentioning while we're here. For this PR, I kept us to using the pattern of useMemo'd elements we've stuck to since #3383.

### DIFF
--- a/src/components/ObservationsFlashList/ObservationsFlashList.tsx
+++ b/src/components/ObservationsFlashList/ObservationsFlashList.tsx
@@ -34,6 +34,9 @@ const { useRealm } = RealmContext;
 
 const AnimatedFlashList = Animated.createAnimatedComponent( CustomFlashList );
 
+// unlike the other FlashList Component props, Separator does _not_ accept an Element
+const ItemSeparator = () => <View className="border-b border-lightGray" />;
+
 interface Props {
   contentContainerStyle?: StyleProp<ViewStyle>;
   // TODO: type data / observations
@@ -198,11 +201,11 @@ const ObservationsFlashList = ( {
     uploadQueue,
   ] );
 
-  const renderItemSeparator = useCallback( ( ) => {
+  const itemSeparatorComponent = useMemo( ( ) => {
     if ( layout === "grid" ) {
       return null;
     }
-    return <View className="border-b border-lightGray" />;
+    return ItemSeparator;
   }, [layout] );
 
   const renderFooter = useCallback( ( ) => (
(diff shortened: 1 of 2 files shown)


src/components/ObservationsFlashList/ObservationsFlashList.tsx
```diff
@@ -32,37 +34,42 @@ const { useRealm } = RealmContext;
...
+  hideRGLabel?: boolean;
+  isConnected: boolean;
+  layout: "list" | "grid";
+  obsListKey: string;
+  onEndReached: () => void;
+  onLayout?: ( event: LayoutChangeEvent ) => void;
+  onScroll?: ( event: NativeSyntheticEvent<NativeScrollEvent> ) => void;
   // this ref is being forwarded to the underlying CustomFlashList and used as an imperative handle
   // so the parent can control behavior like scrolling; it's typed as Function because there's not
   // a good way to capture this otherwise with Flow.
```
abbeycampbell: Let's update the comment here to reflect that this is now typed (just remove the second bit)

abbeycampbell (approved):
Agree that useMemo isn't totally necessary for this type of fix, but it still might be an optimization in some cases: I think this is one of those case by case things.

Approved with one tiny ask

FLGMwt:
> Agree that useMemo isn't totally necessary for this type of fix, but it still might be an optimization in some cases: I think this is one of those case by case things.

ah, sorry, I didn't mean to reject `useMemo` generally, agreed it's still important. I meant specifically exploring the memoization of elements directly passed to the `FlashList` `*Component` props.

src/components/ObservationsFlashList/ObservationsFlashList.tsx
FLGMwt: Good catch, thanks!

PR #3630, opened 2026-05-18
Refactor zustand mmkv direct access mapType

jtklein:
Refactored so that the mapType state is directly read from store in all Map instances (Closes MOB-841), and set from the button switcher (Closes MOB-1377). This removes the need for tracking local state.


FLGMwt (approved):
awesome, thank you!

PR #3631, opened 2026-05-18
initialize useCurrentUser useState value

FLGMwt:
Sample problem this fixes (that I introduced):

```js
const currentUser = useCurrentUser();

useEffect(() => {
  const announcements = !!currentUser
    ? await authenticatedAnnouncements() // API A
    : await unauthenticatedAnnouncements(); // API B
}, [currentUser]);
```

Given a logged in user, this triggers an incorrect call to `API B` right before `API A`. The first effect is triggered on the initial state of `null` and only corrected to `currentUser` on the hook's effect run & listener setup call to `setState`.

Fix: initialize the hook's state (with an [initializer function](https://react.dev/reference/react/useState#parameters) so we're not adding realm hits past the first render).

I also simplified the listener logic in a separate commit. This collection only really cares if it changes and then we want to get the latest correct value. The result is equivalent.

closes MOB-1431

### DIFF
--- a/src/sharedHooks/useCurrentUser.ts
+++ b/src/sharedHooks/useCurrentUser.ts
@@ -22,8 +22,11 @@ const useCurrentUser = ( ): User | null => {
       setCurrentUser( User.currentUser( realm ) || null );
     };
 
-    realmResults.addListener( listener );
+    // User could* have been mutated between state initialization and listener setup, resync state
+    // * FLGMwt: if you're reading this and have better knowledge of Realm's lifecycle, speak up!
+    listener();
 
+    realmResults.addListener( listener );
     return ( ) => {
       realmResults.removeListener( listener );
     };
--- a/src/sharedHooks/useCurrentUser.ts
+++ b/src/sharedHooks/useCurrentUser.ts
@@ -1,6 +1,5 @@
 import { RealmContext } from "providers/contexts";
 import { useEffect, useState } from "react";
-import type Realm from "realm";
 import User from "realmModels/User";
 
 const { useRealm } = RealmContext;
@@ -8,41 +7,27 @@ const { useRealm } = RealmContext;
 const useCurrentUser = ( ): User | null => {
   const realm = useRealm( );
   const [currentUser, setCurrentUser] = useState<User | null>( null );
+
   useEffect( ( ) => {
     const realmResults = realm.objects( User ).filtered( "signedIn == true" );
 
-    // Sets the current user if one is detected, unsets it if not
-    const listener = (
-      collection: Realm.OrderedCollection<User>,
-      changes: Realm.CollectionChangeSet,
-    ) => {
-      if (
-        ( changes.deletions && changes.deletions.length > 0 )
-        || !collection[0]?.isValid( )
-      ) {
-        setCurrentUser( null );
-      } else {
-        setCurrentUser( collection[0] );
-      }
+    // when the signedIn User collection changes, update state to latest currentUser or null.
+    // We don't need to handles changes in the listener because we only ever care about the
+    // single result (or lack of)
+    const listener = () => {
+      setCurrentUser( User.currentUser( realm ) || null );
     };
 
-    // Run the listener on the results for the first time
-    listener( realmResults, {
-      insertions: [],
-      deletions: [],
-      newModifications: [],
-      oldModifications: [],
-    } );
+    realmResults.addListener( listener );
 
-    // Add the listener
-    realmResults?.addListener( listener );
-
-    // Remove the listener when this dismounts
     return ( ) => {
-      realmResults?.removeListener( listener );
+      realmResults.removeListener( listener );
     };
   }, [realm] );
 
+  // isValid() here _only_ protects an invalid Realm being used in the one render pass between
+  // the listener firing on logout and currentUser being setStat'd to null.
+  // async callbacks referencing this `currentUser` return still need to `isValid()`.
   return currentUser?.isValid( )
     ? currentUser
     : null;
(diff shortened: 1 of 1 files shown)


src/sharedHooks/useCurrentUser.ts
```diff
@@ -1,48 +1,37 @@
...
-    listener( realmResults, {
-      insertions: [],
-      deletions: [],
-      newModifications: [],
-      oldModifications: [],
-    } );
-
-    // Add the listener
-    realmResults?.addListener( listener );
+    realmResults.addListener( listener );
```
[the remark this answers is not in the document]
FLGMwt: @copilot Do you mean when the realm effect dependency changes? or when the realm Collection that the listener is added to changes?

I don't think this is a legitimate concern, but tell me what you're thinking.

src/sharedHooks/useCurrentUser.ts
```diff
@@ -1,48 +1,37 @@
...
-      if (
-        ( changes.deletions && changes.deletions.length > 0 )
-        || !collection[0]?.isValid( )
-      ) {
-        setCurrentUser( null );
-      } else {
-        setCurrentUser( collection[0] );
-      }
+    // when the signedIn User collection changes, update state to latest currentUser or null.
+    // We don't need to handles changes in the listener because we only ever care about the
```
FLGMwt:
```suggestion
    // We don't need to handle changes in the listener because we only ever care about the
```

src/sharedHooks/useCurrentUser.ts
```diff
@@ -1,48 +1,37 @@
...
   const realm = useRealm( );
-  const [currentUser, setCurrentUser] = useState<User | null>( null );
+
+  // use a state initializer for the initial value from Realm
+  const [currentUser, setCurrentUser] = useState<User | null>(
+    () => User.currentUser( realm ) || null,
+  );
+
   useEffect( ( ) => {
     const realmResults = realm.objects( User ).filtered( "signedIn == true" );
```
jtklein, line 16: I am curious if realmResults.length is > 1 for any user out there somehow.

src/sharedHooks/useCurrentUser.ts
```diff
@@ -1,48 +1,37 @@
...
-      }
+    // when the signedIn User collection changes, update state to latest currentUser or null.
+    // We don't need to handle changes in the listener because we only ever care about the
+    // single result (or lack of)
+    const listener = () => {
+      setCurrentUser( User.currentUser( realm ) || null );
     };
 
-    // Run the listener on the results for the first time
-    listener( realmResults, {
```
jtklein, line 30: Can we keep this part in? It doesn't seem to me to hurt us here. I don't suppose `realm` will change instance throughout the usage of the hook. If  the user signs out and we delete realm we also re-start the app, so all instances of this will unmount.
I don't know if this was here to take care of state initialization on the first effect call only or if there was a real concern that realm instance changes triggered this effect multiple times.

jtklein (approved):
I think this if fine. I am not sure why past developers decided to invoke the listener on every effect trigger.

I think my bigger concern is why do we need to keep a local state here anyway? This hook is used a lot, so we have lots of listeners and lots of currentUser state. Why do we need a state with the live realm object for every hook? Can we not just return User.currentUser( realm )?

src/sharedHooks/useCurrentUser.ts
FLGMwt: honestly, if we made `true` the primaryKey, it'd solve a lot of this confusion ; D

FLGMwt:
> I think this if fine. I am not sure why past developers decided to invoke the listener on every effect trigger.
> 
> I think my bigger concern is why do we need to keep a local state here anyway? This hook is used a lot, so we have lots of listeners and lots of currentUser state. Why do we need a state with the live realm object for every hook? Can we not just return User.currentUser( realm )?

This was necessary for them to allow react to notice ~~when`currentUser` was `!currentUser`~~ EDIT: when `!deepEqual(prevCurrentUser, currentUser)` (or, for that matter, when a single property changed) https://www.mongodb.com/docs/atlas/device-sdks/sdk/react-native/react-to-changes

I've since tested the _reactive_ useQuery / useObject hooks and they honestly might be the right thing to do instead. I had more "ickyness" before about too many js listeners but that wasn't  informed by data.

src/sharedHooks/useCurrentUser.ts
FLGMwt: I'm nearly certain it was added for initialization. In principle, this PR moves initialization of a value that is synchronously available from first effect run setState to initialState.

I'm thankful I didn't finish responding last night, because stewing on this made me realize: even though the effect runs & listeners are setup immediately after paint, there's still a chance for Realm to _synchronously_ get a write to User during render between this state initialization and the effect run for listener setup. This is tricky!

FLGMwt: I restored the listener call in the effect here: https://github.com/inaturalist/iNaturalistReactNative/pull/3631/changes/7b2e223568bdb05674f8f59996a00ac982cd2e6a 

Thanks for bringing this up!

Since I didn't respond yet to the `[realm]` effect dep: I think we are assumed to treat the realm reference as stable and I think a lot of our app would break if it _wasn't_. I'm still trying to figure out where that falls on the spectrum of `useEffectEvent` concerns because it locally applies to "stable" deps like some of our zustand actions. 

e.g.: _we_ know that we never change the referential integrity of `setExploreView` ([def])(https://github.com/inaturalist/iNaturalistReactNative/blob/7b2e223568bdb05674f8f59996a00ac982cd2e6a/src/stores/createExploreSlice.ts#L14) but it still "annoyingly" has to be an effect/memo dep

PR #3632, opened 2026-05-19
Update existing android detox tests

jtklein:
Fixes some steps in our current Android e2e tests. Part of MOB-939

From what I can see during some testing in this branch and through fixing some additional broken steps with some more mocking, many of the remaining Android e2e test errors are also legit bugs users might experience on real devices. So, I am fine with merging this PR in as is, and from this new baseline define further tickets that fix some of these bugs, and then re-assess the e2e test failures.


e2e/helpers.js
```diff
@@ -19,30 +19,22 @@ function execPromise( command ) {
...
+      "adb push e2e/animal.jpg"
+        + " /sdcard/Android/data/org.inaturalist.iNaturalistMobile/files/e2e_test.jpg",
+    );
   }
 }
 
 export async function iNatE2eBeforeEach( device ) {
-  // device.launchApp would be preferred for an app of our complexity. It does work locally
-  // for both, but on CI for Android it does not work. So we use reloadReactNative for Android.
-  if ( device.getPlatform() === "android" ) {
```
jtklein, line 42: I am removing this now because we no longer run these tests in CI, and if we try to do that again we are facing a new reality..

src/sharedHelpers/installData.e2e-mock
```diff
@@ -1,7 +1,8 @@
-import { useState } from "react";
+import { useCallback, useEffect, useState } from "react";
```
jtklein, line 1: Taken as is from here: https://linear.app/inaturalist/issue/MOB-939/fix-existing-android-detox-tests#comment-d8d65c15

tests/vision-camera/vision-camera.js
```diff
@@ -49,13 +51,37 @@ export class mockCamera extends React.PureComponent {
...
+      return;
+    }
+    frameProcessor.frameProcessor( mockFrame );
   }
 
   // eslint-disable-next-line class-methods-use-this, react/no-unused-class-component-methods
   async takePhoto() {
-    // TODO: this only works on iOS
+    if ( Platform.OS === "android" ) {
+      /*
```
jtklein, line 64: Add a mock for Android to return the new animal image above as a new photo when pressing the shutter button.
Sidenote: I also tried to remove the mock, so that the emulator uses the emulated camera on Android but that was throwing vision-camera errors left and right, so I am fine with this here as it is similar to what we do on iOS already.

.detoxrc.js
```diff
@@ -1,20 +1,6 @@
-const { execSync } = require( "child_process" );
 const fs = require( "fs" );
 const { version } = require( "./package.json" );
 
-function getAvailableAVD( ) {
```
jtklein, line 5: Let us restrict this for now so that it for sure runs locally on the specific device set up for Android detox tests, according to the setup guide linked in code below.

src/sharedHelpers/installData.e2e-mock
FLGMwt: I hope we never ever ever do this in a real component for sharing hook state, but LGTM in e2e ; D

.detoxrc.js
FLGMwt: I'm with you there. A dev can override this if they need, but it makes sense to keep the tooling non-general. thanks for the link!

FLGMwt (approved):

PR #3637, opened 2026-05-19
fix useLocalObservation regression: set listener on filtered 'desired' realm collection

FLGMwt:
I noticed a mistake I made with this. We discussed the collection fidelity here https://github.com/inaturalist/iNaturalistReactNative/pull/3609#discussion_r3229917873 and while that's still correct about `localObservations` / `collection`, the _returned_ & cached mapping is based on a different collection, `filteredObs` which is the result of a filter & sort. 

As a result, a `changes` that references `obs[0]` would, in this case, incorrectly target a `needs_sync` obs sorted to the top of `filteredResults`.

The fix:

We don't actually need to deal with `observations` and `filteredObservations` separately. We can just make `filteredObservations` the "primary" collection and add a listener to that. This keeps us from having to manually reconcile the two.

### DIFF
--- a/src/sharedHooks/useLocalObservations.js
+++ b/src/sharedHooks/useLocalObservations.js
@@ -37,13 +37,12 @@ const useLocalObservations = ( ): Object => {
       // Satisfy the useEffect return type by returning a destructor function.
       return () => {};
     }
-    const localObservations = realm.objects( "Observation" );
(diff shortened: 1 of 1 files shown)


jtklein (approved):
The code changes seem good. Have tested in Debug mode on iOS and seen no errors.

PR #3640, opened 2026-05-20
Add fetchPlaceName mock

jtklein:
Part of MOB-939.

This fixes a fail in the AI camera e2e test. The reason for the fail is an unhandled promise rejection on using geocoding, which I have now mocked away. This is a legit bug for users on Android phones without Play Services, so we should at one point fix the source of the bug and remove this mock. Follow-up ticket here MOB-1433.

This PR makes the aiCamera.e2e.js test pass for me locally.

### DIFF
--- a/tests/vision-camera/vision-camera.js
+++ b/tests/vision-camera/vision-camera.js
@@ -64,7 +64,7 @@ export class mockCamera extends React.PureComponent {
       /*
         On Android, copyAssetsFileIOS is not available. Instead, copy the test
         image that was pushed to the app's external files directory by
-        iNatE2eBeforeAll (via `adb push e2e/maestro/android/test.jpg`).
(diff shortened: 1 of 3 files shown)


tests/vision-camera/vision-camera.js
```diff
@@ -127,7 +127,7 @@ export class mockCamera extends React.PureComponent {
   }
 
   render() {
-    return <View style={style} />;
+    return <View style={style} collapsable={false} />;
```
FLGMwt, line 130: what was the error that this fixed out of curiosity?

FLGMwt (approved):
nice! a placeholder mock makes sense to keep tests moving along

tests/vision-camera/vision-camera.js
jtklein: RN Gesture Handler related.

PR #3641, opened 2026-05-20
Replace android e2e file

FLGMwt:
see separate commits: this is just getting rid of the "old" workflow file and replacing it w/ the "new" one.


jtklein (approved):
Thanks! I have triggered new and old in main to see if it still works before. But I don't think there is a way to trigger from this branch with the new name if I remember correctly.

FLGMwt:
Aye, I believe that's correct. Thanks!

PR #3642, opened 2026-05-20
Type ObsDetails nav params

jtklein:
Adds some types to ObsDetails navigation props.

### DIFF
--- a/src/components/ObsDetails/ObsDetailsContainer.js
+++ b/src/components/ObsDetails/ObsDetailsContainer.js
@@ -30,7 +30,8 @@ const ObsDetailsContainer = ( ): Node => {
     setObsDetailsTab,
   } = useLayoutPrefs( );
   const currentUser = useCurrentUser( );
-  const { params } = useRoute();
+  // const { params } = useRoute<TabStackScreenProps<"ObsDetails">["route"]>( );
+  const { params } = useRoute( );
   const {
     targetActivityItemID,
     uuid,
--- a/src/components/ObsDetails/ObsDetailsContainer.js
+++ b/src/components/ObsDetails/ObsDetailsContainer.js
@@ -12,7 +12,6 @@ import type { Node } from "react";
 import React, {
   useState,
 } from "react";
-import { LogBox } from "react-native";
 import Observation from "realmModels/Observation";
 import {
   useCurrentUser,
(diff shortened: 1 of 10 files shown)


src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -25,20 +24,14 @@ import { OBS_DETAILS_TAB } from "stores/createLayoutSlice";
 
 import ObsDetails from "./ObsDetails";
 
-// this is getting triggered by passing dates, like _created_at, through
-// react navigation via the observation object. it doesn't seem to
-// actually be breaking anything, for the moment (May 2, 2022)
-LogBox.ignoreLogs( [
-  "Non-serializable values were found in the navigation state",
-] );
```
FLGMwt, line 33: Definitely makes sense to remove this since we explicitly decided to _stop_ sending non-serializable data through route params!

I think this warning probably fires on something more problematic than the comment let on: we shouldn't have had a problem passing a `Date` through params, it was probably flagging a live Realm object or something w/ a function.

It'll be good to have these flagged again!

src/components/ObsDetails/ObsDetailsContainer.js
```diff
@@ -25,20 +24,14 @@ import { OBS_DETAILS_TAB } from "stores/createLayoutSlice";
...
-
 const ObsDetailsContainer = ( ): Node => {
   const {
     obsDetailsTab,
     setObsDetailsTab,
   } = useLayoutPrefs( );
   const currentUser = useCurrentUser( );
-  const { params } = useRoute();
+  // const { params } = useRoute<TabStackScreenProps<"ObsDetails">["route"]>( );
+  const { params } = useRoute( );
```
FLGMwt, line 34: Is this meant to be commented out for the next person who comes around to type this file?

FLGMwt (approved):
LGTM!

src/components/ObsDetails/ObsDetailsContainer.js
jtklein: Yes.

src/components/ObsDetails/ObsDetailsContainer.js
jtklein: Yes, one reason I choose to remove it here. We do not want to neglect such errors if they are logged.

PR #3645, opened 2026-05-21
Two small changes

jtklein:
Two small changes for MOB-203, requested [here](https://linear.app/inaturalist/issue/MOB-203/handle-moderatoractions#comment-18128a35).


FLGMwt (approved):
lgtm, reflects product feedback

PR #3647, opened 2026-05-21
PotentialDisagreementSheet not allowing you to choose first value

jtklein:
Closes MOB-1395.

Added a new prop `requireSelectionChange` to be able to not consider the default value being "dirty" so that the user can already submit the radio button sheet with the default value.

Also updated types to reflect current real usage.


sepeterson (approved):
Code looks solid! Tested working locally

PR #3649, opened 2026-05-23
Consolidate ObsDetails screens

jtklein:
Closes MOB-1453

We currently have an issue with the header of ObsDetails not showing in e2e tests. The fact that we have to use a `Stack.Group` in the navigator to switch over isDefaultMode to decide which screen to show makes finding the source of the bug a bit more conplicated.

Anyways, we have a lot of shared code between the two screens still, especially everything to do with fetching the data source for the observation to be displayed is duplicated in both.

This refactor removes the need to switch on navigator level, instead we switch further below in the tree. We can even first check if the obs is only saved locally and display the SavedMatch screen as an early return.


src/components/ObsDetails/ActivityTab/ActivityTab.js
```diff
@@ -24,9 +24,11 @@ const ActivityTab = ( {
   activityItems,
   openAgreeWithIdSheet,
   isConnected,
-  targetItemID,
   onLayoutTargetItem,
 }: Props ): Node => {
+  // const { params } = useRoute<TabStackScreenProps<"ObsDetails">["route"]>( );
+  const { params } = useRoute( );
```
jtklein, line 30: I decided to read the route params in the component that uses it, instead of passing through all the way from screen.

src/components/ObsDetailsSharedComponents/ObsDetailsModeSwitcher.js
```diff
@@ -0,0 +1,123 @@
...
+  openAgreeWithIdSheet: Function,
+  refetchRemoteObservation: Function,
+  refetchSubscriptions: Function,
+  showAddCommentSheet: Function,
+  subscriptions?: Object,
+  wasSynced: boolean,
+  uuid: string,
+};
+
+const ObsDetailsModeSwitcher = ( props: Props ): Node => {
```
jtklein, line 30: My idea for this component is to harbor the logic for which UI to display.
I don't like the big list of prop drilling, but overall I think this is an improvement.

src/components/ObsDetailsSharedComponents/ObsDetailsScreen.tsx
```diff
@@ -11,12 +12,9 @@ import {
   useRemoteObservation,
 } from "sharedHooks";
 
-import SavedMatchContainer from "./SavedMatch/SavedMatchContainer";
-
-const ObsDetailsDefaultModeScreensWrapper = () => {
```
jtklein, line 16: I have based this new screen on the previous default mode screen wrapper, but it is also a combination with the previous ObsDetails/ObsDetailsContainer

My general idea was for this to be the container that takes care of fetching the observation, and the new ObsDetailsContainer be responsible for having the mutation logic.

src/components/ObsDetailsSharedComponents/ObsDetailsScreen.tsx
jtklein, line 64: Even though this was only a part of the default mode wrapper, I am not restricting this to isDefaultMode here, because for a saved only obs in advanced mode we route ObsEdit instead off ObsDetails.

tests/unit/components/ObsDetailsSharedComponents/ObsDetailsScreen.test.js
```diff
@@ -1,12 +1,15 @@
+import { useRoute } from "@react-navigation/native";
 import { screen } from "@testing-library/react-native";
-import ObsDetailsDefaultModeScreensWrapper
-  from "components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper";
+import ObsDetailsScreen from "components/ObsDetailsSharedComponents/ObsDetailsScreen";
```
jtklein, line 3: Moved this test as I have the component

tests/unit/components/ObsDetailsSharedComponents/ObsDetailsContainer.test.js
```diff
@@ -1,6 +1,6 @@
 import { useRoute } from "@react-navigation/native";
 import { fireEvent, screen, waitFor } from "@testing-library/react-native";
-import ObsDetailsContainer from "components/ObsDetailsDefaultMode/ObsDetailsDefaultModeContainer";
+import ObsDetailsContainer from "components/ObsDetailsSharedComponents/ObsDetailsContainer";
```
jtklein, line 3: Have moved this test as I have the component.

tests/integration/ObsDetails.test.js
```diff
@@ -1,7 +1,5 @@
 import { screen, waitFor } from "@testing-library/react-native";
-import ObsDetailsContainer from "components/ObsDetails/ObsDetailsContainer";
-import ObsDetailsDefaultModeScreensWrapper
-  from "components/ObsDetailsDefaultMode/ObsDetailsDefaultModeScreensWrapper";
+import ObsDetailsScreen from "components/ObsDetailsSharedComponents/ObsDetailsScreen";
```
jtklein, line 2: There is only one container now.

jtklein (commented):
Comments only

src/components/ObsDetailsSharedComponents/ObsDetailsModeSwitcher.js
```diff
@@ -0,0 +1,123 @@
...
+      id: OBS_DETAILS_TAB.DETAILS,
+      testID: "ObsDetails.DetailsTab",
+      onPress: () => setObsDetailsTab( OBS_DETAILS_TAB.DETAILS ),
+      text: t( "DETAILS" ),
+    },
+  ];
+
+  const showActivityTab = obsDetailsTab === OBS_DETAILS_TAB.ACTIVITY;
+
+  if ( !observation ) {
```
sepeterson, line 74: when is observation falsy? A loading state?

src/navigation/StackNavigators/TabStackNavigator.tsx
```diff
@@ -223,21 +210,11 @@ const TabStackNavigator = ( { route }: BottomTabProps ) => {
...
-            />
-          )
-          : (
-            <Stack.Screen
-              name="ObsDetails"
-              component={FadeInObsDetailsContainer}
-              options={OBS_DETAILS_OPTIONS}
-            />
-          )}
+        <Stack.Screen
```
sepeterson, line 213: Excellent

sepeterson (approved):
Fantastic refactor! Smoke tested ios/android

src/components/ObsDetailsSharedComponents/ObsDetailsModeSwitcher.js
jtklein: Ah yes, it shouldn't be. This is a duplicate check to what is in ObsDetailsContainer `observationShown &&`

PR #3651, opened 2026-05-24
Fix graphical bug when reordering observations in EvidenceList

nebotron:
Hello team! Thanks so much for all of your work on iNaturalist. It's an incredible project to document nature and help scientists! I love using it, and I'd love to contribute if I can. I read the [contribution guidelines](https://github.com/inaturalist/iNaturalistReactNative/blob/main/CONTRIBUTING.md), but let me know if there is anything I can do better. I know there is no open ticket for this, but it's a [known bug](https://forum.inaturalist.org/t/visual-glitch-when-reordering-images/78805/4), and I wanted to fix it.

There was a graphical bug when drag-and-dropping items in react-native-draggable-flatlist. This was [fixed upstream](https://github.com/computerjazz/react-native-draggable-flatlist/pull/586) in 4.0.3. iNaturalistReactNative currently uses 4.0.1. This PR pulls the upstream patch into iNaturalist's patch.

My first idea was to upgrade react-native-draggable-flatlist to 4.0.3, but that causes a cascading dependency mismatch (log below). The patch is small so pulling it into the [iNaturalist patch](https://github.com/inaturalist/iNaturalistReactNative/commit/0e71fb5062741a1bceabacf49a9052f6b3385bde) seems like the simplest path forward. If you think it would be better for me to upgrade `react-native-draggable-flatlist`, `react-native-worklets`, and `react-native-reanimated` instead of patching, I confirmed that approach works as well, but could potentially have side effects since those are pretty important packages.

```
$ npm install
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: inaturalistreactnative@1.0.22
npm error Found: react-native-worklets@0.7.4
npm error node_modules/react-native-worklets
npm error   react-native-worklets@"^0.7.0" from the root project
npm error
npm error Could not resolve dependency:
npm error peer react-native-worklets@"0.8.x" from react-native-reanimated@4.3.1
npm error node_modules/react-native-reanimated
npm error   react-native-reanimated@"^4.2.2" from the root project
npm error   peer react-native-reanimated@">=2.8.0" from react-native-draggable-flatlist@4.0.3
npm error   node_modules/react-native-draggable-flatlist
npm error     react-native-draggable-flatlist@"^4.0.3" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

Before:

https://github.com/user-attachments/assets/5985f78e-754b-48a2-8953-79a2079d1f2b

After:

https://github.com/user-attachments/assets/7f2b6f8f-441c-4631-b1d7-a8732a10d831

cc @albullington


patches/react-native-draggable-flatlist+4.0.1.patch
```diff
@@ -301,22 +301,6 @@ index 57308fb..ae63cdb 100644
...
 \ No newline at end of file
-diff --git a/node_modules/react-native-draggable-flatlist/lib/module/context/animatedValueContext.js b/node_modules/react-native-draggable-flatlist/lib/module/context/animatedValueContext.js
```
nebotron: This was merged upstream: https://github.com/computerjazz/react-native-draggable-flatlist/commit/c10575df82995129dc158aa913ed7e8e9554d6cc

sepeterson:
Hey @nebotron! This looks great and I was able to test and verify it works on ios/android. One thing: could you regenerate the patch using `npx patch-package react-native-draggable-flatlist` please? As is it's causing a failure during npm install

nebotron:
Thanks for the review! I really appreciate it. I wasn't able to reproduce the `npm install` error locally, but I regenerated the patch via `npx patch-package react-native-draggable-flatlist` and it came out a little bit different, so hopefully this fixes the issue in CI. It seems that I need your approval to run the workflow to verify?

sepeterson:
@nebotron Hmm kind of a puzzling issue. I think now that I've allowed the jest tests for this pr they should run whenever you push to this branch. If you can get them to pass, I'll merge this!

nebotron:
Unfortunately, I still can't reproduce this issue locally, either using a fresh clone of the repo, `rm -rf node_modules`, or `npm ci` to clear the build state. `npm rebuild` works just fine on this branch on my machine.

I have a theory though - I suspect the ["Cache node modules" step](https://github.com/inaturalist/iNaturalistReactNative/actions/runs/26550958872/job/78251239737?pr=3651) is picking up a stale version of node_modules, or a version where the *previous* patch was already applied. I have temporarily disabled caching in the test.yml github action. If that fixes it, that confirms the culprit at least.

If this doesn't work, how would you feel about me trying the alternative approach of upgrading react-native-draggable-flatlist, react-native-worklets, and react-native-reanimated? It would mean a smaller patch to maintain, at least.

Would you mind approving the workflow again? Sorry to bug you, but it seems I need approval every time.

nebotron:
Thank you for approving the workflow! [It passed](https://github.com/inaturalist/iNaturalistReactNative/actions/runs/26612326401/job/78474531407?pr=3651), which I think confirms the theory that there is a bug in the `Cache node modules` step. IMHO, the best way to proceed is to remove the cache. If you look at the [test run](https://github.com/inaturalist/iNaturalistReactNative/actions/runs/26612326401/job/78474531407?pr=3651), reinstalling dependencies without caching takes 1 minute, while unit tests take 4. Adding 1 minute to CI pipelines seems like a reasonable price to pay to avoid a potentially buggy build component.

If you would prefer a different approach, like manually [clearing the cache in the GitHub actions interface](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches), I'm ok with that too. Just let me know how you'd like me to proceed :)

cc @FLGMwt, since you have done some work with node_modules caching

sepeterson:
@nebotron Nice! We want to keep the caching but if you create a new PR with just the patch-package generated patch, I will approve the CI tests, clear any relevant caches, and merge

nebotron:
Fair enough. I have reverted all of the build system changes on this PR. If you clear caches and restart tests, it should be clear to merge.

sepeterson (approved):
Nice. Thanks!

FLGMwt:
Congrats @nebotron on your first contribution! 

> Thank you for approving the workflow! [It passed](https://github.com/inaturalist/iNaturalistReactNative/actions/runs/26612326401/job/78474531407?pr=3651), which I think confirms the theory that there is a bug in the `Cache node modules` step. IMHO, the best way to proceed is to remove the cache. If you look at the [test run](https://github.com/inaturalist/iNaturalistReactNative/actions/runs/26612326401/job/78474531407?pr=3651), reinstalling dependencies without caching takes 1 minute, while unit tests take 4. Adding 1 minute to CI pipelines seems like a reasonable price to pay to avoid a potentially buggy build component.
> 
> If you would prefer a different approach, like manually [clearing the cache in the GitHub actions interface](https://docs.github.com/en/actions/how-tos/manage-workflow-runs/manage-caches), I'm ok with that too. Just let me know how you'd like me to proceed :)
> 
> cc @FLGMwt, since you have done some work with node_modules caching

**tl;dr takeaway: it seems to be that when we edit and existing patch, we should expect to manually clear npm caches on GitHub Actions and `rm -rf node_modules/my-dependency && npm install` locally to make sure the patch runs cleanly**

Thanks @nebotron and @sepeterson for figuring this out! I agree that it doesn't make sense to disable caching for this issue. I have the privilege of hindsight, but my understanding of what happened is something like the following:

`my-dependency` looks like:

```
# index.ts
Alice
Bob
```

We _previously_ had a patch that did this:

```
# my-dependency.+1.0.0.patch
ln1: Alice => Apples
```

Such that after `npm install`, we have the following (on our local filesystem and saved as the `node_modules` cache):

```
# index.ts
Apples
Bob
```

This PR introduced a change that make the patch look like:

```
# my-dependency.+1.0.0.patch
ln1: Alice => Apples
ln2: Bob => Bananas
```

With the desired result of:

```
# index.ts
Apples
Bananas
```

What we saw:
- When we `npm install`'d without node_modules / cache, this worked fine
- When we `npm install`'d _with_ node_modules / cache this broke

Expected cause:
When `node_modules/my-dependency` exists, `npm install` doesn't re-download it so it remains as is. 
On post-install, `patch-package` sees `my-dependency.+1.0.0.patch` and tries to apply it. Its heuristic is something like:
- try to [apply the patch
](https://github.com/ds300/patch-package/blob/master/src/applyPatches.ts#L428C5-L428C72)
- if that fails, _try to [reverse](https://github.com/ds300/patch-package/blob/master/src/applyPatches.ts#L439-L444) the patch_. That is: if the patch can be exactly undone, that means it was already applied
- if it can't _apply_ the patch or _reverse_ the patch, then something's amiss! [so it bails](https://github.com/ds300/patch-package/blob/master/src/applyPatches.ts#L445-L447)

So in our case, both apply and reverse failed because on disk, our cached dependency with first applied version of the patch looked like:

```
# index.ts
Apples
Bob
```

And it couldn't fully turn people into fruit or fruit into people (examples are hard).

-----

Anyhow! This was some good learning for me, I've been meaning to dig a tiny bit into patch-package because I've run into this situation before. We appreciate the opportunity for a better understanding and we can be mindful of this type of issue and how to address it in the future. Namely:

- If we're modifying an existing `.patch` file (without changing the version), we'll need to clear node_modules (at least for that package) and GitHub Actions' npm cache

..........

BONUS

It looks like patch-package@8 added [support for _multiple_ patches](https://github.com/ds300/patch-package/tree/master#creating-multiple-patches-for-the-same-package) for a single dependency that would suit this purpose. That is, it would allow for a patch _per concern_ which would help mitigate this issue and be a lot more pleasant for patch maintenance. It would be something like the following which would be managed separately and applied sequentially:

```
patches/react-native-draggable-flatlist+4.0.1+001+reanimated-3-fix.patch
patches/react-native-draggable-flatlist+4.0.1+002+reordering-bug.patch
```

That's not a suggestion, just noting for the future

PR #3652, opened 2026-05-25
Increase maximum zoom in the media viewer

nebotron:
There are many images in iNaturalist where the subject is very small in the frame. Particularly when ID'ing, 5x zoom is often not enough. Increasing this value is a simple enough change, and aligns with the [2026H1 goal](https://www.inaturalist.org/blog/123394-our-product-goals-for-january-june-2026) of building the identifier community. The mobile app has huge potential to make it easier for identifiers to spend a few minutes doing it on the go.

This is also a good feature for accessibility - I don't have great eyesight, and I imagine I'm not alone among iNaturalist users.

cc @FLGMwt @albullington - seems like you do work here. It seems I don't have permission to add reviewers to a PR?

### DIFF
--- a/src/components/MediaViewer/CustomImageZoom.tsx
+++ b/src/components/MediaViewer/CustomImageZoom.tsx
@@ -9,7 +9,7 @@ import React, {
 import useDeviceOrientation from "sharedHooks/useDeviceOrientation";
 
 const MIN_SCALE = 0.5;
-const MAX_SCALE = 5;
+const MAX_SCALE = 50;
 
 interface Props {
   uri: string;


nebotron:
cc @sepeterson, in case you care to review :)

danrademacher:
@nebotron Thanks for this submission! 

We have limited capacity to review outside PRs (we updated [our docs](https://github.com/inaturalist/iNaturalistReactNative/blob/main/CONTRIBUTING.md) recently to be more clear about that), but I admit this is a VERY small change (1 character, smallest possible PR I suppose!), with potentially helpful impact for a range of use cases, as you note.

For this specific feature, I'm curious how you landed on 50X. Is that a zoom level you have found useful in other contexts?

nebotron:
I figured with a max image size of 2048 pixels, zooming in 50x lets 40 pixels fill the screen, such is enough to see really every detail in the image. I absolutely have seen images where they can't easily be ID'ed with only 5x zoom.

nebotron:
Do you have any interest in reviewing this @danrademacher? Is there a different value you would prefer? I don't have a super strong opinion, but from playing around with it on some typical observations (mostly birds) I think it should be 30 or higher.

danrademacher:
Apologies for the slow response on this. We've been clarifying our practices around how to manage open source contributions. For feature changes like this, even simple ones (vs bug fixes), we need to make space in our planing to review, prioritize, and incorporate into our release schedule.

I added this request to our internal tracking so we can prioritize along with other requests coming in, and then when the time comes, we can reopen this PR, adjust as needed. I expect that in the next few months, though I don't have a specific timeline right now.

PR #3653, opened 2026-05-25
Update react native to 0.83.9

jtklein:
Compiled and tested in Debug and Release mode on iOS and Android, by using the AI camera.


sepeterson (approved):
Smoked test ios/android

PR #3656, opened 2026-05-26
Fix: Media deletion was operating with stale data

jtklein:
Closes MOB-1392

`mediaUris` is referencing the state before the media was deleted, so we have to select the new state after deletion based on mediaUris filtered by the just deleted uri. Or we could update the callback to work with non-stale data, but I don't think a callback is needed here anyway. We are just calling a function in the media viewer modal.


FLGMwt (approved):
I agree, I like the local fix approach over a refactor to fix a specific bug like this

PR #3657, opened 2026-05-26
MOB-882: popto when selecting taxon from taxon details

sepeterson:
Closes [MOB-882](https://linear.app/inaturalist/issue/MOB-882/explore-swiping-back-after-adding-id-from-taxon-page-not-working)

Instead of adding a new screen to the stack, return to a previous one. Mirrors the approach in useNavigateWithTaxonSelected here https://github.com/inaturalist/iNaturalistReactNative/blob/ea5e4bb20bdc6bdedf4c95914d5f46fd14b11915/src/components/Suggestions/hooks/useNavigateWithTaxonSelected.ts#L47

### DIFF
--- a/src/components/TaxonDetails/TaxonDetails.js
+++ b/src/components/TaxonDetails/TaxonDetails.js
@@ -1,7 +1,12 @@
 // @flow
 
 import { refresh, useNetInfo } from "@react-native-community/netinfo";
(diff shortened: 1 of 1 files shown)


abbeycampbell (approved):
Tested on iOS/android and lgtm!

PR #3661, opened 2026-05-27
Use nvmrc to specifiy the version in github actions node setup

jtklein:
@FLGMwt I have seen your android e2e test uses this instead of specifying the version in the workflow file. Is this preferred?

### DIFF
--- a/.github/workflows/crowdin-pull.yml
+++ b/.github/workflows/crowdin-pull.yml
@@ -27,7 +27,8 @@ jobs:
       - name: Install Node.js, NPM and Yarn
         uses: actions/setup-node@v6
         with:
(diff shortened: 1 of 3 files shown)


FLGMwt (approved):
no strong opinions here. I was coming off the ios e2e file and treating the android e2e as fresh since it wasn't live.

I see the appeal of using the .nvmrc as documentation, but w/ CI, idk, maybe it's worth making a separate config as an "intentional" or "visible" change since we don't touch it much?

I'm leaning towards yeah, using nvmrc as source of truth and responding if it breaks rather than making it a separate maintenance thing.

thanks for this!

PR #3662, opened 2026-05-27
Update rozenite to 1.10.0

jtklein:
Also replaces deprecated mmkv- with general storage-plugin.


FLGMwt (approved):
LGTM, nice that they consolidated that

PR #3663, opened 2026-05-27
MOB-168: set needLocation to false when navigating to locationPicker

sepeterson:
Repro is not consistent but this definitely fixes it for me.

Only downside I can see is this will stop the location fetching even if a user doesn't actually select a location from the LocationPicker. I think it's worth the tradeoff for simplicity's sake

### DIFF
--- a/src/components/Match/MatchContainer.tsx
+++ b/src/components/Match/MatchContainer.tsx
@@ -294,7 +294,6 @@ const MatchContainer = ( ) => {
 
   const navToLocationPicker = useCallback( ( ) => {
     stopWatch( subscriptionId );
-    setNeedLocation( false );
     navigation.navigate( "LocationPicker" );
   }, [stopWatch, subscriptionId, navigation] );
 
--- a/src/components/Match/MatchContainer.tsx
+++ b/src/components/Match/MatchContainer.tsx
@@ -294,6 +294,7 @@ const MatchContainer = ( ) => {
 
   const navToLocationPicker = useCallback( ( ) => {
     stopWatch( subscriptionId );
+    setNeedLocation( false );
     navigation.navigate( "LocationPicker" );
   }, [stopWatch, subscriptionId, navigation] );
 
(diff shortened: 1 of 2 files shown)


FLGMwt:
> Only downside I can see is this will stop the location fetching even if a user doesn't actually select a location from the LocationPicker

I'm not sure I understand the implications here. Could you rephrase this as a "before" and "after" for this scenario?

src/components/Match/MatchContainer.tsx
```diff
@@ -294,6 +294,7 @@ const MatchContainer = ( ) => {
 
   const navToLocationPicker = useCallback( ( ) => {
     stopWatch( subscriptionId );
+    setNeedLocation( false );
```
jtklein: This is not required here because we only route to the location picker as a fallback if the user hasn't given location picker. In the default mode we prefer the device location over user input when location permission has been given.

jtklein (approved):
This works for me. Setting needLocation doesn't remove the location but only stops watching for a more precise one. So, in my testing if I go to Location picker from ObsEdit and do not select a new location and go back, I can see the last info that useWatch had retrieved which was not too bad. This might be more problematic in places with little GPS coverage though. So, we should also test this further in production in the field.

sepeterson:
> Could you rephrase this as a "before" and "after" for this scenario?

For sure!

Before:
1. Be on ObsEdit with location fetching
2. Navigate to LocationPicker, do *not* pick a location
3. Navigate back, location is continuing to fetch or has finished and a sufficiently accurate location is set

After:
1. Be on ObsEdit with location fetching
2. Navigate to LocationPicker, do *not* pick a location
3. Navigate back, location fetching was interrupted on nav and the set location is whatever it was when you navigated away

src/components/Match/MatchContainer.tsx
sepeterson: Good point! Will remove

PR #3664, opened 2026-05-27
Mob 1415 third party acknowledgements in about screen

jtklein:
Closes MOB-1415

Using a new dependency `react-native-legal` to create the information about licenses of packages used in the app, and using their deafult function to open the sheet to display them in the app.
On iOS, this also adds a section to the Settings app about Licenses, see: https://incubator.callstack.com/react-native-legal/docs/react-native#license-display-in-settings

IMG_4686
IMG_4687
IMG_4684
IMG_4685


README.md
```diff
@@ -372,6 +372,7 @@ fastlane prod
 1. Write release notes based on commits since the last release. Try to keep them brief but emphasize what's new and what's fixed. Just keep them in a text editor; you'll save them at a later step.
 1. Edit `package.json` and update the `version` per semantic versioning rules: bump the patch version if the only changes were bug fixes, bump minor version if there were new features, and bump the major version if the app was completely re-written or can't import data from previous versions.
 1. `npm install` to set the version in `package-lock.json`
+1. `npx react-native legal-generate` to re-generate the licenses to show in the app (based on current dependencies)
```
FLGMwt: Being extra sensitive to supply chain attacks, I think this could be made an npm script so we can enforce we run against our _own_ `.bin/react-native` (from the rn-cli). If we were to accidentally run this command `npx react-native` outside of a project would (prompt) install & run `react-native`'s bin instead. It happens to say "this doesn't work, you probably wanted rn-cli", but since the packages are owned by different parties, it's at least _one_ extra angle.

FLGMwt (approved):
This is cool! I didn't expect a bespoke package for the UI for this, but I'm a fan.

non-blocking on the `npx` thing

README.md
jtklein: Have added an npm script

PR #3666, opened 2026-05-27
memoize extraData to flashlist

FLGMwt:
came up in discovery of MOB-1458 / MOB-1447. PR-ing separately. Closes MOB-1459, my description copied from there:

> Moving from non-stable pojos to string primitives as list data highlighted an optimization opportunity: we were creating a non-stable reference to extraData which invalidates every render optimization that FlashList might attempt.
> 
> Even if a pojo reference was stable or a uuid identity was stable, we were forcing FlashList to renderItem again because extraData (which informs renderItem) was unstable.
> 
> These need to be wrapped in an object for the FlashList prop so this is a clear case where a memo object wrapper of two separate primitives is warranted.

### DIFF
--- a/src/components/ObservationsFlashList/ObservationsFlashList.tsx
+++ b/src/components/ObservationsFlashList/ObservationsFlashList.tsx
@@ -255,10 +255,10 @@ const ObservationsFlashList = ( {
     t,
   ] );
 
-  const extraData = {
+  const extraData = useMemo( () => ( {
     gridItemWidth,
     numColumns,
-  };
+  } ), [gridItemWidth, numColumns] );
 
   // only used id as a fallback key because after upload
   // react thinks we've rendered a second item w/ a duplicate key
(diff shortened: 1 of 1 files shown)


jtklein (approved):
Your changes re memoization: LGTM. However, `gridItemWidth` is not actually exported from `useGridLayout` so I would like to see a decision at one point if we can remove it from extraData or we need to export it from the hook or we access it from the style object. Doesn't affect your current work because it is forever `undefined`.

FLGMwt:
> Your changes re memoization: LGTM. However, `gridItemWidth` is not actually exported from `useGridLayout` so I would like to see a decision at one point if we can remove it from extraData or we need to export it from the hook or we access it from the style object. Doesn't affect your current work because it is forever `undefined`.

Good catch, on further discovery, we're not actually consuming extraData either. It _seems_ like it was meant to intentionally invalidate the list item renders based on these properties. But `gridItemWidth` we know to be `undefined` and `<FlashList key={numColumns} />` is already doing the work of invalidating the entire list when `numColumns` changes.

Updated PR to remove `extraData` entirely. This has the same performance effect while removing the misleading code.

I also tested omitting `key` and changing numColumns (at the useGridLayout level to get derived styles) and it seems to work fine. Maybe we were previously concerned about a relayout from 2 => ? columns? **In any case**, I'll leave the key as is and just

PR #3668, opened 2026-05-28
MOB-1322: rm removeBottomBorder from stack group

sepeterson:
Fixes not being able to tap the Match screen header edit icon on android.

As near as I can tell the background component added with removeBottomBorder was blocking taps to the edit button. It also seems not to be needed any longer.

<details>
  <summary>Screenies for visual comparison (they're the same)</summary>

  Android before

  

  
  
  Android after

  

  

  

  iOS after
  
  IMG_0169

  IMG_0171

  IMG_0170

  </details>


abbeycampbell (approved):
👍

PR #3670, opened 2026-05-28
Mob 1193 identify resources that can be reused between myobs and

abbeycampbell:
closes [MOB-1193](https://linear.app/inaturalist/issue/MOB-1193/identify-resources-that-can-be-reused-between-myobs-and-explore)

No changes to existing behavior for ExploreV2 debug sheet sort or currently in-prod species sort. Just pulling everything to a shared place and getting the approaches aligned (I included species because it was mentioned in the ticket, and ExploreV2 will need it as well).

There are some exports in `observationsSort.ts` that are currently not imported anywhere, but this will change when we wire up the UI in [MOB-1192](https://linear.app/inaturalist/issue/MOB-1192/add-sort-sheet-and-sort-options-to-explore-v2)


src/sharedHelpers/speciesSort.ts
```diff
@@ -0,0 +1,66 @@
...
+// TODO: ExploreV2 will support additional species sort options
+// (observed by me, not observed by me).
+// MyObs will have these count only, but we'll want to define sort options for explore
+export const MY_OBSERVATIONS_SPECIES_SORT_OPTIONS: SPECIES_SORT[] = [
+  SPECIES_SORT.COUNT_DESC,
+  SPECIES_SORT.COUNT_ASC,
+];
+
+export function useSpeciesSortLabels( ): Record<SPECIES_SORT, SpeciesSortLabels> {
+  const { t } = useTranslation( );
```
sepeterson, line 35: Claude pointed out that useTranslation() makes useSpeciesSortLabels a hook and that usually hooks go in a sharedHooks directory.

Not too fussed about this but wanted to note in case anyone else feels strongly

sepeterson (approved):
Nice work!

src/sharedHelpers/speciesSort.ts
abbeycampbell: oh that's an interesting point. I think I'd argue that the hook isn't really shared and will only ever be used by the sort bottom sheet. keeping it here means that when we add/change sort options (which we are shortly going to do, and might do again in the future if we figure out a way to do taxonomic sorting?), we can touch fewer files and keep the things that need updating co-located.

FLGMwt: Different idea: I think we can defer the translation to the markup and just make this a static object for the translation keys instead of the resolved labels. `COUNT_DESC` 's label translation id will _always_ be `Most-Observed-Default`. The calling component is already `useTranslation`ing

src/sharedHelpers/speciesSort.ts
```diff
@@ -0,0 +1,66 @@
...
+  speciesCounts: T[],
+  sortOptionId: SPECIES_SORT,
+): T[] {
+  const sorted = [...speciesCounts];
+
+  sorted.sort( ( a, b ) => {
+    const diff = a.count - b.count;
+    return sortOptionId === SPECIES_SORT.COUNT_ASC
+      ? diff
+      : -diff;
```
FLGMwt, line 62: that's a neat trick, I hadn't seen that before!

FLGMwt (approved):
LGTM but I'd like to change that hook to a static config object and avoid the custom hook

src/sharedHelpers/speciesSort.ts
abbeycampbell: I actually did that first! but then the commit hook that checks to make sure all translation keys are in use did not pick up that approach, and thought the sort species keys were unused, which is why I switched to this

PR #3671, opened 2026-05-28
Show no-evidence state on SavedMatch after deleting media

jtklein:
Closes MOB-1297

Also removed the Learn More button if there is no taxon to route to, because it crashes the app.

After
IMG_4705
Before
IMG_4707

After
IMG_4706
Before
IMG_4708

### DIFF
--- a/src/components/Match/PhotosSection.tsx
+++ b/src/components/Match/PhotosSection.tsx
@@ -2,6 +2,7 @@ import type { ApiPhoto, ApiTaxon } from "api/types";
 import classnames from "classnames";
 import MediaViewerModal from "components/MediaViewer/MediaViewerModal";
 import {
+  IconicTaxonIcon,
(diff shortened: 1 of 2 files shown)


src/components/Match/PhotosSection.tsx
```diff
@@ -188,6 +189,25 @@ const PhotosSection = ( {
     </View>
   );
 
+  if ( observationPhotos.length === 0 ) {
+    return (
+      <View className="h-[390px]">
```
FLGMwt, line 194: minor: we could at least have a shared var for this repeated `390` but I suppose variable reuse in classNames can get a little ugly

FLGMwt (approved):
> Also removed the Learn More button if there is no taxon to route to, because it crashes the app.

Hey that's a pretty good reason!

PR #3672, opened 2026-05-28
Show iconic taxon in obs list

jtklein:
When there is no media for an observation but a taxon is assigned, the fallback UI should show the iconic taxon. So, when mapping realm obs for the list we also need to get the iconic_taxon_name value.

Have found this while doing sth else, and I don't think there is a ticket for this.

Before
IMG_4709
After
IMG_4704


FLGMwt:
Oh nice! I just ran into this testing the preview render, didn't get as far as tracking it down.

FLGMwt (approved):

PR #3673, opened 2026-05-29
Adjust radius of circle when selecting a location

nebotron:
Today, if I type a location into the location selector, it keeps the same radius (several thousand kilometers) even if the location is much more specific. This change looks at the geojson of the place to find the smallest radius which contains the entire place. This saves a lot of manual pinching and zooming, and makes the location selector more useful.

Before:

before

After:

after

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
         "classnames": "^2.5.1",
         "date-fns": "^3.6.0",
         "date-fns-tz": "^3.2.0",
+        "haversine-distance": "^1.2.4",
         "i18next": "^26.0.4",
         "i18next-fluent": "^2.0.0",
         "i18next-resources-to-backend": "^1.2.1",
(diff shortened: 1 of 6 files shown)


nebotron:
@albullington - would you be willing to review this?

abbeycampbell:
Hello and thank you for your contribution! 
You noticed the same issue as our product team, and we actually had an open ticket to fix this in our current cycle. So, we have a PR landing that takes a slightly different approach, and I'm going to close yours, but I'll explain why if it's helpful.

Our version uses our existing `getMapRegion` helper (which already takes a bbox and returns a padded region) and reads the corners directly from the 5-point rectangle, which let's us avoid that extra dependency and all the extra calculations.
We appreciate the contribution and the thoroughness — the attention to testing in particular :)

nebotron:
Thanks for the response! I appreciate the context, and I'm glad you're working on this :)

I really appreciate what iNaturalist project is accomplishing, and I'd love to help, but it feels silly for me to duplicate effort. Do you think it would be at all feasible for me to get access to your Linear issue tracker, even in a read-only capacity? Let me know if you'd like to discuss this in a non-public forum, like email.

PR #3675, opened 2026-05-29
Add sort button to explore v2

jtklein:
Closes MOB-1191

The ticket says nothing about wiring it up, and what it is supposed to do is not yet possible on ExploreV2, so I assume this is purely cosmetic, i.e. to put the button in it's place.
Even that is  a future TODO though because the Results section does not switch between Obs/Species/Users yet, so probably the placement will have to change later as well.

### DIFF
--- a/src/components/Explore/ExploreV2/ExploreV2DebugSheet.tsx
+++ b/src/components/Explore/ExploreV2/ExploreV2DebugSheet.tsx
@@ -127,7 +127,7 @@ const ExploreV2DebugSheet = ( ) => {
       <INatIconButton
         icon="triangle-exclamation"
         className={classnames(
-          "absolute bottom-5 right-5",
+          "absolute top-5 right-5",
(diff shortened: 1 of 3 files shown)


FLGMwt (approved):
LGTM! The scope tracks with the 1pt estimate but @abbeycampbell can you confirm that's what was intended here?

abbeycampbell:
> LGTM! The scope tracks with the 1pt estimate but @abbeycampbell can you confirm that's what was intended here?

Yep, this ticket is only to slap the sort button where it's needed in ExploreV2. We have a separate ticket to handle the divergences between MyObs and Explore sortsheet UI and add that in, and then there will be further work to wire it up. 
I think ideally this would've happened after adding the header with obs and species tabs, but this is fine and I thing the inline todos tell us what still needs to be done

abbeycampbell (approved):

PR #3676, opened 2026-05-29
Display proper error message if project is invite only

jtklein:
Closes MOB-1189

The actual fix for this bug is small: show an Alert when membership_model is "inviteonly". However, was also migrating everything to TS that I touched here.
I did not migrate all the API calls though because our query wrappers are not typed yet.


src/api/types.d.ts
```diff
@@ -24,11 +24,22 @@ export interface ApiPost {
...
+  field: string;
+  value: string | null;
+}
+
 export interface ApiProject {
-  icon?: string;
-  id?: number;
-  project_type?: "collection" | "umbrella" | ""; // FYI "" means "traditional"
-  title?: string;
+  description: string;
```
jtklein, line 33: According to the API v2 docs icon, title, description, and project_type are nullable, however, on the website you can only create a new project with title and description being mandatory fields. And a fallback icon is given when none is provided. So, I am assuming these fields to be present all the time.

src/components/ProjectDetails/ProjectDetailsContainer.tsx
```diff
@@ -1,83 +1,81 @@
...
     optsWithAuth => fetchProjects( id, {
       fields: {
-        id: true,
-        title: true,
-        icon: true,
+        description: true,
         header_image_url: true,
+        icon: true,
+        id: true,
+        membership_model: true,
```
jtklein, line 41: This is the new field required for the actual bug fix of the ticket.

src/components/ProjectDetails/ProjectDetailsContainer.tsx
```diff
@@ -1,83 +1,81 @@
...
+        place_id: true,
```
jtklein, line 42: Unrelated bug fix: The code below is based on having this key but we were not actually querying for it.

src/components/ProjectDetails/ProjectDetailsContainer.tsx
```diff
@@ -1,83 +1,81 @@
...
         project_type: true,
-        description: true,
-        current_user_is_member: true,
+        rule_preferences: {
```
jtklein, line 44: Unrelated UI fix: ProjectDetails should format date preferences like the project list item does, so we also need to query for these.

src/components/ProjectDetails/ProjectDetailsContainer.tsx
```diff
@@ -1,83 +1,81 @@
...
 
-  const { data: projectPlace } = useAuthenticatedQuery(
+  const { data: projectPlace } = useAuthenticatedQuery<ApiPlace>(
     fetchProjectPlaceQueryKey,
     optsWithAuth => fetchPlace( project?.place_id, {
       fields: "all",
     }, optsWithAuth ),
   );
 
-  const { data: projectMembers } = useAuthenticatedQuery(
```
jtklein, line 54: All we cared for doing this query was it's total_result value, i.e. the number of users that have joined this project. A project does have `user_ids`, an array of the joined user's id. So, we can include this field and get rid of one API query.

src/components/ProjectDetails/ProjectDetails.tsx
```diff
@@ -38,21 +38,39 @@ const LEAVE = "LEAVE";
 
 const PROJECT_URL = `${Config.OAUTH_API_URL}/projects`;
 
-type Props = {
-  project: Object,
-  joinProject: Function,
-  leaveProject: Function,
-  loadingProjectMembership: boolean
+interface Project {
```
jtklein, line 41: Typed as to what fields we actually use in this component here.

src/components/ProjectDetails/ProjectDetails.tsx
```diff
@@ -215,7 +239,13 @@ const ProjectDetails = ( {
             <Button
               level="neutral"
               text={t( "JOIN" )}
-              onPress={() => setOpenSheet( JOIN )}
+              onPress={( ) => {
+                if ( project.membership_model === "inviteonly" ) {
```
jtklein, line 243: This is the actual fix for the ticket.

src/components/ProjectDetails/ProjectDetailsContainer.tsx
```diff
@@ -114,26 +112,30 @@ const ProjectDetailsContainer = ( ): Node => {
...
+    ( _, optsWithAuth ) => joinProject( { id }, optsWithAuth ),
     {
       onSuccess: ( ) => {
         queryClient.invalidateQueries( membershipQueryKey );
       },
       onError: error => {
-        logger.error( "could not join project: ", project.id, error );
+        // project is not undefined here because we call the mutation in the child
+        // which has a !project check before rendering the buttons that call here
+        logger.error( "could not join project: ", ( project as ApiProject ).id, error );
```
sepeterson, line 123: Appreciate the comment here, this is kind of a tricky case. Also not requesting a change but I learned you can assert something is not null or undefined like this `project!.id`

sepeterson (approved):
Looks good!

src/api/types.d.ts
```diff
@@ -24,11 +24,22 @@ export interface ApiPost {
   published_at: string;
 }
 
+interface ProjectRulePreference {
```
sepeterson: Also here's something interesting -- this isn't exported but it gets imported in ProjectDetails and ts seemingly doesn't mind. Weird

PR #3677, opened 2026-05-29
ObsImagePreview: inline render helpers

FLGMwt:
cherry-picked from work in [MOB-1458: Rework useLocalObservations to useLocalObservationUuids](https://linear.app/inaturalist/issue/MOB-1458/rework-uselocalobservations-to-uselocalobservationuuids) 

Doing my best to test downstream regressions for shifting the obs flashlist data flow. Performance testing shows that we were spending a _lot_ of time in ObsImagePreview and PhotoCount. It was unclear whether the callbacks themselves were causing issues. To make sure, I broke the render helpers into helper local components and inline conditions. 

It didn't end up having a material perf impact (turns out, our image count badge is expensive!), but I'm inclined to leave the refactor. We've discussed this a bit, but compared to render helpers, I think factoring Components like this is a bit more [semantic React](https://react.dev/learn/thinking-in-react#step-1-break-the-ui-into-a-component-hierarchy) and we can entirely avoid the "useMemo" / "useCallback" question.

commits broken down into one "concern" at a time

### DIFF
--- a/src/components/ObservationsFlashList/ObsImagePreview.tsx
+++ b/src/components/ObservationsFlashList/ObsImagePreview.tsx
@@ -119,31 +119,6 @@ const ObsImagePreview = ( {
     obsPhotosCount,
   ] );
 
-  const renderSelectable = useCallback( ( ) => {
-    if ( selectable ) {
-      return (
-        <View
-          className={classNames(
-            "flex items-center justify-center",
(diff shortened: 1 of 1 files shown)


jtklein (approved):
This is looking good.
> commits broken down into one "concern" at a time

Thanks, that was helpful.

PR #3678, opened 2026-05-29
Make use translation shim android only

FLGMwt:
cherry-picked from work in [MOB-1458: Rework useLocalObservations to useLocalObservationUuids](https://linear.app/inaturalist/issue/MOB-1458/rework-uselocalobservations-to-uselocalobservationuuids)

we're shimming useTranslation for an Android-only Java exception. I haven't validated if we still need to handle the Android exception, but we can at least skip the shim on iOS since this is called so frequently. This doesn't have a demonstrable impact on the performance of ObsFlashList.

### DIFF
--- a/src/sharedHooks/useTranslation.android.ts
+++ b/src/sharedHooks/useTranslation.android.ts
@@ -1,0 +1,23 @@
+import { useTranslation } from "react-i18next";
+
+console.log( "wrapped" );
+// Wrap react-i18next's useTranslation to catch a mysterious exception that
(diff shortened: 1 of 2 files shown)


jtklein (approved):
Oh, that is good. Can rejoice in the typed t function again.
Wondering if we do need this at all after so much time. Was just quickly checking if anyone had posted an issue with this error message but couldn't find one. So, I'd also be tempted to think we are doing sth wrong if only we have this error. Anyways, for a later time.

PR #3680, opened 2026-05-29
disable flashlist maintainVisibleContentPosition for render issues

FLGMwt:
opens and closes MOB-1464

also uncovered in "gotchas" docs looking at flashlist stuff: https://shopify.github.io/flash-list/docs/known-issues#3-data-re-ordering-can-cause-items-to-move This has annoyed me for a while but never found what I thought was a spacing issue.

My impression from the docs is that this configuration doesn't apply to our use of flashlist at the moment.

Before:

https://github.com/user-attachments/assets/f647dc37-09b9-4c13-8bb1-c76d8933a45f

https://github.com/user-attachments/assets/266daf6a-a5ae-49ad-ac83-79beb98e3f5f

After:

https://github.com/user-attachments/assets/4a650da0-b3e1-4ce2-b402-9cdc40ec2aa0

(lost the grid fixed vid, lemme know if I should shoot another one)

### DIFF
--- a/src/components/SharedComponents/FlashList/CustomFlashList.tsx
+++ b/src/components/SharedComponents/FlashList/CustomFlashList.tsx
@@ -163,7 +163,6 @@ const CustomFlashList = <T, >( props: FlashListProps<T> & { ref?: React.Ref<Flas
   return (
     <FlashList
       ref={internalRef}
-      initialNumToRender={5}
       onEndReached={handleEndReached}
       onEndReachedThreshold={0.2}
       onViewableItemsChanged={handleViewableItemsChanged}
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/FlashList/CustomFlashList.tsx
```diff
@@ -163,7 +163,6 @@ const CustomFlashList = <T, >( props: FlashListProps<T> & { ref?: React.Ref<Flas
   return (
     <FlashList
       ref={internalRef}
-      initialNumToRender={5}
```
FLGMwt, line 166: this appears to be a vestige of previous FlatList use. This is not a FlashList prop.

abbeycampbell: good catch

abbeycampbell (approved):
Nice one! I think this is the fix for some of our other 'blank spaces in lists' tickets too

PR #3682, opened 2026-05-30
Update PostListItem design

jtklein:
Small re-design of PostListItem. Makes extracting the excerpt from the HTML body obsolete.

Closes MOB-1465


sepeterson (approved):
👍

PR #3683, opened 2026-06-01
Update eslint-plugin-testing-library to 7.16.2

jtklein:
Used --fix on one test


tests/integration/navigation/broken/MyObservations.test.js
```diff
@@ -106,13 +106,13 @@ const writeObservationsToRealm = ( observations, message ) => {
   }, message );
 };
 
-const pressIndividualUpload = observation => {
+const pressIndividualUpload = async observation => {
```
FLGMwt, line 109: Seems like we should be awaiting this then: https://github.com/inaturalist/iNaturalistReactNative/blob/main/tests/integration/navigation/broken/MyObservations.test.js#L173

I know it's in a broken / disabled test but nice to do while we're here

FLGMwt (approved):
LGTM on the update, not a big deal if we skip the `await`

tests/integration/navigation/broken/MyObservations.test.js
jtklein: Thanks, I missed that.

PR #3686, opened 2026-06-01
fix first obs UI hang, prime date-fns-tz / intl cache in deferred startup

FLGMwt:
The first launch of ObsDetails hangs for ~100ms because of DateDisplay's call to [formatApiDatetime](https://github.com/inaturalist/iNaturalistReactNative/blob/6d28c3d420902d2a280cafcac384c95da9ac9a88/src/components/SharedComponents/DateDisplay.tsx#L70) which turned out to be yet another downstream issue of Hermes & Intl compatibility: https://github.com/facebook/hermes/discussions/1211. Only the first call per app launch is affected.

This PR does not address the problem, it just adds a deferred task throwaway call to `formatApiDatetime` so that the expensive first call does not affect a user interaction.

React Profiler results on Simulator of DateDisplay initial render. On real device, I was seeing like 133ms for DateDisplay 😬:

|Before 85.6ms |After 2.4ms|
|---|---|
|DateDisplay 85ms on first render|DateDisplay 2.4ms on first render| 

In both cases, subsequent renders are nominal (about 2ms for initial render, .5ms if Screen was recycled for a different obs).

### DIFF
--- a/src/components/hooks/useDeferredStartup.ts
+++ b/src/components/hooks/useDeferredStartup.ts
@@ -12,6 +12,7 @@
 import { cleanupLogFiles } from "components/Developer/logManagementHelpers";
 import { RealmContext } from "providers/contexts";
 import { useEffect } from "react";
+import { useTranslation } from "react-i18next";
 import {
   clearComputerVisionPhotos,
   clearGalleryPhotos,
@@ -19,6 +20,7 @@ import {
   clearRotatedOriginalPhotosDirectory,
   clearSyncedMediaForUpload,
 } from "sharedHelpers/clearCaches";
+import { formatApiDatetime } from "sharedHelpers/dateAndTime";
 import { log } from "sharedHelpers/logger";
 import { logSentinelFiles } from "sharedHelpers/sentinelFiles";
 import getStorageMetrics from "sharedHelpers/storageMetrics";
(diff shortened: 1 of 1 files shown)


src/components/hooks/useDeferredStartup.ts
```diff
@@ -12,13 +12,15 @@
 import { cleanupLogFiles } from "components/Developer/logManagementHelpers";
 import { RealmContext } from "providers/contexts";
 import { useEffect } from "react";
+import { useTranslation } from "react-i18next";
```
jtklein: That's one of the earlier imports in the app run, so we might need the custom wrapper which deals with a possible Android crash. (Suppose we need it at all.)

jtklein (approved):
LGTM

I did not check on the performance tab myself, because I didn't know how to start the app with the performance session recording already.

src/components/hooks/useDeferredStartup.ts
jtklein: import { useTranslation } from "sharedHooks";

FLGMwt: whoops, thanks!

FLGMwt:
> I did not check on the performance tab myself, because I didn't know how to start the app with the performance session recording already.

I did timings in dev to confirm that the 100ms cost was _indeed_ paid during the deferred task is that what you mean? Anyhow, you inspired me, PR in a sec

PR #3688, opened 2026-06-02
halted launch debug tool

FLGMwt:
Thanks for the inspiration @jtklein! https://github.com/inaturalist/iNaturalistReactNative/pull/3686#pullrequestreview-4405216245 

The app has to be launched for us to be able to attach a debugger or create a profiling session. That's annoying when we're interested in the actual launch itself!

This adds a Button control in Rozenite Controls to set a flag and restart the app:
image

On restart in `__DEV__`, we now conditionally set up the app root with wrapper that waits for dev to press a button before rendering the actual app root:
image

[Sample React Performance trace](https://github.com/user-attachments/files/28485737/initial-launch.json):

image

Bonus: adds `__DEV__` global type / eslint key

Creates & closes MOB-1485: Add capability to debug / profile app startup

### DIFF
--- a/src/components/Developer/Developer.tsx
+++ b/src/components/Developer/Developer.tsx
@@ -176,56 +176,54 @@ const DebugTools = () => {
         text="LOG IN AGAIN"
         className="mb-5"
       />
-      { // eslint-disable-next-line no-undef
-        __DEV__ && (
-          <>
-            <Button
-              onPress={() => navigation.navigate( "UILibrary" )}
-              text="UI LIBRARY"
-              className="mb-5"
-            />
-            <Button
-              onPress={() => { throw new Error( "Test error" ); }}
-              text="TEST ERROR"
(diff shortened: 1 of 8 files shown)


src/components/Developer/Developer.tsx
```diff
@@ -176,56 +176,54 @@ const DebugTools = () => {
...
-                } );
-              }}
-              text="TEST INATAPIERROR"
-              className="mb-5"
-            />
-            <Button
-              onPress={() => {
-                throw new INatApiTooManyRequestsError( {
-                  routeName: "TaxonDetails",
+      {__DEV__ && (
```
FLGMwt, line 179: big delta from moving an indentation level. Only change is deleting the eslint comment

jtklein (approved):
Have tested and this works on my end as well. LGTM

PR #3695, opened 2026-06-02
Update formatProjectDate callsites to guard early against null/undefined values

jtklein:
Closes MOB-1391

Started to do some additional TS typing here but that revealed some irregularities in ProjectRequirements.tsx in the parent - child contract, e.g. rule.name is translated string but RuleItem accesses an object based on the untranslated string. So, I am cutting my TS efforts short here which means that the interfaces added here are knowingly incomplete.


src/realmModels/Observation.js
```diff
@@ -30,6 +30,10 @@ class Observation extends Realm.Object {
     icon: true,
     title: true,
     project_type: true,
+    rule_preferences: {
```
jtklein, line 33: Some API queries used this field set, and so were not showing the formatted date-range correctly.

src/components/ProjectList/ProjectListItem.tsx
```diff
@@ -36,8 +41,8 @@ const ProjectListItem = ( { item, isHeader = false }: Props ) => {
     />
   );
 
-  const displayProjectIcon = icon => {
-    const productionIcon = icon?.replace( "staticdev", "static" );
+  const displayProjectIcon = ( ) => {
+    const productionIcon = item!.icon.replace( "staticdev", "static" );
```
FLGMwt: have we changed our opinion on non-null assertions?

previously: https://github.com/inaturalist/iNaturalistReactNative/pull/3257#discussion_r2608159949

jtklein: No, we haven't. I forgot, but changed to use cast as Project now.

src/components/ProjectDetails/ProjectRequirements.tsx
```diff
@@ -14,13 +16,19 @@ import AboutProjectType from "./AboutProjectType";
 import ProjectRuleItem from "./ProjectRuleItem";
 
 const getFieldValue = item => item?.[0]?.value;
+interface Project {
+  icon: string;
+  id: number;
+  project_type: "collection";
```
abbeycampbell, line 22: is it correct that only `collection` project types are allowed here?

abbeycampbell: if not, could we have a shared type between `ProjectRequirements.tsx` and `ProjectListItem.tsx`? They're both subsets of the `ApiProject` interface, so we could even do like: 
```
Pick<ApiProject, "icon" | "id" | "project_type" | "rule_preferences" | "title">
```

src/components/ProjectDetails/ProjectRequirements.tsx
```diff
@@ -242,8 +247,8 @@ const ProjectRequirements = ( ) => {
   }
 
   const dateRule = RULES.find( r => r.name === t( "Date" ) );
-  const { projectDate } = formatProjectDate( project, t, i18n );
-  if ( projectDate !== null ) {
+  if ( project?.rule_preferences ) {
+    const { projectDate } = formatProjectDate( project, t, i18n );
```
abbeycampbell, line 251: Small regression here: previously, when no dates were defined, we'd display 'Any' in the project requirements, just like all the other categories. But now it shows a blank row because we removed the fallback when we have project requirements, but no date requirements. I think we need the check for project dates too:
```
if ( projectDate !== null ) {
    dateRule.inclusions = [...];
  }
```

abbeycampbell (changes requested):
Sorry for delayed review, gotta fix my github notifications

src/components/ProjectDetails/ProjectRequirements.tsx
jtklein: 1) Depends a bit with how you mean allowed. It should be correct because we only ever route to this screen for collection projects, https://github.com/inaturalist/iNaturalistReactNative/blob/3fd95aec9eb6a01ca1aad8e8ea2b13c62588e61f/src/components/ProjectDetails/ProjectDetails.tsx#L213-L214
The screen does not do it's own check though, so if a project id for a non collection project is given it will fetch another type.
2) I am not sure the concept of ApiProject will survive long-term. We currently multiple screens that query for projects from the API. In some cases they use different sets of fields for API V2, so the object returned on each screen has different keys. I assume now this is the reason why we did end up with Api types that have optional keys because we pretend we can't be sure what the API returns because we want to have the type flexible for each screen. The flip-side of which is that I have seen some logic based on the expectation that a key in a response object can be there or not, when in reality it is always present on this particular screen.

src/components/ProjectDetails/ProjectRequirements.tsx
jtklein: Thanks for catching that.

src/components/ProjectDetails/ProjectRequirements.tsx
abbeycampbell: Got it. In that case, I feel like the separate interface definitions where they're needed is totally appropriate

abbeycampbell (approved):

PR #3698, opened 2026-06-03
Use our camera format filters on Android as well

jtklein:
Code part of MOB-1294 (not finished because it requires testing in production.)


FLGMwt:
@jtklein can you offer guidance on what I'm looking for in testing this?

jtklein:
> @jtklein can you offer guidance on what I'm looking for in testing this?

If code looks good, and if the camera looks alright on your Android phones. The ticket will be moved to Testing and we will see during testing if this looks okay on different Android manufacturer's phones.
It will probably look different before and after this PR, but the question is if it is still okay after or even better.

sepeterson (approved):
Tested both standard and cv camera on device ✅

PR #3700, opened 2026-06-03
Enable edgeToEdgeEnabled gradle property

jtklein:
Closes MOB-1277

The actual fix for the ticket above was to make the AddObsButton modal have statusBarTranslucent and navigationBarTranslucent true. In reading up on it I discovered that with the new edgeToEdgeEnabled=true gradle property (RN >0.81) these two Modal props are always true. https://github.com/react-native-community/discussions-and-proposals/discussions/921

So, I decided to flip it to true now also because I previously thought that having edgeToEdgeEnabled=false meant we are fully opt-out of the edge-to-edge display behaviour. However, that was incorrect, on Android 15+ apps are forced into the edge-to-edge display, and this gradle property sets wether Android version below 15 are also enabled to be displayed edge-to-edge. So, by setting to true now we should have a unified experience no matter what Android version. (Could not find information on a lower version bound of this behaviour, but have tested on SDK 30 and seems fine. SO, 87% of users should be fine.) 

The fix is a new gradle property so, you need to make a new native build to see the effect. This change has effects throughout the app, but I'd like to propose to fix them from this new baseline, since edge-to-edge display is enforced in newer Android versions (15+). In other words, this ticket will beget more tickets but only on Android.

Pixel 6 Android 16
Before
Screenshot_20260603-144627

After
Screenshot_20260603-224718

Samsung Android 11
Before
IMG_6785

After (this is correct because we have a different ticket to tackle the three button navigation bar hiding the app)
IMG_6784


android/gradle.properties
```diff
@@ -36,8 +36,8 @@ newArchEnabled=true
 # Use this property to enable or disable the Hermes JS engine.
 # If set to false, you will be using JSC instead.
 hermesEnabled=true
- 
+
 # Use this property to enable edge-to-edge display support.
 # This allows your app to draw behind system bars for an immersive UI.
 # Note: Only works with ReactActivity and should not be used with custom Activity.
-edgeToEdgeEnabled=false
+edgeToEdgeEnabled=true
```
FLGMwt, line 43: 💯

FLGMwt (approved):
Awesome! Looks great!!!!!!

The bar in your last "After" photo is captured by [MOB-1302: 3-button navigation bar covering UI on all screens](https://linear.app/inaturalist/issue/MOB-1302/3-button-navigation-bar-covering-ui-on-all-screens), correct?

jtklein:
Correct

PR #3702, opened 2026-06-03
MOB-1369: refactor useperformance to use an imperative stop

sepeterson:
Apologies for some scope creep here. I tried to keep things limited but wanted the after state to be something I was comfortable with.

Of note:
- Add imperative stop
- Move logging effect into hook
- import `performance` from react-native-performance

### DIFF
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -14,7 +14,6 @@ import LinearGradient from "react-native-linear-gradient";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { VolumeManager } from "react-native-volume-manager";
 import convertScoreToConfidence from "sharedHelpers/convertScores";
-import { log } from "sharedHelpers/logger";
 import { deleteSentinelFile, logStage } from "sharedHelpers/sentinelFiles";
 import { logFirebaseEvent } from "sharedHelpers/tracking";
 import {
@@ -39,8 +38,6 @@ import usePredictions from "./hooks/usePredictions";
 import LocationStatus from "./LocationStatus";
 
 const isTablet = DeviceInfo.isTablet();
-
-const logger = log.extend( "AICamera" );
 
 // const exampleTaxonResult = {
 //   id: 12704,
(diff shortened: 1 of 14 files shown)


src/sharedHooks/usePerformance.ts
```diff
@@ -1,39 +1,43 @@
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useRef } from "react";
+import performance from "react-native-performance";
```
FLGMwt, line 2: I don't know if `react-native-performance` predates RN including the `performance` global. Do you have thoughts about using the global vs the import? I think they implement the same spec.

src/sharedHooks/usePerformance.ts
```diff
@@ -1,39 +1,43 @@
...
-
-      setLoadTime( logMessage );
-    };
-    if ( startTime === 0 ) {
-      setStartTime( global.performance.now() );
-    }
-    if ( !isLoading ) {
-      getPerformanceReport();
+    startTime.current = performance.now( );
+  }, [] );
```
FLGMwt, line 21: if this effect is being called unconditionally, can we use it to initialize the ref? That is: remove the effect and just have `const startTime = useRef( performance.now( ) );`

src/components/Camera/AICamera/AICamera.js
```diff
@@ -142,14 +139,9 @@ const AICamera = ( {
 
   const { t } = useTranslation();
 
-  const { loadTime } = usePerformance( {
-    isLoading: camera?.current !== null,
+  const stop = usePerformance( {
```
FLGMwt, line 142: It'd be nice to give these returns a contextual name so 100 lines late it's clear what we're "stop"ping, along the lines of "stopPerformanceTimer" or "stopCameraTrace" or something.

src/sharedHooks/usePerformance.ts
```diff
@@ -1,39 +1,43 @@
...
+
+  const stop = useCallback( ( ) => {
+    if ( !isDebugMode( ) ) { return; }
```
FLGMwt, line 24: we should minimize direct store access. I think this should get the value from the hook and add it as an effect dependency.

src/sharedHooks/usePerformance.ts
```diff
@@ -1,39 +1,43 @@
...
+      : loadTimeMessage;
+    logger.info( logMessage );
+  }, [screenName] );
+
+  useEffect( ( ) => {
+    if ( isLoading === false ) {
+      stop( );
     }
-  }, [isLoading, startTime, screenName] );
+  }, [isLoading, stop] );
```
FLGMwt, line 38: does `isLoading` still make sense as an argument? Wouldn't this mean that `NetworkService` and `StartupService` immediately trigger this and log `Load Time: 0ms`?

sepeterson: Ya this was busted before I made these changes and I didn't fix it bc it seemed out of scope

PR #3703, opened 2026-06-03
Mob 1437 location picker map should zoom to fill location when

abbeycampbell:
closes [MOB-1437](https://linear.app/inaturalist/issue/MOB-1437/location-picker-map-should-zoom-to-fill-location-when-tapping-a-search)

one call out that on repeated/subsequent searches, especially when going from a lower zoom level to a higher one over a significant distance (think 'Norway' => 'Plano, TX') can result in map tiles not rendering/loading state until the user moves the map. I think in the future, we could just not worry about animating jumps beyond a certain threshold and just reset the view, but I think it's acceptable here because it seems highly unlikely that a user would need to do multiple location selects in wildly different parts of the world when selecting an obs location.

### DIFF
--- a/src/components/LocationPicker/LocationSearch.tsx
+++ b/src/components/LocationPicker/LocationSearch.tsx
@@ -20,6 +20,9 @@ interface Place {
   point_geojson: {
     coordinates: [number, number];
   };
+  bounding_box_geojson?: {
+    coordinates: [number, number][][];
+  };
 }
 interface Props {
   locationName: string;
--- a/src/components/LocationPicker/LocationSearch.tsx
+++ b/src/components/LocationPicker/LocationSearch.tsx
@@ -44,7 +44,7 @@ const LocationSearch = ( {
     ( optsWithAuth: ApiOpts ) => fetchSearchResults( {
       q: locationName,
       sources: "places",
-      fields: "place,place.display_name,place.point_geojson",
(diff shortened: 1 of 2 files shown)


src/components/LocationPicker/LocationSearch.tsx
```diff
@@ -44,7 +44,7 @@ const LocationSearch = ( {
     ( optsWithAuth: ApiOpts ) => fetchSearchResults( {
       q: locationName,
       sources: "places",
-      fields: "place,place.display_name,place.point_geojson",
+      fields: "place,place.display_name,place.point_geojson,place.bounding_box_geojson",
```
jtklein, line 50: Can you add this new field to the Place interface above.

jtklein (approved):
Functionality-wise okay I would say. I do see the loading indicator not disappearing also on the first search if I have no location. So, from 0,0 to something reasonable like Berlin. But I think it is okay to merge and deal with this later if it keeps bothering people.
PS: did a rebase and force push

FLGMwt:
> can result in map tiles not rendering/loading state until the user moves the map

Do you have an idea as to why this is happening? It would be cool to see _what_ causes a user-initiated move to "fix" the issue and maybe we can set that same state on navigate end or something?

But if this fixes the crux of the issue, maybe we can handle that separately.

PR #3704, opened 2026-06-04
MOB-1327: MVP explorev2 header

sepeterson:
We need some sort of header component so we can get to and work on UniversalSearch. This minimal component just shows ExploreV2Subject, ExploreV2LocationState, and the magnifying glass icon.

I also moved the debug icon out of the way

<details>
<summary>screenshot</summary>
Simulator Screenshot - iPhone 17 - 2026-06-04 at 11 43 16

</details>

### DIFF
--- a/src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
+++ b/src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
@@ -75,15 +75,15 @@ const ExploreV2Header = ( ) => {
           : null}
       </View>
       <INatIconButton
-        icon="magnifying-glass"
-        size={20}
-        color={colors.white}
-        className="bg-inatGreen rounded-md"
         onPress={() => navigation.navigate( "UniversalSearch" )}
         accessibilityLabel={t( "Search" )}
         accessibilityHint={t( "Opens-search-interface" )}
         testID="ExploreV2Header.searchButton"
-      />
+      >
+        <View className="w-10 h-10 bg-inatGreen rounded-md items-center justify-center">
+          <INatIcon name="magnifying-glass" size={20} color={colors.white} />
+        </View>
+      </INatIconButton>
     </View>
   );
 };
--- a/src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
+++ b/src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
@@ -53,8 +53,8 @@ const ExploreV2Header = ( ) => {
   const place = locationLabel( state.location, t );
 
   return (
-    <View className="bg-white px-6 py-4 flex-row justify-between items-center">
-      <View className="flex-1 mr-3">
+    <View className="bg-white p-4 flex-row justify-between items-center">
+      <View className="flex-1 mr-5">
         <Heading4 numberOfLines={1} ellipsizeMode="tail">
           {subject}
         </Heading4>
@@ -76,6 +76,7 @@ const ExploreV2Header = ( ) => {
       </View>
       <INatIconButton
         icon="magnifying-glass"
+        size={20}
         color={colors.white}
         className="bg-inatGreen rounded-md"
         onPress={() => navigation.navigate( "UniversalSearch" )}
--- a/src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
+++ b/src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
@@ -1,0 +1,90 @@
+import {
+  Body3,
+  Heading4,
+  INatIcon,
+  INatIconButton,
+} from "components/SharedComponents";
+import { View } from "components/styledComponents";
+import type { TFunction } from "i18next";
+import type { ExploreV2LocationState, ExploreV2Subject } from "providers/ExploreV2Context";
+import {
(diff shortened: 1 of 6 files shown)


src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+import {
+  EXPLORE_V2_PLACE_MODE,
+  useExploreV2,
+} from "providers/ExploreV2Context";
+import React from "react";
+import { useTranslation } from "sharedHooks";
+import colors from "styles/tailwindColors";
+
+interface Props {
+  onPressSearch: () => void;
```
jtklein: The ticket reads "A green search button (magnifying glass) on the right that opens Universal Search" so I don't think this needs to be a prop if it doesn't need custom behaviour. We can navigate directly from the header.

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </View>
+          )
+          : null}
+      </View>
+      <INatIconButton
+        icon="magnifying-glass"
+        color={colors.white}
+        className="bg-inatGreen rounded-md"
+        onPress={onPressSearch}
+        accessibilityLabel={t( "Search" )}
```
jtklein, line 79: I think this needs an a11yHint as well, because it is not obvious from the label what the button press does. As it is not performing a search immediately but navigates to more UI.

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+                numberOfLines={1}
+                ellipsizeMode="tail"
+              >
+                {place}
+              </Body3>
+            </View>
+          )
+          : null}
+      </View>
+      <INatIconButton
```
jtklein, line 77: Default button size of this is 44. Designs show 40x40. I think we landed on 44x44 for all icon buttons as our minimum supported size due to accessibility. So, I don't know if you have to get back to product and request a bigger button in the designs, or if the idea is to have the green square be 40x40 with an invisible outer part of the button stretching to 44x44.
In any case, the default icon size is 18 and the designs need a 20. Depending on the button size decisions above, if you keep a 44x44 green square potentially the ratio is more important and then design wants a 22px icon instead?

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+const ExploreV2Header = ( { onPressSearch }: Props ) => {
+  const { t } = useTranslation( );
+  const { state } = useExploreV2( );
+
+  const subject = subjectLabel( state.subject, t );
+  const place = locationLabel( state.location, t );
+
+  return (
+    <View className="bg-white px-6 py-4 flex-row justify-between items-center">
+      <View className="flex-1 mr-3">
```
jtklein: The gap between the UI item for what the subject is to the magnifying button is 20px in the designs. We can already put this here.

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+const ExploreV2Header = ( { onPressSearch }: Props ) => {
+  const { t } = useTranslation( );
+  const { state } = useExploreV2( );
+
+  const subject = subjectLabel( state.subject, t );
+  const place = locationLabel( state.location, t );
+
+  return (
+    <View className="bg-white px-6 py-4 flex-row justify-between items-center">
```
jtklein: The padding is uniformly 16px in the designs in x and y. Unless there is the left chevron for which the padding increases to 21. (But I think we should push back on this and keep it uniform at 16px left for the chevron as well. So, this View would need p-4.

jtklein: "justify-between" will not work for the version with chevron later?

jtklein (changes requested):
I haven't commented on the what seems to me temporary labels, but I think we should already make the button and container as the design specs are so we don't forget later.

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
sepeterson: This is true but I think we should wait to change it til we take the rest of the ticket up and actually add the chevron

sepeterson:
@jtklein Thanks! I think I addressed all your comments with the one exception

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
sepeterson: I went with designed approach of icon at 20, green square at 40x40 here because the 44x44 accessibility requirement typically concerns touchable area rather than visual size and we can retain that while still matching the design

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
jtklein: ok

jtklein (approved):
LGTM now

PR #3707, opened 2026-06-05
Update maestro e2e ci

jtklein:
We are currently having an error during the unit-test phase of the react-native-device-info library subproject:
Failing task: :react-native-device-info:testDebugUnitTest

We don't need to build unit tests for dependencies during our e2e testing, so, I restricted to only run "assemble" and not an entire "build" task.

Also updated the avd to be the same as in our detox tests.

Run passed here: https://github.com/inaturalist/iNaturalistReactNative/actions/runs/27059434129

### DIFF
--- a/.github/workflows/e2e_android.yml
+++ b/.github/workflows/e2e_android.yml
@@ -92,7 +92,7 @@ jobs:
         cd android
         # for the CI build, we can just target the x86 build for the emulator
         # also, we can exclude linting tasks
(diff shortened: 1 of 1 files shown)


FLGMwt (approved):
wooooooooo, this is awesome!

so much more makes sense for me with the confusing realization that `assemble` is a _subset_ of `build` and not the opposite.

PR #3709, opened 2026-06-08
MOB-1259: simplify login ref

sepeterson:
- Add scrollViewRef to Login and pass as a prop
- Remove children logic from LoginSignUpWrapper and add internalRef as a fallback for consumers that don't pass scrollViewRef


abbeycampbell (approved):
lgtm!

PR #3711, opened 2026-06-08
MOB-1469: pop back to Suggestions instead of pushing new screen

sepeterson:
Attached a video of the new behavior to the ticket: https://linear.app/inaturalist/issue/MOB-1469/back-button-behavior-flow-incorrect-on-both-cameras#comment-cf047523


jtklein (approved):
This fixes the ticket for me.
It does not fix other bugs with the same button, but I guess it should not do that because we have other tickets. E.g. from ObsEdit to Suggestions then back and another back prompts me to discard, even though I could get back to the camera.

PR #3712, opened 2026-06-09
MOB-1369: rm usePerformance

sepeterson:
This hook isn't giving us accurate numbers and doesn't have a clear path to becoming useful


FLGMwt:
+1, good riddance. the original ticket came out of a PR followup to refactor. But why refactor! thanks for identifying that this is poorly used!

FLGMwt (approved):
> +4 -160

^I love a negative delta

yeah, I'm with you. we had the kind of "tell the effect to run" pattern going on here

PR #3720, opened 2026-06-11
Fix broken integration tests

jtklein:
Closes MOB-879

Before every file move, commits "Move file without changes" I waited to see if "tests" run pass in the CI. I tried to split this giant Cursor session into commits that "make sense".

Most of the changes in the test files are refactors of shared steps into helper functions. I believe the actual fixes to make the tests pass again are, a new mock for our Mortal component, changing the way we use fake timers for the tests that involve screen transitions or Modal/BottomSheet usage (AddObsButton, AddObsBottomSheet).


tests/helpers/mockMortalForIntegration.js
```diff
@@ -0,0 +1,9 @@
+// Mortal gates children on navigation focus, which does not run reliably after
+// CommonActions.reset in integration tests. See MOB-879.
+jest.mock( "components/SharedComponents/Mortal", ( ) => {
```
sepeterson, line 3: Was this the biggest blocker for the broken tests?

tests/helpers/addObsBottomSheet.js
```diff
@@ -0,0 +1,192 @@
...
+  for ( let attempt = 0; attempt < 2; attempt += 1 ) {
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      const okButton = await screen.findByText( "OK", {}, { timeout: 2_000 } );
+      // eslint-disable-next-line no-await-in-loop
+      await actor.press( okButton );
+      // eslint-disable-next-line no-await-in-loop
+      await pressObsEditSaveButton();
+    } catch {
+      break;
```
sepeterson, line 166: Worth mentioning claude says this break can cause one of the PhotoImport tests to silently fail and recommends wrapping just the `findByText` line in the try/catch. Appears correct to me but certainly not a blocker

sepeterson (approved):
Kept review high-level but looks great!

PR #3721, opened 2026-06-11
MOB-754: fetch taxon bounding box improvements

sepeterson:
Research revealed the 429s from this endpoint to be much less of an issue than described in the ticket.

In the process, I discovered some low-effort QoL changes that I've PRed here

### DIFF
--- a/src/components/TaxonDetails/TaxonMapPreview.tsx
+++ b/src/components/TaxonDetails/TaxonMapPreview.tsx
@@ -1,4 +1,5 @@
 import { searchObservations } from "api/observations";
+import type { ApiTaxon } from "api/types";
 import {
   Button,
   DetailsMap,
@@ -11,15 +12,18 @@ import {
   View,
 } from "components/styledComponents";
 import React, { useState } from "react";
+import type { RealmTaxon } from "realmModels/types";
 import { useTranslation } from "sharedHooks";
 import useAuthenticatedQuery from "sharedHooks/useAuthenticatedQuery";
 
 import TaxonDetailsTitle from "./TaxonDetailsTitle";
 
+const ONE_HOUR_MS = 3600000;
+
 interface Props {
   observation: object;
   showSpeciesSeenCheckmark: boolean;
-  taxon: object;
+  taxon: ApiTaxon | RealmTaxon;
 }
 
 const TaxonMapPreview = ( {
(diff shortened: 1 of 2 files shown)


src/components/TaxonDetails/TaxonMapPreview.tsx
```diff
@@ -40,13 +44,20 @@ const TaxonMapPreview = ( {
...
       return_bounds: true,
       per_page: 0,
       ttl: -1,
     }, optsWithAuth ),
+    // The bounding box of a taxon's observations changes very slowly, so
+    // avoid refetching on every mount.
+    {
+      enabled: !!taxon.id,
+      gcTime: ONE_HOUR_MS,
+      staleTime: ONE_HOUR_MS,
```
sepeterson, line 59: staleTime keeps the cache "fresh" for 1 hour so it's valid to return for the same queryKey

gcTime keeps the cache from being garbage collected so it's available while still fresh

The response here is very small so I think it should be fine to keep them around for an hour. Here's a sample:
```
{
   "total_results":299350,
   "total_bounds":{"swlat":-54.5642,"swlng":-176.5670,"nelat":74.6297,"nelng":178.3165},
   "page":1,
   "per_page":0,
   "results":[]
}
```

src/components/TaxonDetails/TaxonMapPreview.tsx
```diff
@@ -40,13 +44,20 @@ const TaxonMapPreview = ( {
   const {
     data: obsSearchResponse,
   } = useAuthenticatedQuery(
-    ["fetchTaxonBoundingBox"],
+    ["fetchTaxonBoundingBox", taxon.id],
```
sepeterson, line 47: With the old key you could get a different taxon's cached bounding box resulting in a wonky map. With taxon.id add we can be sure it's unique to that taxon and thus good to keep around for a while

FLGMwt: killer find!

src/components/TaxonDetails/TaxonMapPreview.tsx
FLGMwt: I'm 100% on board w/ setting a non-zero `staleTime` to avoid the aggressive refetch-on-focus behavior.

I think based on https://stackoverflow.com/questions/72828361/what-are-staletime-and-cachetime-in-react-query it looks like the `gcTime` (used to be `cacheTime`) clock doesn't even start until every component holding onto that query key is unmounted. An hour seems like overkill but I suppose in most cases it's "as long as you're using the app don't clear this" and it probbbbbably doesn't hurt?

FLGMwt (approved):
sorry! Left another in-progress review yesterday ><

LGTM : D

PR #3723, opened 2026-06-12
Mob 1460 new myobs screen state foundation

abbeycampbell:
closes [MOB-1460](https://linear.app/inaturalist/issue/MOB-1460/new-myobs-screen-state-foundation)

Establishes a source of truth for MyObs sort and search state so future tickets can use it. Species sort now uses the new Context, but there should be no behavior changes.

Mirrors what we have going on in Explore V2.


src/components/MyObservations/MyObservationsContainer.tsx
```diff
@@ -71,6 +76,8 @@ const MyObservationsContainer = ( ) => {
   const taxaListRef = useRef<FlashListRef<SpeciesCount>>( null );
   const navigateToObsEdit = useNavigateToObsEdit( );
 
+  const { state: myObsState, dispatch: myObsDispatch } = useMyObservations( );
```
jtklein, line 79: myObsState.speciesSort could be destructured if you are expecting more states

src/providers/MyObservationsContext.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from "react";
+import { OBSERVATIONS_SORT } from "sharedHelpers/observationsSort";
```
jtklein, line 2: Does this and SPECIES_SORT need to be imported as type? I see you import SPECIES_SORT as type in MyObsContainer.

src/providers/MyObservationsContext.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface MyObservationsContextValue {
+  state: MyObservationsState;
+  dispatch: ( action: MyObservationsAction ) => void;
+}
+
+const MyObservationsContext = React.createContext<
+  MyObservationsContextValue | undefined
+>( undefined );
+
+interface MyObservationsProviderProps {
```
jtklein: Are you expecting any more props? If not we could use PropsWithChildren directly.

src/providers/MyObservationsContext.tsx
```diff
@@ -0,0 +1,106 @@
...
+    speciesSort: SPECIES_SORT;
+  }
+  | {
+    type: MY_OBSERVATIONS_ACTION.SET_TAXON_SEARCH;
+    searchTaxon: MyObservationsTaxon;
+  }
+  | { type: MY_OBSERVATIONS_ACTION.CLEAR_TAXON_SEARCH };
+
+export const initialMyObservationsState: MyObservationsState = {
+  observationsSort: OBSERVATIONS_SORT.DATE_UPLOADED_NEWEST,
```
jtklein, line 40: I understand that obs sort and taxon will be wired up later?

jtklein (approved):
LGTM
It does feel a bit over the top to test how a reducer works, like to dispatch one action and test that the state is changed according to the action.

src/providers/MyObservationsContext.tsx
abbeycampbell: Correct

src/components/MyObservations/MyObservationsContainer.tsx
abbeycampbell: it will be once I add obs sort!

src/providers/MyObservationsContext.tsx
abbeycampbell: No, SPECIES_SORT is only being used as a ts type in `MyObservationsContainer` but here we need the enum values

PR #3726, opened 2026-06-16
Zooming on media viewer not working on android

jtklein:
Closes MOB-232

This PR:
- replaces FLatList with Carousel (which we already have as a dependency).
- Resolves conflicts between multiple GestureHandlers by restricting 1 finger to pan or swipe-to-close, and 2 fingers to zoom.
- Migrates all I touched to TS

### DIFF
--- a/src/components/MediaViewer/AttributionButton.tsx
+++ b/src/components/MediaViewer/AttributionButton.tsx
@@ -4,7 +4,6 @@ import {
 import React from "react";
 import { Alert } from "react-native";
 import useTranslation from "sharedHooks/useTranslation";
-import colors from "styles/tailwindColors";
 
 interface Props {
   attribution?: string;
@@ -28,7 +27,6 @@ const AttributionButton = ( {
           ? "creative-commons"
           : "copyright"
       }
-      color={colors.white}
       accessibilityLabel={usableAttribution}
       optionalClasses={optionalClasses}
     />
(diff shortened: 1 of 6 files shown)


src/components/MediaViewer/AttributionButton.tsx
```diff
@@ -28,7 +27,6 @@ const AttributionButton = ( {
           ? "creative-commons"
           : "copyright"
       }
-      color={colors.white}
```
jtklein, line 31: There is no such prop

src/components/MediaViewer/MainMediaDisplay.tsx
```diff
@@ -0,0 +1,205 @@
+import SoundContainer from "components/ObsDetailsSharedComponents/Media/SoundContainer";
```
jtklein, line 1: I did a commit to move the file without changes first, but I guess the diff is too big to see this as a TS file migration.

jtklein (commented):
Comments

src/components/MediaViewer/MainMediaDisplay.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+interface PhotoItem {
+  attribution?: string;
+  licenseCode?: string;
+  localFilePath?: string;
+  type: "photo";
+  url: string;
+}
+
+interface SoundItem {
```
sepeterson, line 30: Could MediaSelector and MediaView share SoundItem and PhotoItem? Looks like they're pretty similar

sepeterson (approved):
Tested on device and works well! Just the 1 question then good to merge

src/components/MediaViewer/MainMediaDisplay.tsx
jtklein: Probably, I haven't thought about where to export it from yet.

PR #3729, opened 2026-06-16
filter missing basics on demand for manual sync

FLGMwt:
The scope of #3690 ballooned too much, splitting for better reviewability & testing. Part of MOB-1458

This is part of whittling down what MyObsContainer needs from mapped obs so we can limit to just IDs. We only need the actual resolved `missingBasicsIds` at the moment we're sync-ing for default mode users. Following PRs will remove the dependence on mapped obs for the `numUnuploadedObsMissingBasics` as well (which is used to decide if we want the "missing basics" banner).


abbeycampbell (approved):
👍 sorry for the conflicts. 
I don't think I realized until this PR that we allow advanced mode users to upload with date/location missing.

PR #3739, opened 2026-06-17
MOB-1327 explore header redesign part 1

sepeterson:
In this PR

- Styling to match designs
- Thumbnail images
- Respects common/scientific name preference
- updated test data

Part 2 will contain the UI for the observations/species tabs and some unit tests

<details>
Simulator Screenshot - iPhone 17 - 2026-06-16 at 22 00 04
Simulator Screenshot - iPhone 17 - 2026-06-16 at 21 59 56
Simulator Screenshot - iPhone 17 - 2026-06-16 at 21 53 36

</details>


src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -14,9 +19,11 @@ import {
   useExploreV2,
 } from "providers/ExploreV2Context";
 import React from "react";
-import { useTranslation } from "sharedHooks";
+import { useCurrentUser, useTranslation } from "sharedHooks";
 import colors from "styles/tailwindColors";
 
+const THUMBNAIL_CLASS = "w-[62px] h-[62px] rounded-lg";
```
abbeycampbell, line 25: not for this PR, but it feels like it'd be nice to have this be a shared class at some point, since we use this style in a few places

abbeycampbell (approved):
LGTM!

jtklein (commented):
I think we should re-use the existing components as much as we can here for consistency.
The general idea for this result bar would be to replace the second text line of: UserListItem, TaxonResult, ProjectListItem.

This is one of the places in the app where I would have liked to use the Explore V2 project as a project where we talk through how we want to composite components that re-use existing designs.
I haven't looked through the code but I think we could have something like this for the user info (subclass names are all made up):

In ExploreV2 header:
```
<UserListItem>
  <UserListItem.UserIcon />
  <UserListItem.TextSection>
    <UserListItem.UserName />
    <PlaceName /> (Or whatever because it is only relevant here)
  </UserListItem.TextSection>
</UserListItem>
```

In Advanced Search (as the selected user info), Search Users (in the list as one result) and general in lists that show users:
```
<UserListItem>
  <UserListItem.UserIcon />
  <UserListItem.TextSection>
    <UserListItem.UserName />
    <UserListItem.UserObservations />
  </UserListItem.TextSection>
</UserListItem>
```

sepeterson:
> This is one of the places in the app where I would have liked to use the Explore V2 project as a project where we talk through how we want to composite components that re-use existing designs.

@jtklein I think we can do that! I like your high level approach a lot.

I would like to keep momentum on ExploreV2 going so I think we should split this discussion out from this ticket. A reusable solution should be relatively simple to apply after the fact.

I think this also could be an interesting opportunity to try collaborating and figuring things out via LLM code sketches. I pushed a [very large PR here](https://github.com/inaturalist/iNaturalistReactNative/pull/3748) that was the result of my claude session on this topic. I wouldn't suggest you read directly but rather digest it with your LLM and maybe we can decide on an approach via code as a team

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
sepeterson: Good flag! Followup: https://github.com/inaturalist/iNaturalistReactNative/pull/3749

PR #3742, opened 2026-06-17
MOB-967: add account settings entry point to UserProfile

sepeterson:
Split out the account settings webview logic to a reusable hook and add a pencil icon to UserProfile as a new entrypoint


src/sharedHooks/useNavigateToAccountSettings.ts
```diff
@@ -0,0 +1,103 @@
...
+const SETTINGS_URL = `${Config.OAUTH_API_URL}/users/edit?noh1=true`;
+const FINISHED_WEB_SETTINGS = "finished-web-settings";
+
+interface Options {
+  // Called when the user returns from the web settings view (e.g. to refetch
+  // any profile data that may have changed)
+  onFinish?: ( ) => void;
+}
+
+const useNavigateToAccountSettings = ( { onFinish }: Options = {} ) => {
```
sepeterson, line 26: To limit the scope of this ticket I tried to mirror the functionality as it existed in LoggedinInDefaultSettings as closely as possible

sepeterson:
Mentioned in stand-up that this isn't easily testable in staging. Posted a video of some testing on the linear ticket

src/sharedHooks/useNavigateToAccountSettings.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    navigation.navigate( "FullPageWebView", {
+      title: t( "ACCOUNT-SETTINGS" ),
+      loggedIn: true,
+      initialUrl: SETTINGS_URL,
+      blurEvent: FINISHED_WEB_SETTINGS,
+      clickablePathnames: ["/users/delete"],
+      skipSetSourceInShouldStartLoadWithRequest: true,
+      shouldLoadUrl: url => {
+        async function signOutGoHome() {
```
FLGMwt, line 77: nit but a reminder for me: do we have a consistent style preference for inline function definitions?

src/sharedHooks/useNavigateToAccountSettings.ts
```diff
@@ -0,0 +1,103 @@
...
+          Alert.alert(
+            t( "Account-Deleted" ),
+            t( "It-may-take-up-to-an-hour-to-remove-content" ),
+          );
+          // sign out
+          await signOut( { realm, clearRealm: true, queryClient } );
+          // revert back to default mode
+          setIsDefaultMode( true );
+          // navigate to My Obs
+          navigation.navigate( "ObsList" );
```
FLGMwt, line 87: doesn't signout reset the app? I'm curious why we would need the following actions.

FLGMwt:
> Mentioned in stand-up that this isn't easily testable in staging. Posted a video of some testing on the linear ticket

@sepeterson can you clarify what the limitation is? are we not able to point the webview to staging?

src/sharedHooks/useNavigateToAccountSettings.ts
sepeterson: I believe you're correct that this is unnecessary. It's just a copy/paste of the existing logic into the reusable hook.

sepeterson:
> > Mentioned in stand-up that this isn't easily testable in staging. Posted a video of some testing on the linear ticket
> 
> @sepeterson can you clarify what the limitation is? are we not able to point the webview to staging?

The trouble is that to access the staging site you need to provide HTTP login info first. The webview won't prompt for it like the browser does. You can pass the HTTP login info via the auth header but then you can't pass the JWT so the web view opens a web page where you aren't signed in

src/components/UserProfile/UserProfile.tsx
```diff
@@ -46,10 +49,13 @@ const UserProfile = ( ) => {
...
 
   const fetchId = userId || login;
-  const { data: remoteUser, isError, error }: {
+  const {
+    data: remoteUser, isError, error, refetch: refetchUser,
+  }: {
     data: ApiUser | null;
     isError: boolean;
     error: INatApiError | ErrorWithResponse;
+    refetch: ( ) => void;
```
jtklein, line 58: Just out of curiosity: Why?

jtklein (approved):
This works on my end. Have tested in a production build, UserProfile routes to the full page WebView, changing my bio was promptly reflected in UserProfile when coming back.
I can see that the refactor into shared hook is for the large part copy-paste.

Have not tested on staging, and thusly not tested the account deletion sign out flow. But since it is properly copied over as it was I don't see a reason to check it as well.

src/components/UserProfile/UserProfile.tsx
sepeterson: The hook just needs some way to fetch the user again after the webview is closed

sepeterson:
@jtklein Thanks for testing!

PR #3743, opened 2026-06-17
dedicated hook for "unsynced" and "missing basics" observations

FLGMwt:
Key insight of this whole blasted thing (getting ids only) was to separate two very disparate query sources:
- _all_ Realm observations
- Realm observations _needing sync_

We were previously handing the needing sync concern while we were iterating over _all_ obs which made it tricky to pull apart and optimize _all_ obs to get just IDs.

This PR:
- removes reliance on useLocalObs / mapped Realm observations to calculate and be reactive to unsync'd observations and obs missing basics (location / date)
- removes the zustand key `numUnuploadedObservations`. This can get out of sync and should be derived from the data. This can now be read from the new count hook which is react reactive.

later: "needs_sync" is on the chopping block too. It's a similarly flaky / manually sync'd property on Observation. We should probably switch to just using the filter (now an exported constant filter string).


src/sharedHooks/useObservationCounts.ts
```diff
@@ -0,0 +1,43 @@
...
+      type: Observation,
+      query: observations => observations.filtered( UNSYNCED_FILTER ),
+      keyPaths: KEY_PATHS,
+    },
+  );
+
+  return useMemo(
+    ( ) => ( {
+      unsyncedObservationsCount: unsyncedObs.length,
+      observationsMissingBasicsCount: unsyncedObs
```
abbeycampbell: nitpick: we end up with three similar names for this value across this hook and our components when we destructure it as `numObsMissingBasics` in the MyObsContainer and then again to `numUnuploadedObsMissingBasics` in MyObsSimple. Any reason why we shouldn't have a consistent name all the way through?

tests/realm.setup.js
```diff
@@ -35,6 +35,7 @@ jest.mock( "providers/contexts", ( ) => {
     RealmContext: {
       ...originalModule.RealmContext,
       useRealm: ( ) => global.realm,
+      useQuery: ( ) => [],
```
abbeycampbell, line 38: I think this mock is the cause of the failing tests, manually testing this branch locally shows the expected results.

abbeycampbell (approved):
One nitpick but largely lgtm! Going to go ahead and approve pending passing tests ofc

src/sharedHooks/useObservationCounts.ts
```diff
@@ -0,0 +1,43 @@
...
+  observationsMissingBasicsCount: number;
+}
+
+const useObservationCounts = ( ): ObservationCounts => {
+  const unsyncedObs = useQuery(
+    {
+      type: Observation,
+      query: observations => observations.filtered( UNSYNCED_FILTER ),
+      keyPaths: KEY_PATHS,
+    },
```
jtklein, line 32: Should we add an empty dependency array just for documentation sake?

src/sharedHooks/useObservationCounts.ts
```diff
@@ -0,0 +1,43 @@
+import { RealmContext } from "providers/contexts";
+import { useMemo } from "react";
+import Observation, { UNSYNCED_FILTER } from "realmModels/Observation";
+
+const { useQuery } = RealmContext;
+
+// only re-evaluate unsync'd obs / missing basics when these change
+const KEY_PATHS = [
```
jtklein, line 8: I do like the idea of restricting this to only necessary re-renders. I am afraid though of this drifting away from the definition of UNSYNCED_FILTERS in Observation. SO, if someone adds a different definition of what "unsynced" means there we need to remember to also track that here, or we will see stale data.

jtklein (approved):
From what I can see the only functional change here is the addition of KEY_PATHS, so that the calculation of how many unsynced obs do we have is driven by the query listener and only happening for a subset of key changes, instead of before the calculation happening during the obs mapping in useLocalObservations.
I do think this is okay, but I am still worried that this core functionality is a bit under-tested at the moment while we are making changes to how it works.
PS: Have not looked at the test changes because they are red at the moment.

src/sharedHooks/useObservationCounts.ts
FLGMwt: agreed, better to use existing names.

src/sharedHooks/useObservationCounts.ts
FLGMwt: Good idea, fixed & I'll do that for queries going forward

src/sharedHooks/useObservationCounts.ts
FLGMwt: That's a good callout. I've updated this to reflect the new projects fields with that in mind.

Maybe this could be co-located with the exported filter? I'm on the fence since the keypath here is specific to this query. I'm leaning on leaving it here for now but evaluating with our forward use of the realm hooks.

PR #3744, opened 2026-06-18
Update vision-plugin to 6rc3

jtklein:
Over creative day I had a bit of time to fix some smaller bugs with our vision-plugin while working in the code.

This is the entire diff for this update: https://github.com/inaturalist/vision-camera-plugin-inatvision/compare/5cb5194...main

This includes a big chunk of repo scaffolding updates from earlier this year, but the "client-side" relevant changes are the introduction of a model provider class on iOS and a few minor bugfixes in src/index.tsx and src/lookUpLocation.ts (e.g. few places on the earth used a wrong elevation as input).

The biggest wanted fix is the model provider on iOS:
Before if one session of the app used the AI camera as well as offline cv predictions from a photo we put the cv model into memory twice.
Screenshot 2026-06-18 at 11 44 17

After: Only once per session.
Screenshot 2026-06-18 at 11 11 31

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -3931,7 +3931,7 @@ PODS:
   - VisionCamera/React (4.7.3):
     - React-Core
     - VisionCamera/FrameProcessors
-  - VisionCameraPluginInatVision (6.0.0-rc.2):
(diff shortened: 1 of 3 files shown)


FLGMwt (approved):
very nice! good find

PR #3745, opened 2026-06-18
Use bottom insets instead of absolute positioning for AddObsTooltip

jtklein:
Closes MOB-1277 (Again)

This was reintroduced by adding bottom insets to the CustomTabBar.


sepeterson (approved):
LGTM!

Image

PR #3746, opened 2026-06-18
MOB-1531 add nitro modules dep to test for impact

FLGMwt:
This PR: adds the `react-native-nitro-modules` core dependency to support NitroModules-based native modules. no functional changes, just added to verify independent impact on build / app behavior.

Closes MOB-1531, subissue of MOB-1159

---

A number of our dependencies are moving to implementations dependent on [NitroModules](https://nitro.margelo.com/), an OSS alternative to the RN-official [TurboModules](https://reactnative.dev/docs/turbo-native-modules-introduction) approach for writing native modules (see also: [ExpoModules](https://docs.expo.dev/modules/overview)).

Those libraries include at least:

- react-native-vision-camera
- react-native-audio-recorder-player
- react-native-mmkv
- react-native-sensitive-info
- [prospective?] react-native-nitro-image (as a potential option for https://github.com/candlefinance/faster-image which is now unmaintained)

We expect that non-Nitro versions of those libraries may be left inactive (fine) and not be kept up to date to be compatible with new React Native versions (**less fine**).

In discussion with @jtklein, we've recognized the importance of caution in adopting NitroModules. It introduces a factor of vendor lock-in to a new, "non-official", shared dependency for multiple key packages, _especially_ rn-vision-camer, for which we have the critical [vision-camera-plugin-inatvision](https://github.com/inaturalist/vision-camera-plugin-inatvision) plugin.

As a first step for investigating / validating NitroModules for iNatRN, this PR pulls in the core NitroModules dependency for a build & soak test. We just want to confirm that it existing in the app doesn't break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "react-native-mmkv": "^3.3.3",
     "react-native-modal": "^14.0.0-rc.1",
     "react-native-modal-datetime-picker": "^18.0.0",
+    "react-native-nitro-modules": "^0.35.9",
     "react-native-open-maps": "^0.4.3",
     "react-native-orientation-locker": "github:wonday/react-native-orientation-locker",
     "react-native-paper": "^5.12.3",
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -179,6 +179,35 @@ PODS:
     - nanopb/encode (= 3.30910.0)
   - nanopb/decode (3.30910.0)
   - nanopb/encode (3.30910.0)
+  - NitroModules (0.35.9):
+    - boost
+    - DoubleConversion
+    - fast_float
+    - fmt
+    - glog
+    - hermes-engine
+    - RCT-Folly
+    - RCT-Folly/Fabric
+    - RCTRequired
+    - RCTTypeSafety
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -128,6 +128,7 @@
     "react-native-mmkv": "^3.3.3",
     "react-native-modal": "^14.0.0-rc.1",
     "react-native-modal-datetime-picker": "^18.0.0",
+    "react-native-nitro-modules": "^0.35.9",
```
jtklein, line 131: I don't like that this is a version 0.x

jtklein (approved):
Have compiled and started the app in Debug and Release mode on Android and iOS, and haven't see any issues during building.
Not sure what other kind of review you'd like here.

FLGMwt:
> Have compiled and started the app in Debug and Release mode on Android and iOS, and haven't see any issues during building. Not sure what other kind of review you'd like here.

Yeah, nothing expected, just tagged you since it came out of our conversation. Once it seems fine in a build, I'll take a look at which nitro dependency makes the most sense (probably still mmkv as we discussed, assuming nothing tricky with data migration).

PR #3748, opened 2026-06-18
Reusable list item

sepeterson:
DO NOT MERGE -- this is just for discussion


sepeterson:
This achieves the composable with props thing but if I feed it back to claude without that goal it flags the Contexts as overcomplicating things, which I tend to agree with. Maybe we pass some props? And compose from a set of naive components from CompositeListItem?

PR #3749, opened 2026-06-18
one thumbnail class

sepeterson:


src/components/Notifications/ObservationIcon.tsx
jtklein: Please re-use here as well.

jtklein: Please re-use here as well.

src/components/ProjectList/ProjectListItem.tsx
jtklein, line 66: Not your code, and already broken on main, but I just realized that the thumbnail class is not actually being applied here.

src/components/SharedComponents/DisplayTaxon.tsx
```diff
@@ -32,7 +33,7 @@ const DisplayTaxon = ( {
   const { t } = useTranslation( );
   const currentUser = useCurrentUser( );
 
-  const imageClassName = "w-[62px] h-[62px] rounded-lg";
+  const imageClassName = THUMBNAIL_CLASS;
```
jtklein, line 36: Not your code, but this is giving a TS error in line 72.

jtklein (changes requested):
Great small PR!
Only one thing I would request, to re-use in the other places in ObservationIcon as well for consistency. The other comments you can treat as just comments.
After that feel free to merge on your own time.

PR #3752, opened 2026-06-19
Traditional Projects: API types and wrappers

jtklein:
Closes MOB-1491

What is included:
* Consolidated all different versions of fields we queried for when getting projects into two new const in `fields.ts`. This is a marked divergence from prior code where the fields we query for are either at callsite directly or in the Realm model class that the queried data should be persisted as. I think having the declaration close to the call site is mostly fine but it leads to harder to type data I think. Having the fields definitions close to the realm models contributes to the confusion as to what data type API or realm are we actually passing down as prop in some places.
I'd like to use Projects which are currently solely retrieved from API but in this project have to be added to the realm persisted models as a new blueprint how to divide API and Realm classes.
So, for this PR please review if this new structure makes sense as well.

Other changes:
* Added fields and types that will be required for `project_observation_fields`. You can check the new fields with logging `project` in ProjectDetailsContainer and ProjectRequirements, and `userProjects` in ProjectListContainer.
* Added fields and types that will be required for `project_observations` and differentiated `non_traditional_projects` from it. You can see them by logging `remoteObservation` in ProjectListContainer.
* Endpoints wrapper for ObservationFieldValues (OFV) and ProjectObservations (PO) which will have to be wired up later. If you want to test them from within the app, I was asking an agent to add six debug buttons in the Developer menu to call them with example data, the IDs in the unit tests are real IDs from production that should work. Observation probably has to be one of yours (although I don't know if you it should be possible to add one of my observations to a project as a site admin)

Left some comments in some commit msg

### DIFF
--- a/src/api/fields.ts
+++ b/src/api/fields.ts
@@ -13,21 +13,29 @@ const PROJECT_OBSERVATION_FIELDS_FIELDS = {
   required: true,
 };
 
-export const PROJECT_FIELDS = {
-  description: true,
-  header_image_url: true,
+export const PROJECT_SUMMARY_FIELDS = {
   icon: true,
   id: true,
-  membership_model: true,
-  place_id: true,
-  project_observation_fields: PROJECT_OBSERVATION_FIELDS_FIELDS,
   project_type: true,
   rule_preferences: {
     field: true,
     value: true,
   },
   title: true,
+};
+
+export const PROJECT_DETAIL_FIELDS = {
+  ...PROJECT_SUMMARY_FIELDS,
+  description: true,
+  header_image_url: true,
+  membership_model: true,
+  place_id: true,
   user_ids: true,
+};
+
+export const PROJECT_SUMMARY_POF_FIELDS = {
+  ...PROJECT_SUMMARY_FIELDS,
+  project_observation_fields: PROJECT_OBSERVATION_FIELDS_FIELDS,
 };
 
 export const PROJECT_FIELDS_ALL = "all";
@@ -41,7 +49,7 @@ export const OBSERVATION_FIELD_VALUE_FIELDS = {
 
 export const PROJECT_OBSERVATION_FIELDS = {
   id: true,
-  project: PROJECT_FIELDS,
+  project: PROJECT_SUMMARY_FIELDS,
   project_id: true,
   uuid: true,
 };
--- a/src/api/fields.ts
+++ b/src/api/fields.ts
@@ -31,3 +31,10 @@ export const PROJECT_FIELDS = {
 };
 
 export const PROJECT_FIELDS_ALL = "all";
+
+export const OBSERVATION_FIELD_VALUE_FIELDS = {
+  id: true,
+  observation_field: OBSERVATION_FIELD_FIELDS,
+  uuid: true,
+  value: true,
+};
--- a/src/api/fields.ts
+++ b/src/api/fields.ts
@@ -1,9 +1,9 @@
 const OBSERVATION_FIELD_FIELDS = {
+  allowed_values: true,
+  datatype: true,
+  description: true,
   id: true,
   name: true,
-  datatype: true,
-  allowed_values: true,
-  description: true,
 };
 
 const PROJECT_OBSERVATION_FIELDS = {
--- a/src/api/fields.ts
+++ b/src/api/fields.ts
@@ -6,7 +6,7 @@ const OBSERVATION_FIELD_FIELDS = {
   name: true,
 };
 
-const PROJECT_OBSERVATION_FIELDS = {
+const PROJECT_OBSERVATION_FIELDS_FIELDS = {
   id: true,
   observation_field: OBSERVATION_FIELD_FIELDS,
   position: true,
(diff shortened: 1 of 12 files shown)


src/api/fields.ts
```diff
@@ -0,0 +1,47 @@
...
+  place_id: true,
+  project_observation_fields: PROJECT_OBSERVATION_FIELDS_FIELDS,
+  project_type: true,
+  rule_preferences: {
+    field: true,
+    value: true,
+  },
+  title: true,
+  user_ids: true,
+};
```
abbeycampbell: this adds a bunch of keys we didn't previously use. I'm wondering what the implications are for the two places where we might not need all this extra stuff: `/realmModels/Observation.js` and `ProjectListContainer.tsx`, particularly since I imagine `user_ids` can get quite large for some projects.

Would it make sense to instead make two constants that are scoped to include different key? One for project summaries/overviews, one for project details?

src/api/types.d.ts
```diff
@@ -178,13 +195,24 @@ export interface ApiNotification {
...
+  project_id: number;
+  uuid: string;
+}
+
+// When using OBSERVATION_FIELD_VALUE_FIELDS
+export interface ApiObservationFieldValue {
+  id: number;
+  observation_field: ApiObservationField;
+  uuid: string;
+  value: string;
```
abbeycampbell, line 217: this is typed as `string | number` in the ObservationFieldValueAttributes interface

src/api/fields.ts
```diff
@@ -0,0 +1,47 @@
+const OBSERVATION_FIELD_FIELDS = {
```
sepeterson, line 1: Having the fields here is much clearer. Great change

abbeycampbell (changes requested):

abbeycampbell:
oops sorry @sepeterson didn't see you'd started reviewing this one

src/realmModels/Observation.js
```diff
@@ -73,11 +63,9 @@ class Observation extends Realm.Object {
     private_location: true,
     private_place_guess: true,
     project_ids: true,
-    project_observations: {
-      project: Observation.PROJECT_FIELDS,
-    },
+    project_observations: PROJECT_OBSERVATION_FIELDS,
```
sepeterson, line 66: Doesn't this mean that *every* observation with projects fetched via useRemoteObservation would now be newly returning `PROJECT_OBSERVATION_FIELDS` with nested `PROJECT_FIELDS` with nested `PROJECT_OBSERVATION_FIELDS_FIELDS` with nested `OBSERVATION_FIELD_FIELDS`?

In short the response would get significantly larger for ObsDetails for observations with many projects. Do we need all this info immediately or should there be a separate fetch to get the specifics on project?

tests/jest.setup.js
```diff
@@ -81,6 +81,12 @@ inatjs.observations.search.mockResolvedValue( makeResponse( ) );
 inatjs.observations.speciesCounts.mockResolvedValue( makeResponse( ) );
 inatjs.observations.updates.mockResolvedValue( makeResponse( ) );
 inatjs.users.projects.mockResolvedValue( makeResponse( ) );
+inatjs.project_observations.create.mockResolvedValue( makeResponse( ) );
```
sepeterson, line 84: Is there a reason to mock these globally on setup?

sepeterson:
> oops sorry @sepeterson didn't see you'd started reviewing this one

Haha all good! At least we both flagged the same thing.

sepeterson:
@jtklein sorry about the redundant review! Overall I think this looks great -- just want to be sure on the obs fields thing

src/api/types.d.ts
jtklein: Yes, I believe this is correct: ObservationFieldValueAttributes are write params and the API seems to accept either string or number https://api.inaturalist.org/v2/docs/#/ObservationFieldValues/post_observation_field_values
When querying the model is defined as only string here: https://api.inaturalist.org/v2/docs/#model-ObservationFieldValue

tests/jest.setup.js
jtklein: Yes, I am pretty sure I will need this for follow-up tickets, so just to be sure I don't forget later.

src/realmModels/Observation.js
jtklein: Yes, it did mean that. I have now added different flavors of project query fields.
Thanks for pointing this out. This is indeed a thing I needed to fix in this PR. Better to be conservative now with fetching data until we implement in future work which API calls need to be wired up to retrieve more data to persist projects.
I think with the changes I made now, we should be back at the fields set we queried for in the different screens before this PR. Plus one additional set we are not using yet but will be.

src/api/fields.ts
jtklein: Yes, it does make sense, and I have now added different flavors of project query fields.
Thanks for pointing this out. This is indeed a thing I needed to fix in this PR. Better to be conservative now with fetching data until we implement in future work which API calls need to be wired up to retrieve more data to persist projects.
I think with the changes I made now, we should be back at the fields set we queried for in the different screens before this PR. Plus one additional set we are not using yet but will be.

abbeycampbell (approved):
Thanks! lgtm

PR #3754, opened 2026-06-22
convert StatTab to ts

sepeterson:
Helpful change related to MOB-1327

### DIFF
--- a/src/components/MyObservations/StatTab.tsx
+++ b/src/components/MyObservations/StatTab.tsx
@@ -8,7 +8,13 @@ import { useTranslation } from "sharedHooks";
 
 export const OBSERVATIONS_TAB = "observations";
 
-const StatTab = ( { id, numTotalObservations, numTotalTaxa } ) => {
+interface Props {
+  id: string;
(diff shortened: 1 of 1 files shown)


src/components/MyObservations/StatTab.tsx
```diff
@@ -8,7 +8,13 @@ import { useTranslation } from "sharedHooks";
 
 export const OBSERVATIONS_TAB = "observations";
 
-const StatTab = ( { id, numTotalObservations, numTotalTaxa } ) => {
+interface Props {
+  id: string;
+  numTotalObservations?: number;
```
jtklein, line 13: Having these two numbers as undefined seems consistent with higher order components. The string that we use to display them seems to fall through to the "other" variant when the stat is undefined. So, this typing looks correct to me. Do we want to wait with rendering though until these are defined?

Anyways, looking at this now the entire API of this component and the Tabs.tsx it is propped in seems a bit convoluted.

jtklein (approved):
LGTM

PR #3755, opened 2026-06-22
MOB-1327 Obs and species tabs UI

sepeterson:
In this PR:

- Add Observation and Species tabs for ExploreV2
- Add hook to fetch species count
- refine styling
- some tests


src/appConstants/tabs.ts
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const OBSERVATIONS_TAB = "observations";
```
sepeterson, line 2: There's a couple other places where we can import this that I will add in a follow-up PR

src/sharedHooks/useSpeciesCount.ts
```diff
@@ -0,0 +1,39 @@
...
+  lng?: number;
+  radius?: number;
+  verifiable?: boolean;
+}
+
+interface Options {
+  enabled?: boolean;
+}
+
+const useSpeciesCount = (
```
sepeterson, line 22: There's at least 1 other spot where we can drop this in directly which I will also put in a followup PR

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -105,11 +106,14 @@ const ExploreV2Header = ( ) => {
 
   return (
     <View className="bg-white" testID="ExploreV2Header">
-      <View className="p-4 flex-row items-center">
+      <View className="px-4 pt-4 pb-2 flex-row items-center">
```
jtklein: Not sure where the pb-2 is coming from. In the design I see consistently a padding of 16px between the subject and StatsTabs or the green button and StatsTab without subject.

src/components/SharedComponents/StatTab.tsx
```diff
@@ -1,20 +1,21 @@
...
 import React from "react";
 import { useTranslation } from "sharedHooks";
 
-export const OBSERVATIONS_TAB = "observations";
-
 interface Props {
   id: string;
   numTotalObservations?: number;
   numTotalTaxa?: number;
+  className?: string;
```
jtklein: Because we use nativewind the className prop is referring only to the next styled component and not propped down I believe. So, a prop like so never reaches the part below where it should override the default.
If you need a prop to override below you can use sth like wrapperClassName which we should have examples for in the codebase.
However, I don't see a difference between the MyObs tabs and ExploreV2 tabs in terms of spacing, so not sure why we would need this here.

src/components/Explore/ExploreV2/components/ExploreV2Tabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+const ExploreV2Tabs = ( { observationsCount, speciesCount }: Props ) => {
+  const { t } = useTranslation( );
+
+  const renderTabComponent = useCallback(
+    ( { id }: TabComponentProps ) => (
+      <StatTab
+        id={id}
+        numTotalObservations={observationsCount}
+        numTotalTaxa={speciesCount}
+        className="py-1.5"
```
jtklein: See below

src/components/Explore/ExploreV2/components/ExploreV2Tabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+            onPress: ( ) => undefined, // MOB-1332
+          },
+          {
+            id: SPECIES_TAB,
+            text: t( "Species" ),
+            testID: "ExploreV2Tabs.species",
+            onPress: ( ) => undefined, // MOB-1332
+          },
+        ]}
+        TabComponent={renderTabComponent}
```
jtklein, line 52: I understand you wanted to keep existing code as close to other examples, but I do find the API quite convoluted. We define a Component we want to render for all tabs, prop in all different stats (currently two, later four), then switch based on which id we have to decide which stat to display.
Rendered component could be a per tabs.item based function.

src/components/Explore/ExploreV2/components/ExploreV2Tabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+// switching between observations and species views is deferred to MOB-1332.
+const ExploreV2Tabs = ( { observationsCount, speciesCount }: Props ) => {
+  const { t } = useTranslation( );
+
+  const renderTabComponent = useCallback(
+    ( { id }: TabComponentProps ) => (
+      <StatTab
+        id={id}
+        numTotalObservations={observationsCount}
```
jtklein, line 25: Seeing TS errors about null here and next line.

src/sharedHooks/useSpeciesCount.ts
```diff
@@ -0,0 +1,39 @@
+import type { QueryKey } from "@tanstack/react-query";
+import { fetchSpeciesCounts } from "api/observations";
+import type { ApiParams, ApiResponse } from "api/types";
+import useAuthenticatedQuery from "sharedHooks/useAuthenticatedQuery";
+
+export interface SpeciesCountParams extends ApiParams {
+  taxon_id?: number;
```
jtklein, line 7: All four of these ids can potentially also be arrays of ids according to the API V2 docs. I don't think ExploreV2 entails being able to select multiple subjects of the same kind, so not sure if we need to change anything here. But we could if we want to.

src/sharedHooks/useSpeciesCount.ts
```diff
@@ -0,0 +1,39 @@
...
+interface Options {
+  enabled?: boolean;
+}
+
+const useSpeciesCount = (
+  params: SpeciesCountParams,
+  queryKey: QueryKey,
+  { enabled = true }: Options = {},
+): number | null => {
+  const countParams = { ...params, per_page: 0 };
```
jtklein: I think most (all?) of Explore V1 API calls use a `ttl: -1` param to always get the latest data and no cache on the server.

jtklein (approved):
Only minor changes requested, see inline comments. But also fine if none of these actually apply.

tests/unit/components/Explore/ExploreV2/components/ExploreV2Tabs.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+  it( "renders the provided observation and species counts", () => {
+    renderComponent( <ExploreV2Tabs observationsCount={42} speciesCount={7} /> );
+
+    expect( screen.getByText( "42" ) ).toBeTruthy();
+    expect( screen.getByText( "7" ) ).toBeTruthy();
+  } );
+
+  it( "renders a placeholder when counts are not provided", () => {
+    renderComponent( <ExploreV2Tabs /> );
```
jtklein, line 33: Does that also apply when a count is null or only undefined?

src/components/Explore/ExploreV2/components/ExploreV2Tabs.tsx
sepeterson: Should be fixed 👍

src/sharedHooks/useSpeciesCount.ts
sepeterson: Good to know! I'm going to leave this info in a comment since there aren't plans to use it like this yet

src/sharedHooks/useSpeciesCount.ts
sepeterson: Good spot! Adding

src/components/Explore/ExploreV2/components/ExploreV2Tabs.tsx
sepeterson: Ya I agree. I added a note to the Species tab ticket so I can make a better solution for this with those changes https://linear.app/inaturalist/issue/MOB-1332/species-tab#comment-d5318e58

tests/unit/components/Explore/ExploreV2/components/ExploreV2Tabs.test.js
sepeterson: Yep -- added a test

src/components/SharedComponents/StatTab.tsx
sepeterson: The reason this is necessary is that statTab has a bunch of padding around the text that isn't in the designs. Somehow I didn't notice className was getting magically disappeared. Renamed to wrapperClassName for the merge

PR #3756, opened 2026-06-22
Traditional projects: realm models and schema updates

jtklein:
Closes MOB-1492

### DIFF
--- a/src/realmModels/Observation.js
+++ b/src/realmModels/Observation.js
@@ -26,9 +26,7 @@ export const UNSYNCED_FILTER
   + " || ANY observationPhotos._synced_at == null"
   + " || ANY observationSounds._synced_at == null"
   + " || ANY projectObservations._synced_at == null"
-  + " || ANY observationFieldValues._synced_at == null"
-  + " || ANY projectObservations._synced_at <= projectObservations._updated_at"
-  + " || ANY observationFieldValues._synced_at <= observationFieldValues._updated_at";
+  + " || ANY observationFieldValues._synced_at == null";
 
 // noting that methods like .toJSON( ) are only accessible when the model
 // class is extended with Realm.Object per this issue:
--- a/src/realmModels/Observation.js
+++ b/src/realmModels/Observation.js
@@ -527,6 +527,7 @@ class Observation extends Realm.Object {
       latitude: "double?",
       license_code: { type: "string", mapTo: "licenseCode", optional: true },
       longitude: "double?",
+      observationFieldValues: "ObservationFieldValue[]",
       observationPhotos: "ObservationPhoto[]",
       observationSounds: "ObservationSound[]",
       // date and/or time submitted to the server when a new obs is uploaded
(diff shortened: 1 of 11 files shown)


src/realmModels/Observation.js
```diff
@@ -21,6 +21,15 @@ export const GEOPRIVACY_PRIVATE = "private";
 
 const logger = log.extend( "index.js" );
 
+export const UNSYNCED_FILTER
```
jtklein, line 24: Refactored into this const mostly to avoid merge conflicts with https://github.com/inaturalist/iNaturalistReactNative/pull/3743

src/realmModels/Observation.js
```diff
@@ -527,6 +530,7 @@ class Observation extends Realm.Object {
       latitude: "double?",
       license_code: { type: "string", mapTo: "licenseCode", optional: true },
       longitude: "double?",
+      observationFieldValues: "ObservationFieldValue[]",
```
jtklein, line 531: Actual wiring of OFV and PO from a remote obs into local persistence in a later ticket A4 (MOB-1497)

src/realmModels/Observation.js
```diff
@@ -21,6 +21,15 @@ export const GEOPRIVACY_PRIVATE = "private";
 
 const logger = log.extend( "index.js" );
 
+export const UNSYNCED_FILTER
+  = "_synced_at == null || _synced_at <= _updated_at"
+  + " || ANY observationPhotos._synced_at == null"
+  + " || ANY observationSounds._synced_at == null"
+  + " || ANY projectObservations._synced_at == null"
+  + " || ANY observationFieldValues._synced_at == null"
+  + " || ANY projectObservations._synced_at <= projectObservations._updated_at"
```
sepeterson: I'm out of my wheelhouse a bit here but claude flagged this and the next line as having an unintended behavior that seems correct to me. It seems that as written this would resolve true if *any* projectObservation has _updated_at >= another projectObservation's _synced_at.

Something like `min(_synced_at across all POs) <= max(_updated_at across all POs)`

Its suggested fix is to use a subquery like `SUBQUERY(projectObservations, $po, $po._synced_at == null || $po._synced_at <= $po._updated_at).@count > 0`

sepeterson:
@jtklein Lookin good for the most part! Just the 1 question and then feel free to merge.

src/realmModels/Observation.js
jtklein: Extend my gratitude to your claude. Someone whispered in my ear to be more defensive here than required.
I was trying to be defensive for a later ticket when we can update an OFV after we have already synced the obs. Planning around a child being unsynced while the obs is not marked as unsynced.
In general your claude is probably correct with my wrong statement, but it also reveals that we have a theoretical gap in obsPhotos and obsSounds. If the obs is treated as synced and one of the obsPhotos is once synced (= !null) and then edited and not synced anymore, the obs would count as not needing sync. In practice we always flip the obs as needs sync whenever we do an update on a child. Because it goes through `saveLocalObservationForUpload`, which bumps obs' _updated_at as well, so obs matches `_synced_at <= _updated_at`

So, I have removed this defensive check and will add to the ticket that marks PO/OFVs as unsynced to make sure we also mark the obs as synced on an update to a PO/OFV.

PR #3759, opened 2026-06-22
Mob 1462 search myobs UI

abbeycampbell:
Starts to add UI for [MyObs search](https://linear.app/inaturalist/issue/MOB-1462/search-myobs-ui)

Breaking it up into multiple PRs so it's less overwhelming to review. This one adds Search to the MyObs header behind a feature flag, gates the feature for offline users, and adds navigation to a new search screen, for which I emulated the approach we took for Explore V2, where we give the tab screen it's own stack navigator (most of the big diff here is moving a bunch of logic and UI from MyObservationsContainer to the new MyObservationsResults screen).


src/components/MyObservations/Search/SearchMyObservationsTaxon.tsx
```diff
@@ -0,0 +1,85 @@
+import { useNavigation } from "@react-navigation/native";
+import type { ApiTaxon } from "api/types";
+// TODO: ExploreSearchHeader is a generic back/title/reset button header that previously
+// was only used in explore. It should be promoted to components/SharedComponents now that it's
+// shared. The cross-feature import below works for now but will be fixed in a follow-up.
+import ExploreSearchHeader from "components/Explore/SearchScreens/ExploreSearchHeader";
```
sepeterson, line 6: Appreciate the todo

sepeterson (approved):
Looks great! Very excited for this feature

src/components/AddObsBottomSheet/AddObsButton.js
```diff
@@ -46,7 +46,7 @@ const AddObsButton = ( ): React.Node => {
       // Base trigger condition in all cases:
       // Only show the tooltip if the user has the AI camera as the default button option.
       // Only show the tooltip on MyObservations screen.
-      const onObsList = currentRoute?.name === "ObsList";
+      const onObsList = currentRoute?.name === "MyObservationsResults";
```
FLGMwt, line 49: if we want to fully rename "ObsList" to "MyObsResults" on _both_ sides of the feature flags, I think our var names should come along too. Should this be `isOnMyObsResults`? (I added the `is` because `on{stuff}` sounds like an event handler.

src/sharedHooks/useRozenite.tsx
```diff
@@ -122,6 +126,15 @@ const useRozenite = ( { queryClient, storageAdapters }: RozeniteOptions ) => {
               setTraditionalProjectsEnabled( !traditionalProjectsEnabled );
             },
           },
+          {
+            id: "search-my-observations",
+            type: "toggle",
+            title: "Search My Observations",
+            value: searchMyObservationsEnabled,
+            onUpdate: () => {
+              setSearchMyObservationsEnabled( !searchMyObservationsEnabled );
```
FLGMwt, line 135: not your change, but both of these toggling functions should be based on contemperaneous state: `enabled => !enabled`

FLGMwt (approved):
LGTM! non-blocking naming / setState comments

src/sharedHooks/useRozenite.tsx
FLGMwt: oh no, the worst thing has happened. It's my change! 

Disregard, I'll address in a followup

src/components/AddObsBottomSheet/AddObsButton.js
abbeycampbell: will make this change in my next chunk of this work

src/sharedHooks/useRozenite.tsx
abbeycampbell: setOverride is typed to take a boolean though? maybe I'm confused as to what you mean by `enabled => !enabled`.

`sections` is memoized too, with these enabled values in the dependency array. So I don't think we have to worry about stale values here, the sections should re-compute when `enabled` changes.

FLGMwt: yeah, I'm sorry, I realized afterwards that they were zustand wrappers, not local state. I probably should have exposed them as toggles rather than boolean setters. They're dev-only so I won't worry about them. 

> the sections should re-compute when enabled changes

Even better: I'm wrong! You were definitely right here, not necessary. https://react.dev/reference/react/useState#is-using-an-updater-always-preferred

PR #3765, opened 2026-06-24
Traditional projects: PO and OFV download mapping for remote observations

jtklein:
Closes MOB-1497

I mistakenly added `projectId` to the realm schema for an OFV. I thought we'd need to tie an OFV to a specific PO/project, but they are global pieces of evidence at the time of making an observation. So, this PR also removes the projectId field from OFV realm schema and thusly bumps realm schema version.

We currently have no mechanism to decide before we query for an observation if we need to persist it or not, i.e. if it is created by currentUser or not. E.g. in useRemoteObservation we check if we should upsert in an effect depending on the fetched obs data.
This means that we have to query for PO and OFV wherever we might call `Observation.upsertRemoteObservations`, so fields need to be added to Observation.FIELDS but also to ADVANCED_MODE_LIST_FIELDS. Which does mean we will query for these info also for other users' obs where we currently (and not in scope of Trad Project in app) do not need them.

After OFV and PO in local realm for one obs.
Screenshot 2026-06-24 at 17 33 25
Screenshot 2026-06-24 at 17 48 15


src/realmModels/Observation.js
```diff
@@ -180,6 +189,20 @@ class Observation extends Realm.Object {
     const taxon = obs.taxon
       ? Taxon.mapApiToRealm( obs.taxon, realm )
       : null;
+
+    const observationFieldValues = (
```
jtklein, line 193: Following obsPhotos

src/realmModels/Observation.js
```diff
@@ -208,6 +231,17 @@ class Observation extends Realm.Object {
       return mappedObsSound;
     } );
 
+    const projectObservations = ( obs.project_observations || [] ).map( apiPo => {
```
jtklein, line 234: Following obsPhotos

src/realmModels/Observation.js
```diff
@@ -208,6 +231,17 @@ class Observation extends Realm.Object {
...
+      const mappedPo = ProjectObservation.mapApiToRealm( apiPo );
+      const existingPo = existingObs?.projectObservations?.find(
+        ePo => ePo.uuid === apiPo.uuid?.toLowerCase( ),
```
abbeycampbell: why does this uuid need to be lowercased, but the ones for observationFieldValues, photos, and sounds do not?

abbeycampbell (approved):
Do we have a separate ticket to add a `currentUser` gate before we actually get to the point where traditional projects in the app will allow us to edit these fields? I definitely think we don't want to be storing them locally for other people's observations, right? Is it worth adding a TODO comment somewhere?

This is a nitpick/personal preference, so feel free to disregard (and maybe it is less and less important in a time where we increasingly rely on LLMs to deliver code and perform code review anyway): I always prefer when variable names are written out in plain english, even when it means things get long--I just feel like it helps with readability a lot. so like `mappedPo` becomes `mappedProjectObservation`. I also feels like it allows for more nuance with naming, like it makes it easier to discern at a glance whether i'm thinking in the singular (`mappedProjectObservation`) or if we're dealing with an array of objects (`mappedProjectObservations`). Just a thought but again feel free to ignore.

src/realmModels/Observation.js
jtklein: Have removed

jtklein:
Re var lenght:
For me length of const names is probably highly correlated with breadth of scope they are attached to. So, since all of these new const heres do not leave the mapApiToRealm scope I am fine with that.

jtklein:
> Do we have a separate ticket to add a `currentUser` gate before we actually get to the point where traditional projects in the app will allow us to edit these fields? I definitely think we don't want to be storing them locally for other people's observations, right? Is it worth adding a TODO comment somewhere?

The TradProj POD scope is quite big so maybe I missed it, but I don't think it will be changing how observations are persisted in the app. So, wherever there is a current user gate it will stay like so. So, since I believe, we only persist obs of the current user nothing will change for obs for other users. In other words, we are not planning to persist these fields for other users' obs because we are not persisting other users' obs.
The actual UI to edit the fields for the current user's obs will be behind a login check.

PR #3767, opened 2026-06-25
Traditional projects: joined projects sync to realm

jtklein:
Closes MOB-1496

New persistence code and mapApiToRealm code added by closely following the same code for Observation realm model. With one major deviance, we do not store a _created_at or _synced_at date for Projects, POF, and OF because we currently have no plans of adding project admin functions (CRUD) to the app. The only purpose of this persisted info is to allow to add obs to projects while offline.

As a proof-of-concept, I wired up the ProjectList and Projects screens to persist all projects the current user has joined. This is not final because we probably should decouple fetching all joined projects from a certain UI state and place it on app start instead.

After: Navigating to projects list persists project summary data into realm:
Screenshot 2026-06-25 at 14 40 41
Screenshot 2026-06-25 at 14 43 07

### DIFF
--- a/src/components/Projects/ProjectsContainer.tsx
+++ b/src/components/Projects/ProjectsContainer.tsx
@@ -1,6 +1,8 @@
 import { PROJECT_SUMMARY_FIELDS, PROJECT_SUMMARY_POF_FIELDS } from "api/fields";
 import isEmpty from "lodash/isEmpty";
-import React, { useCallback, useState } from "react";
+import { RealmContext } from "providers/contexts";
+import React, { useCallback, useEffect, useState } from "react";
+import Project from "realmModels/Project";
 import {
   useCurrentUser,
   useLocationPermission,
@@ -20,7 +22,10 @@ export enum TAB_ID {
   NEARBY = "NEARBY"
 }
 
+const { useRealm } = RealmContext;
+
 const ProjectsContainer = ( ) => {
+  const realm = useRealm( );
   const [searchInput, setSearchInput] = useState( "" );
   const currentUser = useCurrentUser( );
   const memberId = currentUser?.id;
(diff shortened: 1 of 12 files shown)


src/realmModels/Project.ts
```diff
@@ -1,6 +1,43 @@
...
+      projectObservationFields: pofs,
+      // Leaving this here explicitly as a reminder to potentially need to refactor project_type
+      // into projectType across the entire codebase for consistency with other data models
+      project_type: apiProject.project_type,
+    };
+
+    return localProject;
+  }
+
+  static upsertRemoteProjects( apiProjects: ApiProjectSummaryWithPOF[], realm: Realm ) {
```
sepeterson, line 26: Will there be a mechanism to delete projects a user has left?

src/components/Projects/ProjectsContainer.tsx
```diff
@@ -62,6 +71,13 @@ const ProjectsContainer = ( ) => {
     enabled: !isEmpty( apiParams ),
   } );
 
+  // Update local copy of a user's joined projects
+  useEffect( () => {
+    if ( currentTabId === TAB_ID.JOINED ) {
+      Project.upsertRemoteProjects( projects, realm );
```
sepeterson: If I am on the JOINED tab and I type in a search, I think this effect is saving all the results to realm. I added this log
```
  useEffect( () => {
    if ( currentTabId === TAB_ID.JOINED ) {
      console.log("Projects effect: ", projects)
      Project.upsertRemoteProjects( projects, realm );
    }
  }, [currentTabId, projects, realm] );
```

I and see this on console after typing a search (I only have 2 joined projects)

Image

I know you mentioned that this logic will likely move to app start but probably makes sense to guard against this now just in case

sepeterson:
@jtklein Looks good! just the one concern on the JOINED tab effect. Feel free to merge after that's addressed

src/components/Projects/ProjectsContainer.tsx
jtklein: Thanks for catching that. You are right. I have reverted the proof-of-concept persisting of projects from ProjectContainer now. We do still have the one in ProjectListContainer to try it out.

src/realmModels/Project.ts
jtklein: Yes.
On JOIN a project will be persisted, on LEAVE it will be removed, and general sync actions TBD.

PR #3768, opened 2026-06-25
MOB-1339 ExploreV2 search autocomplete

sepeterson:
UI etc for universal search autocomplete. Sorry for the big PR. It's mostly tests

There are some default search options that I am deferring to another PR


src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
```diff
@@ -0,0 +1,102 @@
...
+import useCurrentUser from "sharedHooks/useCurrentUser";
+import useTranslation from "sharedHooks/useTranslation";
+import type { UniversalSearchResultItem } from "sharedHooks/useUniversalSearch";
+
+interface Props {
+  result: UniversalSearchResultItem;
+  onPress: ( ) => void;
+}
+
+const resultId = ( result: UniversalSearchResultItem ): number | undefined => {
```
sepeterson, line 18: There's a reusable ID extractor function to be added for this stuff but I gotta add it later

src/components/Explore/ExploreV2/helpers/universalSearchSubject.ts
```diff
@@ -0,0 +1,63 @@
...
+  subject: ExploreV2Subject,
+  commonNameIsPrimary: boolean,
+): string => {
+  switch ( subject.type ) {
+    case "user":
+      return subject.user.login;
+    case "project":
+      return subject.project.title;
+    case "taxon":
+    default:
```
jtklein: Does this need to be the default? If not a { type: "taxon", taxon: Taxon } is passed in accessing subject.taxon.name could throw an exception?

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
```diff
@@ -0,0 +1,109 @@
...
+                count: result.user.observations_count || 0,
+              } )}
+              pressable={false}
+            />
+          </View>
+        );
+      case "project":
+        return (
+          <View className="flex-1">
+            <ProjectListItem item={result.project} />
```
jtklein, line 80: The spacing here between image and text is bigger compared to the others. Not sure if that means we have to adapt the baseline of the shared component or the design here.

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const currentUser = useCurrentUser( );
+
+  const renderContent = ( ) => {
+    switch ( result.type ) {
+      case "taxon": {
+        const photo = result.taxon.default_photo?.url;
+        return (
+          <View className="flex-row items-center flex-1">
+            {photo
+              ? (
```
jtklein, line 41: Don't you want to re-use TaxonResult here?

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    <Pressable
+      accessibilityRole="button"
+      accessibilityLabel={resultLabel( result )}
+      className="flex-row items-center justify-between px-4 py-2 border-b border-lightGray"
+      onPress={onPress}
+      testID={`UniversalSearchResult.${result.type}.${resultId( result )}`}
+    >
+      {renderContent( )}
+      <INatIconButton
```
jtklein, line 102: For TaxonResult we'd have that as a prop to display and hook up the press. For Project and User we do not have that as a prop in the list item component. Makes me wonder if we should consolidate these approaches and add the info button to the list item.

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
```diff
@@ -0,0 +1,109 @@
...
+        );
+      default:
+        return null;
+    }
+  };
+
+  return (
+    <Pressable
+      accessibilityRole="button"
+      accessibilityLabel={resultLabel( result )}
```
jtklein: Have quickly tested this in a screen reader. Nesting these two buttons like so means the accessibility of the info button will be "swallowed", i.e. not visible to the screen reader.
Have not tested my suggestions, and they are only from distant memory, so better to actually test again in the screen reader: If the a11y label is supposed to be the actual text that is inside the Pressable, I think we can just remove this Pressable being accessible. Otherwise you'd have to follow using "link" as a11y type, e.g. like in TaxonResult.

tests/unit/components/Explore/ExploreV2/components/UniversalSearchResult.test.js
```diff
@@ -0,0 +1,115 @@
...
+  from "components/Explore/ExploreV2/components/UniversalSearchResult";
+import initI18next from "i18n/initI18next";
+import React from "react";
+import { renderComponent } from "tests/helpers/render";
+
+jest.mock( "sharedHooks/useCurrentUser", ( ) => ( {
+  __esModule: true,
+  default: jest.fn( ),
+} ) );
+const useCurrentUser = require( "sharedHooks/useCurrentUser" ).default;
```
jtklein: You can also just have an import `import * as useCurrentUser from "sharedHooks/useCurrentUser";` at the top and the mock works.

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -26,26 +39,104 @@ const INPUT_BOX_CLASSES = classnames(
...
+    setSubjectText( subjectToText( subject, commonNameIsPrimary ) );
+    setFilledFromSelection( true );
+    setQueryImmediately( "" );
+    dispatch( { type: EXPLORE_V2_ACTION.SET_SUBJECT, subject } );
+    locationInputRef.current?.focus( );
+  }, [commonNameIsPrimary, dispatch, setQueryImmediately] );
 
   const handleReset = useCallback( ( ) => {
-    setTaxonText( "" );
+    setSubjectText( "" );
```
jtklein, line 104: Unrelated to this PR, but I am just seeing this while testing: the pressable area of the Reset button needs to be bigger I think. Basically, the entire height of the header I believe.

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -26,26 +39,104 @@ const INPUT_BOX_CLASSES = classnames(
...
 
+  const renderResult = useCallback<ListRenderItem<UniversalSearchResultItem>>( ( { item } ) => (
+    <UniversalSearchResult
+      result={item}
+      onPress={( ) => handleSelect( resultToSubject( item ) )}
+    />
+  ), [handleSelect] );
+
+  let body = null;
+  if ( showResults ) {
```
jtklein: I don't see us using this pattern, conditional render of the FlashList in other places much, I think we were running into some issues in the past with it, but I don't really remember what. In most places we therefore have a more elaborate ListEmptyComponent with booleans as to what "empty" component to show. No internet connection, no data found, still loading, fallback UI selection for search not started yet.

jtklein (approved):
No blockers from my perspective for this PR, as the entire V2 is still behind a feature flag. Even though I left some inline comments that I believe need to be changed at one point, however, it is hard to see though what kind of work is still to come, e.g. the a11y problem outlined in an inline comment, I am not sure if that is at all relevant for the final state of ExploreV2.
In other words, it is sometimes hard to see if something is meant to be temporary or not.
One quirkiness: I have not been able to get all three types within one result, but maybe that is just me not knowing what to search for.

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
jtklein: Was running into the same problem on another screen, here as a fix for this already: https://github.com/inaturalist/iNaturalistReactNative/pull/3778

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
sepeterson: Made a ticket for this https://linear.app/inaturalist/issue/MOB-1538/reset-button-tap-target-in-exploresearchheader-is-too-small

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
sepeterson: TaxonResult isn't easily reusable like the other two below. I think long term the solution might be to have a smaller component that is usable here and as a child of TaxonResult

src/components/Explore/ExploreV2/helpers/universalSearchSubject.ts
sepeterson: Changes the defaults to either return an empty string or throw an error. I think I'm leaning towards throwing an error in cases where we have strong filtering/type safety upstream rather than always falling back to ` | null`

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
sepeterson: You're correct that one wasn't reachable by screen reader. I made them siblings in the hierarch for now. I don't fully follow the a11y typing and making things not accessible but I've added a note to https://linear.app/inaturalist/issue/MOB-1340/universal-search-info-i-button to double check the behavior when I work on it

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
sepeterson: This is a good callout, I reused EmptySearchResults to match the other lists

PR #3769, opened 2026-06-25
trust wix / applesimutils brew cask for detox CI

FLGMwt:
CI has started failing with the following:

```
Error: Refusing to load formula wix/brew/applesimutils from untrusted tap wix/brew.
Run `brew trust --formula wix/brew/applesimutils` or `brew trust wix/brew` to trust it.
Error: Process completed with exit code 1.
```

This PR makes homebrew explicitly trust the applesimutils formula which Detox depends on


jtklein (approved):
Thanks for adding the fix.

PR #3770, opened 2026-06-26
MOB-1339 Hooks to support search autocomplete

sepeterson:
Just trying to split up MOB-1339 a little bit

In this PR:

- A hook to support debounce on an effect on value change
- A hook to support universal search's autocomplete


src/sharedHooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,33 @@
+import {
+  useCallback, useEffect, useRef, useState,
+} from "react";
+
+const DEFAULT_DELAY_MS = 400;
+
+// Debounces a value imperatively: `debounce` schedules an update after a delay
+// (replacing any pending one), while `setImmediately` applies a value right away
+// and cancels anything pending. Mirrors the inline debounce in MentionTextInput.
+const useDebouncedValue = <T, >( initialValue: T, delayMs: number = DEFAULT_DELAY_MS ) => {
```
sepeterson, line 10: This hook could potentially be used in a few spots but we'll just do Explore V2 for now

src/sharedHooks/useUniversalSearch.ts
```diff
@@ -0,0 +1,85 @@
...
+    rule_preferences: true,
+  },
+};
+
+const useUniversalSearch = ( query: string ) => {
+  const trimmedQuery = query.trim( );
+  const shouldFetch = trimmedQuery.length > 0;
+
+  const {
+    data: results = [],
```
FLGMwt, line 44: for here and in the queryFn: do we not want a `null` state to differentiate an empty result from a default / error state?

FLGMwt (approved):
LGTM, just question on useQuery results default.

src/sharedHooks/useUniversalSearch.ts
```diff
@@ -0,0 +1,85 @@
...
+    login: true,
+    icon_url: true,
+    observations_count: true,
+  },
+  project: {
+    id: true,
+    title: true,
+    icon: true,
+    project_type: true,
+    rule_preferences: true,
```
jtklein: This is a nested object. Querying with only a true here in the fields set will only retrieve the id, e.g. `{ id: 123 }`. You will need the field and value of a date rule preferences though in order to show the date range of a collection project as shown in the designs. So, you could re-use PROJECT_SUMMARY_FIELDS from fields.tsx which would make the result item an ApiProjectSummary type.

src/sharedHooks/useUniversalSearch.ts
```diff
@@ -0,0 +1,85 @@
...
+          q: trimmedQuery,
+          sources: ["taxa", "users", "projects"],
+          fields: UNIVERSAL_SEARCH_FIELDS,
+        },
+        optsWithAuth,
+      );
+      if ( !response ) { return []; }
+      // Preserve the API's score-sorted order, mapping each result to a tagged union
+      return response.results.reduce<UniversalSearchResultItem[]>( ( acc, result ) => {
+        if ( result.taxon ) {
```
jtklein: I can see that there is prior code in `fetchSearchResults` that does the same, but the API does return the type value already if you want to query for it: https://api.inaturalist.org/v2/docs/#/Search/get_search
So, all results could already have the same shape you are creating anew here.

tests/unit/sharedHooks/useUniversalSearch.test.js
```diff
@@ -0,0 +1,73 @@
+import { waitFor } from "@testing-library/react-native";
+import useUniversalSearch from "sharedHooks/useUniversalSearch";
+import { renderHookInApp } from "tests/helpers/render";
+
+jest.mock( "api/search" );
+const { search } = require( "api/search" );
```
jtklein, line 6: You can also just plainly import here.

jtklein (changes requested):
LGTM. Great unit testing! Love the size of this PR.

One change required though to avoid a small UI bug.

src/sharedHooks/useUniversalSearch.ts
sepeterson: Yes I think we ultimately do but I'm going to defer that change to the ticket where we'll eventually use it: https://linear.app/inaturalist/issue/MOB-1330/no-results-state

src/sharedHooks/useUniversalSearch.ts
sepeterson: Good thing you've been working on these recently! I was able to cleanly reuse `PROJECT_SUMMARY_FIELDS`

src/sharedHooks/useUniversalSearch.ts
sepeterson: This was good cleanup but required some changes in search.ts to use a discriminated union type for results that's worth a quick re-check

jtklein (approved):
Still LGTM

PR #3771, opened 2026-06-26
Mob 1462 search myobs UI 2

abbeycampbell:
Closes [MOB-1462](https://linear.app/inaturalist/issue/MOB-1462/search-myobs-ui)
This includes:
- Searched taxon banner and active state for search icon
- Suppression of announcements and other banners while search is active
- Empty state
- The addition of a helper to handle text that needs to be informed by naming preferences (giving us a way to render a one-line common/scientific name when we don't have access to the full Taxon that `DisplayTaxonName` requires)
- Search screen tweaks including 'Reset' disabling

### DIFF
--- a/src/components/MyObservations/Search/SearchedTaxonBanner.tsx
+++ b/src/components/MyObservations/Search/SearchedTaxonBanner.tsx
@@ -9,6 +9,7 @@ import {
 } from "providers/MyObservationsContext";
 import React from "react";
 import type { RealmUser } from "realmModels/types";
+import { taxonDisplayName } from "sharedHelpers/taxon";
 import { useCurrentUser, useTranslation } from "sharedHooks";
 
 const SearchedTaxonBanner = ( ) => {
@@ -23,14 +24,7 @@ const SearchedTaxonBanner = ( ) => {
 
   if ( !searchedTaxon ) return null;
 
-  // Mirrors DisplayTaxonName: show scientific name when the user has
-  // opted into scientific-first or opted out of common names;
-  // otherwise show common name (falling back to scientific if missing).
-  const preferScientific = currentUser?.prefers_scientific_name_first === true
-    || currentUser?.prefers_common_names === false;
-  const displayName = preferScientific
-    ? searchedTaxon.name
-    : ( searchedTaxon.preferredCommonName || searchedTaxon.name );
+  const displayName = taxonDisplayName( searchedTaxon, currentUser );
 
   return (
     <View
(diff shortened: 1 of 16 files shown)


src/components/MyObservations/Search/SearchedTaxonBanner.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const SearchedTaxonBanner = ( ) => {
+  const { t } = useTranslation( );
+  const { state, dispatch } = useMyObservations( );
+  const { searchedTaxon } = state;
+
+  // The Realm `User` class doesn't declare its schema fields as typed
+  // instance properties (see realmModels/User.ts), so cast to the typed
+  // RealmUser interface to access the user's name-display prefs.
+  const currentUser = useCurrentUser( ) as RealmUser | null;
```
sepeterson, line 23: No change (or response) requested but something I would like to better understand is the relationship between realm models and ts types because I'm not sure we've been consistent

src/sharedHooks/useTaxonSearch.ts
```diff
@@ -96,6 +98,13 @@ const useTaxonSearch = ( taxonQueryArg = "" ) => {
 
       if ( isLoading ) return;
 
+      // Don't fall back to local results until the remote query has actually
+      // run at least once. useAuthenticatedQuery starts with enabled=false
+      // while it resolves the auth state, which makes isLoading=false even
+      // though no remote fetch has happened yet — without this check we'd
+      // flash the "Showing offline search results" callout in that window.
+      if ( shouldFetchRemote && !isFetched ) return;
```
sepeterson: I think this is blocking locally-persisted realm taxa from appearing in offline suggestions. `isFetched` doesn't seem to be flipping to true

Repro: Offline, take a photo -> tap search icon -> type in search for taxon you have already viewed online -> no results

I added these logs
```
      console.log("shouldFetchRemote: " + shouldFetchRemote + " isFetched: " + isFetched)
      if ( shouldFetchRemote && !isFetched ) return;
      console.log("past return")
```
and here's what I see in console

Image

sepeterson:
@abbeycampbell Just the one bug I think! I would say good to merge after it's addressed

src/sharedHooks/useTaxonSearch.ts
abbeycampbell: great catch! Should be fixed by adding an offline check to the conditional

src/components/MyObservations/Search/SearchedTaxonBanner.tsx
abbeycampbell: yeah. We do casts like this in a few places. I think ideally we would actually be typing the User class, so things like `useCurrentUser` actually just give us a return type we can work with

sepeterson (approved):

PR #3774, opened 2026-06-26
Fetch Journal Posts for blog and projects

jtklein:
Closes MOB-1424

Punting fetching of other user's blog posts to another ticket because I think we have no API v2 endpoint for this.


src/api/posts.ts
```diff
@@ -18,4 +16,21 @@ const fetchUserPosts = async (
   }
 };
 
-export default fetchUserPosts;
+const fetchProjectPosts = async (
+  params: Record<string, unknown> = {},
+  opts: Record<string, unknown> = {},
+): Promise<Record<string, unknown> | null | ErrorWithResponse | INatApiError> => {
+  try {
+    return await inatjs.projects.posts( params, opts );
```
jtklein, line 24: We are using the same endpoint in projects api wrappers to get the total count of blog posts in ProjectDetails screen. Maybe I should have better consolidated the two functions.

src/api/projects.js
```diff
@@ -38,15 +38,15 @@ const fetchProjectMembers = async (
   }
 };
 
-const fetchProjectPosts = async (
+const fetchProjectPostsCount = async (
```
jtklein, line 41: Renamed to avoid cache collision.

src/components/Journal/ProjectPosts.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+import PostList from "./PostList";
+
+interface Props {
+  projectIcon?: string;
+  projectId?: number;
+  projectTitle?: string;
+}
+
+const PostsForProjects = ( {
```
jtklein, line 23: This is for the most part an adaptation of ProjectMembers now with the API endpoint swapped out for project.posts from project.members.
The next page fetching logic and UI elements I have moved into ProjectList though to share with the other types of Journal posts screens, i.e. only Blog in this PR, UserPosts yet to come.

src/components/Journal/PostList.tsx
```diff
@@ -1,29 +1,49 @@
...
+  posts: ApiPostForUser[];
+  isFetchingNextPage: boolean;
+  fetchNextPage: ( ) => void;
 }
 
-const PostList = ( {
-  posts,
-}: Props ) => {
-  const renderPost: ListRenderItem<ApiPost> = ( { item } ) => (
+const PostList = ( { posts, isFetchingNextPage, fetchNextPage }: Props ) => {
```
jtklein, line 26: See comment on ProjectPosts

src/components/Journal/Blog.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useMemo,
+} from "react";
+import {
+  useInfiniteScroll,
+  useTranslation,
+} from "sharedHooks";
+
+import PostList from "./PostList";
+
+const Blog = ( ) => {
```
jtklein, line 17: Follows largely ProjectPosts

src/components/Journal/Blog.tsx
```diff
@@ -0,0 +1,66 @@
...
+    [t, totalPosts],
+  );
+
+  useEffect( ( ) => {
+    navigation.setOptions( headerOptions );
+  }, [headerOptions, navigation] );
+
+  if ( !blogPosts ) {
+    return null;
+  }
```
FLGMwt: I'm getting a flash of default gray non-white background when the query starts, I'm guessing on this eager return, do we need a wrapper or to return an empty view w/ bg?

src/components/Journal/Blog.tsx
```diff
@@ -0,0 +1,66 @@
...
+      headerSubtitle: t( "X-JOURNAL_POSTS", {
+        count: totalPosts || 0,
+      } ),
+    } ),
+    [t, totalPosts],
+  );
+
+  useEffect( ( ) => {
+    navigation.setOptions( headerOptions );
+  }, [headerOptions, navigation] );
```
FLGMwt, line 47: not your changes, but we could replace `headerOptions` w/ `t, totalPosts` and simplify this.

src/components/Journal/Blog.tsx
```diff
@@ -0,0 +1,66 @@
...
+    totalResults: totalPosts,
+  } = useInfiniteScroll( queryKey, fetchUserPosts, queryParams, {
+    enabled: true,
+  } );
+
+  const headerOptions = useMemo(
+    () => ( {
+      headerTitle: t( "Blog" ),
+      headerSubtitle: t( "X-JOURNAL_POSTS", {
+        count: totalPosts || 0,
```
FLGMwt, line 39: I don't care for defaulting to an affirmative `0`. We should either have an indeterminate value or not display the subtitle. Currently we're showing `0 JOURNAL POSTS` while in the loading state.

src/components/Journal/PostList.tsx
```diff
@@ -1,29 +1,49 @@
...
+  const { isConnected } = useNetInfo();
+  const footerComponent = useMemo(
+    () => (
+      <InfiniteScrollLoadingWheel
```
FLGMwt, line 30: on empty, this is right up against the header, maybe we can center?

Image

FLGMwt (changes requested):
some small things, but generally LGTM code-wise.

Two things in testing, both show in video:
- I'm not sure what's causing it but there's an inconsistent separator that's not between every two elements.
- Something's up the infinite scroll. I can't get it to load more than the initial posts on first launch no matter how hard I flex the pull up / pull down. It seems that navigating sometimes triggers a request that gets it going? from then on, it seems to infinite scroll as expected.

https://github.com/user-attachments/assets/cf4eaef4-99ea-47c4-9777-f902fbd1c1d9

src/components/Journal/Blog.tsx
jtklein: Have removed the null guard.

src/components/Journal/PostList.tsx
jtklein: This PR/ticket was not concerned with UI only with fetching data. For further tickets the idea was to copy the UI of the ProjectMembers list as is. Since this issue is also apparent there I'd defer that to a standalone ticket if Product deems it necessary.

src/components/Journal/Blog.tsx
jtklein: This PR/ticket was not concerned with UI only with fetching data. For further tickets the idea was to copy the UI of the ProjectMembers list as is. Since this issue is also apparent there I'd defer that to a standalone ticket if Product deems it necessary.

src/components/Journal/Blog.tsx
jtklein: This PR/ticket was not concerned with UI only with fetching data. For further tickets the idea was to copy the UI of the ProjectMembers list as is. Since this issue is also apparent there I'd defer that to a standalone Maintenance ticket fixing it for all instances simultaneously.

jtklein:
This PR/ticket was not concerned with UI only with fetching data. For further tickets the idea was to copy the UI of the ProjectMembers list as is.
I have now added some more commits here already to make this list closer to ProjectMembers' UI.

This should fix the separator issue.

I was not able to reproduce the "I can't get it to load more than the initial posts" bug on my device. My guess would be that your phone fits all the 10 items on screen on initial load and we never trigger an onEndReached. I have increased the per_page to 20. Let me know if that fixes that issue on your end.

jtklein:
Does all thumbs up mean "Approved" now?

FLGMwt:
> Does all thumbs up mean "Approved" now?

Sorry, didn't have a chance to retest before I got sick. Just tested and looks good now!

FLGMwt (approved):

PR #3782, opened 2026-06-30
MOB-1540: two fixes for the media viewer

sepeterson:
Fixes the described crash.

The `key` change was the fix to another buggy behavior I saw where it showed a blank screen after deleting a photo instead of moving to the next one.


jtklein (approved):
LGTM

Can confirm this fixes the crash.

PR #3783, opened 2026-06-30
Taxon mapper

sepeterson:
I attempted a ts conversion of Taxon.js at the same time but it's a real can of worms

This function is mostly to support displaying iconic taxa as search options for ExploreV2. The useIconicTaxa hook returns live realm objects and I wanted a reusable way to get them to POJOs.


jtklein (approved):
I think this is a good thing. Fct looks good and there are a lot of tests.

PR #3785, opened 2026-06-30
potential flakiness test

sepeterson:


FLGMwt (approved):
lgtm, thanks for looking into this!

I think my timers fix was probably a red herring then

PR #3786, opened 2026-07-01
Traditional projects: Project chooser screen

jtklein:
Closes MOB-1502

This adds the base UI for the Add to projects screen. A list of all joined projects locally available. No API query for projects is run here yet, so for testing you'd need to go to your User profile -> Projects to persist them and then the list will be populated.

Have also added fcts to map local realm objects into pojos following Observation. Typing is not too consistent though but that seems more like a broader concern of RealmType v ApiType c PojoType separation not related to this UI PR.


src/components/Projects/helpers/displayDates.ts
```diff
@@ -20,6 +20,13 @@ const formatProjectDate = (
   t: TFunction,
   i18n: i18next,
 ): FormattedProjectDate => {
+  // "" means "traditional" which do not have date ranges
+  if ( project.project_type === "" ) {
```
jtklein, line 24: We only persist traditional project, so calling this format function with a realm project mapped to pojo has no rule_preferences field. To avoid a crash on filtering below I refactored this to be an early return.

src/components/SharedComponents/ViewWrapper.tsx
```diff
@@ -83,6 +83,31 @@ const SharedStackViewWrapper = ( {
     </ScreenShell>
   );
 };
+
+const SharedStackBottomInsetViewWrapper = ( {
```
jtklein, line 87: The insane panoply of headers + nav stacks we have requires yet another type of screen wrapper.

src/components/AddToProjects/AddToProjects.tsx
```diff
@@ -1,11 +1,199 @@
...
+  const onSave = ( ) => {
+    navigation.goBack( );
+  };
+  const disabled = false;
+
+  const renderRightIcon = useCallback(
+    ( item: RealmProject, isSelected: boolean ) => {
+      // Logic if all required fields have been filled out will live in zustand
+      if (
+        projectObservations.some(
```
sepeterson: Getting a crash here for an obs that hasn't been saved to realm. Repro: Import Photo -> ObsEdit -> Add to Projects

I don't totally follow the data flow here so up to you if this needs to be fixed now

<details>
<summary>screenshot</summary>

Image

</details>

tests/factories/LocalProject.js
```diff
@@ -0,0 +1,12 @@
+import { define } from "factoria";
+
+import pofFactory from "./LocalProjectObservationField";
+
+export default define( "LocalProject", faker => ( {
+  description: faker.lorem.paragraph(),
+  icon: faker.image.url(),
+  id: faker.number.int(),
+  projectObservationFields: [pofFactory( "RemoteProjectObservationField" )],
```
sepeterson: should the string in pofFactory be "LocalProjectObservationField" ?

src/realmModels/ObservationField.ts
```diff
@@ -19,6 +20,18 @@ class ObservationField extends Realm.Object {
     return localOf;
   }
 
+  static mapRealmToPojo( realmObservationField: RealmObservationField ) {
+    return {
+      allowedValues: realmObservationField.allowedValues.length > 0
```
sepeterson: If `allowedValues.length === 0` doesn't that mean it's already just []

src/realmModels/Project.ts
```diff
@@ -38,6 +39,19 @@ class Project extends Realm.Object {
     }, "upserting projects in Project" );
   }
 
+  static mapRealmToPojo( realmProject: RealmProject ) {
```
sepeterson, line 42: Nice. There's probably a good tech debt ticket to be had adding a function like this for all the places we're using live realm objects in components

src/components/AddToProjects/AddToProjects.tsx
jtklein: Should be fixed now

src/components/SharedComponents/ViewWrapper.tsx
sepeterson: It's all so complicated

sepeterson (approved):
Great progress! Good to go after any followup you want to do

tests/factories/LocalProject.js
jtklein: Yes, correct.

src/realmModels/ObservationField.ts
jtklein: No, allowedValues is a realm List in the ObservationField RealmObject. So, we have to map it out to a JS array. But we probably don't need to defend against being empty and just run the map on an empty array.

PR #3787, opened 2026-07-01
MOB-1341: extract search-field state into useSearchField hook

sepeterson:
I was working on the location search field and noticed I was exactly replicating most of the state stuff. I figured now would be a good time to add a reusable solution for both.


jtklein (approved):
LGTM.
Search field still works.
Went through the `git diff main > xyz.diff` by deleting what is copied over (renamed) and quickly checking the additional code like the interface.

PR #3790, opened 2026-07-01
fix stale subscriptions in useObservationLocation & useWatchPosition

FLGMwt:
We have some complicated state / effect logic that _seems_ to be smoothing over stale callback closures of state values, most importantly, our watchPosition subscriptionId.

By moving this to useFocusEffect for useObsLoc and useWatchPos, we can inline some of those values to obviate stale close issues. The bonus effect is simplifying the code and removing effect deps which likely exacerbated the buggy implementation.

[EDIT] We now rely on useWatchPosition for focus management. Previously, we were returning a `stopWatch` that was seemingly called before navigating away.

This should fix some hard-to-reproduce issues around "location never stops loading" around error'd location calls and focus. It _probably_ led to some leaked watch subscriptions too.

### DIFF
--- a/src/components/Match/MatchContainer.tsx
+++ b/src/components/Match/MatchContainer.tsx
@@ -289,14 +289,13 @@ const MatchContainer = ( ) => {
   const {
     isFetchingLocation,
     stopWatch,
-    subscriptionId,
     userLocation,
   } = useObservationLocation( { shouldFetchLocation } );
 
   const navToLocationPicker = useCallback( ( ) => {
-    stopWatch( subscriptionId );
+    stopWatch( );
     navigation.navigate( "LocationPicker" );
-  }, [stopWatch, subscriptionId, navigation] );
+  }, [stopWatch, navigation] );
 
   const latitude = currentObservation?.latitude;
   const longitude = currentObservation?.longitude;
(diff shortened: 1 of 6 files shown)


FLGMwt:
pulling the thread on this, the focusEffect removes the need for returning stopWatch at all which is used to "prime" navigations (probably redundantly now).

sepeterson (approved):
Beautiful diff. I tested pretty thoroughly on Match and ObsEdit and had no issues. We'll definitely want to be sure to test in the field but LGTM

jtklein:
I was also halfway through review earlier. Have not tested yet but the code changes looked solid to me.
The only question I had was: Did you have to remove the MAX_POSITION_AGE_MS and handling of it? Without looking a the history this seems like something we would have to add after seeing some problems in production.

FLGMwt:
> I was also halfway through review earlier. Have not tested yet but the code changes looked solid to me. The only question I had was: Did you have to remove the MAX_POSITION_AGE_MS and handling of it? Without looking a the history this seems like something we would have to add after seeing some problems in production.

I could not determine why it was necessary. The `maximumAge` argument option to `watchPosition` should handle this. I confirmed on iOS and Android that I was constantly getting new values milliseconds apart so certainly fresher than 60s ; D 

I understood the check to be added in error but would be happy to add back for safety. I'll hold off on merge til your review, thank you!

PR #3791, opened 2026-07-01
Mob 1388 allow comments with agreeing ids

abbeycampbell:
closes [MOB-1388](https://linear.app/inaturalist/issue/MOB-1388/allow-comments-with-agreeing-ids)

Fix is in the first commit, the rest is a bit of clean-up, and type error fixes for this file while avoiding anything that would balloon and touch too many other files.

I think I'd make the argument that agreeing Ids, whose state lives in `useObsDetailsSharedLogic.ts` and suggesting Ids, whose state is in `IdentificationSheets.tsx` could probably actually share a reducer, since they kind of do the same things. But that would be a tech-debt follow-up, and this just to fixes the current bug.

### DIFF
--- a/src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
+++ b/src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
@@ -1,6 +1,9 @@
 import { useNavigation, useRoute } from "@react-navigation/native";
 import { createComment } from "api/comments";
 import { createIdentification } from "api/identifications";
+import type {
+  AgreeIdentification,
+} from "components/ObsDetailsSharedComponents/hooks/useObsDetailsSharedLogic";
 import AgreeWithIDSheet from "components/ObsDetailsSharedComponents/Sheets/AgreeWithIDSheet";
 import PotentialDisagreementSheet
   from "components/ObsDetailsSharedComponents/Sheets/PotentialDisagreementSheet";
@@ -141,14 +144,19 @@ export const identReducer = ( state: IdentState, action: IdentAction ): IdentSta
     case CLEAR_SUGGESTED_TAXON:
       return { ...state, identTaxon: null };
     case HIDE_SUGGESTED_ID_SHEET:
-      return { ...state, showSuggestIdSheet: false };
+      return {
+        ...state,
+        showSuggestIdSheet: false,
+        identTaxon: null,
+        newIdentification: null,
+      };
     default:
       return state;
   }
 };
 
 interface Props {
-  agreeIdentification: boolean;
+  agreeIdentification: AgreeIdentification | null;
   closeAgreeWithIdSheet: () => void;
   confirmRemoteObsWasDeleted?: () => void;
   handleCommentMutationSuccess: ( data: unknown ) => void;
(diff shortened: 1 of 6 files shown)


src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
```diff
@@ -323,14 +323,14 @@ const IdentificationSheets: React.FC<Props> = ( {
...
-      observation_id: observation.uuid,
+      observation_id: observation?.uuid,
       taxon_id: ident.taxon?.id,
       body: ident.body,
     };
 
     loadActivityItem( );
     createIdentificationMutate( { identification: agreeParams } );
-  }, [createIdentificationMutate, observation.uuid, loadActivityItem] );
+  }, [createIdentificationMutate, observation?.uuid, loadActivityItem] );
```
abbeycampbell, line 332: I assumed since observation is a required prop, there would be no need for the optional chaining here and we could lose the eslint warning about the inferred dependency being different than the source.
However, removing them triggered React compiler-managed memoization for the component. Combined with `useObsDetailsSharedLogic` remapping Observation.mapApiToRealm(remoteObservation) fresh on every update, observation refs became unstable and caused Maximum call stack size errors. Claude says we can fix this by memoizing the mapped observation in `useObsDetailsSharedLogic`, but that should definitely be separate from this bug fix

src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
```diff
@@ -141,14 +144,19 @@ export const identReducer = ( state: IdentState, action: IdentAction ): IdentSta
     case CLEAR_SUGGESTED_TAXON:
       return { ...state, identTaxon: null };
     case HIDE_SUGGESTED_ID_SHEET:
-      return { ...state, showSuggestIdSheet: false };
+      return {
+        ...state,
+        showSuggestIdSheet: false,
+        identTaxon: null,
```
jtklein: Not quite sure why that was necessary. If I open a second New ID or Agree sheet I am seeing a different ID in the new sheet.

src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
```diff
@@ -416,7 +424,10 @@ const IdentificationSheets: React.FC<Props> = ( {
           hidden={showIdentBodySheet}
           loading={isCreateIdPending}
           onPressClose={closeAgreeWithIdSheet}
-          identification={agreeIdentification}
+          identification={{
+            taxon: agreeIdentification.taxon,
+            body: newIdentification?.body,
```
jtklein, line 419: So, this one line is the actual bugfix, we were just not adding the body to the ID?

jtklein (approved):
The fix works: Adding a comment to an agreeing ID is saved.

All cleanup changes look good too.

useObsDetailsSharedLogic was one of the first refactors to remove duplicated code for the Advanced, Default, (and Simple?) Mode variants of ObsDetails. It was initially imported in multiple screens/containers. We have since refactored the screens into one container, and therefore the hook is only imported once now, as is IdentificationSheets. So, if you feel like splitting apart IDSheets logic from the hook is better, feel free to do so. I don't see a blocker in combining into idReducer

src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
abbeycampbell: oh, you're right. the taxon already gets reset via `SET_NEW_IDENTIFICATION`

src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx
abbeycampbell: yep! agreeIdentification only has taxon, so the actual comment body was always undefined.

abbeycampbell:
> useObsDetailsSharedLogic was one of the first refactors to remove duplicated code for the Advanced, Default, (and Simple?) Mode variants of ObsDetails. It was initially imported in multiple screens/containers. We have since refactored the screens into one container, and therefore the hook is only imported once now, as is IdentificationSheets. So, if you feel like splitting apart IDSheets logic from the hook is better, feel free to do so. I don't see a blocker in combining into idReducer

I actually meant it the other way: pulling suggest/comment/disagreement state into `useObsDetailsSharedLogic` so agreeing and suggesting ids share logic there. But actually maybe what you're describing is cleaner? Either way it's a follow-up

PR #3796, opened 2026-07-02
MOB-1332 ExploreV2 species tab

sepeterson:
In this PR:

- new components for the ExploreV2 species view 
- currently selected tab in the context
- tests

There'll be a followup with the logic for the already-observed checkmark

<details>
<summary>screenshot</summary>
Simulator Screenshot - iPhone 17 - 2026-07-09 at 13 55 23

</details>


src/components/SharedComponents/TaxonGridItem.tsx
```diff
@@ -71,7 +73,7 @@ const TaxonGridItem = ( {
         )}
 
         { upperRight && (
```
sepeterson, line 75: This prop was around but unused which is why I can just go ahead and change the styling

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -0,0 +1,94 @@
...
+  taxon: ApiTaxon;
+}
+
+interface Props {
+  enabled: boolean;
+  isConnected: boolean | null;
+  params: SpeciesCountQueryParams;
+}
+
+const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
```
sepeterson, line 30: In this PR, this component is mostly a copy/paste of a subset of the V1 SpeciesView component but it's going to diverge more in a followup

abbeycampbell: In testing this, after tapping on a species grid item and viewing observations of that species, if I visit the species tab again I only see the now-searched for species. Would we want all species to re-populate again on a subsequent visits to this tab, or do we want to provide a quick way refresh the state here? Just wondering; doesn't seem clear from the designs

abbeycampbell (approved):
Looking really good. Just an open question about behavior from me

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
sepeterson: I think this follows the design based on [one frame here](https://www.figma.com/design/MChpvx4ZrKEVVwsWKt4lkI/iNaturalist-Mobile-UI-Design?node-id=28167-85498&m=dev).  Definitely worth confirmation though so I added this and a couple other items on the project to discuss w/ Abhas when he's feeling better: https://linear.app/inaturalist/project/explore-re-design-fba229378ff8/activity#project-update-bb3ce46a

PR #3797, opened 2026-07-02
Location picker types

FLGMwt:
doing some types while looking into MOB-1529.

I'm taking advantage of the fact that `LocationPicker` is its own island to do some type isolation. Two key ones there:

`LocationPickerPlace`: types a `Place` _as used by this feature_. The fields are non-null because we rely on the API to return them. We still need to figure out how we want to handle `api/types` as API descriptions vs app / feature descriptions and how to bridge useQuery => app typings, but this at least draws a line there.

`LocationPickerObservation`: minimal subset of obs props _as used by this feature_. We have a _big_ thing to figure out w/ "how to type obs" throughout the app / API / Realm. This type draws the line at LocationPicker and `const currentObservation = useStore( state => state.currentObservation );` <- however that is defined. Importantly, it highlights that we (probably incorrectly?) use `positional_accuracy` & `positionalAccuracy`. 

Once we have a more holistic obs typing approach and enforcement, it would probably make sense to remove `LocationPickerObservation`. For now, I'm thinking it's helpful as documentation towards that.

### DIFF
--- a/src/components/LocationPicker/LocationPicker.tsx
+++ b/src/components/LocationPicker/LocationPicker.tsx
@@ -8,7 +8,8 @@ import {
 import { SharedStackViewWrapper } from "components/SharedComponents/ViewWrapper";
 import { View } from "components/styledComponents";
 import React from "react";
-import type { LatLng } from "react-native-maps";
+import type { LayoutChangeEvent } from "react-native";
+import type { LatLng, Region } from "react-native-maps";
 import { useTranslation } from "sharedHooks";
 
 import CrosshairCircle from "./CrosshairCircle";
@@ -16,22 +17,23 @@ import DisplayLatLng from "./DisplayLatLng";
 import Footer from "./Footer";
 import LoadingIndicator from "./LoadingIndicator";
 import LocationSearch from "./LocationSearch";
+import type { LocationPickerPlace } from "./types";
 
 interface Props {
   accuracy: number;
-  handleSave: () => void;
+  handleSave: ( ) => void;
   hidePlaceResults: boolean;
   loading: boolean;
   locationName: string;
   initialRegion: LatLng;
-  onCurrentLocationPress: () => void;
-  onMapReady: () => void;
-  onRegionChangeComplete: () => void;
+  onCurrentLocationPress: ( ) => void;
+  onMapReady: ( ) => void;
+  onRegionChangeComplete: ( newRegion: Region ) => void;
   region: LatLng;
   regionToAnimate: object;
-  selectPlaceResult: () => void;
-  updateLocationName: () => void;
-  onMapLayout: () => void;
+  selectPlaceResult: ( place: LocationPickerPlace ) => void;
+  updateLocationName: ( name: string ) => void;
+  onMapLayout: ( event: LayoutChangeEvent ) => void;
 }
 
 const LocationPicker = ( {
--- a/src/components/LocationPicker/LocationPicker.tsx
+++ b/src/components/LocationPicker/LocationPicker.tsx
@@ -1,5 +1,3 @@
-// @flow
-
 import classnames from "classnames";
 import {
   CloseButton,
@@ -9,8 +7,8 @@ import {
 } from "components/SharedComponents";
 import { SharedStackViewWrapper } from "components/SharedComponents/ViewWrapper";
 import { View } from "components/styledComponents";
-import type { Node } from "react";
 import React from "react";
+import type { LatLng } from "react-native-maps";
 import { useTranslation } from "sharedHooks";
 
 import CrosshairCircle from "./CrosshairCircle";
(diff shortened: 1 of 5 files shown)


src/components/LocationPicker/LocationPickerContainer.tsx
```diff
@@ -131,14 +135,14 @@ const reducer = ( state, action ) => {
...
-const LocationPickerContainer = ( ): Node => {
+const LocationPickerContainer = ( ) => {
   const currentObservation = useStore( state => state.currentObservation );
   const updateObservationKeys = useStore( state => state.updateObservationKeys );
   const navigation = useNavigation( );
 
   const [state, dispatch] = useReducer( reducer, initialState );
-  const [radiusToMapHeight, setRadiusToMapHeight] = useState( undefined );
-  const [mapDimensionsRatio, setMapDimensionsRatio] = useState( undefined );
+  const [radiusToMapHeight, setRadiusToMapHeight] = useState<number | null>( null );
```
jtklein, line 144: I personally prefer undefined before something is getting defined, and null whenever I have to set something back to non-existence after it was used for something.

src/components/LocationPicker/types.ts
```diff
@@ -0,0 +1,33 @@
+export interface LocationPickerObservation {
+  latitude?: number;
+  longitude?: number;
+
+  privateLatitude?: number;
+  privateLongitude?: number;
+
+  positional_accuracy?: number;
+  positionalAccuracy?: number;
```
jtklein: I also think that this is probably in error. I don't see a mapping of snake to camel like some other properties of Observation. If I change the access from camel to snake in Map.tsx line 566. The LocationIndicator shows additional UI in the map which I believe is wanted.
So, I'd say to fix this here as well.

src/components/LocationPicker/LocationPicker.tsx
```diff
@@ -9,32 +7,34 @@ import {
...
+  handleSave: ( ) => void;
+  hidePlaceResults: boolean;
+  loading: boolean;
+  locationName: string;
+  initialRegion: LatLng;
+  onCurrentLocationPress: ( ) => void;
+  onMapReady: ( ) => void;
+  onRegionChangeComplete: ( newRegion: Region ) => void;
+  region: LatLng;
+  regionToAnimate: object;
```
jtklein: Region ?

src/components/LocationPicker/LocationPicker.tsx
```diff
@@ -9,32 +7,34 @@ import {
...
-  updateLocationName: Function,
-  onMapLayout: Function,
-};
+interface Props {
+  accuracy: number;
+  handleSave: ( ) => void;
+  hidePlaceResults: boolean;
+  loading: boolean;
+  locationName: string;
+  initialRegion: LatLng;
```
jtklein: TS error below says this should be Region?

src/components/LocationPicker/LocationPicker.tsx
```diff
@@ -9,32 +7,34 @@ import {
...
+  onCurrentLocationPress: ( ) => void;
+  onMapReady: ( ) => void;
+  onRegionChangeComplete: ( newRegion: Region ) => void;
+  region: LatLng;
```
jtklein: Region?

src/components/LocationPicker/LocationPickerContainer.tsx
```diff
@@ -1,26 +1,30 @@
...
 import fetchPlaceName from "sharedHelpers/fetchPlaceName";
 import useStore from "stores/useStore";
 
 import LocationPicker from "./LocationPicker";
+import type { LocationPickerObservation, LocationPickerPlace } from "./types";
 
 const CROSSHAIRRADIUS = 254 / 2;
 
-const setInitialRegion = ( currentObservation, radiusToMapHeight, mapDimensionsRatio ) => {
+const getInitialRegion = (
```
jtklein, line 23: Return can be typed as Region?

src/components/LocationPicker/LocationPickerContainer.tsx
jtklein, line 84: Set via SELECT_PLACE_RESULT and also of Region type.

jtklein (changes requested):
Loving this slow TS migration! Anyways, as written in inline comment I think positionalAccuracy is erroneous and should be fixed here.
Also, a few props should be of Region type.

src/components/LocationPicker/LocationPickerContainer.tsx
FLGMwt: null is compelling for me because:
- it follows the useRef component semantics / idiomatic type
- it's impossible to be "accidentally" null where undefined can intentional or erroneous

With the typescript migration and progressive type enforcement (and discovery), `null` seems like a safer / more explicit state. of course, undef is still ideal for optional properties.

src/components/LocationPicker/types.ts
FLGMwt: Oh shoot, I'm sorry. I had a comment that was roughly `// TODO, investigate discrepancy in potential useCurrentLocation sources` on this that I didn't transfer when I moved to a file. That's what I meant to reference in the PR desc "Importantly, it highlights that we (probably incorrectly?) use positional_accuracy & positionalAccuracy."

I was trying to cut the PR as "descriptive" and defer functional changes identified by that description. That said, since you tested it (thank you!), I don't mind cutting it over.

FLGMwt:
> Loving this slow TS migration! Anyways, as written in inline comment I think positionalAccuracy is erroneous and should be fixed here. Also, a few props should be of Region type.

Aye, sorry I left `Region` a little loose. I've added what I think are the correct-as-used types where appropriate and left TODO comments where there's discrepancies which _may_ require functional changes.

Also, yeah, did that `positional_accuracy` change, ty for your testing!

jtklein (approved):
LGTM

PR #3798, opened 2026-07-03
rn-maps patch: don't ignore region change cbs on large map jumps

FLGMwt:
Fixes MOB-1529

the location picker lets you:
- pan
- search and select an iNat Place
- resolve "Current Location"

the latter two use `animateToRegion` and we display a loading indicator until rn-mapview's MapView's `handleRegionChangeComplete` callback triggers and we resolve the location.

On iOS `handleRegionChangeComplete` wasn't getting triggered for large map distances where the animation is and region change tracking are short-circuited. In my testing, it seems harmless to re-enable the change tracking so that we get our expected callbacks.

A new patch isn't ideal, but it's a lot cleaner than my other option of a timeout "cleanup" in case the callback never triggered (and would need separate changes for place & myLocation).

### DIFF
--- a/patches/react-native-maps+1.27.1.patch
+++ b/patches/react-native-maps+1.27.1.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/react-native-maps/ios/AirMaps/RNMapsMapView.mm b/node_modules/react-native-maps/ios/AirMaps/RNMapsMapView.mm
+index 447520d..8043dcb 100644
+--- a/node_modules/react-native-maps/ios/AirMaps/RNMapsMapView.mm
++++ b/node_modules/react-native-maps/ios/AirMaps/RNMapsMapView.mm
+@@ -60,7 +60,7 @@ - (void)animateToRegion:(NSString *)regionJSON duration:(NSInteger)duration{
+     NSDictionary* regionDic = [RCTConvert dictonaryFromString:regionJSON];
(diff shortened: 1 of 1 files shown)


FLGMwt:
I verified this with the `MapView` handlers by replacing `onRegionChangeComplete={handleRegionChangeComplete}` [here](https://github.com/inaturalist/iNaturalistReactNative/blob/1bcf82f4f7d617ddd188c747360e0f4b81ef8cc3/src/components/SharedComponents/Map/Map.tsx#L533) with:

```
onRegionChangeStart={( ...args ) => console.log( "onRegionChangeStart", ...args )}
onRegionChange={( ...args ) => console.log( "onRegionChange", ...args )}
onRegionChangeComplete={( ...args ) => {
  console.log( "onRegionChangeComplete", ...args );
  handleRegionChangeComplete( ...args );
}}
```

Before:

Selecting a place result or "current location" close to the current map region (when the map visibly animates), all log callbacks are triggered and loading state stops. Selecting a place result or "current location" far away, such as when the current map region is focused on New Mexico and I press "current location" to Minnesota, the map jumps, loading state is stuck and only `onRegionChangeStart` is fired.

After applying patch & rebuilding iOS:

Both cases trigger all handlers, loading state stops.

sepeterson (approved):
works for me

PR #3799, opened 2026-07-03
Traditional projects: post join project sync for offline chooser

jtklein:
Closes MOB-1524

On join success for a traditional project it is persisted to realm.

To test:
New app install
Login
Join a traditional project
Go to the Add to Projects screen and see the newly joined project in the list


sepeterson (approved):
I was able to add obs to a joined project offline

PR #3800, opened 2026-07-03
Correctly track owners_identification_from_vision through match screen

jtklein:
Closes MOB-1167

Do not set indicator for CV used to true for iconicTaxon which is a user choice not based on CV.

Before
Screenshot 2026-07-03 at 12 14 20

After
Screenshot 2026-07-03 at 12 14 34


sepeterson (approved):
nice

PR #3803, opened 2026-07-06
Mob 1292 filled id and comment icons not appearing until you navigate

abbeycampbell:
closes [MOB-1292](https://linear.app/inaturalist/issue/MOB-1292/filled-id-and-comment-icons-not-appearing-until-you-navigate-to)

- removes a hook that was setting all updates as `viewed=true` when the app was foregrounded
- refetch updates when pulling to refresh (so icons will fill when new ids and comments are added)
- add some logic to `useObservationsUpdates` that checks local update state against what the server considers unviewed or not, so if a user views a comment/id on web, we will reflect that in the app as well. Bumped the per_page limit to 100 because I imagine that will better account for power users with lots of activity.


src/components/NetworkService.ts
```diff
@@ -1,14 +1,9 @@
...
-  useIconicTaxa,
-  useObservationUpdatesWhenFocused,
-} from "sharedHooks";
+import { useIconicTaxa } from "sharedHooks";
 
 import useTaxonCommonNames from "./hooks/useTaxonCommonNames";
 import useWorkQueue from "./hooks/useWorkQueue";
 
 const NetworkService = ( ) => {
-  useObservationUpdatesWhenFocused( );
```
sepeterson, line 10: powerful

sepeterson (approved):
Looks good! Was able to test locally and saw the problem fixed.

<details>

Image

</details>

src/components/NetworkService.ts
abbeycampbell: yeah it felt good

PR #3808, opened 2026-07-07
move LAST_CRASH_DATA from zustandmmkv to installdata

FLGMwt:
closes MOB-1375, subtask of MOB-1372

full detail for class of issue this addresses in #3556 

tl;dr: we want to eliminate any _non_ zustand access to the mmkv kvs backing zustand. zustand can't be reactive to those updates and non-zustand consumers can't be reactive to their own `.get`s.


jtklein (approved):
LGTM, only one thing could you export "LAST_CRASH_DATA" from installData.tsx to tighten the usage contract, and make it visible that this is another key used there.

PR #3812, opened 2026-07-08
Enable mobile CV model 2.30

jtklein:
Closes MOB-1555
To enable the CV model 2.30 in the app we need to explicitly allow it in the vision plugin (for reasons).

Diff in the package is only these two commits:
https://github.com/inaturalist/vision-camera-plugin-inatvision/commit/214eb07b94ccf9d65a933a9b330f3c806f2c6f8d https://github.com/inaturalist/vision-camera-plugin-inatvision/commit/0354705c124e3a9fabcca3e5881e64cd93d2f0df

Have tested in Debug and Release mode on Android and iOS with cv model 2.30 included. AI camera works in all cases with and without location (= with or without geomodel). In airplane mode predictions from file also worked. A newly added species according to the blog post also works.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -3960,7 +3960,7 @@ PODS:
   - VisionCamera/React (4.7.3):
     - React-Core
     - VisionCamera/FrameProcessors
-  - VisionCameraPluginInatVision (6.0.0-rc.3):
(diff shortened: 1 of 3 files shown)


sepeterson (approved):

FLGMwt (approved):
update and upstream package change looks good!

PR #3814, opened 2026-07-08
MOB-1291: invalidate user queries on follow/unfollow

sepeterson:
closes MOB-1291

Was able to repro before and this fixed it


jtklein (approved):
Invalidating the query fixes the bug like a charm now that the API also has a fix for it.

PR #3815, opened 2026-07-08
TS migration for some files in Camera

jtklein:


src/components/Camera/StandardCamera/CameraOptionsButtons.tsx
sepeterson, line 67: Seeing a ts error here that we're missing 5 required props. I think they can all be made optional in TabletButtons' prop interface

sepeterson (approved):
Thanks for doing this! No blockers

PR #3816, opened 2026-07-08
MOB-1332: already-seen checkmark for ExploreV2 species tab

sepeterson:


src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -53,6 +55,64 @@ const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
...
+
+  // taxonIdChunks holds the cumulative list of all taxa viewed, sliced into groups of 10.
+  const taxonIdChunks: number[][] = [];
+  for ( let i = 0; i < taxonIds.length; i += SEEN_CHUNK_SIZE ) {
+    taxonIdChunks.push( taxonIds.slice( i, i + SEEN_CHUNK_SIZE ) );
+  }
+
+  // useQueries executes taxonIdChunks.length queries,
+  // relying on React Query's caching so each full chunk actually runs only once
+  const observedIds = useQueries( {
```
sepeterson, line 74: This is a new pattern and hard to follow so I added step-by-step explainers as best I understand it.

sepeterson:
I'm all ears if anyone would prefer not to introduce a new pattern here. It's also an option to just copy/paste the logic for this from SpeciesView

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -53,6 +55,64 @@ const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
...
+    queries: taxonIdChunks.map( chunk => ( {
+      queryKey: ["exploreV2SpeciesCountsSeen", currentUser?.id, chunk.join( "," )],
+      queryFn: ( ) => fetchSpeciesCounts( {
+        user_id: currentUser?.id,
```
jtklein, line 78: Haven't looked at the designs for this question, but is just something I have seen while testing: This query should not be restricted to the filters I have set in Explore correct?
So, when I a searching for species in a specific place, the checkmark is present because I have seen this species somewhere, but not necessarily in this specific place.
I can see that we haven't been doing this in V1, just to make sure this should stay the same.

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -53,6 +55,64 @@ const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
     { enabled },
   );
 
+  const SEEN_CHUNK_SIZE = 10;
```
jtklein: Why 10? Shouldn't we try to make as little queries as required, so go all the way up to max (which I believe is 200 but would need confimation).?

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -53,6 +55,64 @@ const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
...
+        fields: {
+          taxon: {
+            id: true,
+          },
+        },
+      } ),
+      enabled: !!( currentUser && chunk.length > 0 ),
+      // we don't really need to worry about a user's life list changing on this screen
+      staleTime: Infinity,
+      retry: ( failureCount: number, error: unknown ) => reactQueryRetry(
```
jtklein, line 90: Do we already want to create a wrapper for this hook, like we have for useQuery? So that retry logic is centralized?

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -53,6 +55,64 @@ const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
...
+      // the combine callback unions the taxon ids present in the results
+      // (i.e. observed by this user) across all chunks
+      results.forEach( r => {
+        const seen = ( r.data as { results?: SpeciesCountResult[] } | undefined )?.results;
+        seen?.forEach( x => ids.add( x.taxon.id as number ) );
+      } );
+      // results holds every chunk's query state (redundant fetches prevented by cache).
+      // Return a sorted array so this reference stays stable when the observed set is unchanged,
+      // which keeps renderItem memoized.
+      return Array.from( ids ).sort( ( a, b ) => a - b );
```
jtklein: I am a bit confused here. Why make a Set, an Array from this Set, and finally a Set again that is used outside here? Can we not just use `ids` from here?

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
```diff
@@ -53,6 +55,64 @@ const ExploreV2SpeciesView = ( { enabled, isConnected, params }: Props ) => {
...
+  const taxonIdChunks: number[][] = [];
+  for ( let i = 0; i < taxonIds.length; i += SEEN_CHUNK_SIZE ) {
+    taxonIdChunks.push( taxonIds.slice( i, i + SEEN_CHUNK_SIZE ) );
+  }
+
+  // useQueries executes taxonIdChunks.length queries,
+  // relying on React Query's caching so each full chunk actually runs only once
+  const observedIds = useQueries( {
+    queries: taxonIdChunks.map( chunk => ( {
+      queryKey: ["exploreV2SpeciesCountsSeen", currentUser?.id, chunk.join( "," )],
```
jtklein: Do we get any benefits from doing a join on the chunk array here? I haven't seen us using this in other places were we have complicated structures as queryKey elements. I have not found an array in a quick search just now but we are using the entire fields object in some cases.

tests/unit/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.test.js
```diff
@@ -0,0 +1,132 @@
...
+import { ExploreV2Provider } from "providers/ExploreV2Context";
+import React from "react";
+import factory from "tests/factory";
+import { renderComponent } from "tests/helpers/render";
+
+const mockSpeciesCounts = [
+  {
+    count: 12,
+    taxon: {
+      ...factory( "RemoteTaxon" ),
```
jtklein: We have a bit of a mix if using factory's return with in function overrides as the object, and spreading the result into an object. I have no opinion just noting some increasing divergence to the existing codebase, which I also contributed to.

jtklein (approved):
Have tested and this works on my end. I do not get why we do Set -> Array -> Set though for the results.

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
FLGMwt: Yeah, I'm curious why we would need anywhere close to 200 or chunk them anyway. Can't we restrict our query to the scroll window?

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
sepeterson: As far as I'm aware this behavior should stay the same. I'll confirm with Abhas

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
sepeterson: 10 because that's how many observations are returned per page by useInfiniteScroll. I'm pushing a commit to make this more explicit.

Would it be better to fetch 200 right off the bat even if we aren't necessarily going to fetch the taxa results to match?

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
sepeterson: To be honest I doubt we'll need to put useQueries in any other places. Even here I'd prefer a single API call but we have an established pattern of using two to get this info and afaik we don't do that any other places.

I'll add a comment here to split out a wrapper if we end up with another usage of useQueries

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
sepeterson: Yeah this is a little convoluted.

The intention is:

1.  First Set deduplicates ids
2. Convert to a sorted array for the return bc react-query returns a stable reference if the new value is equivalent to the old one but it won't compare Sets
3. Third Set because Set.has is more performant than Array.includes

In practice we shouldn't get duplicates here so I'll remove the first Set

src/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.tsx
sepeterson: No you're totally right. We can just use the array here

tests/unit/components/Explore/ExploreV2/screens/ExploreV2SpeciesView.test.js
sepeterson: I guess I prefer the non-spreading version so I'll use that here and add to my claude docs

PR #3817, opened 2026-07-09
Add logic to fallback to any device we get for all cameras

jtklein:
Closes MOB-1118


abbeycampbell (approved):
the ticket mentions adding extra logging around this (devices finished loading, device found via filter, fallback used) that weren't added. Do we think that's overkill?

jtklein:
Yes, maybe the logging is too defensive for a non existent problem later. We will still get error log on server if there is no device, so we can monitor how often that happens.

PR #3818, opened 2026-07-09
fix useLocationSearch test

sepeterson:
- use waitfor like the other test cases
- clear queryClient in beforeEach


jtklein (approved):
LGTM

PR #3820, opened 2026-07-09
MOB-1341: ExploreV2 Location Defaults

sepeterson:
Default "Nearby" and "Worldwide" location defaults

also a change to how exploreV2 handles needing location permissions


src/components/Explore/ExploreV2/components/LocationDefaultOptions.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React from "react";
+import useTranslation from "sharedHooks/useTranslation";
+import colors from "styles/tailwindColors";
+
+interface Props {
+  onSelectNearby: ( ) => void;
+  onSelectWorldwide: ( ) => void;
+}
+
+const ROW_CLASSES = "flex-row items-center px-[15px] h-[42px] border-b border-lightGray";
```
jtklein: Do we want a fixed height here? I'd rather go with a py 11px. I haven't checked how a11y font size scaling works with a fixed height, but I would assume it breaks.

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -53,6 +60,12 @@ const INPUT_BOX_CLASSES = classnames(
 // single FlatList so we never conditionally mount/unmount it.
 type SearchResultItem = UniversalSearchResultItem | LocationSearchResultItem;
 
+type SelectedLocation =
+  | { type: "place"; place: Place }
+  | { type: "nearby"; lat: number; lng: number; radius: number }
+  | { type: "nearby-needs-permission" }
```
jtklein, line 66: I feel we should separate these concerns. Needing location permission is different enough than what to query for in my mind.

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -130,16 +143,45 @@ const UniversalSearch = ( ) => {
...
+    switch ( next.placeMode ) {
+      case EXPLORE_V2_PLACE_MODE.NEARBY:
+        setSelectedLocation( {
+          type: "nearby", lat: next.lat, lng: next.lng, radius: next.radius,
+        } );
+        break;
+      case EXPLORE_V2_PLACE_MODE.NEEDS_PERMISSION:
+        setSelectedLocation( { type: "nearby-needs-permission" } );
+        break;
+      default:
```
jtklein, line 172: Are you open to add `case EXPLORE_V2_PLACE_MODE.WORLDWIDE` above default as a fallthrough for completeness sake?

src/providers/ExploreV2Context.tsx
```diff
@@ -168,15 +170,19 @@ export type DefaultExploreV2Location =
...
   }
-  return {
-    placeMode: EXPLORE_V2_PLACE_MODE.NEARBY,
-    lat: location.latitude,
-    lng: location.longitude,
-    radius: 1,
-  };
+  // No coordinates, fallback to worldwide if we already have perms
+  const hasPermission = await checkLocationPermissions( );
+  return hasPermission
```
jtklein, line 183: This is getting a bit complicated to maintain comprehension. V2Container is making it's own permission check before this here.

src/providers/ExploreV2Context.tsx
```diff
@@ -168,15 +170,19 @@ export type DefaultExploreV2Location =
...
+    };
   }
-  return {
-    placeMode: EXPLORE_V2_PLACE_MODE.NEARBY,
-    lat: location.latitude,
-    lng: location.longitude,
-    radius: 1,
-  };
+  // No coordinates, fallback to worldwide if we already have perms
+  const hasPermission = await checkLocationPermissions( );
```
jtklein, line 182: Do we need to update the "defaultExploreV2Location" tests?

src/components/Explore/ExploreV2/components/LocationDefaultOptions.tsx
```diff
@@ -0,0 +1,46 @@
...
+      </Pressable>
+      <Pressable
+        accessibilityRole="button"
+        accessibilityLabel={t( "Worldwide" )}
+        accessibilityHint={t( "Filters-observations-to-anywhere-in-the-world" )}
+        className={ROW_CLASSES}
+        onPress={onSelectWorldwide}
+        testID="LocationDefaultOptions.worldwide"
+      >
+        <INatIcon name="globe-outline" size={15} color={colors.darkGray} />
```
jtklein: This is not the same icon as in the designs.

This goes beyond this PR, but I remember chatting with Abhas that it would be cool if we can do the same that iNaturalist Classic on iOS does which is randomly cycling through the world icon with different continents featured.

jtklein (approved):
This works for me, does what is advertised and is still behind a feature flag. So, my verdict is Approved.
However, for my personal taste, mixing concerns of permission status and default search location UI state is too much of a burden for future maintenance. I have no suggestions as to how to make it different. So, I am fine with merging in as is.
 
One bug: without location permission:
Switching from Worldwide back to Nearby still shows the Worldwide Obs an Species numbers in the tabs.

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
sepeterson: Ya I agree. I have an idea that I will follow up on here https://github.com/inaturalist/iNaturalistReactNative/pull/3832

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
sepeterson: Yes! but I will address this (likely rm this function in https://github.com/inaturalist/iNaturalistReactNative/pull/3832)

src/providers/ExploreV2Context.tsx
sepeterson: Agreed. I hope to address this in https://github.com/inaturalist/iNaturalistReactNative/pull/3832 to have just one permission check

src/providers/ExploreV2Context.tsx
sepeterson: This'll change a bit in the followup so I'll address there

src/components/Explore/ExploreV2/components/LocationDefaultOptions.tsx
sepeterson: Good eye! Swapped it out

PR #3822, opened 2026-07-09
Mob 1461 create hook for server ordered observations in myobs

abbeycampbell:
closes [MOB-1461](https://linear.app/inaturalist/issue/MOB-1461/create-hook-for-server-ordered-observations-in-myobs)

Adds the groundwork for sorting MyObs by something other than the current default (and eventually for search & grouped by iconic taxa) in the form of two hooks:

`useServerOrderedObservations` — hook that fetches a page of server-ordered observation uuids for a given sort and upserts the results into Realm. Pagination to be handled in a follow-up.
`useMyObservationsQuery` — wrapper hook that reads the current sort selectionfrom `MyObservationsContext` and picks a data source: the default sort stays fully local so offline experience is preserved, and any other sort uses the aforementioned hook with the locally-unsynced observations merged in and pinned to the top.

Also adds a debug sheet in the flavor or the one we have in ExploreV2, for testing. This new logic isn't wired to anything but that right now; that will happen in [MOB-1479](https://linear.app/inaturalist/issue/MOB-1479/wire-up-observation-sort-in-myobs).

### DIFF
--- a/src/components/MyObservations/hooks/useMyObservationsQuery.ts
+++ b/src/components/MyObservations/hooks/useMyObservationsQuery.ts
@@ -1,0 +1,89 @@
+import { RealmContext } from "providers/contexts";
+import { useMyObservations } from "providers/MyObservationsContext";
+import { useMemo } from "react";
+import { OBSERVATIONS_SORT } from "sharedHelpers/observationsSort";
+import useLocalObservationIds from "sharedHooks/useLocalObservationIds";
+
+import useServerOrderedObservations from "./useServerOrderedObservations";
+
+const { useQuery } = RealmContext;
+
+const NOOP_REFETCH = ( ) => undefined;
+
+interface UseMyObservationsQueryResult {
+  observationIds: { uuid: string }[];
+  isServerAuthoritative: boolean;
+  isLoading: boolean;
+  error: Error | null;
+  refetch: ( ) => void;
+}
+
+// We want to preserve offline behavior for the default sort (created at, desc) so a user can see
+// and interact with their obs offline. This hook uses selected sort to determine whether Realm or
+// the server should be the authoritative source of a user's observations (unsynced obs
+// are always merged in at the top regardless of source).
+
+const useMyObservationsQuery = ( ): UseMyObservationsQueryResult => {
+  const { state } = useMyObservations( );
+  const isDefaultSort = state.observationsSort === OBSERVATIONS_SORT.DATE_UPLOADED_NEWEST;
+
+  const localObservationIds = useLocalObservationIds( );
+
+  const {
+    observationIds: serverObservationIds,
+    isLoading,
+    error,
+    refetch,
+  } = useServerOrderedObservations( {
+    sortBy: state.observationsSort,
+    enabled: !isDefaultSort,
+  } );
+
+  // if we want obs from the server, we'll want to prepend local, unsynced obs to the top
+  const unsyncedObs = useQuery<{ uuid: string }>(
+    {
+      type: "Observation",
+      query: observations => observations
+        .filtered(
+          "needs_sync == true AND "
+          + "(_deleted_at == nil OR _pending_deletion == false OR _pending_deletion == nil)",
+        )
+        .sorted( "_created_at", true ),
+      keyPaths: ["uuid"],
+    },
+    [],
+  );
+
+  const unsyncedObservationIds = useMemo(
+    ( ) => unsyncedObs.map( ( { uuid } ) => ( { uuid } ) ),
+    [unsyncedObs],
+  );
+
+  // dedupe in case any locally unsynced obs also exist in the server results
+  const observationIds = useMemo( ( ) => {
+    if ( isDefaultSort ) return localObservationIds;
+    const unsyncedUuids = new Set( unsyncedObservationIds.map( o => o.uuid ) );
+    return [
+      ...unsyncedObservationIds,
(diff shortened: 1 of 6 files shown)


src/components/MyObservations/hooks/useMyObservationsQuery.ts
```diff
@@ -0,0 +1,89 @@
...
+    isLoading,
+    error,
+    refetch,
+  } = useServerOrderedObservations( {
+    sortBy: state.observationsSort,
+    enabled: !isDefaultSort,
+  } );
+
+  // if we want obs from the server, we'll want to prepend local, unsynced obs to the top
+  const unsyncedObs = useQuery<{ uuid: string }>(
```
abbeycampbell, line 43: @FLGMwt curious your thoughts on this idea: `useLocalObservationIds` and `useMyObservationsQuery` each run their own Realm query to get unsynced observations. So, regardless of whether we are using default sort or not, one of those sets of results will go unused, which means we're querying an extra time for no reason. I think we could get around this by including a `needsSync` boolean alongside the uuid in the `useLocalObservationIds` results. that way here, we could just filter to figure out which ones are actually unsynced rather than querying again.

FLGMwt: Reviewing this today but short answer here: we do want separate queries for these, even if (for now), one or the other is unused.

As it is, the obsId return array value is referentially stable and *only* changes when an id is added or removed. If we included needs sync, it would change when any obs's sync status changed.

We normally shouldn't be concerned with the referential stability of pure data, but since this one informs list render performance, it's worth considering here.

I'll have a think on how we can add a "enabled" option to the Realm hooks.

src/components/MyObservations/hooks/useMyObservationsQuery.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  // dedupe in case any locally unsynced obs also exist in the server results
+  const observationIds = useMemo( ( ) => {
+    if ( isDefaultSort ) return localObservationIds;
+    const unsyncedUuids = new Set( unsyncedObservationIds.map( o => o.uuid ) );
+    return [
+      ...unsyncedObservationIds,
+      ...serverObservationIds.filter( o => !unsyncedUuids.has( o.uuid ) ),
+    ];
+  }, [isDefaultSort, localObservationIds, unsyncedObservationIds, serverObservationIds] );
```
FLGMwt, line 70: `Set` [retains insertion order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#description:~:text=You%20can%20iterate%20through%20the%20elements%20of%20a%20set%20in%20insertion%20order) so you can order/dedupe in the set constructor itself

```
return Array.from(new Set([
  ...unsyncedObservationIds,
  ...serverObservationIds,
])
```

src/components/MyObservations/hooks/useServerOrderedObservations.ts
```diff
@@ -0,0 +1,88 @@
...
+  useEffect( ( ) => {
+    if ( data?.results ) {
+      Observation.upsertRemoteObservations( data.results, realm );
+    }
+  }, [data?.results, realm] );
+
+  const observationIds = useMemo(
+    ( ) => ( data?.results || [] ).map( ( { uuid } ) => ( { uuid } ) ),
+    [data?.results],
+  );
```
FLGMwt: Should these live in the queryFn?

src/components/MyObservations/hooks/useMyObservationsQuery.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import useServerOrderedObservations from "./useServerOrderedObservations";
+
+const { useQuery } = RealmContext;
+
+const NOOP_REFETCH = ( ) => undefined;
+
+interface UseMyObservationsQueryResult {
+  observationIds: { uuid: string }[];
+  isServerAuthoritative: boolean;
```
FLGMwt, line 15: I'm iffy on `Authoritative` here which to me suggests something about data integrity. Are we just answering the question `isDefaultSearch`?

FLGMwt (approved):
omg I'm so excited, this looks great! So clean < 3 < 3 < 3

src/components/MyObservations/hooks/useMyObservationsQuery.ts
abbeycampbell: Interesting, I meant it like, "the server is in charge of the ordering". 
I don't want to have the name include 'Search', because soon "search" will be a part of this as well (searches are also 'server authoritative'), and I think that has the potential to get confusing since this applies to sort as well. but I see what you mean. Maybe `isSourcedFromServer`? or `requiresServer`, something like that. Let me know what you think, I can change it in a follow up or when I add pagination.

src/components/MyObservations/hooks/useMyObservationsQuery.ts
abbeycampbell: yes, but Set dedupes objects by reference, not value, so since we're dealing with `{uuid}` objects we'd want to dedupe on the id strings only, like:

 ```
const orderedUuids = new Set( [
      ...unsyncedObservationIds.map( o => o.uuid ),
      ...serverObservationIds.map( o => o.uuid ),
    ] );
    return Array.from( orderedUuids, uuid => ( { uuid } ) );
```

I kind of think the intention is clearer in the original.

PR #3824, opened 2026-07-10
Traditional projects: joined projects sync triggers on app startup and chooser screen

jtklein:
Closes MOB-1535

Caches joined projects in Realm on app start (deferred task) and when opening the AddToProject screen. Adds syncJoinedProjects helper (fetch, upsert, prune stale + empty list).

Originally the ticket also laid out to paginate through all possible multiple pages of a users joined projects. But I pushed the remaining pagination/online-guard work to a new ticket MOB-1568. Because I wanted to get a review on the following already:

Took the one API call I am using here out into a new file to type: `fetchUserProjects`. In `usersTyped` there is now one API with a typed full api response (not only returning results). I have added the possibility to specify with a generic type what the results array entails. For me this would be my blueprint forward for such a wrapper. So, I'd be happy to get some eyes on it.
Some open questions:
How can I make it clearer that as soon as you add a fields param to opts the returned results shape will differ? I did add a default api result type which I want to be using as the result when there is no fields param set.
I had to add the two possible error return because `handleError` does return these two types in some code paths. Two codepaths are only optional and I am not using the option here. One code path depends on the error thrown by the API, so potentially possible. I could "typecast" the return of handleError as `never` to get rid of the error types. Or I could add handling around if these types are the return in the caller of this wrapper. Neither seems quite as ideal to me. Any other ideas?

### DIFF
--- a/src/api/types.d.ts
+++ b/src/api/types.d.ts
@@ -1,4 +1,8 @@
 // Generic types, please keep alphabetized
+export interface ApiDefaultResult {
+  id: number;
+}
+
 export interface ApiOpts {
   api_token?: string;
 }
--- a/src/api/usersTyped.ts
+++ b/src/api/usersTyped.ts
@@ -2,13 +2,15 @@ import inatjs from "inaturalistjs";
 
 import type { ErrorWithResponse, INatApiError } from "./error";
 import handleError from "./error";
-import type { ApiOpts, ApiParams, ApiResponse } from "./types";
+import type {
+  ApiDefaultResult, ApiOpts, ApiParams, ApiResponse,
+} from "./types";
 
 interface UsersProjectsParams extends ApiParams {
   id: number;
 }
 
-const fetchUserProjects = async <T>(
+const fetchUserProjects = async <T = ApiDefaultResult>(
   params: UsersProjectsParams = {},
   opts: ApiOpts = {},
 ): Promise<ApiResponse<T> | null | ErrorWithResponse | INatApiError> => {
--- a/src/api/usersTyped.ts
+++ b/src/api/usersTyped.ts
@@ -2,8 +2,16 @@ import inatjs from "inaturalistjs";
 
 import type { ErrorWithResponse } from "./error";
 import handleError from "./error";
+import type { ApiParams } from "./types";
 
-const fetchUserProjects = async ( params = {}, opts = {} ): Promise<object> => {
+interface UsersProjectsParams extends ApiParams {
+  id: number;
+}
+
+const fetchUserProjects = async (
+  params: UsersProjectsParams = {},
+  opts = {},
+): Promise<object> => {
   try {
     const response = await inatjs.users.projects(
       params,
--- a/src/api/usersTyped.ts
+++ b/src/api/usersTyped.ts
@@ -2,7 +2,7 @@ import inatjs from "inaturalistjs";
 
 import type { ErrorWithResponse } from "./error";
 import handleError from "./error";
-import type { ApiParams } from "./types";
+import type { ApiOpts, ApiParams } from "./types";
 
 interface UsersProjectsParams extends ApiParams {
   id: number;
@@ -10,7 +10,7 @@ interface UsersProjectsParams extends ApiParams {
 
 const fetchUserProjects = async (
   params: UsersProjectsParams = {},
-  opts = {},
+  opts: ApiOpts = {},
 ): Promise<object> => {
   try {
     const response = await inatjs.users.projects(
--- a/src/api/usersTyped.ts
+++ b/src/api/usersTyped.ts
@@ -1,0 +1,21 @@
+import inatjs from "inaturalistjs";
+
+import type { ErrorWithResponse } from "./error";
+import handleError from "./error";
+
+const fetchUserProjects = async ( params = {}, opts = {} ): Promise<object> => {
+  try {
+    const response = await inatjs.users.projects(
+      params,
+      opts,
+    );
+    return response;
+  } catch ( e ) {
+    return handleError(
+      e as ErrorWithResponse,
+      { context: { functionName: "fetchUserProjects", opts } },
+    );
+  }
+};
+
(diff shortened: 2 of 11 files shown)


src/api/types.d.ts
```diff
@@ -1,4 +1,8 @@
 // Generic types, please keep alphabetized
+export interface ApiDefaultResult {
```
jtklein, line 2: When we query the API without a fields param, we only get the id of the item back.

src/api/usersTyped.ts
```diff
@@ -0,0 +1,29 @@
...
+import type {
+  ApiDefaultResult, ApiOpts, ApiParams, ApiResponse,
+} from "./types";
+
+interface UsersProjectsParams extends ApiParams {
+  id: number;
+}
+
+const fetchUserProjects = async <T = ApiDefaultResult>(
+  params: UsersProjectsParams,
```
jtklein, line 14: Have removed the default empty object here. We were never calling this query without params, if we would have this should have errored out anyways because a GET for an :id driven endpoint like users/{id}/projects will reject when called without an id.

src/components/ProjectDetails/ProjectDetailsContainer.tsx
```diff
@@ -108,10 +108,9 @@ const ProjectDetailsContainer = ( ) => {
     ( _, optsWithAuth ) => joinProject( { id }, optsWithAuth ),
     {
       onSuccess: ( ) => {
-        // Persist traditional projects on join
-        if ( project?.project_type === "" ) {
```
jtklein, line 112: That was an error, the build plan says to persist all projects, not only traditional projects.

src/components/hooks/useDeferredStartup.ts
```diff
@@ -107,6 +109,10 @@ const useDeferredStartup = ( ) => {
       formatApiDatetime( "1970", i18n, { timeZone: "Etc/UTC" } );
       return Promise.resolve();
     } );
+    const id11 = deferTask( "syncJoinedProjects", async () => {
+      const currentUserId = User.currentUser( realm )?.id;
+      await syncJoinedProjects( realm, currentUserId );
```
jtklein, line 117: The sync fct has an early return for logged out users, but it will still leave a time log of how long it ran. I don't mind but maybe I don't see something.

src/sharedHelpers/syncJoinedProjects.ts
```diff
@@ -0,0 +1,56 @@
+import { PROJECT_SUMMARY_POF_FIELDS } from "api/fields";
+import type { ApiProjectSummaryWithPOF } from "api/types";
+import fetchUserProjects from "api/usersTyped";
+import { getJWT } from "components/LoginSignUp/AuthenticationService";
+import type Realm from "realm";
+import Project from "realmModels/Project";
+
+import safeRealmWrite from "./safeRealmWrite";
+
+const deleteNotRemoteProjects = ( remoteProjects: number[], realm: Realm ) => {
```
jtklein, line 10: Copied and adapted from deleteRemoteObservations

src/sharedHelpers/syncJoinedProjects.ts
```diff
@@ -0,0 +1,56 @@
...
+      ? realm.objects( "Project" )
+      : realm.objects( "Project" )
+        .filtered( `NOT (id IN { ${remoteProjects} })` );
+    localProjectsToDelete.forEach( project => {
+      realm.delete( project );
+    } );
+  }, "removing projects from realm that are not remote" );
+};
+
+async function syncJoinedProjects(
```
jtklein, line 23: Following syncRemoteObservations

src/components/AddToProjects/hooks/useSyncJoinedProjects.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const { useRealm } = RealmContext;
+
+const useSyncJoinedProjects = ( ) => {
+  const realm = useRealm( );
+
+  const currentUser = useCurrentUser( );
+
+  useEffect( ( ) => {
+    syncJoinedProjects( realm, currentUser?.id );
```
FLGMwt, line 17: I would prefer to keep `syncJoinedProjects` strict on non-null `currentUserId: number` and guard at the call site here.

src/components/AddToProjects/AddToProjects.tsx
```diff
@@ -35,6 +36,7 @@ const ItemSeparator = ( ) => <View className="border-b border-lightGray" />;
 const AddToProjects = ( ) => {
   const { t } = useTranslation( );
   const navigation = useNavigation( );
+  useSyncJoinedProjects( );
```
FLGMwt, line 39: I'm colored by confusion around the existing "useSync" hooks and what/when they _do_ things.

What do you think about the following API for this concern?:

```
useFocusEffect(useCallback(() => {
  syncJoinedProjects();
}));
```

src/sharedHelpers/syncJoinedProjects.ts
```diff
@@ -0,0 +1,56 @@
...
+  realm: Realm,
+  currentUserId: number | undefined,
+): Promise<void> {
+  if ( !currentUserId ) {
```
FLGMwt: mentioned above but I would like to remove this guard and check at callsite.

FLGMwt (changes requested):

FLGMwt (approved):
on reflection, my null-guard concern didn't warrant "Requesting Changes" because it follows existing patterns.

I still want that conversation, but it oughtn't block this PR

src/sharedHelpers/syncJoinedProjects.ts
jtklein: I do agree

src/components/AddToProjects/hooks/useSyncJoinedProjects.ts
jtklein: Agreed, have done so now.

PR #3825, opened 2026-07-10
Mob 1479 wire up observation sort in myobs UI only

abbeycampbell:
Partial [MOB-1479](https://linear.app/inaturalist/issue/MOB-1479/wire-up-observation-sort-in-myobs)

I decided to put this behind a separate feature flag from search, since there's a strong possibility that sort is ready first and I don't want to keep touching/changing the feature flag names to keep them properly descriptive. I think it will be nice to have the option to test and ship them independently.

### DIFF
--- a/src/components/MyObservations/MyObservationsResults.tsx
+++ b/src/components/MyObservations/MyObservationsResults.tsx
@@ -20,6 +20,7 @@ import { Alert } from "react-native";
 import Observation from "realmModels/Observation";
 import Taxon from "realmModels/Taxon";
 import type { RealmObservation } from "realmModels/types";
+import type { OBSERVATIONS_SORT } from "sharedHelpers/observationsSort";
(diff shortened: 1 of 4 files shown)


src/components/MyObservations/MyObservationsResults.tsx
```diff
@@ -134,6 +135,13 @@ const MyObservationsResults = ( ) => {
     } );
   };
 
+  const setObservationsSortOptionId = ( value: OBSERVATIONS_SORT ) => {
+    myObsDispatch( {
+      type: MY_OBSERVATIONS_ACTION.SET_OBSERVATIONS_SORT,
+      observationsSort: value,
```
jtklein, line 141: Just realized that observationsSort is also a file name, so searching for this gives a lot of results.

jtklein (approved):
I love it when a plan comes together.

PR #3826, opened 2026-07-13
upgrade to nativewind@4

FLGMwt:
Closes MOB-1148

This PR upgrades to [Nativewind@4](https://www.nativewind.dev/blog/announcement-nativewind-v4) (note: v3 was canceled, 4 is the next step from 2). Nativewind 4 is a significant rewrite so this necessarily touches quite a few files. 

I cherry-picked the first commit from a spike I did on this in May, which includes everything necessary to wire up the project on v4. There were a number of things off in this commit around text, maps, but the core interface are functional.

For fixing the rest, this is my first time using Claude for a "big" problem. I thought it would be well-suited at finding the nooks and crannies of the app to apply systemic / migratory changes. I'm sharing the Claude Plan Document and Claude's summary of changes below:

<details>

<summary>Plan for fixing some enumerated styling issues</summary>

# Finish the NativeWind 2 → 4 migration

## Context

The spike commit (`7a999b474`) did the core wiring correctly: babel preset, metro `withNativeWind`, `global.css` with `@tailwind` directives, `nativewind/preset` in tailwind config, and removal of `styled()`. I verified empirically (compiled the real CSS through nativewind's `cssToReactNativeRuntime`, and rendered real Typography components through the repo's actual babel chain in jest) that **the font classes themselves compile and apply correctly** — `font-Lato-*` → `fontFamily: "Lato-*"` end to end, including with the react-compiler plugin active.

**Answer to the open question: nothing needs to move from `tailwind.config.js` to CSS.** NativeWind 4 uses Tailwind 3's JS config; `global.css` stays as just the three `@tailwind` directives. (Moving theme to CSS is a Tailwind v4 / NativeWind v5 thing.)

The reported symptoms map to four real, verified issues:

| Symptom | Root cause |
|---|---|
| Cramped / smaller | NativeWind 4 inlines `rem` at **14px** (v2 used 16px) → all rem-based spacing (`p-*`, `m-*`, `gap-*`, `h-22`, …) shrank 12.5%. Verified: `p-4` compiles to 14 instead of 16. |
| System font shows | Third-party components lost `styled()` and are NOT auto-interop'd in v4 — `className` on them is silently ignored (`BottomSheetTextInput`, `SafeAreaView`, `LinearGradient`, `FasterImageView`). Also a pre-existing bug: `KebabMenu.tsx:43` sets `fontFamily: tailwindFontMedium` (the *class name* `"font-Lato-Medium"`) instead of the font name → kebab menu items always fell back to system font. |
| Wrong weights | v4 resolves conflicting classes by **stylesheet order, not className string order**. Verified via probe: an element with both `font-Lato-Regular` and `font-Lato-Bold` renders **Regular** regardless of string order (Regular is generated last among the Lato keys). Anywhere that appends a font override onto a base class that already has a font class now picks the wrong one. |
| Jest suite broken | The css-interop babel plugin rewrites `React.createElement(...)` to a hoisted `_ReactNativeCSSInterop.createInteropElement(...)` import; `jest.mock` factories in `tests/jest.setup.js:196-208` use `React.createElement` → babel-plugin-jest-hoist rejects the out-of-scope reference. |

### 1. Restore 16px rem base — `metro.config.js`
```js
withNativeWind( configWithRozenite, { input: "./global.css", inlineRem: 16 } )
```
This alone should fix the app-wide cramped look.

### 2. Register third-party components — `src/components/styledComponents.ts`
Core RN components (View, Text, TextInput, Modal, Pressable, ScrollView, Image, ImageBackground, KeyboardAvoidingView…) are auto-registered in v4 — leave those as plain re-exports. `PressableWithTracking` is fine too (it forwards props to core `Pressable` from app-compiled code). Register the rest:

```ts
import { cssInterop } from "nativewind";

const SafeAreaView = cssInterop( UnstyledSafeAreaView, { className: "style" } );
const LinearGradient = cssInterop( UnstyledLinearGradient, { className: "style" } );
const FasterImageView = cssInterop( UnstyledFasterImageView, { className: "style" } );
const BottomSheetTextInput = cssInterop( StyledBottomSheetTextInput, { className: "style" } );
```
(Keep the existing `SafeAreaView === undefined` jest fallback.)

### 3. Wrong weights — fix real double-font-class sites
Most elements carry only one font class because `InatText`'s `{...props}` spread *replaces* the merged className (pre-existing behavior, unchanged by the migration — verified in the probe). So this is an audit, not a rewrite:
- Grep for places that combine a component/base classes containing one `font-Lato-*` with a caller-supplied second `font-*` class via proper `classnames()` merging (e.g. `src/navigation/`, `CustomTabBar`, button/label components — the files matching `props.className` + `classnames`).
- Fix each by removing the redundant base font class or making the override the only font class present.
- Do NOT reorder `fontFamily` keys in `tailwind.config.js` to game the cascade — it can't restore string-order semantics and just moves the problem.

### 4. Pre-existing font bug (drive-by) — `src/components/SharedComponents/KebabMenu.tsx:43`
`fontFamily: tailwindFontMedium` → `fontFamily: fontMedium` (import from `appConstants/fontFamilies`).

### 5. Fix jest
- `tests/jest.setup.js:196-208`: rewrite the two `jest.mock` factories to avoid the createElement rewrite — e.g. `jest.fn( ( { children } ) => children ?? null )`, or `jest.requireActual( "react" ).createElement` if a host element is required (the plugin only rewrites bindings imported via `require("react")`/`import`).
- Add a CSS module mapping so `App.js`'s `import "../../global.css"` doesn't break tests: `moduleNameMapper: { "\\.css$": "<rootDir>/tests/mocks/cssMock.js" }` (new empty-object mock file).
- Run both suites (`npm run test:unit`, integration). Expect some churn: className styles do NOT resolve to style objects in jest anymore (metro injects the compiled CSS; jest has no CSS registered), so snapshots/assertions that relied on tailwind-derived styles may need updating. `tests/unit/components/SharedComponents/DisplayTaxon.test.js` asserts `opacity: 0.5` via `toHaveStyle` — check whether that comes from an inline style (fine) or a class (needs adjustment).
- Contingency if many tests depend on resolved tailwind styles: register compiled CSS in a `setupFilesAfterEnv` using `nativewind/test` helpers (requires adding `@tailwindcss/container-queries` as a devDependency — I already confirmed the helper needs it).

### 6. Housekeeping
- Bump `tailwindcss` `^3.3.2` → `^3.4.17` (nativewind 4's recommended line; peer range allows it).
- Delete the probe file I created: `tests/unit/nativewind4-font-probe.test.tsx`.
- `node_modules` has an unsaved `@tailwindcss/container-queries` install (`--no-save`) — either add it as devDependency (contingency above) or `npm install` to restore.

## Verification

1. `npm run ios` (or android) and check against `main` side by side:
   - **Developer > UI Library > Typography screen** (`src/components/Developer/UiLibrary/Typography.js`) — every Heading/Body variant shows correct Lato weight/size.
   - Spacing/padding on the observations list and detail screens no longer cramped (rem fix).
   - `BottomSheetTextInput`-based inputs and SearchBar show Lato, gradients/safe-area layouts styled again.
   - Kebab menu items show Lato-Medium.
2. Optionally call `verifyInstallation()` from `nativewind` once in dev to confirm the runtime sees the compiled CSS, then remove.
3. `npm run test:unit` and `npm run test:integration` pass.

</details>

<details>

<summary>Summary of changes</summary>

NativeWind 2 → 4 migration summary

Your spike wiring (babel preset, metro withNativeWind, preset in tailwind config, styled() removal) was correct. The problems were all runtime behavior changes between v2 and v4:

The core mental-model shift (explains almost every symptom): NativeWind 2's babel plugin resolved className on every capitalized component into styles at the call site, cascading through wrapper layers as style arrays where later-in-array won. NativeWind 4 only styles components explicitly registered with the interop, className rides along as a plain prop, conflicting classes resolve by stylesheet order instead of string order, and inline styles beat classes.

What was fixed:
1. rem base 16→14 shrank all spacing 12.5% → inlineRem: 16 in metro config ("cramped" look)
2. Third-party components lost styling → cssInterop() registrations for SafeAreaView, LinearGradient, FasterImageView, BottomSheetTextInput, and MapView (the broken maps — flex-1 was dead so the map had no size)
3. Typography defaults dropped when callers passed className → all 15 Typography components + ScreenShell + Button now merge with tailwind-merge so later classes win per property, restoring v2 semantics (fixed wrong weights, system fonts, black-instead-of-darkGray, and the "overexposed" modals where bg-white was beating bg-black)
4. Inline textAlign: "left" in InatText beat text-center classes → default skipped when an alignment class is present (Licenses link, button labels, modal text)
5. space-x/y-* polyfills removed in v4 → converted all 34 usages to gap-* (login pill spacing, DQA screens, sheets)
6. className dead on SVG components → arrow on Me tab wrapped in a positioned View
7. Function styles dropped by the interop → INatIconButton uses active:opacity-95/opacity-50 classes; also now forwards className
8. Animated.View + className unreliable → 6 call sites converted to inline styles
9. Metro crashed on tailwind hot updates → bumped to nativewind 4.2.6 / css-interop 0.2.6 (upstream fix); tailwindcss bumped to 3.4.x

Jest: CSS is compiled once in globalSetup and injected per worker (tests/nativewind.setup.js), so classNames resolve to real styles in tests like they did in v2; two jest.mock factories needed jest.requireActual("react") to dodge the interop's createElement rewrite; snapshots re-blessed after verifying diffs were value-identical.

Drive-by fix: KebabMenu was passing the class name "font-Lato-Medium" as a fontFamily value (kebab menus have been system-font for a while).

</details>

<details>

<summary>Context for twMerge / classnames</summary>

why use twmerge in some cases like @src/components/SharedComponents/ViewWrapper.tsx and classnames() in some places like @src/components/SharedComponents/Typography/Body1.tsx ?

Short answer: twMerge is doing conflict resolution, classnames is just string concatenation — and on this NativeWind 4 branch that difference actually matters.

twMerge = conflict resolution. In NativeWind 4, when one element has two conflicting classes (bg-white bg-black), the winner is decided by stylesheet order, not the order in the className string — so plain concatenation no longer lets a caller's class beat a component's default. twMerge fixes that by actually parsing the Tailwind classes and dropping the earlier one of each conflicting pair, so "later in the string wins" behavior is restored deterministically. That's why it's used in the components that own overridable defaults and render the actual native element: ViewWrapper's ScreenShell (bg-white must lose to a caller's bg-black), InatText, Button.

classnames = cheap concatenation (and conditionals). Body1 doesn't render a Text itself — it just prepends text-base + the font class and hands the combined string to InatText. InatText then runs the entire final string through twMerge before it hits the real <Text>. So the conflict resolution still happens exactly once, at the leaf; the intermediate layer only needs to join strings in the right order (caller last). Running twMerge at every layer would be redundant work. classnames also supports the { "some-class": condition } object syntax that ~99 files in the app rely on, which twMerge doesn't do.

Rule of thumb the branch follows: twMerge at the point where a className finally lands on a native element and defaults must be overridable; classnames everywhere upstream of that. The comments in InatText.tsx:24-26 and ViewWrapper.tsx:28-30 document the why.

One thing to keep an eye on: this only stays safe as long as every classnames-composed string eventually flows through a twMerge-ing leaf like InatText. A wrapper that concatenates defaults and passes straight to a raw styled View/Text would silently lose caller overrides — a merge conflict of the kind git can't help you with.

</summary>


src/components/AddObsBottomSheet/AddObsBottomSheet.tsx
```diff
@@ -31,7 +31,7 @@ const AI_CAMERA_SUPPORTED = ( Platform.OS === "ios" && majorVersionIOS >= 11 )
   || ( Platform.OS === "android" && Platform.Version > 21 );
 
 const GREEN_CIRCLE_CLASS = "bg-inatGreen rounded-full h-[36px] w-[36px] mb-2";
-const ROW_CLASS = "flex-row justify-center space-x-4 w-full flex-1";
+const ROW_CLASS = "flex-row justify-center gap-x-4 w-full flex-1";
```
FLGMwt, line 34: > space-x/y-* polyfills removed in v4 → converted all 34 usages to gap

src/components/SharedComponents/Map/Map.tsx
```diff
@@ -39,6 +40,10 @@ import LocationIndicator from "./LocationIndicator";
 import ObscuredLocationIndicator from "./ObscuredLocationIndicator";
 import SwitchMapTypeButton from "./SwitchMapTypeButton";
 
+// className support: third-party components need explicit registration with
+// nativewind 4 for className to have any effect
+cssInterop( MapView, { className: "style" } );
```
FLGMwt, line 45: > Third-party components lost styling → cssInterop() registrations for SafeAreaView, LinearGradient, FasterImageView, BottomSheetTextInput, and MapView

src/components/SharedComponents/Typography/Body3.tsx
```diff
@@ -1,12 +1,13 @@
...
+import classnames from "classnames";
 import React from "react";
 import type { TextProps } from "react-native";
 
 import InatText from "./InatText";
 
 const Body3 = ( props: TextProps ) => (
   // eslint-disable-next-line react/jsx-props-no-spreading
-  <InatText className={`text-xs ${tailwindFontMedium}`} {...props} />
+  <InatText {...props} className={classnames( "text-xs", tailwindFontMedium, props.className )} />
```
FLGMwt, line 10: See `Context for twMerge / classnames` in PR description for further explanation

src/components/App.tsx
```diff
@@ -1,5 +1,7 @@
 // @flow
 
+import "../../global.css";
```
FLGMwt, line 1: https://www.nativewind.dev/docs/getting-started/installation#5-import-your-css-file

src/components/styledComponents.ts
```diff
@@ -19,29 +19,32 @@ import {
 import UnstyledLinearGradient from "react-native-linear-gradient";
 import { SafeAreaView as UnstyledSafeAreaView } from "react-native-safe-area-context";
 
-const View = styled( UnstyledView );
-const KeyboardAvoidingView = styled( UnstyledKeyboardAvoidingView );
+// Core react-native components are registered with nativewind automatically;
+// third-party components need explicit cssInterop registration for className
+// to have any effect
```
FLGMwt, line 31: FYI on styled vs cssInterop

tests/nativewind.setup.js
```diff
@@ -0,0 +1,28 @@
+// Makes className resolve to styles in jest the way it does in the app.
```
jtklein, line 1: I am not so happy about this file and the other jest setups. This seems like a lot of extra setup required just so that we can work with this one package. I couldn't find any reference to jest in the nativewind docs. So, I am wondering how we get to this file content?
Makes me a bit nervous to not find official guidelines on how to setup nativewind with jest.

src/components/SharedComponents/Map/Map.tsx
jtklein: We could add it to StyledComponent, seems to me the only cssInterop call outside of the pattern.

jtklein (changes requested):
Update related code changes and migrations seem all fine to me. But I do see quite a few regressions in the UI, so I'll post before and after screenshots in Slack for things to fix.

jtklein (changes requested):
Still seeing wrong styles in the buttons inStandardCamera and SoundRecorder:

Image

Image

Image

tests/nativewind.setup.js
FLGMwt: Sorry for the delay in getting back to this.

The change is necessary because the package fundamentally changed how it works.

Before: the babel transform wrapped the classname'd components with nativewind wrappers that did tailwind => js styles interpolation within the implementation of the wrapper component
Now: it does more of a tailwind => css => rn-css-interop which is now _baked into metro_. Since jest doesn't use metro, this nativewind setup recreates that relationship. I couldn't get this to work with the exported `nativewind/test` helpers because they _don't_ use react-testing-library's render. So I deferred to Claude on this what reverse-engineered the same approach from the internals. I included some of the rationale of this in the new agent doc. Let me know if you think more of this belongs in the file itself.

I agree it's not ideal but I think it's fair that we sometimes have extra setup for key packages (such as Realm).

FLGMwt:
@jtklein latest commit addresses the "pill" button issues in media.

On testing, the `?` issue is present in TestFlight already so I don't think this PR affects that. I'll test in prod soon.

Screenshot 2026-08-11 at 10 52 56 AM

jtklein:
> @jtklein latest commit addresses the "pill" button issues in media.
> 
> On testing, the `?` issue is present in TestFlight already so I don't think this PR affects that. I'll test in prod soon.

I don't see an issue with the ?

FLGMwt:
> > @jtklein latest commit addresses the "pill" button issues in media.
> > On testing, the `?` issue is present in TestFlight already so I don't think this PR affects that. I'll test in prod soon.
> 
> I don't see an issue with the ?

Oh, I saw the `?` in the context of the other UI issues that looked wonky and confused myself. I do not see an issue with `?` in comparison with designs / TestFlight / Prod.

jtklein:
> @jtklein latest commit addresses the "pill" button issues in media.
> 
> On testing, the `?` issue is present in TestFlight already so I don't think this PR affects that. I'll test in prod soon.
> 
> Screenshot

Screenshot looking good. Do you want me to test again?

jtklein (approved):
Haven't tested the last screenshot again, but did not find anything else in the prior testing round, so if that is fixed I'd say it's good to go, and we fix anything else forward.

PR #3833, opened 2026-07-14
Mob 1565 pagination for server ordered observations

abbeycampbell:
closes [MOB-1565](https://linear.app/inaturalist/issue/MOB-1565/pagination-for-server-ordered-observations)

Adds server-ordered pagination so we can finish wiring this to the sort UI.

### DIFF
--- a/src/components/MyObservations/MyObsServerOrderedDebugSheet.tsx
+++ b/src/components/MyObservations/MyObsServerOrderedDebugSheet.tsx
@@ -101,8 +101,10 @@ const DebugSheetContent = ( { onClose }: DebugSheetContentProps ) => {
     observationIds,
     isServerAuthoritative,
     isLoading,
+    isFetchingNextPage,
     error,
     refetch,
+    fetchNextPage,
   } = useMyObservationsQuery( );
 
   return (
@@ -147,6 +149,14 @@ const DebugSheetContent = ( { onClose }: DebugSheetContentProps ) => {
         {observationIds.map( ( { uuid }, index ) => (
           <ObservationRow key={uuid} uuid={uuid} index={index} />
         ) )}
+        {isServerAuthoritative && (
+          <DebugButton
+            label={isFetchingNextPage
+              ? "Loading more…"
+              : "Load more"}
+            onPress={( ) => !isFetchingNextPage && fetchNextPage( )}
+          />
+        )}
       </ScrollView>
     </View>
   );
--- a/src/components/MyObservations/hooks/useMyObservationsQuery.ts
+++ b/src/components/MyObservations/hooks/useMyObservationsQuery.ts
@@ -9,13 +9,16 @@ import useServerOrderedObservations from "./useServerOrderedObservations";
 const { useQuery } = RealmContext;
 
 const NOOP_REFETCH = ( ) => undefined;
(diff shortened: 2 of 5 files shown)


src/components/MyObservations/hooks/useMyObservationsQuery.ts
```diff
@@ -75,14 +80,20 @@ const useMyObservationsQuery = ( ): UseMyObservationsQueryResult => {
...
+      ? false
+      : isFetchingNextPage,
     error: isDefaultSort
       ? null
       : error,
-    // since we never fetched for default sort, we don't need to refetch.
+    // since we never fetched for default sort, we don't need to refetch or paginate.
     // pagination is still handled by useInfiniteObservationsScroll
     refetch: isDefaultSort
       ? NOOP_REFETCH
```
FLGMwt, line 92: I think this is appropriate for the scope of the new change but:

do you think that the default search would benefit from aligning with this model? It would be cool to remove `useInfObsScroll` or at least consolidate it for this concern.

src/components/MyObservations/MyObsServerOrderedDebugSheet.tsx
```diff
@@ -147,6 +149,14 @@ const DebugSheetContent = ( { onClose }: DebugSheetContentProps ) => {
         {observationIds.map( ( { uuid }, index ) => (
           <ObservationRow key={uuid} uuid={uuid} index={index} />
         ) )}
+        {isServerAuthoritative && (
+          <DebugButton
+            label={isFetchingNextPage
+              ? "Loading more…"
+              : "Load more"}
+            onPress={( ) => !isFetchingNextPage && fetchNextPage( )}
```
FLGMwt, line 157: Would prefer this to be `disabled={isFetchingNextPage}`

src/components/MyObservations/hooks/useMyObservationsQuery.ts
abbeycampbell: I agree it would be nice to not have two different pagination approaches. I think we could look into it, but it's kinda tricky because default sort has different concerns, being a live Realm query. Since Realm can be written to by other processes running in the background, things might get out of sync, but maybe since this is MyObs specific it would be okay? I can make a ticket to look into it. My goal with this for now is to have these changes be as additive as possible, and try not to change how the current flows are working at the same time

src/components/MyObservations/MyObsServerOrderedDebugSheet.tsx
abbeycampbell: heard and very valid feedback, but this is just the debug sheet and it's getting deleted in my next PR anyways. no user will ever interact with it! and we'll be auto-refetching on scroll for the actual list, no button

FLGMwt (approved):
LGTM, feedback addressed/deferred

src/components/MyObservations/hooks/useMyObservationsQuery.ts
FLGMwt: I think I need to wrap my head around our new approach to the non-default infObsScroll. I'll defer this concern to the wire-up which i think is soon to follow this < 3

PR #3835, opened 2026-07-14
add user id to list fields

abbeycampbell:
closes [MOB-1559](https://linear.app/inaturalist/issue/MOB-1559/tapping-into-my-own-observations-sometimes-shows-the-kebab-menu)

We just didn't have the user id available in the local stub version of an obs, and the first time we visit, we default to this local version while we get the remote one. So this check returns false:

```
const belongsToCurrentUser = (
  observation?.user?.id === currentUser?.id
  || ( !observation?.user && !observation?.id )
);
```
until the next time we visit (and `useRemoteObservation` has had a chance to upsert missing data to realm).

### DIFF
--- a/src/realmModels/Observation.js
+++ b/src/realmModels/Observation.js
@@ -111,6 +111,9 @@ class Observation extends Realm.Object {
       rank_level: true,
     },
     time_observed_at: true,
+    user: {
+      id: true,
+    },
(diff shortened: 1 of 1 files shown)


jtklein (approved):
With this fix I can not replicate the error.
So, LGTM

PR #3836, opened 2026-07-15
Traditional projects: validation module

jtklein:
Closes MOB-1499

Adds client-side validation for traditional project observation fields (POFs), following the logic from the classic apps.

`validateProjectFieldsForObservation` validates all POFs of the given projects against the observation's OFVs, returning which project/OF has an error. OFVs are global per observation, so two projects sharing a field read the same OFV, and each failing project reports its own error. I have also added a more lightweight direct validation of a single value which I thought I need while the user is typing in an input field. I did not need it yet but I am also not quite sure yet if I will later, so I'll leave it as a separate fct: `validateProjectFieldValue`.

Wired into the Add to Projects screen: the per-project "all required fields filled" banner and the per-field pass/fail icons are now driven by real validation results (replacing placeholders) and update live as values change.


tests/unit/helpers/validateProjectFieldsForObservation.test.js
```diff
@@ -0,0 +1,533 @@
...
+        projectObservationFields: [{
+          required: true,
+          obsField: {
+            allowedValues: [],
+            id: 10,
+          },
+        }],
+      };
+      const mockObservation = {
+        observationFieldValues: [{ id: 10, value }],
```
sepeterson: should this have `obsFieldId: 10` ?

sepeterson (approved):
Cool! Contrary to my first impression this isn't a huge diff, just a very well-tested one

<details>
<summary>screen</summary>

Image

</details>

tests/unit/helpers/validateProjectFieldsForObservation.test.js
jtklein: Indeed it should. Thanks, great catch!

PR #3838, opened 2026-07-15
expand tappable area with padding & negative margin

abbeycampbell:
closes [MOB-1538](https://linear.app/inaturalist/issue/MOB-1538/reset-button-tap-target-in-exploresearchheader-is-too-small)

<details>
<summary>Screenshots</summary>
Before:
IMG_4640

After:
IMG_4641

</details>

### DIFF
--- a/src/components/SharedComponents/SearchHeader.tsx
+++ b/src/components/SharedComponents/SearchHeader.tsx
@@ -37,12 +37,13 @@ const SearchHeader = ( {
       <Heading4 className="flex-1 wrap text-center">{headerText}</Heading4>
       <Pressable
         className={classnames(
-          "items-end py-5 px-5 -my-5 -mr-5",
+          "items-end",
           { "opacity-50": resetDisabled },
         )}
         onPress={onReset}
         disabled={resetDisabled}
         accessibilityRole="button"
+        hitSlop={20}
         accessibilityLabel={t( "Reset-verb" )}
         testID={`${testID}.reset`}
       >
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/SearchHeader.tsx
```diff
@@ -37,7 +37,7 @@ const SearchHeader = ( {
       <Heading4 className="flex-1 wrap text-center">{headerText}</Heading4>
       <Pressable
         className={classnames(
-          "w-[50px] items-end",
```
jtklein, line 40: I think you'll need to keep a max width as well? I am seeing the button not wrapping anymore for long words.

Image

Unless you want to roll this ticket into this PR. But I don't think it's actionable yet: https://linear.app/inaturalist/issue/MOB-1536/increase-reset-button-width-to-avoid-header-collision

jtklein (changes requested):
Merging like so will change existing wrapping behaviour in languages with longer words here like German.

src/components/SharedComponents/SearchHeader.tsx
abbeycampbell: oh great point about long translations. This PR doesn't actually change anything about wrapping behavior though; turns out I inadvertently killed that in this commit when I wrapped Body3 in a Pressable: https://github.com/inaturalist/iNaturalistReactNative/pull/3771/changes/ad725ea1991f7ac958573fd187e004269176398f

I am happy to call 1536 done as well, except that I'm not totally clear on Abhas's ask for that ticket. I can work that out with him.

For this issue though, long translations are indeed pushing the header text to the left. I switched to using a hitslop prop instead of my negative margins approach, and now this happens to a much lesser degree, since hitslop extends the pressable area but won't interfere with the layout of other stuff (however if the translation is very long, it will eventually cause the header to wrap).
IMG_4644
IMG_4645

jtklein (approved):
Ah,  I see. So, I think with the header wrapping then both tickets are Done here.

PR #3843, opened 2026-07-16
View a user's journal posts from their profile

jtklein:
Closes MOB-1588

### DIFF
--- a/src/api/posts.ts
+++ b/src/api/posts.ts
@@ -10,7 +10,9 @@ const fetchBlogPosts = async <T = ApiDefaultResult>(
   opts: ApiOpts = {},
 ): Promise<ApiResponse<T> | null | ErrorWithResponse | INatApiError> => {
   try {
-    return await inatjs.posts.for_user( params, opts );
+    const response = await inatjs.posts.for_user( params, opts );
+    if ( !response ) { return null; }
+    return response;
   } catch ( e ) {
     return handleError(
       e as ErrorWithResponse,
@@ -28,7 +30,9 @@ const fetchProjectPosts = async <T = ApiDefaultResult>(
   opts: ApiOpts = {},
 ): Promise<ApiResponse<T> | null | ErrorWithResponse | INatApiError> => {
   try {
-    return await inatjs.projects.posts( params, opts );
+    const response = await inatjs.projects.posts( params, opts );
+    if ( !response ) { return null; }
+    return response;
   } catch ( e ) {
     return handleError(
       e as ErrorWithResponse,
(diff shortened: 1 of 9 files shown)


src/components/Journal/UserPosts.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import PostList from "./PostList";
+
+interface Props {
+  userIcon?: string;
+  userId: number;
+  userLogin?: string;
+}
+
+const UserPosts = ( {
```
jtklein, line 23: Copy of ProjectPosts swapped to use users.{id}.posts

src/api/posts.ts
```diff
@@ -1,27 +1,38 @@
...
+import type {
+  ApiDefaultResult, ApiOpts, ApiParams, ApiResponse,
+} from "api/types";
 import inatjs from "inaturalistjs";
 
-const fetchUserPosts = async (
-  params: Record<string, unknown> = {},
-  opts: Record<string, unknown> = {},
-): Promise<Record<string, unknown> | null | ErrorWithResponse | INatApiError> => {
+const fetchBlogPosts = async <T = ApiDefaultResult>(
```
jtklein, line 8: Renamed the fct to avoid confusion with the new one below.
This call here is "fetch Posts for the current user"
Below new call is "fetch posts from a user"

I think fetchBlogPosts is breaking the current convention of describing what we are fetching slightly by naming it what we are fetching for but I find it understandable and different enough from below.

src/api/posts.ts
```diff
@@ -1,27 +1,38 @@
 import type { ErrorWithResponse, INatApiError } from "api/error";
 import handleError from "api/error";
+import type {
+  ApiDefaultResult, ApiOpts, ApiParams, ApiResponse,
```
jtklein: Following the pattern established in usersTyped

src/navigation/types.ts
```diff
@@ -327,8 +327,9 @@ export type BaseTabStackParamList = {
   };
   // From UserProfile
   // {
-  //   userLogin: user?.login,
-  //   journalPostsCount: user?.journal_posts_count,
+  //   userId: user?.id,
+  //   userIcon: user?.login,
+  //   userLogin: user?.icon_url,
```
abbeycampbell: these are swapped (totally inconsequential unless someone is mindlessly copy/pasting down the line or something)

src/api/posts.ts
```diff
@@ -30,7 +41,28 @@ const fetchProjectPosts = async (
   }
 };
 
+interface UserPostsParams extends ApiParams {
+  id: number;
+}
```
abbeycampbell: since this is exactly the same as `ProjectPostsParams`, could we make one generic interface that handles both?

abbeycampbell (approved):
tested on iOS and android; looks great and meets the ticket criteria. I have a nitpick and a suggestion which you are free to ignore

src/navigation/types.ts
jtklein: Thanks! Done that.

src/api/posts.ts
jtklein: Thanks! Done that.

PR #3845, opened 2026-07-16
MOB-1341: Explore V2 locations permissions simplification

sepeterson:
- Move all permissions handling to ExploreResults
- For nearby mode, keep the fetched location just in ExploreResults


abbeycampbell (approved):
nice!

PR #3847, opened 2026-07-17
On Android use only pinch-to-zoom, no pan-to-zoom

jtklein:
Closes MOB-1580


sepeterson (approved):
Tested on device and it's definitely an improvement

PR #3848, opened 2026-07-17
exploreV2 prototype cleanup

sepeterson:
Clean up orphaned files from @jtklein's prototype

Intentionally leaving FilterModalV2 so I can reuse it for AdvancedSearch


abbeycampbell (approved):
love a diff like this one

Image

PR #3849, opened 2026-07-17
MOB-1557: pojo from useCurrentUser

sepeterson:
get a plain object from useCurrentUser

There was also one spot where the live Realm object was used so I switched it to User.currentUser


abbeycampbell (approved):
lgtm! very good stuff

In the ticket comments Johannes also mentioned removing "the react-native patch that prevents us from seeing the "Accessing object that has been invalidated" error" and to be honest I'm not totally sure where he means. maybe if we do want to do that it could be a separate follow-up?

sepeterson:
> In the ticket comments Johannes also mentioned removing "the react-native patch that prevents us from seeing the "Accessing object that has been invalidated" error" and to be honest I'm not totally sure where he means. maybe if we do want to do that it could be a separate follow-up?

Ya good catch I had forgotten about that! I did test it without the patch and didn't see any issues but as I'm not 100% sure on the surface area, I spun off a dedicated ticket: https://linear.app/inaturalist/issue/MOB-1601/remove-the-react-native08310-patch-suppressing-accessing-object-which

PR #3850, opened 2026-07-17
run npm run icons and clean up share extension

sepeterson:
`npm run icons` is also adding the icon font to the share extension so I had to manually rm those changes.

I tried to update the script to not do that but couldn't get it working within a reasonable amount of time


abbeycampbell (approved):
tested and iOS and android, things looked right to me

PR #3852, opened 2026-07-17
Mob 1479 wire up observation sort in myobs

abbeycampbell:
closes [MOB-1479](https://linear.app/inaturalist/issue/MOB-1479/wire-up-observation-sort-in-myobs)

Finishes sort for new MyObs, still behind feature flag (apart from sort persistence, which will be handled in a follow-up)


sepeterson (approved):
Unable to test yet but code looks great!

PR #3856, opened 2026-07-21
MOB-1346: ExploreV2 advanced search state

sepeterson:
- State and query param construction prep for Advanced search
- Clean up debug sheet, prototype filters modal


src/i18n/strings.ftl
```diff
@@ -81,6 +81,8 @@ Adds-your-vote-of-disagreement = Adds your vote of disagreement
 Advanced-Mode = Advanced Mode
 # Link to the advanced search screen with more search filters
 Advanced-Search = Advanced Search
+# Header title of the Advanced Search screen
+ADVANCED-SEARCH = ADVANCED SEARCH
```
sepeterson, line 85: not used in this PR but will be in the very next one. The commit hook added the npm translate output and I figured it was simple to include than try to mess with untangling that

src/components/Explore/ExploreV2/helpers/buildQueryParams.ts
```diff
@@ -77,6 +82,15 @@ const buildExploreV2QueryParams = (
     }
   }
 
+  // Advanced Search filters
+  Object.assign( params, filtersToApiParams( state.filters, viewerId ) );
```
FLGMwt: I think it's confusing to mix spreading and `.assign` when we're building an eventual `params` result. Can we switch this to spreading with an appropriately named new variable?

We only have one other use of `Object.assign` and that's on a [React prop value](https://github.com/inaturalist/iNaturalistReactNative/blob/4b8561f45bfc63d00b59854b55b9d60f452fa3d0/src/components/ObsDetails/MasonryLayout.js#L28) which we _definitely_ shouldn't be mutating. I'll write up a debt ticket to fix that.

sepeterson: Thanks! Changed it

FLGMwt (approved):
👍 thanks!

PR #3857, opened 2026-07-21
jwt retry backoff

FLGMwt:
Closes MOB-1623 (retroactively created)

we have a key flaw in our authservice code where we hold a shared _jwt_ promise BUTTTTTT we don't care if it's failed or not. We'll wait until the shared promise is resolved but we're happy to retry it again if it immediately fails and fails and fails.

This sometimes results in a broken app state which is only resolved by quitting and restarting, pointing to the issue being instance-based.

This PR introduces module-level state for tracking JWT retries and implements exponential (0s, 2s, 4s, 8s ....) backoff for retries, replacing our sometimes 20-times-a-second failed JWT retries.

### DIFF
--- a/src/components/LoginSignUp/AuthenticationService.ts
+++ b/src/components/LoginSignUp/AuthenticationService.ts
@@ -66,12 +66,23 @@ const authCache: AuthCache = {
   cacheTimeout: 5000,
 };
 
+// module-level tracking of jwt failures. This works around a
+// fault where different parties were competing for retries
+// during downtime or cascading auth failure, sometimes requring
+// user-initiated app restart to heal.
+let jwtRefreshFailureCount = 0;
+let jwtRefreshFailedAt: number | null = null;
+const JWT_REFRESH_BACKOFF_BASE_MS = 5_000;
+const JWT_REFRESH_BACKOFF_MAX_MS = 60_000;
+
 /**
- * Clear cache for isLoggedIn.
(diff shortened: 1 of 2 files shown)


src/components/LoginSignUp/AuthenticationService.ts
```diff
@@ -415,6 +451,10 @@ const getJWT = async (
     // If a refresh is already in-flight, return that shared promise instead
     // of starting a second concurrent request.
     if ( !jwtRefreshPromise ) {
+      // skip the fetch (and return null) if we're still waiting on backoff
+      if ( jwtRefreshBackoffRemainingMs() > 0 ) {
+        return null;
```
sepeterson: I'm wondering if we could return `jwtToken ?? null` here. 

Case I'm thinking of:
1. user's JWT just turned 26 minutes old, they have a spotty connection
2. next JWT fetch fails, maybe a few times so the backoff is up to 60 seconds
3. User momentarily gets a decent connection, but they're in the middle of the backoff window so they're passing `null` for their auth token

Essentially don't throw a potentially valid JWT away just because we failed to fetch a new one

sepeterson: Also if the comment at the top of this file is to be believed 

> Actually they expire in 24 hours, but ideally they would expire every 30 mins, so might as well be futureproof.

sepeterson (approved):
hell yeah

PR #3858, opened 2026-07-21
MOB-1340: universal search result info (i) button

sepeterson:
Nav to taxon/user/project details


src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
```diff
@@ -34,6 +36,28 @@ const resultLabel = ( result: UniversalSearchResultItem ): string => {
...
+          navToDetail: ( ) => navigation.navigate( "ProjectDetails", { id: result.project.id } ),
+        };
+      default:
+        return {
+          hint: t( "Navigates-to-taxon-details" ),
+          navToDetail: ( ) => navigation.navigate( "TaxonDetails", { id: result.taxon.id } ),
+        };
+    }
+  };
+  const { hint: infoHint, navToDetail } = infoButtonProps( );
```
FLGMwt: I think it would be cool to extract this to a file-level helper rather than an effective iife / single use function. I definitely like the convenience of a single return rather than a bunch of mutated `let`s, but if the param list isn't monstruous, can we pass those deps? `resultId()` and `resultLabel()` are already demonstrating this.

FLGMwt (approved):

src/components/Explore/ExploreV2/components/UniversalSearchResult.tsx
sepeterson: Works for me! I'm actually going to turn it into *two* helpers :)

PR #3861, opened 2026-07-23
MOB-1334: ExploreV2 species sort

sepeterson:
The ticket had more options but we didn't know about API limitations

Just matching MyObs for now


abbeycampbell (approved):
🤙

PR #3862, opened 2026-07-23
MOB-1554: explorev2 support unknown in universal search

sepeterson:
the `indentified: false` bit is new to the app but it matches web


abbeycampbell (approved):
lgtm

abbeycampbell:
I am curious though, why totally delete the default search options test?

sepeterson:
> I am curious though, why totally delete the default search options test?

I thought it was getting out of scope with the only changes in that file being to add an extra case to the IconicTaxonChooser callback. There's a ticket further down the road to audit ExploreV2 tests and I wanna add a more holistic test for DefaultSearchOptions then https://linear.app/inaturalist/issue/MOB-1525/audit-screen-level-unit-tests-add-integration-tests

PR #3867, opened 2026-07-23
MOB-1602 prevent pivot card from crashing first launch

FLGMwt:
This fixes the local source of a reliable crash on _initial_ login in default mode when a user has at least one remote obs. 

image

FYI @abbeycampbell I think I tracked it down to this change: https://github.com/inaturalist/iNaturalistReactNative/commit/a38414a0e569b84105c6e2479bc0c4d51cc54211#diff-07ff17bf25fd2c24f5bfad7f70d54783e036fb7df3268eb9e167d8d8dcfee224L446. I think with that change, we have now have a render pass of MyObsSimple where obsIds doesn't yet have the upserted API results. 

This eventually settles and with a small check to guard the property access, everything seems fine. Just wanted to bring to your attention for if you think we need to look further into this (and keep the loading state while obsIds.length hasn't reflected API results)

### DIFF
--- a/src/components/MyObservations/MyObservationsSimple.tsx
+++ b/src/components/MyObservations/MyObservationsSimple.tsx
@@ -519,11 +519,13 @@ const MyObservationsSimple = ( {
             imageComponentOptions={{
               onImageComponentPress: handlePivotCardGridItemPress,
               accessibilityHint: t( "Navigates-to-observation-details" ),
-              imageComponent: (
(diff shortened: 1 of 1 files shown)


sepeterson (approved):
Nice! Ran through repro steps and didn't see a crash.

abbeycampbell (approved):
my bad! thanks for catching this

PR #3868, opened 2026-07-23
fix observation owner display for "Others" notifications.

FLGMwt:
|Before Notification shows "Unknown" for all obs in "Others" notifications|After shows owning user login when available|
|---|---|
|image|image|

https://github.com/inaturalist/iNaturalistReactNative/pull/3835 added a `user` fields to the obs API request where we weren't asking for it before. This would have been fine except that when the Notifications tab grabs obs, it overrides the full `user` fields object to request id & `login` (the username). Since were spreading the default fields _after_ this full object, it got clobbered.

Fix: Now that the default obs fields have a `user` fields object, spread over the defaults instead of fully clobbering them

### DIFF
--- a/src/sharedHooks/useInfiniteNotificationsScroll.ts
+++ b/src/sharedHooks/useInfiniteNotificationsScroll.ts
@@ -73,11 +73,11 @@ async function fetchObsByUUIDs(
     uuids,
     {
       fields: {
+        ...Observation.ADVANCED_MODE_LIST_FIELDS,
         user: {
-          id: true,
+          ...Observation.ADVANCED_MODE_LIST_FIELDS.user,
           login: true,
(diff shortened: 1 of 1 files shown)


abbeycampbell (approved):
whoops, I'll try to remember to check for instances of this next time I add any fields to a request

FLGMwt:
> whoops, I'll try to remember to check for instances of this next time I add any fields to a request

this sort of thing _should_ have been innocuous but the downstream notifs code misbehaved. I only noticed this after switching back to main from the branch but TS actually calls this out! The following would have shown up after your change as a new error. But! we don't have repo-level ts checking and we're progressively permitting errors so there's no way for us to catch these:
image

PR #3869, opened 2026-07-23
MOB-1543: allow rest of upload queue to complete if one fails

sepeterson:
Sets up separate AbortControllers per observation as well as a higher-level instance to cover the whole queue.

Also include new integration tests to cover the bug. These are kind of intensive but I think provide good cover

<details>
<summary>video showing queue continuing after an error</summary>

https://github.com/user-attachments/assets/5745eb41-2b80-4c5b-8607-286809037d6a

</details>

### DIFF
--- a/src/components/MyObservations/hooks/useUploadObservations.ts
+++ b/src/components/MyObservations/hooks/useUploadObservations.ts
@@ -23,7 +23,7 @@ import {
 } from "uploaders/utils/progressTracker";
 
 export const MS_BEFORE_TOOLBAR_RESET = 5_000;
-const MS_BEFORE_UPLOAD_TIMES_OUT = 60_000 * 5;
+export const MS_BEFORE_UPLOAD_TIMES_OUT = 60_000 * 5;
 
 const { useRealm } = RealmContext;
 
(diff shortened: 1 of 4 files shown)


abbeycampbell (approved):
This change looks good to me. I do kinda wish I had looked at this earlier and we couldve gotten this merged before tuesday, for two full weeks of testing. But maybe it's enough to call out to QA'ers that we specifically need testing around batch uploads. fwiw I tested a bunch by shortening the timeout and having claude mock-up a failure.
I do think this fix (separate abort controllers for each obs) sets us up to potentially do concurrent uploads in the future? Which I think would be nice

src/components/MyObservations/hooks/useUploadObservations.ts
```diff
@@ -98,16 +98,29 @@ export default ( canUpload: boolean ) => {
...
+    const abortCurrentUpload = ( ) => obsAbortController.abort( );
+    let timeoutID: ReturnType<typeof setTimeout> | undefined;
     try {
-      const timeoutID = setTimeout( ( ) => {
-        abortController.abort( );
-      }, MS_BEFORE_UPLOAD_TIMES_OUT );
-      await uploadObservation( observation, realm, { signal: abortController.signal } );
-      clearTimeout( timeoutID );
+      abortController?.signal.addEventListener( "abort", abortCurrentUpload );
+      if ( abortController?.signal.aborted ) {
```
jtklein: I am struggling a bit with why this if clause is necessary. Wouldn't we just want to return early if we have an all abort signal? But then also we do have the same check before calling this function here: https://github.com/inaturalist/iNaturalistReactNative/blob/8064181dcdd2bc039eba5ee45dd71011a1828fc7/src/components/MyObservations/hooks/useUploadObservations.ts#L190-L191
Granted, with a realm lookup in between check and this line, so maybe nit immediately before here.

jtklein (approved):
Haven't looked at the tests too closely, but code changes seem fine to me.

src/components/MyObservations/hooks/useUploadObservations.ts
sepeterson: I think you're right & this is unreachable as everything between this and the L190 check are synchronous. I'll rm it

sepeterson:
> This change looks good to me. I do kinda wish I had looked at this earlier and we couldve gotten this merged before tuesday, for two full weeks of testing. But maybe it's enough to call out to QA'ers that we specifically need testing around batch uploads. fwiw I tested a bunch by shortening the timeout and having claude mock-up a failure. I do think this fix (separate abort controllers for each obs) sets us up to potentially do concurrent uploads in the future? Which I think would be nice

Thanks for testing! I will definitely make sure we surface observation uploads as an important thing to test w/ next week's build

PR #3804, opened 2026-07-06
MOB-1339 ExploreV2 Default search subject options

sepeterson:
Adds default search subject options for ExploreV2 including

- iconic taxa
- your own user
- stub of species I haven't observed

PR #3872, opened 2026-07-24
Mob 1482 add a map view to myobs

abbeycampbell:
closes [MOB-1482](https://linear.app/inaturalist/issue/MOB-1482/add-a-map-view-to-myobs)

### DIFF
--- a/src/components/Explore/ExploreV2/screens/ExploreResults.tsx
+++ b/src/components/Explore/ExploreV2/screens/ExploreResults.tsx
@@ -180,9 +180,9 @@ const ExploreResults = ( ) => {
                       testID="ExploreV2ObservationsList"
                     />
                     <ObservationsViewBar
-                      hideMap
                       layout={layout}
                       updateObservationsView={writeLayoutToStorage}
+                      viewOptions={["grid", "list"]}
                     />
                   </>
                 )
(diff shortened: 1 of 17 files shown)


src/components/Explore/ExploreV2/screens/ExploreResults.tsx
```diff
@@ -180,9 +180,9 @@ const ExploreResults = ( ) => {
                       testID="ExploreV2ObservationsList"
                     />
                     <ObservationsViewBar
-                      hideMap
                       layout={layout}
                       updateObservationsView={writeLayoutToStorage}
+                      viewOptions={["grid", "list"]}
```
sepeterson, line 185: nice

tests/unit/components/Explore/ObservationsViewBar.test.js
```diff
@@ -0,0 +1,46 @@
+import { fireEvent, render, screen } from "@testing-library/react-native";
```
sepeterson: Nit but going forward I think we want to prefer userEvent over fireEvent when possible

https://github.com/inaturalist/iNaturalistReactNative/pull/3804#discussion_r3543730821

sepeterson (approved):
Tested working on ios. The search -> map repositioning interaction is so slick

abbeycampbell:
> Tested working on ios. The search -> map repositioning interaction is so slick

thanks! not totally sure if we want it to fully zoom out again if you reset the search while zoomed in, but my assumption for now is that the zoom level should reflect whatever the results currently are

PR #3874, opened 2026-07-24
Mob 1331 offline state

sepeterson:
A bit cleaner in SemanticDiff

Adds an offline state for ExploreResults


jtklein (approved):
This works great on my end.
Just one note: the offline notice is also shown instead of the  "Needs permission" state which I think is in accordance to the designs but I am not entirely sure from the Figma file alone.

sepeterson:
> the offline notice is also shown instead of the "Needs permission" state which I think is in accordance to the designs but I am not entirely sure from the Figma file alone.

Yes I think that's the better way considering the alternate user experience of Request Permissions -> Grant Permissions -> Show Offline Notice

That said, I'll put it on my list to confirm w/ Abhas at our next meeting!

PR #3876, opened 2026-07-25
Mob 1596 deleting a photosound and pressing back on a savedunsynced

abbeycampbell:
closes [MOB-1596](https://linear.app/inaturalist/issue/MOB-1596/deleting-a-photosound-and-pressing-back-on-a-savedunsynced-observation)

We were deleting files from disk, but the Realm observations still had references to them. Saving re-writes rewrites the embedded `observationPhotos/observationSounds` lists  so this issue is only caused by tapping back with unsaved changes. (There are some separate  but related ux issues around navving back vs saving, discarding changes, and what changes are even discardable that I'm gonna open a separate ticket for, but that will need product input).

### DIFF
--- a/src/components/ObsEdit/EvidenceList.js
+++ b/src/components/ObsEdit/EvidenceList.js
@@ -187,7 +187,7 @@ const EvidenceList = ( {
       await ObservationSound.deleteLocalObservationSound(
         realm,
         uriToDelete,
-        currentObservation.uuid,
+        currentObservation?.uuid,
       );
       afterMediaDeleted( uriToDelete );
     }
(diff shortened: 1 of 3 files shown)


src/components/ObsEdit/EvidenceList.js
```diff
@@ -187,7 +187,7 @@ const EvidenceList = ( {
       await ObservationSound.deleteLocalObservationSound(
         realm,
         uriToDelete,
-        currentObservation.uuid,
+        currentObservation?.uuid,
```
sepeterson: Is this new optional chain necessary? It seems like deleteLocalObservationSound shouldn't accept undefined here

sepeterson (approved):
No blockers! Tested on ios and fixes the issue!

src/components/ObsEdit/EvidenceList.js
abbeycampbell: good point, I think you're right. This was just a quick fix for that eslint warning, but I'm just gonna revert and leave it alone to keep this PR scoped to bugfix only

jtklein (approved):

PR #3878, opened 2026-07-27
TypeScript migration TaxonMedia and children

jtklein:


src/components/TaxonDetails/TaxonMedia.tsx
```diff
@@ -1,53 +1,52 @@
...
+  onChangeIndex: ( newIndex: number ) => void;
+  photos: PhotoItem[];
+  tablet: boolean;
 }
 
 const TaxonMedia = ( {
   loading,
   onChangeIndex,
   photos = [],
-  sounds = [],
```
jtklein, line 37: No HOC actually passed sounds in.

src/components/ObsDetails/MasonryLayout.tsx
```diff
@@ -5,35 +5,79 @@ import React, { useEffect, useState } from "react";
 import { Image } from "react-native";
 import Photo from "realmModels/Photo";
 
+interface PhotoItem {
+  id: number;
+  url: string;
+  localFilePath: string;
+  attribution: string;
+  hidden: boolean;
```
websoty, line 13: The previous Flow type allowed id, localFilePath, and attribution to be missing, and it didn’t require hidden. Are these fields guaranteed for every taxon photo now? Otherwise, should they remain optional to avoid making the TypeScript type stricter than the actual data?

src/components/ObsDetails/MasonryLayout.tsx
```diff
@@ -5,35 +5,79 @@ import React, { useEffect, useState } from "react";
...
+
 const numColumns = 2;
 const spacing = 6;
 
-const photoUrl = photo => Photo.displayLocalOrRemoteLargePhoto( photo );
+const photoUrl = ( photo: PhotoItem ): string => (
+  Photo.displayLocalOrRemoteLargePhoto( photo )
+);
+
+const isSound = ( item: MediaItem ): item is SoundItem => !!item.file_url;
```
sepeterson:
```suggestion
const isSound = ( item: MediaItem ): item is SoundItem => "file_url" in item;
```

sick type narrowing! Could we also do it like this? fixes a local ts error for me

src/components/TaxonDetails/TaxonMedia.tsx
sepeterson: Should be able to drop this guard if `onChangeIndex` is no longer optional

sepeterson, line 88: Also seeing a ts error here since ActivityIndicator doesn't specify a className prop. Can we drop it or is this needed for nativewind magic?

sepeterson (approved):
ty! no blockers!

websoty (approved):

src/components/TaxonDetails/TaxonMedia.tsx
jtklein: I didn't address this because I think the nativewind v4 update PR in flight will fix this for all instances.

src/components/ObsDetails/MasonryLayout.tsx
jtklein: I don't know if they are guaranteed for every taxon photo that is propped in here because we don't really support tablet layout anymore and haven't tested this component in a while.
However, this was a deliberate choice because whatever we receive here is propped down as is into  PhotoContainer which is typed as to require these keys.

src/components/ObsDetails/MasonryLayout.tsx
jtklein: I have seen that suggested on my end as well, but I didn't want to change syntax too much. I am not really sure what the benefit is, it seems to me to be only in TS.

PR #3880, opened 2026-07-27
add offline handling for myobs map view

abbeycampbell:
closes [MOB-1609](https://linear.app/inaturalist/issue/MOB-1609/offline-experience-for-map-view)

### DIFF
--- a/src/components/MyObservations/MyObservationsSimple.tsx
+++ b/src/components/MyObservations/MyObservationsSimple.tsx
@@ -1,3 +1,4 @@
+import { refresh } from "@react-native-community/netinfo";
 import { useNavigation, useRoute } from "@react-navigation/native";
 import type { FlashListRef } from "@shopify/flash-list";
 import type { ViewOption } from "components/Explore/ObservationsViewBar";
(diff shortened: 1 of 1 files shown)


src/components/MyObservations/MyObservationsSimple.tsx
```diff
@@ -427,9 +435,7 @@ const MyObservationsSimple = ( {
             : (
               <>
                 { layout === "map"
-                  ? (
-                    <MyObservationsMapView userId={currentUser?.id} />
-                  )
+                  ? renderMapView( )
```
FLGMwt, line 438: minor matter of taste, non-blocking: if this new offline notice is exclusive to the map screen, maybe the eager return could live in `MyObservationsMapView` instead?

FLGMwt (approved):
LGTM!

src/components/MyObservations/MyObservationsSimple.tsx
abbeycampbell: We'll want the same thing I think for the grouped by iconic taxa view. So barring new product asks for offline behavior, I think I'll keep this approach for now, but if something changes, I like the idea of moving it!

PR #3881, opened 2026-07-28
MOB-1595: exploreV2 map view

sepeterson:
Map view for explore v2 and a bunch of tests

I'm going to handle the "Redo search in map area" feature in a 2nd PR

<details>
<summary>screenshot</summary>
Simulator Screenshot - iPhone 16e - 2026-07-28 at 22 26 48

</details>


jtklein (approved):
This looks all good to me.

What I have tested and works well:
Not having location permission switching to Worldwide then granting from the current location button.
Searching for a place in universal search does zoom to the place (but it might be faster?; I saw a long loading state for that).
Loading state looks good.
From what I have tested bounds and region zooming looks good and correct.

PR #3885, opened 2026-07-28
MOB-1573 feature flags for testing only

FLGMwt:
closes MOB-1573

separate from the defaults which drive whether a feature is permanently enabled or not, this PR introduces a list for feature flags to be defaulted on _only_ for logged in iNaturalist users with the role `admin`. For safety, it also uses `DeviceInfo.DeviceInfo.getInstallerPackageName` to determine if we're running in a TestFlight build. This prevents use from exposing non-released code to production builds.

This _does_ mean that `useFeatureFlag` now depends on Realm and the React component tree. I played around with syncing the user-informed resolved value to zustand, but I think that that will complicate future remote config work so I settled with resolving it downstream of zustand.


src/components/Developer/Developer.tsx
```diff
@@ -256,6 +259,31 @@ const PathStats = () => {
   );
 };
 
+// Temporary diagnostic use of admin/testflight config.
+const TestFlightAdminFeatureFlagTest = () => {
+  console.log( "first" );
```
abbeycampbell: nit as this will be deleted soon: errant log

src/sharedHooks/useFeatureFlag.ts
```diff
@@ -1,14 +1,31 @@
...
 
+import useCurrentUser from "./useCurrentUser";
+
+const isTestFlightBuild = Platform.select( {
+  ios: DeviceInfo.getInstallerPackageNameSync() === "TestFlight",
+  default: false,
+} );
+
 const useFeatureFlag = ( featureFlagKey: FeatureFlag ) => {
+  const userIsAdmin = useCurrentUser()?.roles.includes( "admin" );
```
abbeycampbell: the current user only gets roles written to it at login, and I don't think we refresh it after that, so an admin who's already signed in when this change becomes available in testflight won't see the enabled feature flags they're supposed to see until they log out and back in. We should either add a roles refresh into the app start or foreground, or just be sure we call it out to the team that they'll need to log out and back in to properly test.

abbeycampbell (changes requested):
Tested locally by adding some feature flags to the `flagsEnabledForAdminsInTestFlight` array and hardcoding the `isTestFlightBuild` to true, and everything works as expected!
One nit, one request, and a couple of extra thoughts:
- I do think it would be great if we could find a way to show the true enabled state on the debug screen, or otherwise have some place where an admin user can have visibility into what is currently 'turned on' for them. However I imagine this whole shebang will be replaced by a remote config at some point, so maybe it doesn't matter.
- Part of me wonders if we should just go ahead and add MyObs Sort (and maybe more) to this, so it will be auto-enabled for admins in next weeks internal build, or if that's jumping the gun and we really need to go through a week of internal testing with this as-is? Either way, question for later and a follow-up change.

src/sharedHooks/useFeatureFlag.ts
FLGMwt: Oh, yes, I meant to call that out, thank you! Yeah, understood that we'd need to have internal staff logout and in if they want the automatically-enabled features. It seemed that a "migration" step would be more trouble than it's worth for the temporary nature of this.

abbeycampbell (approved):

FLGMwt:
> I do think it would be great if we could find a way to show the true enabled state on the debug screen, or otherwise have some place where an admin user can have visibility into what is currently 'turned on' for them

You're definitely right here. Latest commit resolves by giving `useFeatureFlagForDebug` (used by the debug screen) awareness of the testFlightAdmin config. I was overthinking the relationship w/ FeatureFlags & Rozenite. Moving Rozenite under our providers (including Realm, which was the blocker) makes more sense for its functionality anyway.

PR #3887, opened 2026-07-28
MOB-1165 language picker doublescroll

FLGMwt:
closes MOB-1165

First two commits are improvements made during investigation:
- useState() the mapped picker options themselves instead of storing locales in state, also use a stateInitializer
- skip `BottomSheetScrollView` when `!scrollEnabled`

The jumping / scrolling Picker issue didn't happen when I rendered the same picker outside of BottomSheet so tracked it down to that. BottomSheet uses a Portal and a useEffect on the children that causes an extra render. Picker's `selected` value is controlled _but_ native iOS has a "sync" step which isn't tolerant of this extra render. The solution: isolate the controlled part of the Picker to its own component so that the controlled `selection` doesn't cross the boundary between the Portal's parent `render` function and the Portal's children. (This replaces the patch I tried first and now have removed since we can fix it in userland)


sepeterson (approved):
Pulled locally and works great!

Also did a check across a bunch of different BottomSheet usages and everything works as before

PR #3889, opened 2026-07-29
MOB-1466 extract journal post image preview

FLGMwt:
Closes MOB-1466

Follows the pattern of iNat Classic iOS to use the first `img` in the Journal/Blog post as preview image in the post list. I think regex works fine for this, but I'm open to suggestions of if we should use a legit xml / html parsing lib for safer validation. Here's a playground for the pattern w/ an example journal post: https://regexr.com/8no7d (also linked as comment).

It maintains the fallback to the iNat logo image for posts without images.

image

### DIFF
--- a/src/components/Journal/PostListItem.tsx
+++ b/src/components/Journal/PostListItem.tsx
@@ -34,15 +34,16 @@ const PostListItem = ( {
       : item.parent.icon_url ?? null;
   }, [item.body, item.parent.icon_url] );
 
-  if ( !item ) {
-    return null;
-  }
-
   return (
     <View className="flex-row items-center mx-3 my-2">
       {imgSrc && (
         <Image
           source={{ uri: imgSrc }}
+          // Passing a key here is not ideal for FlashList, but we do actually want to avoid
+          // recycling on images. When Image gets a new `source` prop, it continues displaying
+          // the previously-loaded image until the new source is loaded, which leads to "stale"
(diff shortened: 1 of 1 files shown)


sepeterson:
Seeing this odd behavior where the thumbnails from previous rows are being shown instead of the placeholder

<details>
<summary>video</summary>

https://github.com/user-attachments/assets/a9b3df90-aa24-4bff-a50b-e9b30da257d8

</details>

Claude claims it could be corrected by using the <FasterImageView> component used in ObsImage but I haven't tested that

jtklein (approved):
The regex, i.e. seeing the image oft the blog posts, works great.

I am also seeing the previous posts image lingering around.
That is an issue we also had in the observation lists some time ago, and I think related to FlashList's item recycling and resulting limitations on what kind of internal state an item can have. I can't point to commits or issues that fixed the problem on the MyObs screen though now. But we do have the "image loading" overlay on the MyObs list now which might have solved it.
FWIW, I don't think it is a big deal for this feature here. So, any solution to this can also be a later ticket. Possibly to extract a correct behaviour of an Image inside a flash list item to a shared component/hook

sepeterson (approved):

FLGMwt:
Thanks y'all! I fixed this by adding a `key` to the `Image`. It's especially noticeable with these journal posts because we're probably always loading a large image from the post and the Image component doesn't un-render the "old" source while it's loading the new source.

PR #3891, opened 2026-07-29
MOB-1342 universal search green search button

sepeterson:
new search button on universal search

<details>
<summary>screenshot</summary>
Simulator Screenshot - iPhone 17 Pro - 2026-07-29 at 11 17 23

</details>

### DIFF
--- a/src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
+++ b/src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
@@ -334,7 +334,7 @@ const UniversalSearch = ( ) => {
           ListEmptyComponent={listEmptyComponent}
         />
       </View>
-      <ButtonBar>
+      <ButtonBar containerClass="bg-white border-t border-lightGray">
         <Button
           level="focus"
           onPress={handleSearch}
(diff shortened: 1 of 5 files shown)


FLGMwt:
LGTM, but it's odd to me that it's under the keyboard, I would never have known it was there. I suppose I would have expected it to float up along with the keyboard. But maybe it makes more sense w/ the non-keyboard interactions in advanced filters?

As implemented, it matches the Figma perfectly, but Figma omits the spilling Obs button. I think in practice, they're a little cozy together.

|Keyboard Open|Keyboard Closed|
|---|---|
|Screenshot 2026-07-29 at 2 20 41 PM|Screenshot 2026-07-29 at 2 20 43 PM-1|

FLGMwt (approved):
LGTM!

sepeterson:
> I suppose I would have expected it to float up along with the keyboard. But maybe it makes more sense w/ the non-keyboard interactions in advanced filters?

Ya I think it can't float bc we're running out of real estate for the autocomplete results. Then after you select location, keyboard dismisses and you have this CTA nudge at the bottom

PR #3892, opened 2026-07-30
MOB-1621 use screen tracking

FLGMwt:
closes MOB-1621

Previously, firebase didn't like that we were using the reserved namespacing prefix in `firebase_screen` so it discarded those attrs. 

We now explicitly use `@react-native-firebase/analytics`'s `logScreenView` for screen tracking. 

We intentionally shifted away from this to satisfy a deprecation warning here: https://github.com/inaturalist/iNaturalistReactNative/commit/86430591fcc38aad73510fd795975db6490b35a1 However, we now have an owner of the Analytics concern and we would like these to present as best as possible to give an accurate landscape. 

I've created MOB-1630 to deal with the eventual package upgrade and handle the deprecation warnings. Those warnings are ignored in LogBox at least for now.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ import "react-native-get-random-values";
 import "react-native-url-polyfill/auto";
 
 import { BottomSheetModalProvider } from "@gorhom/bottom-sheet";
-import { Alert, AppRegistry } from "react-native";
+import { Alert, AppRegistry, LogBox } from "react-native";
 import { getCurrentRoute } from "navigation/navigationUtils";
 import zustandMMKVBackingStorage from "stores/zustandMMKVBackingStorage";
 import {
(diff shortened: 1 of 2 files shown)


jtklein (approved):
Have tested this with firebase Debug mode enabled and this works for me.
I do see multiple "screen_view" events being logged there, some have the screen name some don't, but at least I do see one log for each transition with the correct screen name.
So, I'd say this a workable state for data analyses, but maybe the ticket is not yet Done (because most events will still have no screen name).

sepeterson (approved):

FLGMwt:
Ty for review y'all! MOB-1633 created for tracking the dupe / missing events. 

I'm pulling this forward to enabled our analytics testing but I agree we should address this problem (that was previously happening but swallowed).

PR #3893, opened 2026-07-30
MOB-1346: explore search screen prep for ExploreV2 advanced search

sepeterson:
- decouple ExploreLocationSearch from ExploreV1
- some ts conversions

### DIFF
--- a/src/components/Explore/SearchScreens/ExploreTaxonSearch.tsx
+++ b/src/components/Explore/SearchScreens/ExploreTaxonSearch.tsx
@@ -9,6 +9,7 @@ import React, {
   useCallback,
   useState,
 } from "react";
+import type { ListRenderItem } from "react-native";
 import type { RealmTaxon } from "realmModels/types";
 import useTaxonSearch from "sharedHooks/useTaxonSearch";
 import useTranslation from "sharedHooks/useTranslation";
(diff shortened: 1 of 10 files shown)


src/components/Explore/SearchScreens/ExploreTaxonSearch.tsx
```diff
@@ -47,20 +46,25 @@ const ExploreTaxonSearch = ( {
...
-      testID={`Search.taxa.${taxon.id}`}
-    />
-  ), [
-    onPressInfo,
-    onTaxonSelected,
-  ] );
+  const renderItem = useCallback(
+    // no-unused-prop-types bug
+    // eslint-disable-next-line react/no-unused-prop-types
+    ( { item: taxon, index }: { item: RealmTaxon; index: number } ) => (
```
jtklein: We can do this here:
```
  const renderItem: ListRenderItem<RealmTaxon> = useCallback(
    ( { item: taxon, index } ) => (
```

and also as prop in TaxonSearch

ListRenderItem from "react-native" because downstream is only a FlatList

jtklein (approved):
Lookin good!
Only one minor type suggestion.

src/components/Explore/SearchScreens/ExploreTaxonSearch.tsx
sepeterson: Ya that's great! Thanks!

PR #3894, opened 2026-07-30
update AGENTS.md to encourage honest ts typing

sepeterson:
AGENTS.md update to try to give claude permission to leave ts errors that are outside the scope of the current work.

I was seeing output that relied on dishonest casting to achieve zero ts errors


jtklein (approved):
Seems alright.

PR #3896, opened 2026-07-31
Mob 1615 grouped by iconic taxa ordering logic

abbeycampbell:
closes [MOB-1615](https://linear.app/inaturalist/issue/MOB-1615/grouped-by-iconic-taxa-ordering-logic)

### DIFF
--- a/src/api/observationsTyped.ts
+++ b/src/api/observationsTyped.ts
@@ -1,0 +1,29 @@
+import type { ErrorWithResponse, INatApiError } from "api/error";
+import handleError from "api/error";
+import type { ApiOpts, ApiParams, ApiResponse } from "api/types";
+import inatjs from "inaturalistjs";
+
+export interface IconicTaxonCountResult {
+  count: number;
+  taxon: { name: string } | null;
+}
+
+export interface IconicTaxaCountsParams extends ApiParams {
+  user_id?: number;
+}
+
+const fetchIconicTaxaCounts = async (
+  params: IconicTaxaCountsParams = {},
+  opts: ApiOpts = {},
+): Promise<ApiResponse<IconicTaxonCountResult> | null | ErrorWithResponse | INatApiError> => {
+  try {
+    return await inatjs.observations.iconicTaxaCounts( params, opts );
+  } catch ( e ) {
+    return handleError(
+      e as ErrorWithResponse,
+      { context: { functionName: "fetchIconicTaxaCounts", opts } },
+    );
+  }
+};
+
+export default fetchIconicTaxaCounts;
--- a/src/api/observationsTyped.ts
+++ b/src/api/observationsTyped.ts
@@ -9,6 +9,10 @@ export interface IconicTaxonCountResult {
   taxon: { name: string } | null;
 }
 
+// Accepts only a user_id for now, but could be widened to accept a login string or an array of
+// ids or logins. This endpoint was created to assist with building the grouped-by-iconic-taxa
+// view for My Observations, so for now, narrowing to number only is a deliberate simplification
+// for what this feature actually needs.
 export interface IconicTaxaCountsParams extends ApiParams {
   user_id?: number;
 }
--- a/src/api/observationsTyped.ts
+++ b/src/api/observationsTyped.ts
@@ -1,5 +1,6 @@
 import type { ErrorWithResponse, INatApiError } from "api/error";
 import handleError from "api/error";
+import { ICONIC_TAXA_COUNTS_FIELDS } from "api/fields";
 import type { ApiOpts, ApiParams, ApiResponse } from "api/types";
 import inatjs from "inaturalistjs";
 
@@ -17,7 +18,10 @@ const fetchIconicTaxaCounts = async (
   opts: ApiOpts = {},
 ): Promise<ApiResponse<IconicTaxonCountResult> | null | ErrorWithResponse | INatApiError> => {
   try {
-    return await inatjs.observations.iconicTaxaCounts( params, opts );
+    return await inatjs.observations.iconicTaxaCounts(
+      { fields: ICONIC_TAXA_COUNTS_FIELDS, ...params },
+      opts,
+    );
   } catch ( e ) {
     return handleError(
       e as ErrorWithResponse,
--- a/src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
+++ b/src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
@@ -3,7 +3,7 @@ import type { IconicTaxaGroupCount } from "sharedHelpers/iconicTaxaGroupOrder";
 import { orderIconicTaxaCounts } from "sharedHelpers/iconicTaxaGroupOrder";
 import { useAuthenticatedQuery, useCurrentUser } from "sharedHooks";
 
-interface Options {
+interface IconicTaxaObservationCountsOptions {
   enabled?: boolean;
 }
 
@@ -11,7 +11,7 @@ interface Options {
 // returns them ordered most-observed to least-observed, falling back to
 // orderIconicTaxaCounts' tie-break rules
 const useIconicTaxaObservationCounts = (
-  { enabled = true }: Options = {},
+  { enabled = true }: IconicTaxaObservationCountsOptions = {},
 ): IconicTaxaGroupCount[] => {
   const currentUser = useCurrentUser( );
   const params = { user_id: currentUser?.id, ttl: -1 };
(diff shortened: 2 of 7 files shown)


src/api/observationsTyped.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { ApiOpts, ApiParams, ApiResponse } from "api/types";
+import inatjs from "inaturalistjs";
+
+export interface IconicTaxonCountResult {
+  count: number;
+  taxon: { name: string } | null;
+}
+
+export interface IconicTaxaCountsParams extends ApiParams {
+  user_id?: number;
```
jtklein, line 17: This seems correct to me according to the API v2 docs, albeit more restricted than what is possible: "Include observations observed by users that have this ID or login"
So, it could also be the login string instead of the ID number, and it could also be an array of ids or logins. Don't know if we should make a comment here that this is restrictive to our current usage.

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
```diff
@@ -0,0 +1,38 @@
+import fetchIconicTaxaCounts from "api/observationsTyped";
+import type { IconicTaxaGroupCount } from "sharedHelpers/iconicTaxaGroupOrder";
+import { orderIconicTaxaCounts } from "sharedHelpers/iconicTaxaGroupOrder";
+import { useAuthenticatedQuery, useCurrentUser } from "sharedHooks";
+
+interface Options {
```
jtklein: Can we name this more specific to the name of the hook?

src/sharedHelpers/iconicTaxaGroupOrder.ts
```diff
@@ -0,0 +1,74 @@
...
+  ANIMALIA = "animalia",
+  OTHER = "other",
+}
+
+// This is the tie-break order for the grouped-by-iconic-taxa view: when two or more
+// categories have the same observation count, they'll be ordered as they appear here.
+// This order is relevant to the myobs view only; IconicTaxonChooser.tsx has a different
+// order. Maybe someday we'll unify them, but for now they're separate.
+//
+// OTHER is the same as Explore's iconic taxon filter "unknown" (will include bacteria, viruses,
```
jtklein, line 25: Haven't there been recent changes on the web not to combine true "Unknowns" (no data) with bacteria and the likes anymore?

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
```diff
@@ -0,0 +1,38 @@
...
+// returns them ordered most-observed to least-observed, falling back to
+// orderIconicTaxaCounts' tie-break rules
+const useIconicTaxaObservationCounts = (
+  { enabled = true }: Options = {},
+): IconicTaxaGroupCount[] => {
+  const currentUser = useCurrentUser( );
+  const params = { user_id: currentUser?.id, ttl: -1 };
+
+  const { data } = useAuthenticatedQuery(
+    ["useIconicTaxaObservationCounts", params],
```
jtklein: Instead of `params` we should use `currentUser.id` in the array. The array of queryKey should be treated as the dependency array to the query and this requires stable refs like a useEffect array. The params object above get's recreated on every hook call.

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
```diff
@@ -0,0 +1,38 @@
...
+    async optsWithAuth => {
+      const response = await fetchIconicTaxaCounts( params, {
+        api_token: optsWithAuth.api_token ?? undefined,
```
jtklein: That is probably not required. Is that some defensive agent's work? I don't see us doing that in any other useAuthenticatedQuery.

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
```diff
@@ -0,0 +1,38 @@
...
+      } );
+      // Narrow out fetchIconicTaxaCounts' declared error cases so `results`
+      // is only ever accessed on a real response.
+      if ( !response || response instanceof Error || !( "results" in response ) ) {
+        throw response;
```
jtklein: If `response` would be null we'd be throwing a null, right? From what I remember we should be only ever throwing real Errors, but happy to learn more here.
Anyways, if response is null and not one of the error types, I think we should not just throw, you could create an error and throw it instead.
But I am a bit worried we create different approaches to API calls where we should probably try to consolidate all our approaches into a standardized API call and response handling approach.

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
```diff
@@ -0,0 +1,38 @@
...
+      }
+      return response;
+    },
+    { enabled: enabled && !!currentUser },
+  );
+
+  return orderIconicTaxaCounts( data?.results ?? [] );
```
jtklein, line 27: results should be in data, according to the check above.

jtklein (changes requested):
Have tested and this works for me. So, overall looks solid. I am a bit worried of creating a non-shared understanding of how API calls are handled in the app.
But, I am also not really sure how we can have a `response` of `null` from inatjs. So, I am not sure what kind of recommendation to make for handling null. I believe myself I have treated it mostly in the following cascade: errors need to be thrown and handled by the global error handlers, null needs to get through to the UI components and fallbacks or UI error state needs to appear locally only, responses are fine.

The only requested change is the queryKey, though. Everything else more the start of a conversation.

src/sharedHelpers/iconicTaxaGroupOrder.ts
abbeycampbell: Yes, but those changes were web UI only. we just changed the parameters that get triggered by the web UI, and we didn’t actually change what the endpoint returns for `iconic_taxa=unknown`. From Patrick:

> Yes we didn't change anything about iconic taxa, just what parameters some web UI elements relate to. There are a few clades that don't correspond to iconic taxa - these are "unknown, e.g. Bacteria, Viruses. There are other things that are "unknown" - observations ID'd to Life and observations without and IDs at all. All those things would fall into the bucket of iconic taxon unknown

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
abbeycampbell: yep, this this some claude ish that wasn't intentional on my part, but it seems like it was trying to address a type-mismatch that's coming up because we're only just starting to have useAuthenticatedQuery callers for properly typed api wrappers. 
`api_token` is typed as `string | null` but `ApiOpts.api_token` is typed `api_token?: string;`, so this line is meant to normalize api_token to the actually expected `string | undefined`. The better fix would be updating `ApiOpts.api_token` to explicitly accept string | null, but that's a bigger change that'll touch more stuff. I can make a maintenance/tech debt ticket for it.

src/api/observationsTyped.ts
abbeycampbell: yes you're correct. I'm just trying to keep things scoped to the needs of this specific feature, especially since this endpoint was added simply to help with this view. I'll add a comment for now to make this clearer.

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
abbeycampbell: I think this is fair feedback because it makes it more obvious what should trigger a refetch, but I will point out that we currently do the same thing in `useSpeciesCount.ts` so if this is desired we should at some point make the same change there. Also, I don't think the queryKeys work the same way as a useEffect dependency array, as long as the new object has the same values, it should produce the same cache key. https://tanstack.com/query/latest/docs/framework/react/guides/query-keys#query-keys-are-hashed-deterministically

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
abbeycampbell: This is more of claude trying to solve our type issues in the caller, and I think the move here is just to follow the conventions we use everywhere else, and accept the type errors, as you've done with the user project work. We can address the issues with the return type in a different ticket that's actually about that.

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
abbeycampbell: https://linear.app/inaturalist/issue/MOB-1637/widen-apioptsapi-token-to-accept-string-or-null

jtklein: OK, thanks

jtklein (approved):

PR #3897, opened 2026-08-01
Mob 1610 observation and species counts should reflect search results

abbeycampbell:
closes [MOB-1610](https://linear.app/inaturalist/issue/MOB-1610/observation-and-species-counts-should-reflect-search-results)


jtklein (approved):
LGTM

PR #3900, opened 2026-08-03
MOB-1595: ExploreV2 Redo search in map area

sepeterson:
Add the "Redo search in map area" feature for ExploreV2. Most of the diff is tests


src/components/Explore/ExploreV2/components/ExploreV2MapView.tsx
```diff
@@ -23,23 +29,46 @@ const WORLDWIDE_REGION: Region = {
...
+const regionFromBounds = ( bounds: ApiTotalBounds ): Region => {
+  const latitudeDelta = Math.abs( bounds.nelat - bounds.swlat );
+  const longitudeDelta = Math.abs( bounds.nelng - bounds.swlng );
+  return {
+    latitude: bounds.nelat - ( latitudeDelta / 2 ),
+    longitude: bounds.nelng - ( longitudeDelta / 2 ),
+    latitudeDelta,
+    longitudeDelta,
+  };
+};
```
abbeycampbell: is there a reason we couldn't re-use the existing `getMapRegion` helper for this?

abbeycampbell (approved):
lgtm! one question about a function that seems mostly like a dupe.

I did notice while testing that zooming directly in/out did not trigger the 'redo search' state. But I checked and that's how it works in explore today, so that's existing functionality. 

https://github.com/user-attachments/assets/d3c9fcb6-6f52-4866-b50a-25373cd167e6

sepeterson:
> I did notice while testing that zooming directly in/out did not trigger the 'redo search' state. But I checked and that's how it works in explore today, so that's existing functionality.

Yes good catch! I found that in practice you're almost always also panning while you're zooming so I left it as is. Should be changeable independently down the line if it comes up during testing

src/components/Explore/ExploreV2/components/ExploreV2MapView.tsx
sepeterson: They do the same thing, but with regionFromBounds returning exact bounds without any padding. I was able to combine them in mapHelpers!

PR #3903, opened 2026-08-04
Traditional projects: zustand project state on observation pojo

jtklein:
Closes MOB-1498

Wire Add-to-Projects chooser SAVE into Zustand observation flow: seed selection from existing POs, persist projectObservations and staged synced-PO deletions via updateObservationKeys, and disable SAVE when selection is unchanged.
- Add buildProjectObservationSelection helper (preserve/reuse POs, create new ones, stage synced removals, merge prior delete lists across chooser visits)
- Extend observation flow with currentObservation.projectObservationUuidsToDelete (Zustand-only, not Realm)
- Unit tests for helper + chooser persistence behavior

One thing I'd like to point out here is that this does introduce a new pattern which is in contrast to previous code:
When deleting an ObsPhoto or ObsSound we immediately send the delete call to the server (without even awaiting it), and delete the local versions. Here, we do sth more similar to the Classic apps which is to create an array of IDs of POs that need to be deleted. And then do the actual delete later on sync. A later ticket shall deal with adding this information to realm so that it persist over sessions. (Which thinking about it now potentially makes this Zustand key here completely unnecessary).


tests/unit/sharedHelpers/buildProjectObservationSelection.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+    it( "removes re-selected PO uuid from a prior ObsEdit session delete list", ( ) => {
+      const reselectedPo = factory( "LocalProjectObservation", {
+        projectId: 10,
+        uuid: "reselected-po-uuid",
+        _synced_at: new Date( ),
+      } );
+
+      const result = buildProjectObservationSelection(
+        [reselectedPo],
```
sepeterson, line 86: So I'm still iffy on the whole picture but to point to a single line that has an issue, claude asserts that in the actual usage in AddToProjects.tsx, this call would actually be `buildProjectObservationSelection([], ["reselected-po-uuid"], new Set( [10] ));` becuase reselectedPo would have been removed from currentObservation.projectObservations prior.


<details>
<summary>claude's summary if you want to dig in further</summary>

buildProjectObservationSelection.ts:49–57 — re-selection reconciliation keys on PO uuid, but selection identity is projectId.

A uuid only enters the delete list at the same moment its PO is dropped from projectObservations — both are driven off selectedProjectIds in one pass (line 32 pushes only selected, line 45 stages only unselected). So on any later chooser visit, line 33's find misses and line 37 mints a new uuid. keptUuids and priorUuidsToDelete are therefore disjoint by construction, and the filter on line 54 is a no-op in every reachable state — the behavior described in the comment on lines 52–53 never happens.

Repro: deselect a synced project → SAVE → reopen the chooser → re-select it → SAVE. Result is a new PO for that project plus the old PO's uuid still staged for deletion.

Two things to check independently:

- Call the helper twice, feeding its own output back in the way onSave does. Then compare with the tests at buildProjectObservationSelection.test.js:78 and :95 — both pass a PO that is simultaneously in existingProjectObservations and in existingUuidsToDelete, which the helper can't produce, and which happens to be the only input shape where line 54 fires.
- The swap also loses the synced PO's uuid/id/_synced_at, so Observation.needsSync() (Observation.js:602) stays true for that obs indefinitely, since no code path uploads POs yet.

No user-visible symptom today (flag off, delete list has no consumer). It becomes destructive when the delete is wired up, since a create and a delete would then be queued for the same (observation, project) pair.
</details>

sepeterson (approved):
All looks good from a code quality lens!

I wasn't able to vet that claude contention to my satisfaction so I'll leave it up to you if you want to address it now or if it'll be satisfied by future changes

tests/unit/sharedHelpers/buildProjectObservationSelection.test.js
jtklein: Thanks so much for this. I think this critique is correct, and I was also not sure this PO uuid array approach would stick around.
I will fold the fix for this problem into the next ticket. https://linear.app/inaturalist/issue/MOB-1508/tpod-c1-obsedit-save-persists-poofv-to-realm

PR #3904, opened 2026-08-04
MOB-1591: ExploreV2 keyboard shouldn't cover autocomplete results

sepeterson:
So you can always scroll to and select the last element

### DIFF
--- a/src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
+++ b/src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
@@ -1,3 +1,4 @@
+import { BottomTabBarHeightContext } from "@react-navigation/bottom-tabs";
 import { useNavigation } from "@react-navigation/native";
 import classnames from "classnames";
 import DefaultSearchOptions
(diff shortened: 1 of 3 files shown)


src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -87,6 +91,17 @@ const UniversalSearch = ( ) => {
   const commonNameIsPrimary = currentUser?.prefers_common_names !== false
     && currentUser?.prefers_scientific_name_first !== true;
 
+  const { keyboardHeight, keyboardShown } = useKeyboardInfo( );
+  const tabBarHeight = useContext( BottomTabBarHeightContext ) ?? 0;
```
jtklein, line 95: Haven't seen anyone use this before in the codebase.

jtklein (approved):
This works well on my end.
Not sure we needed to add testing for that.

My biggest concern here is to introduce divergent patterns of how we handle the "same" problem on different screens. From recent memory I know we do it a bit differently on TaxonSearch (granted there is no bottom UI so it needs to be not identical to here).

sepeterson:
> My biggest concern here is to introduce divergent patterns of how we handle the "same" problem on different screens. From recent memory I know we do it a bit differently on TaxonSearch (granted there is no bottom UI so it needs to be not identical to here).

Ya I agree with the diverging patterns concern. That said, I prefer the approach of exact measuring with the BottomTabBarHeightContext over adding an arbitrary amount of extra padding as TaxonSearch does. Maybe we can make a shared hook to consolidate in the future. Made this ticket: https://linear.app/inaturalist/issue/MOB-1648/shared-keyboard-spacing-hook

PR #3907, opened 2026-08-06
Tapping x in the standard camera is not working when photos taken

jtklein:
Closes MOB-1597

Because we did switch to resetting the nav stack on starting a new camera session some time ago, there is no goBack for this screen. So, we have to re-use what we do when we press the X button without having taken photos.


sepeterson (approved):
Looks good! Tested on ios

PR #3908, opened 2026-08-06
First iteration of a journal post details screen

jtklein:
Closes MOB-1454

Tapping a post in any journal list now opens a PostDetails screen with the title, date, and rendered HTML body. Rows became pressable (list owns navigation, item stays presentational), UserText gained an optional contentMargin so images fit their container, and its HTML pipeline is memoized.
Includes a RemotePost factory and basic unit tests. Haven't spent much time on these as this entire feature is still likely to change.

Why UserText over a WebView: it's already `sanitize-html`'d, with its tag allowlist explicitly kept in sync with the Rails Post::ALLOWED_TAGS. It renders real RN text nodes, so font scaling and screen readers work for free, and mentions/links already route correctly.
Reusing FullPageWebView as it currently is implemented wasn't reusable here immediately and also would have needed a new instance of another screen.

Limitation: iframe/embed are stripped, so embedded video doesn't render.

IMG_4952

### DIFF
--- a/src/components/Journal/PostList.tsx
+++ b/src/components/Journal/PostList.tsx
@@ -7,7 +7,7 @@ import {
   InfiniteScrollLoadingWheel,
 } from "components/SharedComponents";
 import { View } from "components/styledComponents";
-import React, { useMemo } from "react";
+import React, { useCallback, useMemo } from "react";
 
 import PostListItem from "./PostListItem";
 
@@ -45,8 +45,9 @@ const PostList = ( {
 
   const emptyComponent = useMemo( () => <ActivityIndicator size={50} />, [] );
 
-  const renderPost: ListRenderItem<ApiPostForUser> = ( { item } ) => (
-    <PostListItem item={item} />
+  const renderPost: ListRenderItem<ApiPostForUser> = useCallback(
+    ( { item } ) => <PostListItem item={item} />,
+    [],
   );
 
   return (
--- a/src/components/Journal/PostList.tsx
+++ b/src/components/Journal/PostList.tsx
@@ -1,4 +1,5 @@
 import { useNetInfo } from "@react-native-community/netinfo";
+import { useNavigation } from "@react-navigation/native";
 import type { ListRenderItem } from "@shopify/flash-list";
 import type { ApiPostForUser } from "api/types";
 import {
@@ -7,6 +8,7 @@ import {
   InfiniteScrollLoadingWheel,
 } from "components/SharedComponents";
 import { View } from "components/styledComponents";
+import type { TabStackScreenProps } from "navigation/types";
 import React, { useCallback, useMemo } from "react";
 
 import PostListItem from "./PostListItem";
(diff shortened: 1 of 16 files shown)


src/components/Journal/PostList.tsx
```diff
@@ -38,8 +45,17 @@ const PostList = ( { posts, isFetchingNextPage, fetchNextPage }: Props ) => {
 
   const emptyComponent = useMemo( () => <ActivityIndicator size={50} />, [] );
 
-  const renderPost: ListRenderItem<ApiPostForUser> = ( { item } ) => (
-    <PostListItem item={item} />
+  const onPressPost = useCallback( ( post: ApiPostForUser ) => {
+    navigation.navigate( "PostDetails", {
```
websoty, line 49: Would it make sense to pass only the post ID here and retrieve the post in PostDetails instead? This could prevent the details screen from displaying stale data if the post gets updated

src/components/SharedComponents/UserText.tsx
```diff
@@ -99,9 +100,10 @@ const LINKIFY_OPTIONS: Opts = {
   },
 };
 
-interface Props extends React.PropsWithChildren {
+interface Props extends PropsWithChildren {
   text: string;
   htmlStyle?: object;
+  contentMargin: number;
```
sepeterson, line 106: Should contentMargin be optional since it has a default value?

sepeterson (approved):

src/components/Journal/PostListItem.tsx
```diff
@@ -34,8 +37,17 @@ const PostListItem = ( {
       : item.parent.icon_url ?? null;
   }, [item.body, item.parent.icon_url] );
 
+  const handlePress = useCallback( ( ) => onPress( item ), [item, onPress] );
+
   return (
-    <View className="flex-row items-center mx-3 my-2">
+    <Pressable
+      accessibilityRole="button"
+      accessibilityLabel={item.title}
```
sepeterson, line 45: Also we should probably add the date in the accessibilityLabel if it's within this pressable

src/components/Journal/PostList.tsx
jtklein: That would be standard practice. But I believe we do not actually have an endpoint to get a Journal Post by ID.

src/components/SharedComponents/UserText.tsx
jtklein: Decidedly not because this fixes a bug with the current user text in all instances it is used. I did not want to extent the scope of this ticket to also pass in an appropriate contentMargin in the other circumstances we use UserText. But I do want to see at least a TS error in other places we use it so that we fix it at one point.

src/components/Journal/PostListItem.tsx
jtklein: Just tested this in VoiceOver, and I think this is fine as is.

PR #3911, opened 2026-08-07
Mob 1616 new UI for grouped by iconic taxa view small grid and

abbeycampbell:
closes [MOB-1616](https://linear.app/inaturalist/issue/MOB-1616/new-ui-for-grouped-by-iconic-taxa-view-small-grid-and-collapsible)

very sorry, this turned out to be kinda big. many of the file touches are related to new icons, snapshot updates, and translations strings. No need to pay much attention to `MyObservationsGroupedByIconicTaxaView.tsx` either, that's mostly placeholder UI and logic that will be coming out (It'll show you just your local realm obs divvied up between taxa catagories for now, with no regard for accuracy; this is just about the UI).


src/i18n/strings.ftl
```diff
@@ -552,6 +558,8 @@ Filters-observations-to-anywhere-in-the-world = Filters observations to anywhere
 # Accessibility hint for the "Nearby" default option in the location search
 Filters-observations-to-your-current-location = Filters observations to your current location.
 # Status when an item has been flagged
```
sepeterson: Looks like this comment get separated from its relevant string `Flagged = Flagged`

Also an issue at 687, 728, 987, 1026

sepeterson:
Also seeing that the view buttons can cover obs or collapsed sections if you scroll all the way down

<details>
<summary>screenshot</summary>
IMG_0210
IMG_0209
</details>

sepeterson (approved):
Unironically love a collapsable section

I think the fluent comments should get fixed before merge but the overlap is up to you

src/components/SharedComponents/CollapsibleSectionHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+        className="flex-row items-center justify-between px-[15px] py-[10px] bg-white"
+        onPress={onToggle}
+      >
+        <View className="flex-row items-center">
+          <INatIcon name={icon} size={18} />
+          <Body3 className="ml-[8px]">{title}</Body3>
+        </View>
+        <View className="flex-row items-center">
+          <Body3 className="mr-[10px]">{t( "Intl-number", { val: count } )}</Body3>
+          <INatIcon
```
websoty, line 42: Should these icons be reversed? Usually an expanded section shows caret-up and a collapsed one shows caret-down

src/components/MyObservations/MyObservationsGroupedByIconicTaxaView.tsx
```diff
@@ -1,23 +1,237 @@
...
+  const iconicTaxaCounts = useIconicTaxaObservationCounts( );
+  const titlesByCategory = useMemo( ( ) => iconicTaxaGroupTitles( t ), [t] );
+
+  // Temporary testing set up which just distributes the list of local obs across every
+  // category so we can mock up multiple sections.
+  const observationsByCategory = useMemo( ( ) => {
+    const buckets = new Map<ICONIC_TAXA_GROUP, typeof localObservationIds>(
+      ICONIC_TAXA_GROUP_ORDER.map( category => [category, []] ),
+    );
+    localObservationIds.forEach( ( item, index ) => {
```
websoty, line 85: Is this temporary distribution expected to ship with this PR? Right now observations are assigned to iconic taxa categories based on their index rather than their actual tax

src/components/MyObservations/MyObservationsGroupedByIconicTaxaView.tsx
```diff
@@ -1,23 +1,237 @@
...
+  }, [] );
+
+  // Sections are ordered most-observed to least-observed, which is the order iconicTaxaCounts
+  // comes back in.ICONIC_TAXA_GROUP_ORDER is the tie-break order that hook sorts by.
+  const rows = useMemo( ( ) => iconicTaxaCounts.flatMap( ( { category, count } ): GroupedRow[] => {
+    const observations = observationsByCategory.get( category ) ?? [];
+    if ( observations.length === 0 ) return [];
+
+    const isOpen = !closedCategories.has( category );
+    const headerRow: GroupedRow = {
```
websoty, line 115: Are we intentionally mixing remote category counts with the temporary local observation buckets here? The header count can represent a completely different set of observations than the tiles rendered below it

src/sharedHelpers/observationUploadStatus.ts
```diff
@@ -0,0 +1,27 @@
...
+// values ObservationsFlashList's renderItem computes inline
+const getObservationUploadStatus = (
+  realm: Realm,
+  uploadQueue: string[],
+  totalUploadProgress: TotalUploadProgressItem[],
+  uuid: string,
+) => {
+  const obsNeedsSync = Observation.isUnsyncedObservation( realm, { uuid } );
+  const obsUploadState = totalUploadProgress.find( o => o.uuid === uuid );
+  const uploadProgress = obsNeedsSync
```
websoty, line 19: Should an unsynced observation with no obsUploadState return undefined instead of 0? ObsPressable explicitly uses undefined to mean “not currently uploading”, so defaulting to 0 here may change the upload-status UI

src/components/SharedComponents/SmallGrid.tsx
```diff
@@ -0,0 +1,125 @@
...
+      ? lastTileSize
+      : tileSize;
+    const marginTop = isFirstRowInSection
+      ? 0
+      : SMALL_GRID_GAP;
+
+    return (
+      <View
+        className="items-start"
+        style={{ width: slotWidth, height: tileSize, marginTop }}
```
websoty, line 102: Should the last tile use the same value for width and height? With lastTileSize as the width but tileSize as the height, the last column renders slightly rectangular while the other tiles are square

src/i18n/strings.ftl
abbeycampbell: oops, great catch

src/components/SharedComponents/CollapsibleSectionHeader.tsx
abbeycampbell: just following the design here, but I can bring it up just in case it was a mistake

src/components/MyObservations/MyObservationsGroupedByIconicTaxaView.tsx
abbeycampbell: yep! see the PR description, this is just about adding the UI and all this stuff is temporary

src/components/SharedComponents/SmallGrid.tsx
abbeycampbell: yep, this deliberate. The product ask is that there is always a 3px gap and the tiles meet the edge of the screen, and those can't both be true with perfectly square tiles if the screen width doesn't divide evenly into the number of columns plus the gaps. I let the last column absorb the space rather than having a gap on the right edge; I found it to be practically imperceptible on several screen sizes.

src/sharedHelpers/observationUploadStatus.ts
abbeycampbell: yes correct. I already ran into this issue myself while trying to coalesce to a number like `uploadProgress ?? 0`. Looking into this further in my next PR but for now this doesn't touch any behavior that's accessible to users. We have a performance fix coming for `isUnsyncedObservation` that will interact with this as well

PR #3912, opened 2026-08-07
Traditional projects: ObsEdit save persists PO and OFV to realm

jtklein:
Closes MOB-1508

This PR:
- Replace MOB-1498s projectObservationUuidsToDelete with in-array _pendingRemoval soft-delete (POs + cleared synced OFVs).
- On SAVE, saveLocalObservationForUpload writes active embeds and Realm _pending_deletion tombstones; selective dirty for changed embeds only.
- Need Realm schema bump; v72: _pending_deletion on PO/OFV embeds.

### DIFF
--- a/src/components/AddToProjects/AddToProjects.tsx
+++ b/src/components/AddToProjects/AddToProjects.tsx
@@ -142,17 +142,13 @@ const AddToProjects = ( ) => {
   }, [] );
 
   const onSave = useCallback( ( ) => {
-    const {
-      projectObservations,
-      projectObservationUuidsToDelete,
-    } = buildProjectObservationSelection(
+    const { projectObservations } = buildProjectObservationSelection(
       currentObservation?.projectObservations,
       currentObservation?.projectObservationUuidsToDelete,
       selectedProjectIds,
     );
     updateObservationKeys( {
       projectObservations,
-      projectObservationUuidsToDelete,
     } );
     navigation.goBack( );
   }, [
--- a/src/components/AddToProjects/AddToProjects.tsx
+++ b/src/components/AddToProjects/AddToProjects.tsx
@@ -144,7 +144,6 @@ const AddToProjects = ( ) => {
   const onSave = useCallback( ( ) => {
     const { projectObservations } = buildProjectObservationSelection(
       currentObservation?.projectObservations,
-      currentObservation?.projectObservationUuidsToDelete,
       selectedProjectIds,
     );
     updateObservationKeys( {
--- a/src/components/AddToProjects/AddToProjects.tsx
+++ b/src/components/AddToProjects/AddToProjects.tsx
@@ -60,6 +60,7 @@ const AddToProjects = ( ) => {
 
   const initialSelectedProjectIds = new Set(
     ( currentObservation?.projectObservations ?? [] )
+      .filter( po => !po._pendingRemoval && !po._pending_deletion )
       .map( po => po.projectId ),
   );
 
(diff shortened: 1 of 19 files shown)


src/components/AddToProjects/hooks/useObservationFieldValue.ts
```diff
@@ -13,27 +26,36 @@ const useObservationFieldValue = ( obsFieldId: number ) => {
...
 
-  const existingOfv = observationFieldValues.find(
-    ofv => ofv.obsFieldId === obsFieldId,
-  );
+  const existingOfv = findActiveOfv( observationFieldValues, obsFieldId );
   const value = existingOfv?.value ?? "";
 
   const setValue = useCallback( ( newValue: string | null ) => {
-    const index = observationFieldValues.findIndex( ofv => ofv.obsFieldId === obsFieldId );
+    const existingRow = observationFieldValues.find( ofv => ofv.obsFieldId === obsFieldId );
```
websoty: Should we use findActiveOfv here as well? Otherwise existingRow may resolve to a pending-removal/deleted OFV while the displayed value is based on the active one

src/components/AddToProjects/hooks/useObservationFieldValue.ts
```diff
@@ -3,6 +3,19 @@ import ObservationFieldValue from "realmModels/ObservationFieldValue";
 import type { RealmObservationFieldValuePojo, RealmObservationPojo } from "realmModels/types";
 import useStore from "stores/useStore";
 
+function isActiveOfv( ofv: RealmObservationFieldValuePojo ): boolean {
+  return !ofv._pendingRemoval && !ofv._pending_deletion;
```
sepeterson: Should we be checking these in ObservationFieldValue.ts `findForObsField` function too? and maybe reusing that for findActiveOfv?

src/realmModels/Observation.js
```diff
@@ -280,6 +280,41 @@ class Observation extends Realm.Object {
...
+        _updated_at: now,
+      };
+    }
+
+    const isNew = !existingEmbed;
+    const wasReactivated = existingEmbed?._pending_deletion && !embed._pending_deletion;
+    if ( isNew || wasReactivated ) {
+      return {
+        ...embed,
+        _synced_at: null,
```
sepeterson, line 303: It seems like we use _synced_at to mean "is this on the server" and a reactivated PO should be, right? So should we keep the existing _synced_at? And we'll know it needs sync bc `_synced_at < _updated_at`?

src/components/AddToProjects/AddToProjects.tsx
```diff
@@ -60,6 +60,7 @@ const AddToProjects = ( ) => {
 
   const initialSelectedProjectIds = new Set(
     ( currentObservation?.projectObservations ?? [] )
+      .filter( po => !po._pendingRemoval && !po._pending_deletion )
```
sepeterson, line 63: Maybe there should be a shared helper that checks these since it seems to happen in a few places

sepeterson (approved):
Looks good overall! Just some open questions that you can choose to pursue or not. We can do more testing when there are UI paths for all this stuff

src/components/AddToProjects/AddToProjects.tsx
jtklein: I don't mind being explicit for now.

src/components/AddToProjects/hooks/useObservationFieldValue.ts
jtklein: Have consolidated a bit. But can not outright only find active in all cases. We'd still need to find all OFVs when checking if we need to re-enable a field for example.

src/components/AddToProjects/hooks/useObservationFieldValue.ts
jtklein: No, so that re-entering a value after a clear restores the same uuid.

PR #3914, opened 2026-08-07
add myobs sort, search, and map view feature flags to enabled for admins array

abbeycampbell:
[MOB-1655](https://linear.app/inaturalist/issue/MOB-1655/add-new-myobs-features-to-admin-testing-pool)

if we stick with the current release plan, we will want to have new myobs features under test by admins in the internal build available ~Aug 12-25, so this is a PR to enabled that for the parts that seem ready (small grid view tbd, but seeming unlikely). 

One call out from the testing notes of Ryan's ticket:

> We may want to do a production release with a no-op flag to make sure that the "AppStore" vs "TestFlight" check works as expected in our app. Like an "Installer source: [App Store | Testflight]" label in Debug

So, if that's the case, maybe we want to wait an additional cycle to ensure everything is working as expected in production? Open to thoughts, but teeing up the PR just in case we want to proceed.

### DIFF
--- a/src/stores/createFeatureFlagSlice.ts
+++ b/src/stores/createFeatureFlagSlice.ts
@@ -33,6 +33,9 @@ export enum FeatureFlag {
 
 export const flagsEnabledForAdminsInTestFlight = [
   FeatureFlag.TestFlightAdminMessageEnabled,
+  FeatureFlag.SearchMyObservationsEnabled,
+  FeatureFlag.SortMyObservationsEnabled,
+  FeatureFlag.MyObservationsMapViewEnabled,
 ];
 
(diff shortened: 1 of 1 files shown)


FLGMwt:
> So, if that's the case, maybe we want to wait an additional cycle to ensure everything is working as expected in production? Open to thoughts, but teeing up the PR just in case we want to proceed.

I say we include it.

If there's an issue with the admins-only Feature Flag (if the diagnostic message incorrectly shows up in prod), we'll still have another regular mid-cycle build opportunity to remove myObs from that config.

FLGMwt (approved):
wooooooo!

websoty (approved):

PR #3918, opened 2026-08-10
Link to observation, project, and taxon in-app from UserText

jtklein:
Closes MOB-440

ProjectDetails is not setup to handle slugs as nav params, ObsDetails is not setup to handle numeric IDs as nav params, so I added an inline lookup to get the numeric ID and UUID, respectively (following existing code).
Then we can route to the details screens as required by the ticket.

### DIFF
--- a/src/sharedHelpers/inatUrlNavigation.ts
+++ b/src/sharedHelpers/inatUrlNavigation.ts
@@ -1,4 +1,6 @@
 import type { NavigationProp, ParamListBase } from "@react-navigation/native";
+import { searchObservations } from "api/observations";
+import { getJWT } from "components/LoginSignUp/AuthenticationService";
 
 const ALLOWED_HOSTS = [
   "www.inaturalist.org",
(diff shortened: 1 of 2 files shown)


src/sharedHelpers/inatUrlNavigation.ts
```diff
@@ -0,0 +1,113 @@
...
+import { searchObservations } from "api/observations";
+import { searchProjects } from "api/projects";
+import { getJWT } from "components/LoginSignUp/AuthenticationService";
+
+const ALLOWED_HOSTS = [
+  "www.inaturalist.org",
+  "inaturalist.org",
+  "api.inaturalist.org",
+  "staging.inaturalist.org",
+];
```
FLGMwt, line 11: non-blocker: we could add staging-lite & staging1-4

FLGMwt (approved):
Really slick idea! works for me against staging

PR #3923, opened 2026-08-11
Remove decompress package

jtklein:
Has a security warning, and we don't even seem to be using it anywhere.

Compiled and tested in Debug and Release mode on Android and iOS.


FLGMwt (approved):
Good find! This prompted me to search for other unused deps, I'll PR a followup.

PR #3925, opened 2026-08-11
MOB-1662 add osx 27 to gemfile platforms

FLGMwt:
Closes MOB-1662

I cannot overstate how much I regret the whim of committing to osx beta without understanding the irreversible consequences.

anyway, here's ~wonderwall~ something that we'd need to do eventually anyway


jtklein (approved):
lgtm

PR #3926, opened 2026-08-11
flatten logging _extraData for readability

FLGMwt:
Closes MOB-1463 to clean up triage.

instead of:

```
{ 
  extraData: {
    id: 1,
    obsCount: 100,
  }
}
 ```

We're now _logging_

```
{
  _extraData: true,
  id:1,
  obsCount: 100
}
```

In both cases, we're only ever sending the following to `api/log`:

```
{ 
  id: 1, 
  obsCount: 100 
}
```


jtklein (approved):
LGTM
Tested by sending a feedback from a production build. Extras still appeared.

PR #3927, opened 2026-08-11
Update react-native-sensitive-info to v6

jtklein:
Closes MOB-1638 (also closes MOB-1160)

Summary
- Upgrade react-native-sensitive-info 5.6 → 6.1.5 (Nitro) and adapt auth storage wrappers/tests to the v6 API.
- On startup, if Realm still has a signed-in user but secure tokens are missing, send them through the existing “Please log in again” flow.

Test plan
- Fresh login / sign-out still stores and clears tokens
- Upgrade path: signed-in Realm + empty secure store → re-login screen for the same user

### DIFF
--- a/src/components/hooks/useDeferredStartup.ts
+++ b/src/components/hooks/useDeferredStartup.ts
@@ -10,6 +10,8 @@
  * runtime can interleave user interactions between them.
  */
 import { cleanupLogFiles } from "components/Developer/logManagementHelpers";
+import { isLoggedIn } from "components/LoginSignUp/AuthenticationService";
+import { navigationRef } from "navigation/navigationUtils";
 import { RealmContext } from "providers/contexts";
 import { useEffect } from "react";
 import User from "realmModels/User";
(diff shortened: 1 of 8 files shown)


src/components/hooks/useDeferredStartup.ts
```diff
@@ -84,6 +86,13 @@ const useDeferredStartup = ( ) => {
   const { i18n } = useTranslation( );
 
   useEffect( ( ) => {
+    const id0 = deferTask( "checkLoggedInStatuses", async () => {
```
websoty, line 89: Could this miss the re-login flow if navigationRef isn't ready when this deferred task runs? Since this is a one-time check, it looks like wouldn't retry once navigation becomes ready
maybe I'm wrong ...

sepeterson (approved):
Code looks good. 

I did find that the upgrade path on android seems to require sign-in again. Probably good to get this out now. iOS went fine

PR #3928, opened 2026-08-11
Remove extraneous deps 11 08 26

FLGMwt:
followup from #3923


abbeycampbell (approved):
Seems good to me. One ask: can you fix the link to the comment in the description so that it points to the right place? for posterity :)

PR #3930, opened 2026-08-11
widen gap between search and sync icons

abbeycampbell:
closes [MOB-1664](https://linear.app/inaturalist/issue/MOB-1664/increase-spacing-between-sync-button-and-search-button)


websoty (approved):

jtklein (approved):
Code change is harmless.

So, 7px because both buttons are 44? So (44-20):2+(44-26):2+x = 28

PR #3933, opened 2026-08-12
Private location not shown in ObsEdit

jtklein:
Closes MOB-1603

Fix Obs Edit showing "No Location" for private observations after upload/sync.

After sync, true coordinates live in privateLatitude/privateLongitude while public lat/lng are null. ObsEdit only read the public fields. This aligns edit UI, upload mapping, and list sync with the existing detail/LocationPicker fallback pattern.

### DIFF
--- a/src/components/LocationPicker/types.ts
+++ b/src/components/LocationPicker/types.ts
@@ -1,10 +1,8 @@
 export interface LocationPickerObservation {
   latitude?: number;
   longitude?: number;
-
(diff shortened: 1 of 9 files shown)


FLGMwt (approved):
LGTM. Confirming that I reproduced the bug on main, switched to this branch and the location was present

and good idea on the missing location helper!

websoty (approved):

PR #3935, opened 2026-08-12
MOB-1223: scroll to top in my observations after syncing remote

sepeterson:
- return hasNewObservations from sync helper
- optional callback on sync hook
- if hasNewObservations and we're not using a list ordered by the server, scroll to the top

<details>
<summary>recording</summary>

https://github.com/user-attachments/assets/46447894-05ea-4e2b-9d60-8c161f07d169

</details>


jtklein (approved):
LGTM. Solid code.

PR #3936, opened 2026-08-12
Mob 1663 tapping reset in search in myobs should not dismiss search

abbeycampbell:
closes [MOB-1663](https://linear.app/inaturalist/issue/MOB-1663/tapping-reset-in-search-in-myobs-should-not-dismiss-search-screen)

Allowing `SearchBar` to honor an externally changed value will give us a behavior change for the `LocationPicker`, which I've explained and raised to product in the linear ticket (IMO, good changes & got Abhas's blessing)


sepeterson (approved):
Looks good! Tested on ios

PR #3937, opened 2026-08-13
iOS share extension photos are copied and never deleted

jtklein:
Closes MOB-1640

Delete iOS share-extension originals from the App Group after resizing into photoUploads, so shared photos no longer accumulate as redundant full-res copies.

Also some TS migrations, and cleanup of no longer used code.

If you checkout https://github.com/inaturalist/iNaturalistReactNative/pull/3937/commits/21edcd149c94483b242d90051088fa08e1397c06 there is a patch to show the AppGroup Shared folder in rozenite for testing that there are no more photos in it.

Before:
Screenshot 2026-08-13 at 12 21 23
After
Screenshot 2026-08-13 at 12 44 21


sepeterson (changes requested):
Pretty sweet addition to the rozenite file viewer!

I was able to use it to find a couple bugs:

1. Import photos as a single observation -> Suggestions -> tap back button -> import again -> error (photos are no longer in app group)

<details>
<summary>recording</summary>

https://github.com/user-attachments/assets/d72e4a63-c28d-4a0e-8ea5-8be43e3e5e6c

</details>

2. Import photos into app -> delete in importer -> photos are not cleaned up

<details>
<summary>recording</summary>

https://github.com/user-attachments/assets/43d7083f-2025-4ec0-9c05-c18bf2fb689e

</details>

src/components/PhotoImporter/GroupPhotosContainer.tsx
```diff
@@ -1,42 +1,59 @@
...
+
+  const { t } = useTranslation( );
+  const [selectedObservations, setSelectedObservations] = useState<GroupedPhoto[]>( [] );
   const [isCreatingObservations, setIsCreatingObservations] = useState( false );
   const totalPhotos = groupedPhotos
     .reduce( ( count, current ) => count + current.photos.length, 0 );
 
-  useEffect( ( ) => {
-    navigation.setOptions( {
+  const headerOptions = useMemo(
```
jtklein, line 41: The header was stuck at text not being translated for me in some very quick app starts.

src/components/PhotoImporter/GroupPhotosContainer.tsx
```diff
@@ -136,7 +153,17 @@ const GroupPhotosContainer = ( ): Node => {
 
   const navBasedOnUserSettings = async ( ) => {
     setIsCreatingObservations( true );
-    const newObservations = await Promise.all( groupedPhotos.map(
+    const updatedGroupedPhotos = await moveSharedGroupedPhotos(
```
jtklein, line 156: Actual fix of the bug in the ticket. Move the iOS files into the DocumentDirectory after which they are treated the same as other files that are imported from photo library.

src/sharedHooks/useShare.ts
```diff
@@ -1,7 +1,15 @@
 import { useEffect } from "react";
 import ShareMenu from "react-native-share-menu";
 
-const useShare = ( onShare: ( ) => void ): void => {
+export interface SharedData {
```
jtklein, line 4: This does introduce a TS error further down the code path because the helpers in GroupPhotos seem to also expect a timestamp property on photos imported from the gallery. We re-use the helpers for photos shared here but I haven't had time to look into if react-native-share-menu returns that property. (Ideally we would import the type directly from the library and not have to create our own type).
I didn't see any bugs because of it, so potentially we can also just make the property optional. But that all felt a bit too scope creepy.

src/stores/createLayoutSlice.ts
```diff
@@ -14,22 +16,54 @@ export enum MAP_TYPES {
   HYBRID = "hybrid",
 }
 
-const createLayoutSlice = set => ( {
+export interface LayoutSlice {
```
jtklein, line 19: Can't believe I am the first person to touch something that imports this file since we decided that whatever we touch we migrate to TS but there you go.

jtklein:
Alright. Those were not bugs, that was me not realizing that I am not dealing with the second code path of > 1 photo.
So, had to massively increase the PR with additional handling of removing the photo files in the Group-photos-flow. Few more files that needed TS migrations.
By migrating the group photos screens to TS I have also seen that we are dealing with a similar problem in importing from the gallery by moving the files to the app's Document folder were they are cleaned using the deferred startup cleanup services.
I have tried moving them immediately when receiving something being shared, but that enters into a race with the clean-up processes on app start. So, sharing a photo on iOS with moving it into the gallery folder will compete with the gallery folder being deleted on app start. It would have been my preferred way because then we would treat importing from gallery and sharing into the app the same after the initial file move. Alas, it will require to have a special check if we are starting the app with something shared and thereby don't want to delete, or delete only files older than x or something more elegant I couldn't think of.
So, my compromise here is to move the files into the gallery folder upon creating the observations, i.e. currently on pressing "SAVE X OBSERVATION" in the group photos screen. After that they are essentially treated and cleaned up the same way photos imported from the gallery are. The file move requires and update to the importer state, in case the user comes back to the importer screen. This did result in UI glitch of photos being not shown until file move has finished in the screen. So, I added a loading indicator which incidentally also fixes some existing bugs of being able to mash the + button while the user has accepted observations.

sepeterson (approved):
Sorry about the big scope increase but this solution looks solid to me. Passed all testing scenarios

PR #3939, opened 2026-08-13
MOB-1346: advanced search screen

sepeterson:
I'll follow up with tests in a second PR. This one's long enough

As it is, you can't really tell you've got an advanced search applied on the results screen but that will change in https://linear.app/inaturalist/issue/MOB-1348/advanced-search-mode


src/components/Explore/ExploreV2/helpers/advancedSearchOptions.ts
```diff
@@ -0,0 +1,186 @@
...
+  value: ValueT;
+}
+const option = <ValueT extends string | number>(
+  value: ValueT,
+  label: string,
+): [ValueT, FilterOption<ValueT>] => [value, { label, value }];
+
+// `label` / `text` are the two lines of each row in the sort sheet; `labelCaps`
+// is the single-line, all-caps version in the dropdown button
+export const getSortByValues = ( t: TFunction ) => ( {
```
sepeterson, line 25: This file is mostly copied from ExploreV1's FilterModal.tsx

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
-  const { t } = useTranslation( );
+  const navigation
+    = useNavigation<ExploreStackScreenProps<"AdvancedSearch">["navigation"]>();
+  const { t } = useTranslation();
+  const iconicTaxa = useIconicTaxa();
+  const currentUser = useCurrentUser();
+
+  const { state: v2State, dispatch: dispatchV2 } = useExploreV2();
+  const [initialDraft] = useState( ( ) => draftFromV2State( v2State ) );
+  const [draft, dispatch] = useReducer( advancedSearchReducer, initialDraft );
```
sepeterson, line 89: Like UniversalSearch, this screen holds a draft of future state locally and only commits it to the context when the user applies the search

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+      };
+    case "SET_SORT":
+      return { ...draft, sortBy: action.sortBy };
+    // Within the scope of advanced search, we only set a taxon as the subject
+    // and all other subjects go in the filter blob
+    // and if a non-taxon subject came in initially, it has to be excluded via withoutSubjectOfType
+    case "SET_USER":
+      return {
+        ...withFilters( draft, { user: action.user, excludeUser: null } ),
+        subject: withoutSubjectOfType( draft, "user", "unobserved" ),
```
sepeterson: Lines like this only exist to translate an initial ExploreV2State set in UniversalSearch with a non-taxon subject. Not the most elegant part of this PR

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+    case "SET_DATE_OBSERVED_EXACT":
+      return withFilters( draft, {
+        dateObserved: DATE_OBSERVED.EXACT_DATE,
+        observed_on: action.observedOn,
+        d1: null,
+        d2: null,
+        months: null,
+      } );
+    case "SET_DATE_OBSERVED_RANGE":
+      return withFilters( draft, {
```
jtklein, line 175: This used to have a `isValidDateFormat` check before settig to state. Is that something we can drop or have replaced with something else somewhere?
Also some other actions had that, I'll leave some Comments.

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+        lrank: action.lrank === TAXONOMIC_RANK.none
+          ? null
+          : action.lrank,
+      } );
+    case "SET_DATE_OBSERVED_ALL":
+      return withFilters( draft, {
+        dateObserved: DATE_OBSERVED.ALL, observed_on: null, d1: null, d2: null, months: null,
+      } );
+    case "SET_DATE_OBSERVED_EXACT":
+      return withFilters( draft, {
```
jtklein, line 167: isValidDateFormat

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+      } );
+    case "SET_LRANK":
+      return withFilters( draft, {
+        lrank: action.lrank === TAXONOMIC_RANK.none
+          ? null
+          : action.lrank,
+      } );
+    case "SET_DATE_OBSERVED_ALL":
+      return withFilters( draft, {
+        dateObserved: DATE_OBSERVED.ALL, observed_on: null, d1: null, d2: null, months: null,
```
jtklein: Please add newlines

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+      return withFilters( draft, {
+        dateObserved: DATE_OBSERVED.MONTHS,
+        observed_on: null,
+        d1: null,
+        d2: null,
+        months: action.months,
+      } );
+    case "SET_DATE_UPLOADED_ALL":
+      return withFilters( draft, {
+        dateUploaded: DATE_UPLOADED.ALL, created_on: null, created_d1: null, created_d2: null,
```
jtklein: Newlines

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+    case "SET_DATE_OBSERVED_MONTHS":
+      return withFilters( draft, {
+        dateObserved: DATE_OBSERVED.MONTHS,
+        observed_on: null,
+        d1: null,
+        d2: null,
+        months: action.months,
+      } );
+    case "SET_DATE_UPLOADED_ALL":
+      return withFilters( draft, {
```
jtklein, line 191: isValidDateFormat

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+        dateUploaded: DATE_UPLOADED.ALL, created_on: null, created_d1: null, created_d2: null,
+      } );
+    case "SET_DATE_UPLOADED_EXACT":
+      return withFilters( draft, {
```
jtklein, line 198: isValidDateFormat

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
+import type { ApiProjectSummary, ApiUser } from "api/types";
+import type {
+  ESTABLISHMENT_MEAN,
+  MEDIA,
+  PHOTO_LICENSE,
+  REVIEWED,
+} from "providers/ExploreContext";
+import {
```
jtklein, line 8: Some imported as type some not, is there a reason?

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
jtklein: Alright, I think I have understood why that is here now. Not the biggest fan of this. Thinking back I believe that was one of the reasons of the Explore V1 context to not split out any "substate" like a subject and just dump everything into a giant reducer.
It did have it's own problems it seems now: comparing V1 to your V2 reducer I can see a lo of drift in V1, meaning some of the action.keys are not even used anymore by any dispatch. So, maybe it was harder keeping track of what to pass in.
Anyway, not asking you to change anything here. Happy to move these things forward and just see where we end up with V2 once everything is tied up and we have tested the UI state for selecting and de-selecting several combinations of subject and subject-like-but-only-a-filter params.

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+        location: { placeMode: EXPLORE_V2_PLACE_MODE.WORLDWIDE },
+      };
+    case "SET_SORT":
+      return { ...draft, sortBy: action.sortBy };
+    // Within the scope of advanced search, we only set a taxon as the subject
+    // and all other subjects go in the filter blob
+    // and if a non-taxon subject came in initially, it has to be excluded via withoutSubjectOfType
+    case "SET_USER":
+      return {
+        ...withFilters( draft, { user: action.user, excludeUser: null } ),
```
jtklein: Potential question: Do we need the entire object here or only the ID (same for project)

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
jtklein: Doesn't the comment of "only taxon is a subject" mean that when calling SET_PROJECT we also need to remove a potential user subject, and here we need to remove a project subject?
Or could we just remove the subject all together in initializing the Advanced Search?

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
```diff
@@ -0,0 +1,233 @@
...
+          ? true
+          : draft.filters.casual,
+      } );
+    case "SET_REVIEWED":
+      return withFilters( draft, { reviewedFilter: action.reviewedFilter } );
+    case "SET_PHOTO_LICENSE":
+      return withFilters( draft, { photoLicense: action.photoLicense } );
+    case "RESET":
+      return {
+        subject: null,
```
jtklein: Is that true? If we start and Advanced Search session with a subject in universal search and hit RESET do we want the subject be removed?

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
+  const hasError = observedEndBeforeStart || uploadedEndBeforeStart || noMonthsSelected;
 
   return (
-    <ViewWrapper testID="AdvancedSearch">
-      <View className="flex-1 items-center justify-center p-4">
-        <Body2>{t( "ADVANCED-SEARCH" )}</Body2>
-      </View>
-    </ViewWrapper>
+    <TopAndBottomInsetViewWrapper testID="AdvancedSearch">
+      <SearchHeader
```
jtklein, line 300: Don't we have to wire up resetDisabled when initial state hasn't changed yet?

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
+        headerText={t( "ADVANCED-SEARCH" )}
+        onClose={() => navigation.goBack()}
```
jtklein: Haven't looked at the V2 tickets too closely. But comparing this to V1 don't we also want to copy over the WarningSheet? Meaning, if a user selects a bunch of filters here, but presses the back arrow instead of the SEARCH button we'd need to warn that all those filters will not be applied this way.

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
+                } else {
+                  filterByIconicTaxonUnknown();
+                }
+              } else if ( taxon?.name?.toLowerCase() === taxonName ) {
+                updateTaxon( null );
+              } else {
+                const selectedTaxon = iconicTaxa
+                  ?.filtered( "name ==[c] $0", taxonName );
+                const iconicTaxon = selectedTaxon && selectedTaxon.length > 0
+                  ? Taxon.mapRealmToPojo( selectedTaxon[0] )
```
jtklein, line 353: Nice!

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
+const ALL_MONTHS = new Array( 12 ).fill( 0 ).map( ( _, i ) => i + 1 );
+
+const isoToday = ( ) => formatObsFieldDate( new Date( ) );
 
 const AdvancedSearch = ( ) => {
-  const { t } = useTranslation( );
+  const navigation
+    = useNavigation<ExploreStackScreenProps<"AdvancedSearch">["navigation"]>();
+  const { t } = useTranslation();
+  const iconicTaxa = useIconicTaxa();
```
jtklein, line 84: Nice!

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
+
+  const subjectUser: ApiUser | null = subject?.type === "user"
+    ? subject.user
+    : null;
+  const displayUser = user || excludeUser || subjectUser;
+  const subjectProject: ApiProjectSummary | null = subject?.type === "project"
+    ? subject.project
+    : null;
+  const displayProject = project || subjectProject;
+  const displayUserCountText = displayUser?.observations_count === undefined
```
jtklein: That seems like defensive overkill to me. If we don't have a user we wouldn't have to display the text. A user from the API should have an observations_count.

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -1,17 +1,761 @@
...
+              dropdown
+              onPress={() => {
+                setOpenSheet( "photoLicense" );
+              }}
+              accessibilityLabel={t( "View-photo-licensing-info" )}
+            />
+          </View>
+        </View>
+      </ScrollView>
+      <ButtonBar>
```
jtklein: UI for this is not correct: there is no drop shadow and there is too much space at the bottom.
For a fix, I think you could copy the ButtonBar styling from the UniversalSearch screen, right?

jtklein (changes requested):
Only a few minor requested changes inline.

The bigger review: I have also taken some time to compare ExploreContext to the new advancedSearchReducer plus ExploreV2Context. From what I can see most funcionality has been ported successfully. One thing to point out though:

V2 has no snapshoting and discarding back to.
When building Explore V1 the ask from product was that after pressing the “Accept” buttons in Advanced filters the results ideally should not need to load visibly but are already fetched in the background. That’s why all that apparatus with `makeSnapshot`, `differsFromSnapshot`, `isNotInitialState`, `discardChanges` exists. So, contrary to V2 where you make a draft and commit on “Accept” which means we will see a loading state when coming back. V1 was only closing the filters modal on “Accept” to the results screen with already loaded results. Which then in turn necessitated the machinery to make a snapshot of the state before entering the FilterModal, so that on "Discard" we can revert back to said snapshot. There is a trade-off between results delay and number of API calls made which was decided in the past by product. I haven't looked at all ExploreV2 tickets, so I am not sure if this change now is a decided change to be made. Anyways, I am not opposed to this way, the state flow is a lot easier to comprehend. So, if we prefer this way I am fine with it. If that has not come up yet in discussion with Product, we should at least keep them informed about the change.

When setting a place, we used to store place_guess as well. Do you think this is fully taken care of with place.display_name?

Very grateful for your mapping of realm objects to plain JS objects.

Unrelated to this code, but can we rename 
buildExploreV2QueryParams
  viewerId to currentUserId

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
sepeterson: Yes I agree. Removing non-taxon subject on init lets us remove these

sepeterson:
Thanks for the review! Regarding the background loading, I hadn't even noticed that and will confirm with Abhas the next time I meet w/ him.

> When setting a place, we used to store place_guess as well. Do you think this is fully taken care of with place.display_name?

I believe so but there is a confusing naming conflict here. It looks like place_guess in explore is just a display string derived from display_name and then deleted at mapParamsToAPI.js:120

> Unrelated to this code, but can we rename buildExploreV2QueryParams viewerId to currentUserId

Done 👍

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
sepeterson: I believe we are safe to drop it. All the dates that make it here are from `formatObsFieldDate( date )` where date is a Date instance

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
sepeterson: Yes the type imports are only used as types and the normal import enums have a value used in the JS side (DATE_OBSERVED.ALL, etc)

src/components/Explore/ExploreV2/helpers/advancedSearchReducer.ts
sepeterson: I think so. V1 clears the subject even if you enter with one

sepeterson: The taxon, I should say

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
sepeterson: Yes good idea, I wired it to be disabled only if state matches the fixed default state under RESET

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
sepeterson: Probably a good idea. Easier in V2 since we don't have to worry about snapshots. I just set it to add the extra modal step if differsFromInitial is true

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
sepeterson: Yep I was able to remove the ternary by moving it inside the JSX block where displayUser is no longer potentially falsy

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
sepeterson: Yep, silly oversight by me. Should be fixed now

sepeterson:
Okay that should be everything addressed!

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
jtklein: Ah, yes, better given the fact that RESET should remove subject as well, see above.

jtklein (approved):
All your changes look good and address my previous requested changes.

I did find one more thing now re-reviewing. It doesn't seem to be introduced by the new changes but I guess I have missed it the first time. When opening a bottom sheet it should be attached to the bottom of the screen, on AdvancedSearch it is wrongly attaching to the top of the tabs. I don't think this is a blocker for this PR and we can fix forward in another ticket.

Similarly, the DATE UPLOADED bottom sheet are covered by the UI around the insertion point. This is in my view a result of one of the previous PRs that refactored the sheets to be inside the component. So, also not to fix here but in a follow-up I guess.

Image

Image

sepeterson:
Thanks @jtklein I just needed to rm the insideModal props since this isn't a modal

PR #3940, opened 2026-08-13
MOB-1672 defer realm clear on logout to fix crash

FLGMwt:
closes MOB-1672

The "invalidated realm" issue seems to happen all the time now on Log Out (I think it used to be dependent on what you'd navigated to before logging out). It also now results in an unrecoverable native error instead of the js error that we could wrap in the Restart screen. We still have things to figure out with MOB-139 / MOB-1601. But this at least smooths over the user-facing problem.

Before:
On signOut, we delete Realm, but somewhere right after that before the RNRestart() is able to complete, Realm errors out because somewhere (callbacks, React Query, async stuffs?) we hold references to those deleted Realms.

After:
On signOut, we set a mmkv flag to delete realm on next "startup". This is explicitly done in `index.js` before anything has a chance to setup / use Realm and create a "invalid" object link.


FLGMwt:
@jtklein curious your thoughts on this since you last worked on reworking signOut around this issue. 

I tried an idea a while ago to navigate to an empty `/logout` navigator / screen & `Mortal` the rest, but of course that still didn't solve callbacks / async with references.

jtklein (commented):
I am not a big fan of this approach. What if sth goes wrong between logout and re-open. Can we end up in a state where the realm delete on re-open does not happen, and end up with two realm files.
I prefer the previous doing one thing after the other in line. At least the realm delete is the last step, so that after the crash in UI we don't expect anything other than a restart to happen.

FLGMwt:
Closing per feedback, thanks @jtklein!

With realm/typescript changes I'm stewing on, I feel better about us maybe finally tracking down our realm handles that are causing this reliable error / crash.

PR #3942, opened 2026-08-14
Remove unused code

jtklein:
Was porting this logic to Seek. These fct are not imported anywhere it seems to me.

@FLGMwt not sure if you had planned to wire them up at one point.


FLGMwt (approved):
ah, I probably should have done initially but if we're not seeing obvious issues from legacy log files after all this time, I approve deleting them.

EDIT, clarification: they _were_ previously used, I just missed deleting these helpers when I deleted the consuming code

PR #3943, opened 2026-08-14
MOB-1346 Advanced search tests

sepeterson:


tests/unit/components/Explore/ExploreV2/helpers/advancedSearchReducer.test.js
```diff
@@ -0,0 +1,498 @@
...
+    expect( draft.subject ).toBeNull( );
+    expect( draft.filters ).toEqual( defaultExploreV2Filters );
+  } );
+
+  it( "keeps an unknown subject", ( ) => {
+    const draft = draftFromV2State( makeV2State( { subject: { type: "unknown" } } ) );
+
+    expect( draft.subject ).toEqual( { type: "unknown" } );
+  } );
+} );
```
abbeycampbell: I would argue that all of this is also tested in `AdvancedSearch.test,js`

tests/unit/components/Explore/ExploreV2/helpers/advancedSearchReducer.test.js
```diff
@@ -0,0 +1,498 @@
...
+      expect( newDraft.subject ).toEqual( { type: "taxon", taxon: TAXON } );
+    } );
+
+    it( "clears the subject when the taxon is removed", ( ) => {
+      const draft = makeDraft( { subject: { type: "taxon", taxon: TAXON } } );
+
+      const newDraft = advancedSearchReducer( draft, { type: "SET_TAXON", taxon: null } );
+
+      expect( newDraft.subject ).toBeNull( );
+    } );
```
abbeycampbell: maybe also already covered by the screen's remove-taxon and iconic taxon tests?

tests/unit/components/Explore/ExploreV2/helpers/advancedSearchReducer.test.js
```diff
@@ -0,0 +1,498 @@
...
+    it( "includes casual observations when filtering for captive", ( ) => {
+      const newDraft = advancedSearchReducer( makeDraft( ), {
+        type: "SET_WILD_STATUS",
+        wildStatus: WILD_STATUS.CAPTIVE,
+      } );
+
+      expect( newDraft.filters.wildStatus ).toEqual( WILD_STATUS.CAPTIVE );
+      expect( newDraft.filters.casual ).toBe( true );
+    } );
+
```
abbeycampbell: this one is also the same as the screen test at line 312

abbeycampbell (approved):
looks mostly good to me. I would say there is some duplication going on between the reducer tests and the screen tests, that I will leave up to you whether or not to address since it's not harmful. 

imo this kinda thing is more likely now that tests are fast and cheap to produce; we get a lot more coverage but run the risk of over-testing/testing the same things multiple times. I'm not even sure how much we care? larger discussion for another day maybe

sepeterson:
> looks mostly good to me. I would say there is some duplication going on between the reducer tests and the screen tests, that I will leave up to you whether or not to address since it's not harmful.

Thanks for pointing that out! Probably harmless but I am removing them bc I wouldn't have included them if I had noticed

PR #3945, opened 2026-08-14
Mob 1589 myobs state should persist during a user session

abbeycampbell:
handles the sort, search, and map view parts of [MOB-1589](https://linear.app/inaturalist/issue/MOB-1589/myobs-state-should-persist-during-a-user-session). Expanded and collapsed header state for the grouped by iconic taxa view will follow once that work is actually done/wired up.

This moves MyObs state into a new `createMyObservationsSlice`, and folds in the old `createMyObsSlice` dropping two dead count fields nothing called.


src/stores/createMyObsSlice.ts
```diff
@@ -1,16 +0,0 @@
-const createMyObsSlice = ( set, get ) => ( {
-  // Stores y offset of MyObs so we can scroll back to the user's position
-  // when returning from the NoBottomTabStackNavigator (MyObs will be trashed
-  // when navigating to ObsEdit, so we lose scroll position)
-  myObsOffset: 0,
-  myObsOffsetToRestore: 0,
-  resetMyObsOffsetToRestore: ( ) => set( { myObsOffsetToRestore: 0 } ),
-  setMyObsOffset: newOffset => set( { myObsOffset: newOffset } ),
-  setMyObsOffsetToRestore: ( ) => set( { myObsOffsetToRestore: get( ).myObsOffset } ),
-  numOfUserObservations: 0,
```
jtklein, line 10: Only the setters seem unused to me. The two numbers are still used, right?
https://github.com/inaturalist/iNaturalistReactNative/blob/1b1e7189fea5a4123ba23a97f34ba18712e57316/src/components/MyObservations/MyObservationsResults.tsx#L357-L358

src/stores/createMyObservationsSlice.ts
```diff
@@ -0,0 +1,90 @@
+import type { Region } from "react-native-maps";
+import type { ICONIC_TAXA_GROUP } from "sharedHelpers/iconicTaxaGroupOrder";
+import { OBSERVATIONS_SORT } from "sharedHelpers/observationsSort";
+import { SPECIES_SORT } from "sharedHelpers/speciesSort";
+import type { StateCreator } from "zustand";
+
+// Everything My Observations needs to remember for the length of a session:
+// sort, search, view state, and scroll position. It lives here rather than
+// in React because the MyObs component tree gets destroyed constantly: Mortal
```
jtklein, line 9: Both the Mortal and resetting navigation where earlier workarounds of trying to cope with the performance problems. So, this does feel a but like a patch on top of a patch. Is there a point where we can get back to rely on navigation default, i.e. MyObs staying intact in the navigation history with local state?

src/components/MyObservations/MyObservationsResults.tsx
```diff
@@ -84,6 +85,8 @@ const MyObservationsResults = ( ) => {
   const startAutomaticSync = useStore( state => state.startAutomaticSync );
   const myObsOffsetToRestore = useStore( state => state.myObsOffsetToRestore );
   const setMyObsOffset = useStore( state => state.setMyObsOffset );
+  const clearMyObservationsViewState: MyObservationsSlice["clearMyObservationsViewState"]
```
jtklein, line 88: The `MyObservationsSlice["clearMyObservationsViewState"]` here and similar in other places is not really needed if the slice is typed? I have not used this in other places and inference seems fine.

src/stores/createMyObservationsSlice.ts
```diff
@@ -0,0 +1,90 @@
...
+      return { myObservations };
+    }
+    // TBD: what to do re: expanded header set when a search becomes active.
+    // do we clear them, or should they re-appear as they were before?
+    return { myObservations, myObservationsMapRegion: null };
+  } ),
+
+  myObservationsMapRegion: null,
+  setMyObservationsMapRegion: region => set( { myObservationsMapRegion: region } ),
+  myObservationsClosedIconicTaxaCategories: new Set( ),
```
jtklein, line 84: This is not wired up on purpose?

src/stores/createMyObservationsSlice.ts
```diff
@@ -0,0 +1,90 @@
...
+  myObservations: MyObservationsState;
+  updateMyObservations: (
+    _updater: ( _previous: MyObservationsState ) => MyObservationsState
+  ) => void;
+
+  // State that only relevant while its view is the active view. Kept out
+  // of myObservations because the map writes on every pan, and everything
+  // reading myObservations would re-render along with it. Cleared when the user
+  // switches views.
+  myObservationsMapRegion: Region | null;
```
jtklein, line 45: Okay, so this is not relevant to the AC in the ticket, but rather with the discussion in the comment section.
I fear this approach might bloat into a bunch of edge-cases and navigation complexities. A lá "if you navigate from here to here under this circumstances then don't reset the region, but if you have changed this one thing only then do reset". But I guess we will see.

jtklein (changes requested):
The refactor from context to zustand slice seems sound to me.
Requested changes because I don't think we can remove numOfUserObservations and numOfUserSpecies.

Haven't looked at tests here because of time crunch.

P.S.: This was more painful to review as it needs to be: smaller commits make a big difference, letting me know which feature flags I need to turn on in the PR description, detailed notes of which pieces of code are merely copied over into another file, are all things that could have saved some time I think.

src/stores/createMyObservationsSlice.ts
FLGMwt: I looked into this a bit yesterday: I think the answer is yes, but it seems out of scope for MyObs. MyObs is now adapting the same zustand/Mortal/BottomTab context pattern necessary for Explore now that it has similar functionality.

I think we'll need a reckoning with nav / context / Mortal, but that probably needs to happen holistically along with Explore.

src/stores/createMyObsSlice.ts
FLGMwt: I think I know what happened here: we _were_ previously reading this `numOfUserObservations` from zustand in one or two places that now use derived data. What's left is exclusively mmkv direct access to these two values. So I think removing the setters and properties from the zustand slice reflects our actual use. 

MOB-1374 (subtask of MOB-1372, "fix improper mmkv access") is responsible for migrating those direct mmkv accesses out. That ticket would likely _reintroduce_ those setters / values to zustand but only once the access patterns are ironed out.

abbeycampbell: That makes sense. Claude told me they were not used, so I opted to remove them to keep the new slice file clean. I will add them back, so we can be back in the state we would have been in for MOB-1374

src/stores/createMyObservationsSlice.ts
abbeycampbell: I thought I was going to have the small grid view done before finishing this ticket, so I was preparing to include the needed changes for that. I'll remove for now.

src/stores/createMyObservationsSlice.ts
abbeycampbell: > a point where we can get back to rely on navigation default

I would love this. I believe you are correct, but Ryan is also correct in that this is out of scope for the MyObs work

src/stores/createMyObservationsSlice.ts
abbeycampbell: Yeah, I was using the ticket as a place to work out with product what the persistence needs were exactly, with the intention of executing them all at once. I should have updated the ticket itself with all persistence asks. I hear you as far as edge cases go, but I think it will be ok to address anything that comes up in follow-up tickets.

src/stores/createMyObservationsSlice.ts
abbeycampbell: I actually decided I want to leave this in, since the small grid view should be wired up this week. But I added a TODO, to acknowledge it's not currently used.

src/components/MyObservations/MyObservationsResults.tsx
abbeycampbell: I think you're right. I'll investigate this in a follow-up/when I actually wire the small-grid view. If that's cool with you.

jtklein (approved):
Thanks for putting the two zustand slice states back in. I don't see the connection to mmkv yet but LGTM.

PR #3948, opened 2026-08-17
Mob 1679 show banners and announcements in the myobs map view

abbeycampbell:
closes [MOB-1679](https://linear.app/inaturalist/issue/MOB-1679/show-banners-and-announcements-in-the-myobs-map-view)

Small change that would probably also be nice to include in the MyObs release so anyone that has the map view as their preferred view isn't missing announcements. I decided the banners are pretty meaningless in this view, so we care about announcements only; more context in ticket. There's still one open product question, but I think it would be fine to include this as is and address in a follow-up if we do want to make the map view scrollable.

### DIFF
--- a/src/components/MyObservations/Announcements.tsx
+++ b/src/components/MyObservations/Announcements.tsx
@@ -46,11 +46,15 @@ type WebshellProps = ComponentProps<typeof Webshell>;
 
 const AutoheightWebView
 = ( webshellProps: WebshellProps ) => {
-  const { autoheightWebshellProps } = useAutoheight( {
+  const { autoheightWebshellProps, contentSize } = useAutoheight( {
     webshellProps,
   } );
-  // eslint-disable-next-line react/jsx-props-no-spreading
-  return <Webshell {...autoheightWebshellProps} />;
+  return (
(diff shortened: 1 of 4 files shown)


sepeterson (approved):
Looks good!

<details>
<summary>screenshot</summary>

Image

</details>

src/components/MyObservations/Announcements.tsx
```diff
@@ -46,11 +46,15 @@ type WebshellProps = ComponentProps<typeof Webshell>;
...
     webshellProps,
   } );
-  // eslint-disable-next-line react/jsx-props-no-spreading
-  return <Webshell {...autoheightWebshellProps} />;
+  return (
+    <View style={{ height: contentSize.height }}>
+      {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+      <Webshell {...autoheightWebshellProps} />
+    </View>
+  );
```
FLGMwt, line 57: mentioning just in case we need to revert to OldMyObs, I have confirmed that this _does_ still work with OldMyObs

FLGMwt (commented):
LGTM, working as expected, but does seem like it would be annoying with a non-dismissible announcement.

PR #3949, opened 2026-08-18
MOB-1574 fix judder during navigation to Taxon Search

FLGMwt:
SuggestionsScreen => TaxonSearchScreen judders during the screen transition as the latter slides in. This seems to be because the component is rendered w/ the text input autoFocused necessarily before the screen transition starts. We can instead wait until the screen transition is finished and _then_ focus to raise the keyboard once we're settled.

https://github.com/user-attachments/assets/6627eab3-51af-45dd-9513-b9150fc96474

note use of useEffectEvent here: to keep the previous logic, we need to check `query`, but we _don't_ want it to be an effect dependency for setting up the listener.

### DIFF
--- a/src/components/Match/MatchTaxonSearchScreen.tsx
+++ b/src/components/Match/MatchTaxonSearchScreen.tsx
@@ -78,6 +78,7 @@ const MatchTaxonSearchScreen = ( ) => {
 
   return (
     <TaxonSearch
+      focusAfterTransition
       isLoading={isLoading}
       isLocal={isLocal}
       query={taxonQuery}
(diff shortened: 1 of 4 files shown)


abbeycampbell (changes requested):
Can you add a `focusAfterTransition` prop to the `TaxonSearch` in `SearchMyObservationsTaxon`? then we're good ;)

FLGMwt:
> Can you add a `focusAfterTransition` prop to the `TaxonSearch` in `SearchMyObservationsTaxon`? then we're good ;)

ty! fixed

abbeycampbell (approved):
thanks!

PR #3950, opened 2026-08-18
MOB-1406 ExploreV2 recent searches

sepeterson:
Show 10 most recent searches for both subject and location fields in universal search. Searches are stored in a new zustand slice and persisted.

Also, exhaustive unit tests

<details>
Simulator Screenshot - iPhone 17 Pro - 2026-08-18 at 17 14 12

Simulator Screenshot - iPhone 17 Pro - 2026-08-18 at 20 48 31

</details>

### DIFF
--- a/src/components/Explore/ExploreV2/components/RecentSearches.tsx
+++ b/src/components/Explore/ExploreV2/components/RecentSearches.tsx
@@ -1,0 +1,53 @@
+import UniversalSearchResult
+  from "components/Explore/ExploreV2/components/UniversalSearchResult";
+import { subjectToResult }
+  from "components/Explore/ExploreV2/helpers/universalSearchSubject";
+import type { UniversalSearchResultItem }
+  from "components/Explore/ExploreV2/hooks/useUniversalSearch";
+import { View } from "components/styledComponents";
+import type { ExploreV2Subject } from "providers/ExploreV2Context";
+import React from "react";
+import useCurrentUser from "sharedHooks/useCurrentUser";
+import type { ExploreRecentSearchesSlice } from "stores/createExploreRecentSearchesSlice";
+import { subjectKey } from "stores/createExploreRecentSearchesSlice";
+import useStore from "stores/useStore";
+
+interface Props {
+  onSelectSubject: ( _subject: ExploreV2Subject ) => void;
+}
+
+interface RecentRow {
+  subject: ExploreV2Subject;
+  result: UniversalSearchResultItem;
+}
+
+const RecentSearches = ( { onSelectSubject }: Props ) => {
+  const currentUser = useCurrentUser( );
+  const subjects = useStore(
+    ( state: ExploreRecentSearchesSlice ) => state.exploreRecentSearches.subjects,
+  );
+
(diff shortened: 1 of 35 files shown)


src/components/Explore/ExploreV2/components/RecentSearches.tsx
```diff
@@ -0,0 +1,53 @@
...
+const RecentSearches = ( { onSelectSubject }: Props ) => {
+  const currentUser = useCurrentUser( );
+  const subjects = useStore(
+    ( state: ExploreV2RecentSearchesSlice ) => state.exploreRecentSearches.subjects,
+  );
+
+  const rows = subjects
+    .map( subject => ( { subject, result: subjectToResult( subject ) } ) )
+    // Only subjects that have a search-result form get a row (no "unknown" or "unobserved")
+    .filter( ( row ): row is RecentRow => row.result !== null )
```
sepeterson, line 33: This step is a little annoying because it's not really functional but narrows the type. Left it in bc it should be harmless and the upstream type fix was getting out of hand

src/components/Explore/ExploreV2/components/RecentSearches.tsx
```diff
@@ -0,0 +1,53 @@
...
+    .filter( ( { subject } ) => !(
+      subject.type === "user" && subject.user.id === currentUser?.id
```
sepeterson, line 35: It's a little silly but I do think this is necessary. One would think you could just filter out currentUser on the add but I intentionally made this feature to persist results through user/login state so I do think it's correct to filter here

jtklein: Not sure what you mean by "persist results through user/login state" the persisted-zustand get's wiped on logout here: https://github.com/inaturalist/iNaturalistReactNative/blob/f6c055b98844f22f95fbfe7f591de4cc1d211b82/src/components/LoginSignUp/AuthenticationService.ts#L267-L268

What does work well is: when logged out search for myself, then log in -> I do see my login only at the top and not again a second time at the recent search results.

When I logout all searches are cleared.

src/stores/useStore.js
```diff
@@ -79,6 +81,10 @@ const useStore = create( persist(
       // Vestigial un-namespaced values in the layout slice
       isAdvancedUser: state.isAdvancedUser,
       obsDetailsTab: state.obsDetailsTab,
+      exploreRecentSearches: {
```
jtklein: Could you move this below layout? I think the ordering is more a historical one here.

jtklein (approved):
Works great! Not sure if I understand the one comment, but in testing the results work as I'd expect to see them.

Haven't looked at tests.

src/components/Explore/ExploreV2/components/RecentSearches.tsx
sepeterson: > Not sure what you mean by "persist results through user/login state" the persisted-zustand get's wiped on logout here:

Yep good catch you're totally right. That step's actually just supporting the logged out w/ recent searches -> logged in flow

PR #3954, opened 2026-08-19
Replace helper fct result with direct lat long check

jtklein:
Closes MOB-1686

Another victim of the RealmObs vs UI = PojoObs divide that is untyped. When starting a new obs by importing a photo we do not have a realm representation yet and do not have the result of the missing_coords fct check in the object. The true culprit of the bug was me applying poor coding standards though, switching from a state that checks for false = !hasLocation to a true = missingCoords, so that seeing an undefined was flowing the wrong path.

### DIFF
--- a/src/components/ObsEdit/EvidenceSection.js
+++ b/src/components/ObsEdit/EvidenceSection.js
@@ -57,7 +57,7 @@ const EvidenceSection = ( {
 
   const latitude = currentObservation?.privateLatitude || currentObservation?.latitude;
   const longitude = currentObservation?.privateLongitude || currentObservation?.longitude;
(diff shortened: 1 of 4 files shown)


FLGMwt (commented):
LGTM in behavior

does this mean we're rejecting the recently introduced realm property for this question?

jtklein:
>does this mean we're rejecting the recently introduced realm property for this question?

No. The property is still there and still used. It was never present in an obs started in ObsEdit which starts as a zustand UI POJO before it get's saved to Realm. So, for some flows the property was undefined and for some flows it was either true or false. It is still used in Realm-only flows.

PR #3955, opened 2026-08-20
MOB-1653 reintroduce max age for Android location checks

FLGMwt:
MOB-1653

I haven't been able to reproduce the stale loc issue, but this seems to be the cause of Tony's issue: I too eagerly removed a maxAge Android "fix" when I last refactored useWatch / geolocation. That wasn't important for the listener but it _does_ apparently impact the cache age of subsequent `getLastLocation` calls.


jtklein (approved):
Code LGTM

Haven't been able to test this the same way the ticket outlines the bug in the app.

PR #3957, opened 2026-08-20
MOB-1348 Explore V2 Advanced search mode

sepeterson:
Has no relation to the existing advanced/default mode split.

This just says that after a user applies AdvancedSearch, their app shows the filters button and navigates to AdvancedSearch from ExploreResults until the next session, with the reset being a function of the slice not persisting.

Seen in the 1st screenshot, it's a bit visually confusing when there's no taxa but a user and/or project from AdvancedSearch. I'm going to discuss this w/ Abhas on Tuesday.

<details>
Simulator Screenshot - iPhone 17 - 2026-08-20 at 15 32 30
Simulator Screenshot - iPhone 17 Pro - 2026-08-20 at 14 58 59

</details>


src/stores/createExploreV2AdvancedSearchSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+export interface ExploreV2AdvancedSearchSlice {
+  exploreV2AdvancedSearch: {
+    advancedSearchMode: boolean;
+    setAdvancedSearchMode: ( _advancedSearchMode: boolean ) => void;
+  };
+}
+
+// advancedSearchMode has to be global state so it stays
+// between different instances of the exloreV2 navigator
+const createExploreV2AdvancedSearchSlice: StateCreator<ExploreV2AdvancedSearchSlice> = set => ( {
```
jtklein, line 12: At this point in the project this might as well be ExploreV2Slice entirely?

src/stores/createExploreV2AdvancedSearchSlice.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface ExploreV2AdvancedSearchSlice {
+  exploreV2AdvancedSearch: {
+    advancedSearchMode: boolean;
+    setAdvancedSearchMode: ( _advancedSearchMode: boolean ) => void;
+  };
+}
+
+// advancedSearchMode has to be global state so it stays
+// between different instances of the exloreV2 navigator
```
jtklein, line 11: I have seen that in testing but wanted to flag it: The advanced search mode should stay between different Explore screens. So, navigating to Advanced Filters, Taxon Details, opening another Explore screen in the stack should have it be in Advanced Search Mode?

src/components/Explore/ExploreV2/helpers/countFilters.ts
```diff
@@ -0,0 +1,28 @@
...
+  ...defaultExploreV2Filters,
+  user: null,
+  excludeUser: null,
+  project: null,
+};
+
+type CountedKey = keyof typeof countedDefaults;
+
+// compare against default filters and count differences
+function countFilters( filters: ExploreV2Filters ): number {
```
jtklein, line 14: Could you double-check we still need this feature in ExploreV2, then I can review this file. We did have a more detailed calculation of the "number" of filters in the past, some things counted others didn't. Before I check I'd like to make sure we still need this.

src/components/Explore/ExploreV2/components/ExploreV2Header.tsx
```diff
@@ -204,19 +216,36 @@ const ExploreV2Header = ( ) => {
...
+              ? t( "Filters" )
+              : t( "Search" )}
+            backgroundColor={colors.inatGreen}
+            icon={advancedSearchMode
+              ? "sliders"
+              : "magnifying-glass"}
+            onPress={() => navigation.navigate( searchScreen )}
+            testID="ExploreV2Header.searchButton"
+          />
+          {filterCount > 0 && (
```
jtklein, line 237: I thought we are getting rid of the NumberBadge in Explore V2? Can you double-check with Product? I don't see it in the designs anymore. (Apart from the text description what counts as an integer increase of this number).

jtklein (approved):
As the code is now I would Approve. Even though I have my doubts we still need the number badge I am certain it will be a pretty niche feature, so for my taste we can move fast with it.

src/stores/createExploreV2AdvancedSearchSlice.ts
sepeterson: I am meeting w/ Abhas tomorrow and will confirm then

src/stores/createExploreV2AdvancedSearchSlice.ts
sepeterson: There's also createExploreV2RecentSearchesSlice which needs to be separate since it's persisted and this one doesn't so I think it make sense to have more specific names

src/stores/createExploreV2AdvancedSearchSlice.ts
sepeterson: Confirmed expected behavior

src/components/Explore/ExploreV2/helpers/countFilters.ts
sepeterson: I took some creative liberty and added the number badge because for sometimes it was unclear that you had anything applied beyond a location if it wasn't a typical subject. Confirmed with Abhas today that this is fine

PR #3958, opened 2026-08-20
Traditional projects: on SAVE field validation gates

jtklein:
Closes MOB-1509

Summary
- Gate AddToProjects SAVE with C2 required-field validation; show inline pass/fail per project row
- Invalid SAVE → Missing info sheet (LEAVE / KEEP EDITING); LEAVE keeps only completed projects and their OFVs

Not changed
- Back / swipe-back → plain pop, no commit, no sheet (only SAVE persists selection)
- Back-navigation parity deferred to [MOB-1691](https://linear.app/inaturalist/issue/MOB-1691) (C3d)

Test plan
- Toggle projects → SAVE → ObsEdit shows updated project count
- Select project with empty required field → SAVE → sheet; KEEP EDITING stays; LEAVE saves completed projects only
- Change selection → back without SAVE → ObsEdit unchanged


sepeterson (approved):
Looks good!

<details>
<summary>screenshot</summary>

Image

</details>

PR #3959, opened 2026-08-20
update @shopify/flashlist to 2.3.2

abbeycampbell:
[MOB-1677](https://linear.app/inaturalist/issue/MOB-1677/upgrade-shopifyflash-list-220-232)

Did a bunch of testing on iOS and android and didn't notice regressions. Fast scrolling MyObs lists on android is a pretty terrible experience, but it's like that on main as well.


jtklein (approved):
Have tested most if not all surfaces of flashlist in the app in Debug and Release mode and haven't found any issues. On MyObs it does look much better and I haven't seen any vanishing cells on my device.

PR #3963, opened 2026-08-21
Fix: not importing location from photos in android

jtklein:
Closes MOB-1656 (for now)

Fix Android gallery imports dropping photo GPS by requesting media-location permission and preserving EXIF when the image picker copies photos (react-native-image-picker PR #2365 copyExif). Also extracts react-native-permissions helpers into sharedHelpers/permissions.ts.

Asking for read-write permission before launching the image picker solves this for me on my device.

I am not an expert in the react-native-image-picker code, and so I am not even sure we need the additional code in the patch. I took it from a PR to the library that wants to solve the same issue but I have not done any research if that is the best solution for this in the library scope. I'd prefer to update the image-picker library to keep up with these things but it hasn't seen any new releases in a year.

We still need the force-old picker boolean because PhotoPicker does not provide EXIF data from what I have seen.


src/components/SharedComponents/PermissionGateContainer.tsx
```diff
@@ -9,71 +8,20 @@ import React, {
...
   Permission,
   PermissionStatus,
 } from "react-native-permissions";
 import {
   checkMultiple,
-  PERMISSIONS,
   requestMultiple,
   RESULTS,
 } from "react-native-permissions";
+import { permissionResultFromMultiple } from "sharedHelpers/permissions";
```
jtklein, line 21: Extracted all permission related helpers into their own file to decouple from this UI logic here.

src/sharedHelpers/permissions.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export async function requestWriteMediaPermission( ) {
+  if ( WRITE_MEDIA_PERMISSIONS.length === 0 ) return true;
+  const result = permissionResultFromMultiple(
+    await requestMultiple( WRITE_MEDIA_PERMISSIONS ),
+  );
+  return result === RESULTS.GRANTED;
+}
+
+export async function requestReadWriteMediaPermissions( ) {
```
jtklein, line 136: This fct is the only thing that is new. Rest is copied over from PermissionGate

patches/react-native-image-picker+8.2.1.patch
```diff
@@ -84,6 +84,101 @@ index 21970d5..b1cd9e9 100644
...
++            ExifInterface.TAG_MAKE,
++            ExifInterface.TAG_MODEL,
++            ExifInterface.TAG_SOFTWARE,
++            ExifInterface.TAG_ARTIST,
++            ExifInterface.TAG_GPS_TIMESTAMP,
++            ExifInterface.TAG_LENS_MODEL,
++            ExifInterface.TAG_F_NUMBER,
++            ExifInterface.TAG_SUBSEC_TIME,
++            ExifInterface.TAG_SUBSEC_TIME_DIGITIZED,
++            ExifInterface.TAG_SUBSEC_TIME_ORIGINAL
```
sepeterson, line 161: We might consider adding `ExifInterface.TAG_DATETIME_ORIGINAL` and `ExifInterface.TAG_GPS_H_POSITIONING_ERROR` because those both can be read in parseExif.ts.

jtklein: Thanks! That's a great find, I have now added those two tags.

patches/react-native-image-picker+8.2.1.patch
sepeterson, line 57: On my emulator I have two options for photo upload, "Media picker" and "Photos"

<details>
<summary>screenshot</summary>

Image

</details>

I'm guessing Media picker is the new version you mentioned in standup bc any photos I choose via that route don't import exif locations. If I choose Photos, it works fine.

What do you think about removing line 57 of the patch here? If I do that, it doesn't show Media picker as an option and just kicks me straight into Photos.

sepeterson, line 38: Oh but this means android 13+ can only select a single photo via the "old" route?

jtklein: I don't want to change the patch as much anymore. I do see the same thing on my device: Left does not give EXIF data, right does give EXIF data after the fix. Before none gave EXIF data.
If I enable the app to use the new Android PhotoPicker I get no EXIF after the fix.
So, there is an improvement there compared to before.

At this point I'd just like to keep this patch as is, and turn around to Product and say, listen we probably have to expect the default for an Android user to be that we won't have a location (and datetime) for an imported photo, how should the app look like to accommodate for this?

jtklein: Not sure what you mean by that. I haven't changed this code in this PR.

jtklein: Do you think you have found the reason for https://linear.app/inaturalist/issue/MOB-1645/cant-select-multiple-photos-for-import-from-google-photos?

sepeterson: Sounds good, it's a product decision.

sepeterson: Perhaps, I haven't tested it. It certainly seems to say right there that EXTRA_ALLOW_MULTIPLE is only true if the Android version is < Tiramisu

sepeterson (approved):
Importing from Google Photos works well

PR #3965, opened 2026-08-24
MOB-1121: react-native-screens to 4.25.2

sepeterson:
We can't go to latest right now because it depends on a RN upgrade. Smoke tested on both ios/android. Screenshots on the linear ticket

Seek update here https://github.com/inaturalist/SeekReactNative/pull/1003


jtklein (approved):
LGTM

PR #3966, opened 2026-08-24
Mob 1347 advanced search observers identifiers tabs

sepeterson:
Adds Observers/Identifiers tabs for ExploreV2 advanced search only. Mirrors the useSpeciesCount implementation for showing the counts. The results themselves should be the same as the old views.

<details>
<summary>recording</summary>

https://github.com/user-attachments/assets/54247efe-e5da-4338-bcfb-a8577f04c556

</details>


src/components/SharedComponents/Tabs/Tabs.tsx
```diff
@@ -29,70 +29,123 @@ interface Props {
...
+  const tabWidth = width * SCROLLABLE_TAB_WIDTH_RATIO;
+  const scrollableTabStyle = { width: tabWidth };
+  const scrollViewRef = useRef<ScrollView>( null );
+  const activeIndex = tabs.findIndex( tab => tab.id === activeId );
+  const lastScrolledIndex = useRef<number | null>( null );
+
+  useEffect( ( ) => {
+    if ( !scrollable || activeIndex < 0 ) return;
+    const maxOffset = Math.max( tabWidth * tabs.length - width, 0 );
+    // Center the active tab, or scroll as far as it'll go for the first & last
```
sepeterson, line 62: I took some creative liberty here to scroll the selected tab into view.

src/api/observations.js
```diff
@@ -208,19 +208,25 @@ const deleteRemoteObservation = async (
   }
 };
 
-const fetchObservers = async ( params: Object = {} ) : Promise<?Object> => {
+const fetchObservers = async (
+  params: Object = {},
+  opts: Object = {},
```
sepeterson: Not really sure about adding opts as an arg here. The case for is that fetchObservers and fetchIdentifiers already get called with useAuthenticatedQuery and I'm not aware of a good reason to strip out the api_token if a user is signed in

src/api/types.d.ts
```diff
@@ -4,7 +4,7 @@ export interface ApiDefaultResult {
 }
 
 export interface ApiOpts {
-  api_token?: string;
+  api_token?: string | null;
```
sepeterson, line 7: api_token actually is set as null for many of the explore endpoints when the user is signed out

src/api/observations.js
sepeterson: Also I looked into a ts conversion of this file but I think it'll need its own PR

src/components/SharedComponents/Tabs/Tabs.tsx
jtklein: Looks good.

src/components/Explore/ExploreV2/screens/ExploreV2UsersView.tsx
```diff
@@ -0,0 +1,80 @@
...
+    fetchUsers,
+    {
+      ...params,
+      ...extraParams,
+    },
+    { enabled },
+  );
+
+  return (
+    <ExploreFlashList
```
jtklein, line 67: If we are keeping ExploreFlashList now might be a good time to migrate it to TS. Maybe not this PR though.

src/components/Explore/ExploreV2/screens/ExploreResults.tsx
```diff
@@ -198,6 +218,93 @@ const ExploreResults = ( ) => {
...
+                  testID="ExploreV2ObservationsList"
+                />
+              )}
+            <ObservationsViewBar
+              layout={layout}
+              updateObservationsView={writeLayoutToStorage}
+              viewOptions={["map", "grid", "list"]}
+            />
+            {!showMap && (
+              <SortButton
```
jtklein, line 268: I do prefer having this in two places now.

src/api/observations.js
jtklein: We have started to migrate the single API calls to TS by moving them to `observationsTyped.ts` from here. You could just move these two and re-use the typing blueprint we have setup with `usersTyped.ts`. I'd appreciate if part of this PR but I won't request it as a change.

src/components/Explore/ExploreV2/hooks/useUserTabCounts.ts
```diff
@@ -0,0 +1,45 @@
...
+  identifiersCount: number | null;
+}
+
+const useUserTabCounts = (
+  params: ExploreV2BaseQueryParams,
+  { enabled = true }: Options = {},
+): UserTabCounts => {
+  const countParams = { ...params, per_page: 0, ttl: -1 };
+
+  const { data: observersData } = useAuthenticatedQuery<ApiResponse<object>>(
```
jtklein: Example of how we used a typed API wrapper in a useAuthQuery format: https://github.com/inaturalist/iNaturalistReactNative/blob/e6659beab2168cc3d4fb91502914de096be8ea09/src/components/ProjectList/ProjectListContainer.tsx#L42-L43

src/api/types.d.ts
jtklein: I am unsure about this change. Yes, some API calls can be unauthenticated. But also, some API calls will forever error out if we do send a null here. So, is the point of the interface to allow any possible value, or do we want more fine-grained single endpoint interface? I don't have an answer here.

jtklein (approved):
"New" tabs look good, scrolling is nice! Not sure if this was ready for review because it said the base branch is out of whack but hey, LGTM.

src/api/observations.js
sepeterson: Oh nice! Moved these to observationsTyped since it was quick enough

src/api/types.d.ts
sepeterson: I'm not sure either, this just reflects that we *do* set api_token to null sometimes. I'm not sure if it makes it to the API though or is simply omitted

src/components/Explore/ExploreV2/screens/ExploreV2UsersView.tsx
sepeterson: I'll follow up

sepeterson:
@jtklein Thanks again for the good review!

PR #3968, opened 2026-08-24
Traditional projects: OFV and PO upload

jtklein:
Closes MOB-1510

Adds a project-children upload step to the observation upload pipeline. After media is attached, dirty observation field values (OFVs) and project observations (POs) sync to the API in sequence — OFVs first, then POs — with parallel uploads within each phase. New code is massively following obs media upload step.

Before: One obs project state on staging (website)
Screenshot 2026-08-24 at 16 15 08

After: filling out the "Add to Projects" form in the app and pressing obs sync => POs and OFVs all uploaded
Screenshot 2026-08-24 at 16 19 27
Screenshot 2026-08-24 at 16 19 41

### DIFF
--- a/src/api/observationFieldValues.ts
+++ b/src/api/observationFieldValues.ts
@@ -25,11 +25,11 @@ const createObservationFieldValue = async <T = ApiDefaultResult>(
   opts: ApiOpts = {},
 ): Promise<ApiResponse<T> | null | ErrorWithResponse | INatApiError> => {
   try {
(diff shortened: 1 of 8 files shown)


FLGMwt (approved):
removing from project / PO deletion is _not_ yet handled, right? that's in MOB-1511?

Otherwise, looks good to me! I found a UI issue while testing but filing that separately.

nice how clean this was following the existing sync patterns 😎

PR #3970, opened 2026-08-24
Remove log for experiment of API v2 fields call

jtklein:
We were logging this to understand performance problems, but didn't help find an obvious culprit and wasn't very actionable, so let's get rid of it for now.


FLGMwt (approved):
thank you!!!!!!!

PR #3971, opened 2026-08-24
Update rozenite to 1.13.0

jtklein:
Copying the same update from Seek.

Tested by compiling and running the app in Debug mode on Android and iOS, and using some panels.


FLGMwt (approved):
tested, works for me w/o issues.

I'm excited for the new import/export goodies on the file-system-plugin! 

also, peeked at the upcoming in rozenite and ooooooh new UI in v2 and a standalone app?

PR #3972, opened 2026-08-24
MOB-1698 default initial value for obs-project-field

FLGMwt:
Found while testing #3968

Before: default value only reflected after scrolling in picker

https://github.com/user-attachments/assets/c5c53f50-0eac-4a54-b03a-a487fd0eac7c

After: default value picked up without scrolling. does not "save" default if you dismiss picker by tapping out.

https://github.com/user-attachments/assets/a12154d9-1ba4-451a-993a-fad44a2b54eb


jtklein (approved):
Thanks for this fix! Works as advertised, LGTM

PR #3974, opened 2026-08-25
Traditional projects: deletion sync for POs and OFVs

jtklein:
Closes MOB-1511

Remark for testing: There is a bit of a known bug that only removing an OFV (by removing all text from an input) does not enable the SAVE button. So, you'd need to also select a new project for SAVE to get enabled. This should not affect the delete sync though.

PR includes:
Syncs removal of traditional project links (POs) and field values (OFVs) to the server before create/update uploads.
- Adds: DELETE tombstoned POs, then OFVs; 404/403 treated as success; local Realm embeds removed after sync; skips remote DELETE for never-synced rows
- Calls deleter from observationUploader immediately before uploadProjectChildren
- Extends UNSYNCED_FILTER so observations with pending PO/OFV deletions stay in the upload queue

Before:
Screenshot 2026-08-25 at 11 10 37
Screenshot 2026-08-25 at 11 44 24

After: Removing a project from the Add to Project screen (which was previously uploaded):
Screenshot 2026-08-25 at 11 11 15

After: Removing an observation field value from one of the inputs of a project which was previously synced to server:
Screenshot 2026-08-25 at 11 47 48

### DIFF
--- a/src/api/observationFieldValues.ts
+++ b/src/api/observationFieldValues.ts
@@ -56,13 +56,13 @@ const updateObservationFieldValue = async <T = ApiDefaultResult>(
   }
 };
 
-const deleteObservationFieldValue = async (
+const deleteObservationFieldValue = async <T = ApiDefaultResult>(
   id: string, // uuid
   opts: ApiOpts = {},
-): Promise<Record<string, unknown> | null | ErrorWithResponse | INatApiError> => {
+): Promise<ApiResponse<T> | null | ErrorWithResponse | INatApiError> => {
   try {
-    const { results } = await inatjs.observation_field_values.delete( { id }, opts );
-    return results;
+    const response = await inatjs.observation_field_values.delete( { id }, opts );
+    return response;
   } catch ( e ) {
     return handleError(
       e as ErrorWithResponse,
--- a/src/api/projectObservations.ts
+++ b/src/api/projectObservations.ts
@@ -37,13 +37,13 @@ const createProjectObservation = async <T = ApiDefaultResult>(
   }
 };
 
-const updateProjectObservation = async (
+const updateProjectObservation = async <T = ApiDefaultResult>(
   params: ProjectObservationUpdateParams,
   opts: ApiOpts = {},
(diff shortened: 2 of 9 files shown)


FLGMwt:
I'm not sure DELETE OFV is working. If I join, leave, then join, I'm seeing my previously saved OFV. This persists across relaunches _and_ login/logout so it seems to still exist on the server. I see the DELETE PO, but never a DELETE OFV:

image


https://github.com/user-attachments/assets/39532a5e-7ee3-49a3-97c1-94cd5c4e7957

jtklein:
> I'm not sure DELETE OFV is working. If I join, leave, then join, I'm seeing my previously saved OFV. This persists across relaunches _and_ login/logout so it seems to still exist on the server. I see the DELETE PO, but never a DELETE OFV:

Yeah, sorry, maybe some more explanation: This is generally expected for your flow.

Removing an observation from a traditional project only tombstones the PO. It does not tombstone the OFVs. OFV delete is wired only when you explicitly clear a synced field value. So, you'd need to go to Add to Projects and remove all text from an input field. To verify OFV DELETE: after upload, clear the field (not just deselect the project), save, upload. You should see DELETE OFV.

Product-wise OFVs are global per observation. One Obs-ObsField pair can apply to multiple projects that share the same field. Deleting the OFV when one project is deselected shouldn't wipe the answer still needed for another selected project. iOS Classic was more aggressive: when a PO was deleted, the OFVs were deleted alongside; Android Classic is not deleting OFVs.

FLGMwt:
> Product-wise OFVs are global per observation. One Obs-ObsField pair can apply to multiple projects that share the same field. Deleting the OFV when one project is deselected shouldn't wipe the answer still needed for another selected project. iOS Classic was more aggressive: when a PO was deleted, the OFVs were deleted alongside; Android Classic is not deleting OFVs.

GOTTTT it, that's right. They're Obs-ObsFields, not ProjectObs-ObsFields. It'll stick this time.

FLGMwt (approved):
with OFV delete clarification, changes LGTM!

PR #3977, opened 2026-08-25
Fix CI: lightningcss native dependency

jtklein:
Stop caching node_modules across workflows and always run npm ci, relying on actions/setup-node's per-OS npm tarball cache instead.

This fixes iOS CI failures where Metro could not load lightningcss.darwin-arm64.node because Ubuntu-filled node_modules caches were reused on macOS and npm rebuild did not install missing optional native packages.


FLGMwt (approved):
Thank you! Yeah it looks like we were double dipping on our old cache implementation and the caching that's now baked into the new setup-node actions

PR #3978, opened 2026-08-25
MOB-1542 fallback to GMT with correct offset when timezone is unsupported

FLGMwt:
The app misrepresents datetimes for certain time zones due to limited `Intl` support in Hermes which they've said they're unlikely to work on. https://github.com/facebook/hermes/issues/1601 https://github.com/marnusw/date-fns-tz/issues/306

For these unsupported timezones / locale strings, date-fns-tz gives us back our local time, but with a "default" GMT short time zone string.

Example:

https://www.inaturalist.org/observations/348588955

This obs occurred at 14:00 IST ("India Standard Time", IANA: "Asia/Kolkata"). Asia/Kolkata has a +5:30 utc offset, no DST. The website correctly reports this in `en-US` as "2:00 PM IST" but mobile shows "2:00 PM GMT" which is a _different instant in time_, incorrectly representing 7:30 PM local IST.

---

_thankfully_ we can detect this error case because the obs from the server includes both the IANA tz and the `time_observed_at` as a Date Time **_Offset_**: `{ observed_time_zone: "Asia/Kolkata", time_observed_at: "2024-01-10T14:00:25+05:30" }`.

If we format a DTO with a non-"00:00" offset and get back a GMT short timezone, we know this is wrong. That is: `{ observed_time_zone: "Etc/GMT", time_observed_at: "2024-01-10T14:00:25+05:30" }` is an _invalid combination of tz & DTO_.

So in this error case, we instead fallback to just formatting it as "Etc/GMT".

Result: we're at least referencing the correct instant in time even though we can't report the localized short time zone.

|Before (wrong instant)|After|
|---|---|
|image|image|

### DIFF
--- a/src/sharedHelpers/dateAndTime.ts
+++ b/src/sharedHelpers/dateAndTime.ts
@@ -4,6 +4,7 @@
 // names that are specific to particular views,
 // like "myObsDateFormat"
 
+import type { Locale } from "date-fns";
 import {
   differenceInDays,
   differenceInHours,
@@ -301,13 +302,17 @@ const ZONE_ABBREVIATION_TOKEN = /z+/;
 // won't extend that data, so detect the failure per call: a bare "GMT" abbreviation for a zone
 // whose real offset (computed via a separate, unaffected code path) isn't
 // actually zero.
-function timeZoneAbbreviationIsUnreliable( date: Date, timeZone: string ): boolean {
-  const offset = formatInTimeZone( date, timeZone, "xxx" );
+function timeZoneAbbreviationIsUnreliable(
+  date: Date,
+  timeZone: string,
+  locale: Locale,
+): boolean {
+  const offset = formatInTimeZone( date, timeZone, "xxx", { locale } );
   if ( offset === "+00:00" ) {
     // Actually UTC/GMT, so a "GMT" abbreviation would be correct
     return false;
   }
-  return formatInTimeZone( date, timeZone, "zzz" ) === "GMT";
+  return formatInTimeZone( date, timeZone, "zzz", { locale } ) === "GMT";
 }
 
 interface FormatDateStringOptions {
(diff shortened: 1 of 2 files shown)


jtklein (changes requested):
I can confirm this works for the obs linked above https://www.inaturalist.org/observations/348588955

For one of my observations I do still see the GMT without offset but ut was taken in CEST. https://www.inaturalist.org/observations/393171121
So, website: Aug 14, 2026 · 3:30 PM CEST
is app: 8/14/26 3:30 PM GMT
2026-08-14T15:30:00+02:00 from server.
Is this something we can add here?

FLGMwt:
image

For me in `en-US`, this shows as GMT+2 as I would expect. Can you share your locale?

jtklein:
>Can you share your locale?

en-DE

FLGMwt:
> For one of my observations I do still see the GMT without offset but ut was taken in CEST. https://www.inaturalist.org/observations/393171121 So, website: Aug 14, 2026 · 3:30 PM CEST is app: 8/14/26 3:30 PM GMT 2026-08-14T15:30:00+02:00 from server. Is this something we can add here?

mk, that should default to `enUS` as my screenshot above and it looks correct on `de` and `es` as well. I'm not sure how you're getting a different result. Can you share a screenshot?

|es|de|
|---|---|
|IMG_0035|IMG_0034|

jtklein:
This is how this branch looks on my test device, immediately after a login:
IMG_5085
IMG_5084
IMG_5086

jtklein:
const zonedOffset = formatInTimeZone( date, timeZone, "zzz" );
is "CEST" for those

FLGMwt:
> const zonedOffset = formatInTimeZone( date, timeZone, "zzz" )

oh thank you!!! I think that clears it up. The "unreliable" check didn't factor in the i18n locale so it hit the device default which doesn't necessarily resolve to the same TZ result.

@jtklein can you check again with this latest change? Thank you for your patience < 3

jtklein (approved):
Yes, the last fix also works on my end (and is hopefully universal for other devices/locales/timezones combinations).

PR #3979, opened 2026-08-25
ignore firebase migration logs

FLGMwt:
we'll do this in the actual migration. for now, they're really annoying! thanks for the callout @jtklein 

reference: https://rnfirebase.io/migrating-to-v22\#switching-off-warning-logs


jtklein (approved):
Thanks for fixing this. LGTM

PR #3980, opened 2026-08-25
MOB-1122 update react-compiler@1.0.0

FLGMwt:
closes MOB-1122

Doesn't seem like there were any necessary changes / regressions on our side which makes sense since we were on the last rc before this 1.0. As far as I can tell, the only relevant changes between the RC and 1.0 were related to internal use of zod validation.

The following shows that we're still getting the "Memo ✨" badges in the Components Dev Tools tab that denote components automatically memoized by react-compiler.

image
image

#3574 already got us square with the latest version of eslints react hooks, a companion to react-compiler


jtklein (approved):
Feels a bit anticlimactic but heck yeah, LGTM

PR #3981, opened 2026-08-25
Traditional projects: per project field form polish

jtklein:
Closes MOB-1550

This PR adds:
- Sort per-project observation fields by position when mapping Realm to the chooser.
- Test that a selected project with an empty required field shows the pencil icon.
- Dismiss the keyboard on scroll; pass validationResult as FlashList extraData so check/pencil icons update with OFVs.

We specifically do not add expand/collapse animation because I didn't get anywhere meaningful in the timebox.


abbeycampbell (approved):
This looks good to me! on iOS + android.

Will we just make a separate ticket for the collapse/expand animation?

PR #3985, opened 2026-08-26
MOB-1713 disable android font padding for glyphs

FLGMwt:
Fixes MOB-1713, a regression from the nativewind@4 upgrade.

note clipping of ✅ icon in lower right

|Before|After|
|---|---|
|image|image|

This is a general fix for our glyph based icons on Android because this also manifested in more subtle layout regressions.

Camera Icon:
|Before (lens off center)|After (lens centered)|iOS Reference (lens centered)|
|---|---|---|
|image|image|image|

Menu layout:
|Before (extra space) |After (less space)|Side by Side split (right is After)
|---|---|---|
|image|image|image|


jtklein (approved):
Great PR, LGTM!

PR #3988, opened 2026-08-26
MOB-1721 use url parsing to fix GH Security regex issue

FLGMwt:
addresses "High severity" GH security issue reported here: https://github.com/inaturalist/iNaturalistReactNative/pull/3987#discussion_r3866861127

I'm also fine just suppressing this issue since the relevant values are config driven.


sepeterson (approved):
Maps still work locally

PR #3989, opened 2026-08-27
MOB-1605 bottom padding to menu items to account for overflow/overlap

FLGMwt:
closes MOB-1605 "can't scroll to last menu option"

I wasn't able to replicate the last option being unselectable, but I shimmed some menu items & padding to force a situation where it's at least awkward with the ObsButton overlap and the BottomTabs obscuring the BottomSeparator:

image

Fix is just to provide some bottom padding to the items container that's generous to account for about half a menu item and the bottom overlap

Before:

https://github.com/user-attachments/assets/a2928820-9cfe-443e-bc87-bb0fc92affe6

After:

https://github.com/user-attachments/assets/0f87fe1c-41a3-4a65-a08c-4c18efda5e2e

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -334,7 +334,7 @@ const Menu = ( ) => {
         </Pressable>
 
         {/* Menu Items */}
-        <View>
+        <View className="pb-[128px]">
           {Object.entries( menuItems ).map( ( [key, item] ) => (
             <MenuItem
               key={key}
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -272,8 +272,7 @@ const Menu = ( ) => {
   }, [currentUser, isConnected, layoutPrefs, realm, t] );
 
   return (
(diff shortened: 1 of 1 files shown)


abbeycampbell (approved):
this works and accomplishes the ticket ask so I'm approving it. I do think there might be a nicer solution though; being able to scroll the entire Menu screen upwards means that the header can interact in an unfortunate way with the notch/stuff at the top of the screen:
Image

What about if we change the outmost `ScrollView` to a regular old `view`, and the `View` that wraps the MenuItems to a `ScrollView`? so the header stays put, and only the menu options scroll?

FLGMwt:
Good callout, confirmed that with Abhas.

iOS:

https://github.com/user-attachments/assets/98112248-4e44-43d8-9c54-a11c7b8c9160

Android:

https://github.com/user-attachments/assets/1e66326d-d8b7-43ca-bf9d-8d7c8ebfdedc

abbeycampbell:
@FLGMwt looks great!

PR #3992, opened 2026-08-27
DateTimePicker never stores the picked date

jtklein:
Closes [MOB-1551](https://linear.app/inaturalist/issue/MOB-1551) which wasn't really a user-facing bug because the library modal stays open and keeps an inner state of the first date picked and so the date= of the second modal was never applied. Removed the React state to not confuse us.

### DIFF
--- a/src/components/SharedComponents/DateTimePicker.tsx
+++ b/src/components/SharedComponents/DateTimePicker.tsx
@@ -22,7 +22,6 @@ const DatePicker = ( {
   onDatePicked,
   toggleDateTimePicker,
 }: Props ) => {
-  const [selectedDateNoTime, setSelectedDateNoTime] = React.useState<Date | undefined>( undefined );
   const [isTimeVisible, setisTimeVisible] = React.useState( false );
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/DateTimePicker.tsx
```diff
@@ -84,7 +82,6 @@ const DatePicker = ( {
       onCancel={_toggleDateTimePicker}
       onConfirm={selectedDate => {
         if ( mode === "datetime" ) {
-          setSelectedDateNoTime( selectedDateNoTime );
```
FLGMwt, line 87: we don't need an `onDatePicked` equivalent for this branch?

jtklein: No. We first show the date picker then the timer picker. Date is stored in the modal internally it seems, so when you choose a time in the second picker the correct date is sent to the last onDatePicked callback.

abbeycampbell (approved):
lgtm

jtklein:
Thanks @abbeycampbell

PR #3993, opened 2026-08-27
DateTimePicker on Android shows two date pickers

jtklein:
Closes MOB-1635, also closes MOB-315.

Fixes a bug where on Android we were showing the date picker twice. The modal was closing after the first pick and recreated for the second pick, so without isTimeVisible.
On some manufacturer there was only one picker showing and on close no further interaction was possible, instead of one picker for date and then one for time.

Before right, after left:

https://github.com/user-attachments/assets/1c1d88d3-a192-45c5-a354-422a92fb4332

### DIFF
--- a/src/components/SharedComponents/DateTimePicker.tsx
+++ b/src/components/SharedComponents/DateTimePicker.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { Platform } from "react-native";
 import DateTimePicker from "react-native-modal-datetime-picker";
 
 type PickerMode = "date" | "time" | "datetime";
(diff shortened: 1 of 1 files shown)


src/components/SharedComponents/DateTimePicker.tsx
```diff
@@ -32,6 +33,25 @@ const DatePicker = ( {
     toggleDateTimePicker( );
   };
 
+  if ( Platform.OS === "android" && mode === "datetime" ) {
```
FLGMwt, line 36: the prop diff is difficult to read and while I trust the fix, it adds to the component being hard to read. This is an android override of the existing "datetime" case, right?

We're always returning `<DateTimePicker />` but with a special-case mix of props. How would you feel about something like:

```
baseProps = {...}
modeProps = mode === 'datetime' ? {...} : mode === 'time' ? {...} : {...}

return <DateTime {...baseProps} {...modeProps} />
```

we have prop spreading eslint-disabled, but I think this is a good case to challenge that (especially with typescript)

FLGMwt: ^not expecting that from this PR.

but @jtklein, I'm unclear on what this change is actually doing/fixing. Can you add a comment for this special case?

jtklein: The change is fixing the bug as described in the ticket, and shown in the video above. Before you were not able to pick a date and time on Android. Now it works with the "datetime" variant of the library used here. What do you mean by "special case"?

abbeycampbell (approved):
LGTM on my samsung

jtklein:
Thanks @abbeycampbell

PR #3995, opened 2026-08-27
Mob 1618 wire up grouped by iconic taxa view

abbeycampbell:
closes [MOB-1618](https://linear.app/inaturalist/issue/MOB-1618/wire-up-grouped-by-iconic-taxa-view)
Apologies again for a large PR! A good chunk of the diff is tests though. Also we're no longer in a rush to review MyObs stuff, so take your time.

- `MyObservationsGroupedByIconicTaxaView` — wired this up and took the liberty of adding sticky headers; otherwise this view is going to have pretty limited utility for users with more than a couple hundred obs in a category. also has pull-to-refresh, and on sort change we will scroll-you-to-top and start over (i actually don't know if this is the preferred UX but it felt cleaner than trying to maintain the current position and reorder any currently expanded categories).
- `useIconicTaxaSectionObservations` — paginates each section independently. Models per-category pagination as one `useQueries` over a flat list of (category, page) descriptors, since `useInfiniteQuery` can't be called per category. Only one request is ever in flight, and activation walks down categories one at a time as the user scrolls or collapses.
- `useUnsyncedObservationIdsByIconicTaxon` — pins locally-saved observations to the top of the section they belong in.
- `SmallGridObsItemContainer` — connected wrapper owning per-observation upload state, with selectors narrow enough that uploading re-renders only the cell actually being uploaded.
- `SmallGrid`— extended with section headers, full-width span rows for per-section loading states, sticky headers, and uniform cell heights.
- `MyObservationsSmallGridSearchResults` — now an active search while in the small grid view will render a flat grid.
- `useStateResetOn` — a shared hook for resetting state when a key changes; so changing sort auto-reopens every section.

There will be follow-ups for: 
- More elegant upload and deletion handling; rn obs just disappear until you refresh
- Persisting collapsed section state through bottom tab navigation
- Performance improvements: this all works pretty well on my personal account with ~900 obs, but starts to lag after scrolling 1/3-1/2 of test_user_10k's obs, and memory reaches 700MB+. It's still usable on my iPhone 12 pro, but definitely slower. 
I also had claude do a comparison of the grid, list, and small grid views on a Samsung Galaxy S9. This was measured on a release build with my ~900-observation account using FlashList's `useBenchmark` (JS FPS, 3 runs per view):

| View       | JS FPS by rows loaded              |
|------------|------------------------------------|
| Small grid | 37.1 at 55 rows → 11.8 at 75rows → 7.8 at 95 rows     |
| Grid       | 14.7 at 209 rows → 13.9 at 269 rows → 12.5 at 329 rows  |
| List       | 7.1 at 369 rows → 6.8 at 389 rows → 6.9 at 409 rows     |

the new view loses ~79% of its frame rate over a 73% row increase, compared to the existing grid which loses ~15% over a larger one. So I'll need to make some improvements there.

A separate but larger question is that this view makes it easier to scroll through a bunch of your obs quickly and cache them in Realm. Is it time to think about setting a limit there? Some way to clear the cache of the obs least likely to be viewed?


src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+      // a tunnel. staleTime only narrows the window that can happen in; these close it. This
+      // view refreshes on pull-to-refresh and after an upload instead.
+      // Turning off reconnect also means a section that failed while offline won't quietly
+      // recover, which is why the error row it renders has to offer a retry.
+      refetchOnReconnect: false,
+      refetchOnWindowFocus: false,
+      retry: ( failureCount: number, error: unknown ) => reactQueryRetry(
+        failureCount,
+        error,
+        { queryKey: ["useIconicTaxaSectionObservations", category] },
```
sepeterson: reuse QUERY_KEY here?

sepeterson:
One thing that cropped up in claude review is that when you have X pages loaded and then do a pull-to-refresh it does X api calls for the refetch. Not a blocker I don't think but maybe worth a followup ticket.

sepeterson (approved):
Works well ✅ 
Looks great ✅ 
Creative solution to a complicated pagination situation ✅ 

<details>
<summary>screens</summary>

Image
Image

</details>

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+  } ), [orderedCounts, pageIndexesByCategory, pagesByCategory] );
+
+  const results = useQueries( {
+    queries: descriptors.map( ( { category, page } ) => ( {
+      queryKey: [
+        QUERY_KEY,
+        currentUser?.id,
+        sortParams,
+        category,
+        page,
```
FLGMwt: may not matter here so non-blocker but it affects cache invalidation: queryKeys should follow a hierarchy of specificity so I believe this should be:

```
category,
sortParams,
page,
```

src/components/MyObservations/hooks/useIconicTaxaObservationCounts.ts
```diff
@@ -8,24 +8,36 @@ interface IconicTaxaObservationCountsOptions {
...
+  const { data, isLoading, refetch } = useAuthenticatedQuery(
     ["useIconicTaxaObservationCounts", currentUser?.id],
     optsWithAuth => fetchIconicTaxaCounts( params, {
       api_token: optsWithAuth.api_token ?? undefined,
     } ),
     { enabled: enabled && !!currentUser },
   );
 
-  return useMemo( ( ) => orderIconicTaxaCounts( data?.results ?? [] ), [data] );
+  const counts = useMemo( ( ) => orderIconicTaxaCounts( data?.results ?? [] ), [data] );
```
FLGMwt: if nothing external impacts `orderIconicTaxaCounts`, we can do the `order` call inside of the queryFn and remove this intermediate memo.

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+          per_page: PER_PAGE,
+          fields: Observation.ADVANCED_MODE_LIST_FIELDS,
+          ttl: -1,
+        }, { api_token: apiToken } );
+        const observations = response?.results || [];
+        try {
+          Observation.upsertRemoteObservations( observations, realm );
+        } catch ( upsertError ) {
+          logger.error( "Failed to upsert iconic taxa section observations", upsertError );
+        }
```
FLGMwt: I don't think we can fail silently here. the queryFn informs the items and their fixed layout but depend on local obs. I tested throwing instead of upserting and it resulted in sections filled w/ whitespace

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+  retryCategory: ( category: ICONIC_TAXA_GROUP ) => void;
+  refreshSections: ( ) => void;
+}
+
+// Paginates each iconic taxa section independently.
+//
+// useInfiniteQuery can't be called once per category, so this
+// models pagination as one useQueries over a flat list of (category, page) descriptors built
+// from state. Requesting a page means bumping a number; React Query does the rest, and pages a
+// user has already scrolled past stay cached when they scroll back up.
```
FLGMwt: > useInfiniteQuery can't be called once per category

it can be, it would look like rendering some hook-only siblings of `<SmallGrid />` that give data/handlers to the parent by category.

I'm quite a bit late on surfacing this so I don't mean it as a suggestion. `useIconicTaxaSectionObservations` seems to work really well. I think there are risks with rolling our own infinite scroll implementation though so if we ever need that as an option, I just wanted to clarify that we _can_ do this.

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+  const advanceFrontier = useCallback( ( ) => {
+    const next = orderedCounts.find(
+      ( { category, count } ) => count > 0 && !pagesByCategory[category],
+    );
+    if ( !next ) return;
+    setPages( { ...pagesByCategory, [next.category]: 1 } );
+  }, [orderedCounts, pagesByCategory, setPages] );
+
+  // See the Result type above for the full set of cases this decides between
+  const nearingEndOfSection = useCallback( ( category: ICONIC_TAXA_GROUP ) => {
```
FLGMwt, line 221: I've hit some intermittent issues with overscrolling "jumping" my frontier ahead a few gaps. I think maybe at the very least, we need to guard against `section` being no further than the frontier and make sure all previous sections are either collapsed or `!hasMore`

https://github.com/user-attachments/assets/aecea7aa-2347-4bb2-9062-1e262803c7eb

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  const refreshSections = useCallback( ( ) => {
+    // Marks every page stale and refetches only the ones currently mounted; the rest reload
+    // when the user scrolls back to them. This is deliberately invalidateQueries rather than
+    // refetchQueries, which matches inactive queries too and would fan out into one request
+    // per cached page — the same thing the refetchOn* flags above exist to prevent.
+    //
+    // Pages are keyed by category, so a refresh that reorders sections doesn't invalidate
+    // what's already loaded; there's no reason to make the user fetch it all again.
+    queryClient.invalidateQueries( { queryKey: [QUERY_KEY] } );
```
FLGMwt: This doesn't match the comment and has a greater blast zone than expected. 

I don't want the screen I'm refreshing right now to affect another instance of `useIconicTaxaSectionObservations`. This marks _every_ query stale so if I go to an instance w/ a different sort, all of its queries will refire. I think this goes against the intention of staleTime & the other refetch options.

I think what we want is: `queryClient.invalidateQueries( { queryKey: [QUERY_KEY, currentUser?.id, sortParams] } );` which should reflect the current state of the screen.

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
```diff
@@ -0,0 +1,275 @@
...
+      .map( index => queryResults[index] )
+      .filter( Boolean );
+    const highestPage = pagesByCategory[category] ?? 0;
+
+    // Offset pagination can repeat an observation across pages if the underlying set shifts
+    // between requests, and a repeated uuid would be a duplicate key in the list
+    const uuids = [...new Set( pages.flatMap( page => page.data?.uuids ?? [] ) )];
+    // Until the first page lands we only know the count, which is close enough to decide
+    // whether there's more to ask for
+    const totalResults = pages[0]?.data?.totalResults ?? count;
```
FLGMwt: do we need to read `totalResults` at all for server counts? I thought that's what count was for.

FLGMwt: And this can be handled as a separate ticket, but it doesn't reflect local-only obs which prevents the section from opening with only one saved obs:

https://github.com/user-attachments/assets/8c2fb7a0-065f-4553-826f-3d84d3f8446c

FLGMwt (changes requested):
Hey, sorry for delay, this was a big read!

I'm hesitant about rolling our own useInfiniteQuery implementation / state, but I'm okay with it being scoped to one feature.

RC on two things and I'm okay to make tickets for things like the frontier skipping and local obs bugs:
- the refresh / invalidate queries needs to be scoped to "this screen"'s queries to respect the intent of of `stale time`, `refetchOnWindowFocus` settings
- we can't silence the Realm upserts, the page just doesn't work if this fails

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
abbeycampbell: 👍 dropping the try/catch

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
abbeycampbell: updated this comment to reflect

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
abbeycampbell: good catch, gonna make this one its own follow-up ticket

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
abbeycampbell: good point, updating the comment too

src/components/MyObservations/hooks/useIconicTaxaSectionObservations.ts
abbeycampbell: Agree, and this means we get to delete a field from the cached payload

I haven't been able to repro local obs not showing up in a previously 0-count category, but I will look into it more as I keep working on this.

abbeycampbell:
Thanks y'all!
I think I handled all the feedback. 
Follow-up tickets to come:
- Don't advance to a new section when the previous one has more stuff
- Pull-to-refresh costs one request per loaded page
- Sections aren't keeping up with local changes (uploads and deletions)
- Performance fiddling
- Persistence

abbeycampbell:
I'm going to merge this since I have 1 approval, it's behind a feature flag, and I don't want to leave it sitting for a whole week while I'm out. All follow-ups have been captured in tickets. @FLGMwt hit me up if you have further concerns and we can make more tickets!

FLGMwt:
> I'm going to merge this since I have 1 approval, it's behind a feature flag, and I don't want to leave it sitting for a whole week while I'm out. All follow-ups have been captured in tickets. @FLGMwt hit me up if you have further concerns and we can make more tickets!

sounds good, thank you!

PR #3997, opened 2026-08-27
Mob 1629 images flash onoff as user selects them for cv

abbeycampbell:
closes [MOB-1629](https://linear.app/inaturalist/issue/MOB-1629/images-flash-onoff-as-user-selects-them-for-cv)

On Android, overflow: hidden + a border radius makes RN clip children against a path derived from the border box. So when border width on this view changed, the photo got clipped entirely, and came back if it was re-selected and the border returned.


sepeterson (approved):
Works well! Tested on android emulator

PR #3999, opened 2026-08-28
Mob 1693 show iconic taxa instead of empty species grid

abbeycampbell:
closes [MOB-1693](https://linear.app/inaturalist/issue/MOB-1693/show-iconic-taxa-instead-of-empty-species-grid)

<details>
<summary>screenshot</summary>

IMG_5558

</details>

### DIFF
--- a/src/components/MyObservations/SimpleTaxonGridItem.tsx
+++ b/src/components/MyObservations/SimpleTaxonGridItem.tsx
@@ -1,5 +1,5 @@
 import classNames from "classnames";
-import { Body4, DisplayTaxonName } from "components/SharedComponents";
+import { Body4, DisplayTaxonName, IconicTaxonIcon } from "components/SharedComponents";
 import INatIconButton from "components/SharedComponents/Buttons/INatIconButton";
 import {
   Image, LinearGradient, Pressable, View,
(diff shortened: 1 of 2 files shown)


src/components/MyObservations/SimpleTaxonGridItem.tsx
```diff
@@ -53,13 +53,23 @@ const SimpleTaxonGridItem = ( {
...
+      </View>
+      { source?.uri && (
+        <Image
+          source={source}
+          className="grow aspect-square"
+          testID="TaxonGridItem.photo"
+          accessibilityIgnoresInvertColors
+          fadeDuration={0}
+        />
+      ) }
```
FLGMwt, line 72: why are we unconditionally showing the iconic icon behind?

abbeycampbell: This is just the same approach we already take with `ObsImage`/`ObsImagePreview`. this also has the benefit of letting us avoid fiddling with error and loading states or deal with re-rendering tiles where the photo fails, because there's always a fallback.

sepeterson (approved):
Looks good!

<details>
<summary>screenshot</summary>

Image

</details>

PR #4000, opened 2026-08-28
Alert for server 422 surfacing

jtklein:
Closes MOB-1512

The ticket is overly scoped to include some things the legacy apps do to show dedicated UI for a failed obs attachment to projects. We have this ticket to provide a similar parity milestone (=testable) version. However, in the final version of the feature we would ideally very seldomly trigger this error because we do massive client-side validation beforehand. In other words, in the final product we should do everything we can to not allow attachment if the server would throw a 422.
So, for this ticket, the only thing that is needed to close it is the new Alert on a 422.

You can currently test this PR by trying to add an observation without coordinates to this test project https://staging.inaturalist.org/projects/test-project-of-unfathomable-awesomeness the ruleset of the project does not allow obs without coordinates, and we are currently not validating that on the client yet. We will be but for now that is one way to trigger a server 422.

I have also added a new property to the Observation realm to persist a simple text message of an error that occurred when uploading/syncing the obs. In debug mode I added showing this single obs-relevant error in the Advanced Mode's obs list view. I think this will be helpful to anyone QAing the app in the field.
E.g.: a 422 error on the top observation while uploading persists like this until the next sync try:
IMG_5088


src/uploaders/utils/errorHandling.ts
```diff
@@ -20,6 +20,21 @@ Object.defineProperty( RecoverableError.prototype, "name", {
   value: "RecoverableError",
 } );
 
+export function formatUploadErrorMessage(
```
jtklein, line 23: No changes, only extracted the code below.

src/uploaders/utils/persistUploadError.ts
```diff
@@ -0,0 +1,36 @@
+import type Realm from "realm";
```
jtklein, line 1: Maybe this file is a bit overkill in terms of compartmentilization of code. I could have also just called safeRealmWrite directly where needed.

src/uploaders/observationUploader.ts
```diff
@@ -201,6 +211,13 @@ async function uploadObservation(
       + ": Project children upload failed",
       error,
     );
+    if (
+      error instanceof INatApiError
+      && error.status === 422
+    ) {
+      const displayMessage = formatUploadErrorMessage( error );
+      Alert.alert( "Error", displayMessage );
+    }
```
abbeycampbell, line 220: Suggestion: since the point is that the project attachment failed and not the observation, we could add something like "Couldn't be added to project: " here. The ticket AC mentions showing the actual project name but I kinda think that's overkill since the hope is validating for projects ahead of time will let us avoid this entirely

src/uploaders/utils/persistUploadError.ts
```diff
@@ -0,0 +1,36 @@
...
+
+function persistObservationUploadError(
+  realm: Realm,
+  realmObservation: RealmObservation,
+  uploadErrorMessage: string,
+): void {
+  if ( !realm || realm.isClosed ) {
+    return;
+  }
+  safeRealmWrite( realm, ( ) => {
```
abbeycampbell: Suggestion: we could wrap both writes here in a try/catch that just logs? These run as step one of a four-step catch block, and a throw inside a catch isn't caught by that same catch, so a failure here would skip `addUploadError`, the deleted-obs cleanup, and the re-login prompt. I think that could mean we'd show success on a failed upload even outside of the 422 path. tbh I'm not sure how likely it is that we're trying to write to an invalidated realm object, but `realmSync.ts` has a check for this so maybe it's worth being careful

abbeycampbell (approved):
Couple suggestions. Approving since it works as expected for me and I think that there is more product-side thinking/decision-making to do about how to actually surface these errors (if they're reachable) anyways

src/uploaders/observationUploader.ts
jtklein: I will not do this now because we might not ever hit this Alert in the end. Also, this can error for POs and OFVs, so we'd need to either track what errored or move this somewhere else.

src/uploaders/utils/persistUploadError.ts
jtklein: Have added a catch to log errors from realm write

PR #4002, opened 2026-08-31
MOB-1705 map view bugs

sepeterson:
Some narrow solutions for some buggy behavior I was seeing on the map view. Inline comments should cover each change pretty well.

There is a change to Map.tsx that has a wider impact. Seemed to be fine in testing.

### DIFF
--- a/patches/react-native-maps+1.27.1.patch
+++ b/patches/react-native-maps+1.27.1.patch
@@ -1,15 +1,19 @@
 diff --git a/node_modules/react-native-maps/ios/AirMaps/AIRMapUrlTile.m b/node_modules/react-native-maps/ios/AirMaps/AIRMapUrlTile.m
-index f77f627..ed9ff22 100644
+index f77f627..1822bfa 100644
 --- a/node_modules/react-native-maps/ios/AirMaps/AIRMapUrlTile.m
 +++ b/node_modules/react-native-maps/ios/AirMaps/AIRMapUrlTile.m
-@@ -10,6 +10,22 @@
+@@ -10,6 +10,26 @@
  #import <React/UIView+React.h>
  #import "AIRMapUrlTileCachedOverlay.h"
  
-+// On iOS, a plain MKTileOverlayRenderer intermittently never paints its tiles after
-+// the overlay is attached to the map, leaving the tile layer invisible until the user
-+// pans or zooms. Rendering through a subclass that overrides the draw path avoids it.
-+// Upstream: https://github.com/react-native-maps/react-native-maps/issues/5598
++// On iOS, a plain MKTileOverlayRenderer intermittently never paints its tiles, leaving
++// the tile layer invisible until the user pans or zooms. MapKit invalidates tile overlays
++// when other overlays are added or removed, which our maps do constantly. Rendering
++// through a subclass avoids it -- but note the overrides below are pure super calls:
++// they look like dead code and are not, and nobody has established why subclassing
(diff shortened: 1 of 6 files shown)


src/components/SharedComponents/Map/Map.tsx
```diff
@@ -132,7 +132,14 @@ const Map = ( {
   } );
   const { isDebug } = useDebugMode( );
   const { screenWidth, screenHeight } = useDeviceOrientation( );
-  const [currentZoom, setCurrentZoom] = useState( 0 );
+  // on iOS, onRegionChangeComplete does not fire when the map mounts with initialRegion
+  // Until the map reports a region of its own, derive the zoom from the region we're mounting with.
```
sepeterson, line 136: I should add this is important because the zoom level decides whether we show the green pins or the orange tiles. Occasionally on ios, the zoom number was bad and we were showing tiles at a zoom level that would typically show pins.

patches/react-native-maps+1.27.1.patch
```diff
@@ -1,3 +1,39 @@
...
+ #import "AIRMapUrlTileCachedOverlay.h"
+ 
++// On iOS, a plain MKTileOverlayRenderer intermittently never paints its tiles after
++// the overlay is attached to the map, leaving the tile layer invisible until the user
++// pans or zooms. Rendering through a subclass that overrides the draw path avoids it.
++// Upstream: https://github.com/react-native-maps/react-native-maps/issues/5598
++@interface AIRMapUrlTileRenderer : MKTileOverlayRenderer
++@end
++@implementation AIRMapUrlTileRenderer
++- (BOOL)canDrawMapRect:(MKMapRect)mapRect zoomScale:(MKZoomScale)zoomScale {
```
websoty: Maybe it would be worth adding a short comment here explaining why you override this method while just forwarding to super
It could help prevent someone from removing it as redundant in the future

patches/react-native-maps+1.27.1.patch
```diff
@@ -1,3 +1,39 @@
...
+--- a/node_modules/react-native-maps/ios/AirMaps/AIRMapUrlTile.m
++++ b/node_modules/react-native-maps/ios/AirMaps/AIRMapUrlTile.m
+@@ -10,6 +10,22 @@
+ #import <React/UIView+React.h>
+ #import "AIRMapUrlTileCachedOverlay.h"
+ 
++// On iOS, a plain MKTileOverlayRenderer intermittently never paints its tiles after
++// the overlay is attached to the map, leaving the tile layer invisible until the user
++// pans or zooms. Rendering through a subclass that overrides the draw path avoids it.
++// Upstream: https://github.com/react-native-maps/react-native-maps/issues/5598
```
jtklein: Not quite sure how you arrived at this piece of code to me. The issue linked seem to be only for base map tiles to me, not url tiles, and also didn't make any code suggestions.

jtklein (approved):
Not quite sure how you arrived at the patch changes but overall I can confirm that the three bugs in the ticket are fixed.

patches/react-native-maps+1.27.1.patch
sepeterson: I just pointed claude at it and it assembled the patch from a couple different bug reports.
And then I linked the wrong issue thread. Going back, the sources it actually cited are:
https://developer.apple.com/forums/thread/125631
https://github.com/briancoyner/MKOverlayRendererBug

I'll update the comment

PR #4010, opened 2026-09-02
MOB-1354 defer sentinel logfile deletion

FLGMwt:
Closes MOB-1354

low priority, I don't think this has a functional implication, it was just annoying me enough to take a stab.

We have these sentinel files to track steps/stages during a obs create. We write to the file adding each stage. We try to do this in a "fire-and-forget" manner. We delete the file once the flow is complete.

We introduced a sneaky regression earlier this year that meant a write "fire-and-forget" coincided with a delete "fire-and-forget". The race condition causes a reliable error for writing to a just-deleted file.

The file is non-critical so instead of fixing this by awaiting the sentinel operations, I swapped the deletion out for a "complete" stage.

Now: instead of deleting sentinels on flow completion (which conflicts with in-flight writes), we mark the sentinel as a new "complete" state. The existing sentinel cleanup process on deferred startup is updated to filter these out of reporting but still delete them.

### DIFF
--- a/src/components/Camera/AICamera/AICamera.js
+++ b/src/components/Camera/AICamera/AICamera.js
@@ -14,7 +14,7 @@ import LinearGradient from "react-native-linear-gradient";
 import { useSafeAreaInsets } from "react-native-safe-area-context";
 import { VolumeManager } from "react-native-volume-manager";
 import convertScoreToConfidence from "sharedHelpers/convertScores";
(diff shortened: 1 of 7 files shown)


websoty (approved):

sepeterson (approved):
Looks good! No longer seeing the error logbox on device

PR #4011, opened 2026-09-02
MOB-1705: various map view bugs

sepeterson:
A suite of bugfixes for map bugs found in exploratory testing. I'll add context in file comments


src/components/Explore/ExploreV2/components/ExploreV2MapView.tsx
```diff
@@ -32,12 +32,33 @@ const WORLDWIDE_REGION: Region = {
...
+  const lngSpan = Math.abs( nelng - swlng );
+  const coversGlobe = latSpan >= 120 || lngSpan >= 300;
+  const collapsedOnOneAxis = Math.min( latSpan, lngSpan ) < 0.001
+    && Math.max( latSpan, lngSpan ) > 1;
+  return coversGlobe || collapsedOnOneAxis;
+};
+
+const regionForResults = ( bounds: ApiTotalBounds, placeMode: EXPLORE_V2_PLACE_MODE ) => (
+  placeMode === EXPLORE_V2_PLACE_MODE.WORLDWIDE && isGlobalBounds( bounds )
+    ? WORLDWIDE_REGION
```
sepeterson, line 47: ExploreV1 uses fixed worldwide bounds on all worldwide searches but for this I chose to use the bounds from the API. But for truly global distributions it was framing the range weirdly so I opted to move to the fixed bounds for that case.

src/components/SharedComponents/Map/Map.tsx
```diff
@@ -340,14 +340,6 @@ const Map = ( {
     ? `${TILE_URL}/points/{z}/{x}/{y}.png?${queryString}&color=%2374ac00`
     : `${TILE_URL}/grid/{z}/{x}/{y}.png?${queryString}`;
 
-  // In Android, MapView does not reliably process tileUrlTemplate changes.
```
sepeterson, line 343: The negative diff here is for an old android workaround that I believe to be no longer necessary. Tested on an android 12 device and android 15 emulator

src/providers/ExploreV2Context.tsx
```diff
@@ -164,6 +168,7 @@ export const initialExploreV2State: ExploreV2State = {
   speciesSortBy: SPECIES_SORT.COUNT_DESC,
   filters: defaultExploreV2Filters,
   activeTab: OBSERVATIONS_TAB,
+  appliedSearchCount: 0,
```
sepeterson, line 171: Useful to differentiate successive searches without having to do more complicated compares

src/components/SharedComponents/Map/helpers/mapHelpers.ts
```diff
@@ -85,16 +85,28 @@ export function latitudeDeltaToMeters(
   return latitudeDelta * metersPerDegreeLatitude( latitude );
 }
 
+// Width in degrees of a bounding box, measured eastward from swlng to nelng.
+// A box crossing the antimeridian is encoded swlng > nelng
+export function longitudeSpan( swlng: number, nelng: number ): number {
```
sepeterson, line 90: Prior to this there was an ExploreV1 bug that didn't move the map for searches at places that spanned the antimeridian which includes significant places like Fiji and the The United States

src/components/SharedComponents/Map/Map.tsx
jtklein: I agree that we might not need this anymore given our multiple updates to react-native-maps since the introduction. Have tested on my Pixel and could not see any issues with distribution maps not rendering immediately.

src/components/SharedComponents/Map/helpers/mapHelpers.ts
jtklein: I also see this behaving different in ExploreV2 now.

src/providers/ExploreV2Context.tsx
jtklein: If you say so.

jtklein (approved):
LGTM

PR #4016, opened 2026-09-02
prevent redo search in map area from also opening an observation

abbeycampbell:
closes [MOB-1571](https://linear.app/inaturalist/issue/MOB-1571/tapping-redo-search-in-map-area-in-android-also-opens-an-observation)

This was happening because the button was positioned outside its container's bounds. React still registered button taps, but Android's native touch handling skipped the empty container entirely and passed the same tap through to the map underneath, so we did both the search redo and obs taps on the map. Also added `box-none` so only the button itself handles touches.

@sepeterson this seems to be working just fine in explore v2, but we might want to add the `box-none` since trying to pan around the redo search button does not work. v minor but I thought I'd flag, since the code touched here is about to be irrelevant

### DIFF
--- a/src/components/Explore/MapView.tsx
+++ b/src/components/Explore/MapView.tsx
@@ -208,21 +208,6 @@ const MapView = ( {
 
   return (
     <View className="flex-1 overflow-hidden h-full">
-      <View className="z-10">
-        {showRedoSearchButton && (
-          <View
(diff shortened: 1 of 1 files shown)


websoty (approved):

jtklein (approved):
LGTM, and fixes the bug in question.

PR #4017, opened 2026-09-02
MOB-1622 upload photos at q90

sepeterson:
Reconsidered the realm approach we looked at in standup. Not quite 1 line but looking better.

To recap the old system was 
```
source image -> maxWidth=2048 q100 JPEG for upload -> maxWidth=640 q100 JPEG for CV

```
and this pr is
```
source image -> maxWidth=2048 q100 JPEG for local persistence -> maxWidth=640 q100 JPEG for CV
                        ↳ maxWidth=2048 q90 JPEG for upload
```

### DIFF
--- a/src/realmModels/ObservationPhoto.ts
+++ b/src/realmModels/ObservationPhoto.ts
@@ -38,12 +38,11 @@ class ObservationPhoto extends Realm.Object {
     return localObsPhoto;
   }
 
-  static mapPhotoForUpload( photo: RealmPhoto ) {
-    const uri = Photo.getLocalPhotoUri( photo.localFilePath );
+  static mapPhotoForUpload( photo: RealmPhoto, uploadUri?: string | null ) {
     return {
       file: new FileUpload( {
-        uri,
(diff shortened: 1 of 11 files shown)


src/realmModels/ObservationPhoto.ts
```diff
@@ -38,12 +38,11 @@ class ObservationPhoto extends Realm.Object {
...
   }
 
-  static mapPhotoForUpload( photo: RealmPhoto ) {
-    const uri = Photo.getLocalPhotoUri( photo.localFilePath );
+  static mapPhotoForUpload( photo: RealmPhoto, uploadUri?: string | null ) {
     return {
       file: new FileUpload( {
-        uri,
-        name: uri?.split( "/" ).pop( ),
+        uri: uploadUri || Photo.getLocalPhotoUri( photo.localFilePath ),
```
FLGMwt, line 44: this only has one caller, right?

is this meant as a fallback if the resize fails? if so, can we add a comment explaining the two sources.

Better yet, maybe it makes more sense to move the default to where we resolve uploadUri

PR #4018, opened 2026-09-02
Mob 1704 explorev2 saved searches

sepeterson:
Adds a feature to save searches made to ExploreV2. Unfortunately no designs to reference since Abhas & I kind if just reasoned it out on a couple calls. You can see what figma references there are and some key screenshots of the UI in the linear ticket.

Always added via a star icon on the results screen, and can be applied and deleted from both UniversalSearch and AdvancedSearch.


src/components/ObsDetailsSharedComponents/Sheets/SuggestIDSheet.tsx
```diff
@@ -30,6 +30,7 @@ const SuggestIDSheet = ( {
   onPressClose,
 }: Props ): React.ReactNode => (
   <BottomSheetV2
+    contentClassName="pt-0"
```
sepeterson, line 33: Since I was messing around with BottomSheetV2. This better-matched the prior appearance

src/stores/createExploreV2SearchesSlice.ts
```diff
@@ -37,17 +38,33 @@ export const addRecent = <T>(
     ...items.filter( existing => keyFn( existing ) !== keyFn( item ) ),
   ].slice( 0, cap );
 
-export interface ExploreV2RecentSearchesSlice {
```
sepeterson, line 40: Repurposed this slice since it's already doing something similar

src/components/Explore/ExploreV2/helpers/savedSearchKey.ts
```diff
@@ -0,0 +1,78 @@
...
+    project: String( filters.project?.id ?? "" ),
+  };
+
+  return ( Object.keys( parts ) as ( keyof ExploreV2Filters )[] )
+    .sort( )
+    .map( field => `${field}=${parts[field]}` )
+    .join( ";" );
+};
+
+const savedSearchKey = ( search: SavedSearchIdentity ): string => [
```
sepeterson, line 70: All this is to make it so a search that you put together manually but matches one that was already saved will show the star icon already filled in.

If I were starting over, I think I might just assign it a uuid and say duplicates aren't a big deal but this does work fine.

src/components/Explore/ExploreV2/screens/AdvancedSearch.tsx
```diff
@@ -155,35 +164,12 @@ const AdvancedSearch = ( ) => {
...
-        break;
-      default: {
-        // Exhaustiveness check: ts fails if a new placeMode is added without a case.
-        const _exhaustive: never = location;
-        break;
-      }
-    }
-    dispatchV2( { type: EXPLORE_V2_ACTION.SET_SORT, sortBy } );
-    dispatchV2( { type: EXPLORE_V2_ACTION.SET_FILTERS, filters } );
+    dispatchV2( {
```
sepeterson, line 167: incidental but this makes a lot more sense for this sceen imo

src/components/SharedComponents/Sheets/BottomSheetV2.tsx
```diff
@@ -98,22 +102,22 @@ const BottomSheetV2 = ( {
     <BottomSheetModal
       backdropComponent={renderBackdrop}
       enableDynamicSizing
+      // disable swipe down to close
+      enableContentPanningGesture={false}
```
abbeycampbell, line 106: adding this mean that we're also disabling swipe to close for the only other consumer of the V2 bottomsheet, `SuggestIDSheet.tsx`. IMO that's totally fine since people still have plenty of ways to dismiss the sheet, but just calling it out

src/components/Explore/ExploreV2/helpers/savedSearchKey.ts
```diff
@@ -0,0 +1,78 @@
+import type { ExploreV2Filters, ExploreV2LocationState } from "providers/ExploreV2Context";
+import { EXPLORE_V2_PLACE_MODE } from "providers/ExploreV2Context";
+import type { SavedSearch } from "stores/createExploreV2SearchesSlice";
+import { subjectKey } from "stores/createExploreV2SearchesSlice";
+
+// The parts of a search that decide whether two saved searches are the same one. Sort is
```
abbeycampbell, line 6: this whole file is slick.

I do worry a little bit about omitting sort in case some people people are out there mainly using grid/list views but that feels like an edge case, I assume most people are looking at the map a lot. If we did ever want to key off of sort too, we could always just add like "oldest observed" "newest uploaded" in the saved search list next to filters

abbeycampbell (approved):
This is so sick and I'm so excited to actually use it!

A couple notes but they're not really about the code, which looks great to me, more about product stuff: The main hiccup I had while playing around with this is that if I'm in universal search, and select a previously saved search that has filters, I'll be in advanced search incidentally for the rest of my session with no way to get back. I think this is totally expected behavior, but it was confusing to me and I wonder if there's some user-friendly way (in general, not just where saved searches are concerned) to get back to universal search if I so choose. More of a note for Abhas and definitely not blocking. I think this is going to be a huge value add for V2 explore

src/components/Explore/ExploreV2/helpers/savedSearchKey.ts
sepeterson: thanks! Since we do already store sort on a saved search, I'll add it to the key since now's the easiest time to make changes here

src/components/SharedComponents/Sheets/BottomSheetV2.tsx
sepeterson: Ya in the moment I thought it was nbd but really there's no reason not to preserve the old behavior so I'll fix this

sepeterson:
> This is so sick and I'm so excited to actually use it!
> 
> A couple notes but they're not really about the code, which looks great to me, more about product stuff: The main hiccup I had while playing around with this is that if I'm in universal search, and select a previously saved search that has filters, I'll be in advanced search incidentally for the rest of my session with no way to get back. I think this is totally expected behavior, but it was confusing to me and I wonder if there's some user-friendly way (in general, not just where saved searches are concerned) to get back to universal search if I so choose. More of a note for Abhas and definitely not blocking. I think this is going to be a huge value add for V2 explore

Thanks for reviewing! Yes I think there are a few quality-of-life change changes to ExploreV2 & saved searches. Hopefully we will be able to do some followup improvements

PR #4019, opened 2026-09-02
use start/end instead of left/right

abbeycampbell:
closes [MOB-1669](https://linear.app/inaturalist/issue/MOB-1669/explore-layout-button-group-shadow-looks-off-in-rtl)


jtklein (approved):
Excellent nice little digestible PR, LGTM

Image

PR #4020, opened 2026-09-02
Update faker to v9

jtklein:
Updating faker to the next major version.

No changes were required, but one deprecation notice to a renamed function call.


sepeterson (approved):
tests passing locally and on CI

PR #4021, opened 2026-09-02
MOB-1570 fix exclude user search params

FLGMwt:
closes MOB-1570 & MOB-1626

We were trying to pass the `excludeUser` options object (`{id: 1234}`) as a query string. For whatever reason, that made the tiling server return unstyled pins. The fix is to pull out the id pass it as the _actual_ search field on the API:  `not_user_id`. 

We were making this work by doing client side filtering, but that's not actually necessary since they're filtered by the API. I tried to keep those changes as minimal as possible.

This mirrors the correct implementation in ExploreV2's https://github.com/inaturalist/iNaturalistReactNative/blob/463329092343cfae3fd7a2292fdafa9a03fee768/src/components/Explore/ExploreV2/helpers/filtersToApiParams.ts#L109, but I think still worth correcting.

Here's this working now:

https://github.com/user-attachments/assets/6f03c2cc-4fc3-4793-9dc0-4248421e9687

### DIFF
--- a/src/components/Explore/ObservationsView.js
+++ b/src/components/Explore/ObservationsView.js
@@ -3,22 +3,14 @@
 import {
   useNetInfo,
 } from "@react-native-community/netinfo";
-import {
-  searchObservations,
-} from "api/observations";
 import useInfiniteExploreScroll from "components/Explore/hooks/useInfiniteExploreScroll";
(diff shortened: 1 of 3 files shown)


sepeterson (approved):
handsome diff!

set the flag and didn't appear as an observer in a known location

<details>
<summary>screen</summary>

Image
</details>

PR #4022, opened 2026-09-03
MOB-1619 temporary workaround for Android 36+ back dismissing app

FLGMwt:
closes MOB-1619 

Pulls a workaround from the RN issue on this here: https://github.com/facebook/react-native/issues/54887

can confirm on a Pixel 8 Pro, API 36+ I can reproduce issue on main and it is fixed with this. All other application navigation seems unaffected.

This is fixed in RN 0.84. I would be fine to skip this workaround and just wait for RN 0.84, but I'm not sure how long that'll be (I created a stub ticket to track it though)


jtklein (approved):
This does fix the bug on my device with Android 17.

LGTM I'd say

but we do need to make sure to track the removal of the patch as a ticket as well

FLGMwt:
> but we do need to make sure to track the removal of the patch as a ticket as well

aye aye. Added a note to the upgrade ticket (MOB-1740) for this.

PR #4023, opened 2026-09-03
MOB-1683 add cancel ref to watchPosition Geolocation callback

FLGMwt:
fixes MOB-1683

fix: don't give unguarded setStates to external callbacks

similar to useObservationLocation, we need a cancel ref to avoid setStates after the host is unmounted.

### DIFF
--- a/src/sharedHooks/useWatchPosition.ts
+++ b/src/sharedHooks/useWatchPosition.ts
@@ -3,7 +3,7 @@ import type {
   GeolocationResponse,
 } from "@react-native-community/geolocation";
 import { useFocusEffect } from "@react-navigation/native";
-import { useCallback, useState } from "react";
+import { useCallback, useRef, useState } from "react";
 
 // Please don't change this to an aliased path or the e2e mock will not get
 // used in our e2e tests on Github Actions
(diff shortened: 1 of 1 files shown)


websoty (approved):

src/sharedHooks/useWatchPosition.ts
```diff
@@ -43,10 +45,11 @@ const useWatchPosition = ( options: {
         clearWatch( id );
         id = null;
       }
-      setIsWatching( false );
+      if ( !cancelledRef.current ) setIsWatching( false );
```
sepeterson: I think we can skip the check here right? success() and failure() already return early if cancelled and stop() doesn't otherwise leave as a callback

We also need `setIsWatching( false )` to run in the cleanup. With these changes if I create an obs, am on ObsEdit, and then tap into the location picker while the isWatching is true, it's stuck if I nav back.

<details>
<summary>recording</summary>

https://github.com/user-attachments/assets/c0e5a820-8ff5-4a86-a028-f763803e9e0f

</details>

sepeterson (approved):
Unfortunately I wasn't able to verify the fix because I was having trouble replicating the original issue but diff looks pretty safe to me.

Good to go if we're able to fix that one bug

src/sharedHooks/useWatchPosition.ts
FLGMwt: omg good catch, thank you! I was a little eager on the "prevent callback" thing but you're right that this is already guarded

PR #4024, opened 2026-09-03
Traditional Projects: Keep synced project metadata on re-select; skip child upload without uuid

jtklein:
Closes MOB-1513

## Summary
- Re-selecting a tombstoned PO/OFV before upload keeps `uuid` / `id` / `_synced_at`, so we do not POST a membership the server still has.
- PO/OFV create/update still uses the observation uuid (API v2). If create/update returns no uuid, skip child create; deletions still run.

## Test plan
- [ ] Remove a project from a synced obs, SAVE, re-add the same project, SAVE, upload — no extra PO POST
- [ ] New obs with projects uploads OFVs then POs using the observation uuid

Other items of the ticket are not needed

### DIFF
--- a/src/uploaders/observationUploader.ts
+++ b/src/uploaders/observationUploader.ts
@@ -190,7 +190,9 @@ async function uploadObservation(
     const childrenStartTime = Date.now( );
     const childOptions = { ...opts, api_token: apiToken };
     await syncProjectChildDeletions( observation, childOptions, realm );
-    await uploadProjectChildren( obsUUID, observation, childOptions, realm );
+    if ( obsUUID ) {
+      await uploadProjectChildren( obsUUID, observation, childOptions, realm );
+    }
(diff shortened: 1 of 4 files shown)


src/uploaders/observationUploader.ts
```diff
@@ -190,7 +190,9 @@ async function uploadObservation(
     const childrenStartTime = Date.now( );
     const childOptions = { ...opts, api_token: apiToken };
     await syncProjectChildDeletions( observation, childOptions, realm );
-    await uploadProjectChildren( obsUUID, observation, childOptions, realm );
+    if ( obsUUID ) {
```
websoty: Should syncProjectChildDeletions also depend on obsUUID, or is it intentionally safe to run when the observation upload returns no UUID?

sepeterson (approved):
Looks good from a code quality perspective. Seems a bit esoteric to test so I'll trust the functional change is correct

PR #4026, opened 2026-09-03
Mob 1330 explore v2 empty results state

FLGMwt:
fixes MOB-1330

sorry this held over so long @sepeterson, I tossed my earlier refactor attempt which didn't make sense to do before removing v1 & old myobs anyway. Learned from it tho and was able to simplify anyway.

Confirmed this don't break any other empty states or Explore V1

https://github.com/user-attachments/assets/f5044571-a6af-4bf4-b091-6e3a1e4d7485

### DIFF
--- a/src/components/ObservationsFlashList/ObservationsFlashList.tsx
+++ b/src/components/ObservationsFlashList/ObservationsFlashList.tsx
@@ -27,6 +27,7 @@ import {
   useNavigateToObsEdit,
   useTranslation,
 } from "sharedHooks";
+import { HALF_GUTTER } from "sharedHooks/useGridLayout";
 import useStore from "stores/useStore";
 
 import ObsPressable from "./ObsPressable";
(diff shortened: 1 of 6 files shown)


src/components/ObservationsFlashList/ObservationsFlashList.tsx
```diff
@@ -217,36 +215,51 @@ const ObservationsFlashList = ( {
...
+
   const contentContainerStyle = useMemo( ( ) => {
-    if ( layout === "list" ) { return contentContainerStyleProp; }
+    // flexGrow so the content container fills the available height when
+    // there's no data -- otherwise it shrinks to fit only the empty state,
+    // and the centering below has no room to center within.
+    // Grid's extra padding (item gutters, tab bar clearance) only matters
+    // when there are grid items to space out -- skip it while empty so the
+    // empty state doesn't need to compensate for padding it doesn't need.
+    if ( layout === "list" || isEmpty ) {
```
websoty: Since layout === "list" already returns above, could this condition just check isEmpty?

src/components/ObservationsFlashList/ObservationsFlashList.tsx
```diff
@@ -217,36 +215,51 @@ const ObservationsFlashList = ( {
...
 
   const emptyContent = useMemo( ( ) => {
-    const showEmptyScreen = showObservationsEmptyScreen
-      ? null
-      : (
-        <Body3 className="self-center mt-[150px]">
-          {t( "No-results-found-try-different-search" )}
-        </Body3>
-      );
+    const emptyContentStyle = {
```
sepeterson: why not make this a className?

sepeterson (approved):
looks good!

src/components/ObservationsFlashList/ObservationsFlashList.tsx
FLGMwt: ah yup! artifact of iteration before the simplification.

fixed, & consolidated the ternary under one `View`

PR #4028, opened 2026-09-04
Traditional projects: Hide required fields success banner when none are

jtklein:
Closes MOB-1742

Came up when iterating on the design files. When there are no required field do not show the text saying that there are required fields.
Only added one condition the rest is indentation.

### DIFF
--- a/src/components/AddToProjects/AddToProjects.tsx
+++ b/src/components/AddToProjects/AddToProjects.tsx
@@ -217,44 +217,50 @@ const AddToProjects = ( ) => {
   ] );
 
   const renderExpanded = useCallback(
-    ( item: RealmProject, projectValid: boolean ) => (
-      <View className="bg-lightGrayOpaque">
-        {!projectValid
-          ? (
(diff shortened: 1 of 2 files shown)


src/components/AddToProjects/AddToProjects.tsx
```diff
@@ -217,44 +217,50 @@ const AddToProjects = ( ) => {
...
-              <Body3 className="ml-2.5">
-                {t( "All-required-fields-have-been-filled" )}
-              </Body3>
-            </View>
+    ( item: RealmProject, projectValid: boolean ) => {
+      const hasRequiredFields = item.projectObservationFields.some( pof => pof.required );
+
+      return (
+        <View className="bg-lightGrayOpaque">
+          {hasRequiredFields && (
```
websoty: Since both branches render the same layout, could we keep a single banner and conditionally set only the icon, color, and text based on projectValid? It would reduce duplication

sepeterson (approved):
Ran w/ the Unidentifiable Plant Damage project and verified no banner is shown

PR #4030, opened 2026-09-04
MOB-1594: take icon greenness from selected state

sepeterson:
ticket was originally for a different bug that's no longer an issue but this came up during investigation so I just fixed it


jtklein (approved):
LGTM

PR #4032, opened 2026-09-06
Update react-native-reanimated and react-native-worklets

jtklein:
Closes MOB-1752

Compiled and tested in Debug and Release mode on Android and iOS by checking all animations we have that use reanimated.

For some reason, the first time opening the camera screen was leading to a completely not-interactive black screen after the camera had already loaded. So, I removed the fade in component wrapper which fixed this bug for me.


FLGMwt (approved):
LGTM, no issues I could perceive. Can't say I noticed the difference removing FadeInView either.

PR #4033, opened 2026-09-07
MOB-1753 nitro modules 0.36.5

FLGMwt:
closes MOB-1753

updates nitro-modules to 0.36.5 and pins it to make nitro dependency updates intentional while they're new and while we're new to them

doesn't seem to cause any problems with our app or `react-native-sensitive-info`, our only as-of-yet nitro dependency.


jtklein (approved):
LGTM

Compiled and tested in Debug and Release mode on Android and iOS by logging in and donig stuff that needs an auth token from sensitive info. (Also nuking token in Debug mode through Rozenite control worked).

PR #4036, opened 2026-09-08
ease debug menu navigation

FLGMwt:
Does two things:
- allows direct navigation to the debug menu after the "secret" version taps
- allows disabling debug mode from the debug menu itself 

https://github.com/user-attachments/assets/38c62538-41ec-49b3-a92f-ce2e9483b54b


jtklein (approved):
LGTM

PR #4042, opened 2026-09-09
MOB-1660: zoom changes + test

sepeterson:
fixes https://linear.app/inaturalist/issue/MOB-1660/mobile-bug-tapping-zoom-button-doesnt-zoom-in

update the zoom label -> zoom value mapping to account for more types of device

Please, if folks have an android device with multiple cameras or a recent iPhone pro, this is an important one to test

### DIFF
--- a/src/components/Camera/hooks/useZoom.ts
+++ b/src/components/Camera/hooks/useZoom.ts
@@ -1,3 +1,4 @@
+import clamp from "lodash/clamp";
 import indexOf from "lodash/indexOf";
 import last from "lodash/last";
 import {
@@ -45,11 +46,6 @@ const useZoom = ( device: CameraDevice ): object => {
   }, [device?.physicalDevices] );
   const minZoom = device?.minZoom ?? 1;
   const neutralZoom = device?.neutralZoom ?? 2;
-  // this maxZoom zooms to 3x magnification on an iPhone 15 Pro
-  // currently the camera viewport is different than the photo taken
-  // so the photo taken with this zoom looks accurate compared with the native camera
-  // photo taken, but the camera preview looks a little too small
-  const maxZoomWithButton = neutralZoom ** 2.5;
   const maxZoomWithPinch = Math.min( device.maxZoom ?? 1, MAX_ZOOM_FACTOR );
   const initialZoom = !device?.isMultiCam
     ? minZoom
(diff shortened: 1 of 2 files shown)


src/components/Camera/hooks/useZoom.ts
```diff
@@ -58,7 +54,28 @@ const useZoom = ( device: CameraDevice ): object => {
...
+  // ios/android anchor their zoom scales differently, which is why neutralZoom differs:
+  //   - Android puts the wide lens at 1.0 and the ultra-wide below it (minZoom ~0.5-0.67),
+  //     so neutralZoom is always 1 and the labels already are the zoom factors.
+  //   - iOS puts the widest lens at 1.0, so on any iPhone with an ultra-wide the default lens
+  //     sits at 2.0 and neutralZoom is 2. Without an ultra-wide it is 1.
+  // The clamp keeps requests inside the device's real range
+  //
+  // notably 3x is that threshold for handing off to the telephoto lens on the 15 Pro
+  // but for more recent Pros, that threshold is 5x so maybe we want to account for that someday
+  const zoomButtonValues = useMemo(
```
sepeterson, line 71: this is all needlessly complicated so I did my best to explain the reasoning for it in the comment above

src/components/Camera/hooks/useZoom.ts
```diff
@@ -45,11 +46,6 @@ const useZoom = ( device: CameraDevice ): object => {
   }, [device?.physicalDevices] );
   const minZoom = device?.minZoom ?? 1;
   const neutralZoom = device?.neutralZoom ?? 2;
-  // this maxZoom zooms to 3x magnification on an iPhone 15 Pro
-  // currently the camera viewport is different than the photo taken
-  // so the photo taken with this zoom looks accurate compared with the native camera
-  // photo taken, but the camera preview looks a little too small
-  const maxZoomWithButton = neutralZoom ** 2.5;
```
sepeterson, line 52: for the life of me I cannot figure out why this number is an exponent and why this was the value chosen. On android neutralZoom is always 1, which of course always resolves to 1

The number we're looking for here on ios is usually 6, which below we compute with `neutralZoom * zoomButtonOptions[2]`

FLGMwt:
Works well for me on Pixel 8 Pro with the callout that the "3x" within the app seems to result in roughly the same zoom level as "5x" in the native camera app. Even so, it takes quite a bit more pinch zooming to force the telephoto lens switch. The switch to the UW for 0.5 works as expected.

I am seeing `capture/focus-canceled The focus operation has been canceled by a new focus request` logged from rn-vision-camera when playing with zoom, I'm guessing around the lens switches? I don't think it impacts the camera though so probably not worth looking into, especially with the v5 on the horizon.

FLGMwt:
image

https://github.com/user-attachments/assets/34fbb50b-78e9-40d0-a419-738f18a374aa


This is a little different than the original bug, I don't know if this needs addressing: on a LG G6, I only ever see 1 as a zoom level. Tapping the 1 doesn't have any effect while so I have to manually pinch zoom. On tapping 1 after zooming, we are correctly "resetting" to default zoom at least.

FLGMwt (approved):
LGTM if that G6 behavior is acceptable. The experience w/ the Pixel 8 Pro seems solid and preferable to prioritize over the G6 if it comes to that.

jtklein:
>This is a little different than the original bug, I don't know if this needs addressing: on a LG G6, I only ever see 1 as a zoom level. Tapping the 1 doesn't have any effect while so I have to manually pinch zoom. On tapping 1 after zooming, we are correctly "resetting" to default zoom at least.

We only offer the .5x and 3x buttons if your phone has a ultra-wide angle lens and telephoto lens, respectively.

jtklein:
This still works well on the iPhone 17 Pro. The button toggles nicely between 0.5, 1 and 3. (I don't have an Android with 3 cameras)

sepeterson:
@jtklein @FLGMwt thank you both for testing!

> Works well for me on Pixel 8 Pro with the callout that the "3x" within the app seems to result in roughly the same zoom level as "5x" in the native camera app. Even so, it takes quite a bit more pinch zooming to force the telephoto lens switch. The switch to the UW for 0.5 works as expected.

Ya there's definitely some more tuning to be done to reach parity with the native camera, complicated by our preview being a different size. For this, I'm going to close it at "fixed on android, produces plausible zooms at the different button values"

FLGMwt:
> We only offer the .5x and 3x buttons if your phone has a ultra-wide angle lens and telephoto lens, respectively.

Sounds good, thanks for the clarification! minor enhancement I can see is disabling the "1x" tap when we're _at_ 1x, but we've got plenty else to worry about ; D

PR #4046, opened 2026-09-09
finish js => tsx for UI library

FLGMwt:
as w/ #3996, these aren't critical files so no urgency on review. 

Just getting them out of the way to remove noise for the final TS push.


jtklein (approved):
Nice! LGTM

PR #4048, opened 2026-09-10
add rn-nitro-image (dep only)

FLGMwt:
closes MOB-1754

Per ticket, we're not actually cutting over to this for image components as of now. We're just pulling it for react-native-vision-camera@5 support. See also https://github.com/inaturalist/SeekReactNative/pull/1017

Package only doesn't seem to have impact on build. I ran into some stuff while spiking the component cutover, but will document those separately since we're not doing that right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "react-native-mmkv": "^3.3.3",
     "react-native-modal": "^14.0.0-rc.1",
     "react-native-modal-datetime-picker": "^18.0.0",
+    "react-native-nitro-image": "^0.15.1",
     "react-native-nitro-modules": "0.36.5",
     "react-native-open-maps": "^0.4.3",
     "react-native-orientation-locker": "github:wonday/react-native-orientation-locker",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -127,6 +127,7 @@
     "react-native-mmkv": "^3.3.3",
     "react-native-modal": "^14.0.0-rc.1",
     "react-native-modal-datetime-picker": "^18.0.0",
+    "react-native-nitro-image": "^0.15.1",
```
jtklein: Please pin to this exact version like the ticket asks for. At least for this combination of dependencies we know that VC 5 builds in some of our apps.

jtklein (changes requested):
Please pin

package.json
FLGMwt: Fixed, sorry about that.

jtklein (approved):
Thanks for the pin.

Was able to compile and smoke test this on Android and iOS in Debug and Release mode, so LGTM.

PR #4049, opened 2026-09-10
MOB-1741 ExploreV2 - seed UniversalSearch text fields from already-applied search

sepeterson:
Since the ticket was cobbled together from a slack convo, the details are conflicted but the intent is for UniversalSearch to open prepopulated with the applied Subject/Location of the prior search.

What this looks like in practice is when you open UniversalSearch, the subject field populates with whatever was the subject prior, and you start with that text selected. When you begin typing, the subject is cleared and autocomplete results start. If you select one of the defaults, it's just replaced. Location field mirrors this

This is useful when trying to search for a different taxon in the same location and things like that

<details>

Simulator Screenshot - iPhone 17 Pro Max - 2026-09-10 at 15 06 19
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-10 at 15 06 26
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-10 at 15 06 37
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-10 at 15 06 44

</details>


src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -116,27 +117,60 @@ const UniversalSearch = ( ) => {
...
+    locationText: state.location.placeMode === EXPLORE_V2_PLACE_MODE.WORLDWIDE
+      ? ""
+      : locationLabel( state.location, t ),
+  } ) );
+  // iOS drops a selection applied while the field is becoming first responder,
+  // so selectTextOnFocus does nothing for the focus that happens at mount (see
+  // facebook/react-native#30585, #44307). The public `selection` prop holds it
+  // instead. Control is released as soon as the user acts so the caret behaves
+  // normally after that.
+  const [subjectSelection, setSubjectSelection] = useState<
```
sepeterson, line 139: There is a prop, `selectTextOnFocus` that should do this for us automatically but it doesn't work properly on ios for a text field that gets focused on mount

Instead we gotta do this and then set it `undefined` as soon as the user wants to change the selection

src/components/Explore/ExploreV2/screens/UniversalSearch.tsx
```diff
@@ -180,56 +209,55 @@ const UniversalSearch = ( ) => {
...
-          place: selectedLocation.place,
-        } );
-        break;
-      case "nearby":
-        dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_NEARBY } );
-        break;
-      default:
-        dispatch( { type: EXPLORE_V2_ACTION.SET_LOCATION_WORLDWIDE } );
-    }
+    dispatch( {
```
sepeterson, line 232: Collapsed this to a single case just because

jtklein (approved):
This works well on my end. Haven't seen anything noteworthy. The iOS special input field handling debt is unfortunate. But LGTM

PR #4050, opened 2026-09-10
bump concurrent-ruby@1.3.8 manually to make dependabot go away

FLGMwt:
I don't know how https://github.com/inaturalist/iNaturalistReactNative/pull/3757 ended up with a mix of 1.3.7 & 1.3.8, but I pushed to .8 for safety: https://github.com/ruby-concurrency/concurrent-ruby/releases/tag/v1.3.8 changes seem inconsequential between the two

validated on fresh `cd ios && bundle install && bundle exec pod install` then rebuilds. Everything seems fine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ ruby ">= 2.6.10"
 # Exclude problematic versions of cocoapods and activesupport that causes build failures.
 gem "activesupport", ">= 6.1.7.5", "!= 7.1.0"
 gem "cocoapods", ">= 1.13", "!= 1.15.0", "!= 1.15.1"
-gem "concurrent-ruby", "< 1.3.4"
+gem "concurrent-ruby", "< 1.3.8"
 gem "xcodeproj", "< 1.26.0"
 
 # Ruby 3.4.0 has removed some libraries from the standard library.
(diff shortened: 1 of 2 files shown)


websoty (approved):

Gemfile
```diff
@@ -7,7 +7,7 @@ ruby ">= 2.6.10"
 # Exclude problematic versions of cocoapods and activesupport that causes build failures.
 gem "activesupport", ">= 6.1.7.5", "!= 7.1.0"
 gem "cocoapods", ">= 1.13", "!= 1.15.0", "!= 1.15.1"
-gem "concurrent-ruby", "< 1.3.4"
+gem "concurrent-ruby", "< 1.3.8"
```
sepeterson, line 10: if we want it at 1.3.8 wouldn't that be "~> 1.3.8" ?

jtklein (commented):
Have you looked into the reasons why the RN init template pins this to < 1.3.4, I assume some build failures from the comment above.
We'd need to make sure I can still build the app after this PR.

FLGMwt:
aye, thanks for the eyes, @.sepeterson & @.jtklein. I'm closing w/o followup. I posted it to mirror the dependabot change but so it could run privileged CI

---

I validated this change with a fresh iOS build and CI's e2e iOS Release build demonstrate the build runs clean.

That said,

I generated a fresh `rn@0.87.1` latest template and it still has the `'concurrent-ruby', '< 1.3.4'`. If they're still pinning this to `< 1.3.4` and that's still a few versions ahead of this, I'd rather ignore the dependabot for now and close this rather than investigate further.

PR #4051, opened 2026-09-10
use an obs uuid/pk when we're doing a single item Realm query for isUnsync'd

FLGMwt:
closes MOB-1654

I swear I had done this before but maybe I just did the `UNSYNCED_FILTER` and missed following up on `isUnsyncedObservation`. Sorry about that, and good catch @.abbey!

### DIFF
--- a/src/realmModels/Observation.js
+++ b/src/realmModels/Observation.js
@@ -480,8 +480,8 @@ class Observation extends Realm.Object {
   };
 
   static isUnsyncedObservation = ( realm, obs ) => {
-    const obsList = Observation.filterUnsyncedObservations( realm );
-    const unsyncedObs = obsList.filtered( `uuid == "${obs.uuid}"` );
+    const unsyncedObs = realm.objects( "Observation" )
+      .filtered( `uuid == "${obs.uuid}" && ( ${UNSYNCED_FILTER} )` );
     return unsyncedObs.length > 0;
   };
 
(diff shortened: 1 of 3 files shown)


src/realmModels/Observation.js
```diff
@@ -480,8 +480,8 @@ class Observation extends Realm.Object {
   };
 
   static isUnsyncedObservation = ( realm, obs ) => {
-    const obsList = Observation.filterUnsyncedObservations( realm );
-    const unsyncedObs = obsList.filtered( `uuid == "${obs.uuid}"` );
+    const unsyncedObs = realm.objects( "Observation" )
+      .filtered( `uuid == "${obs.uuid}" && ( ${UNSYNCED_FILTER} )` );
```
FLGMwt, line 484: @jtklein I was poking at the differences between this and `realm.objectForPrimaryKey(obs.uuid)?.needsSync()` the perf difference is negligible so I'm fine on the smaller diff. But it surfaced that `Obs.needsSync` is missing PO/OFV deletion that `UNSYNCED_FILTER` captures.

Am I correct in thinking that that's fix-worthy? I can do that here or do in a separate PR. I'm thinking it's just adding `_pending_deletion` checks to `PO.needsSync` & `OFV.needsSync`

sepeterson (approved):
smoke tested on android, adding and syncing a few obs and all worked as expected

PR #4052, opened 2026-09-10
fix multiple selection from GPhotos by removing intent condition

FLGMwt:
closes MOB-1645

Diff is unhelpful (SemanticDiff is also wonky) but the key change is removing the `Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU` guard before setting `Intent.EXTRA_ALLOW_MULTIPLE`, the one that GPhotos (and ostensibly other modern pickers?) handles. I confirmed it's harmless to have the multiple intents: this works on pre-and-post Android 13 devices (flexing my trusty LG G6).

We definitely _do_ need to keep the `TIRAMISU` guard on the later logic so I've left that untouched.

Before: System Browser works w/ multiple photos, but GPhotos only allows 1 photo and "shares" as soon as I tap the first one

https://github.com/user-attachments/assets/160f6657-bddd-49e0-8ac8-7711923f11f4

After: both allow selecting multiple

https://github.com/user-attachments/assets/7a27af2a-7df0-48b1-b865-4197f75c08e3


FLGMwt:
on the diff, VS Code file diff does it justice:

image

sepeterson (approved):
Tested working on emulator. 

My photos did come in rotated though 🤔 Feel like we've fixed that issue like 3 times already

<details>
<summary>
screenshot
</summary>

Image

</details>

FLGMwt:
> My photos did come in rotated though

@sepeterson uh oh. is that only with this change? or is that present on `main` with GPhotos in the emulator for you?

sepeterson:
> is that present on `main` with GPhotos in the emulator for you?

Ya just tested and it is pre-existing on the latest release

PR #4053, opened 2026-09-10
MOB-1764: parity between UniversalSearch and AdvancedSearch user defaults

sepeterson:
Makes it so a user on AdvancedSearch has the same defaults a user on UniversalSearch would wrt users

<details>
<summary>
screens
</summary>
Simulator Screenshot - iPhone 16 Pro Max - 2026-09-10 at 17 26 51
Simulator Screenshot - iPhone 16 Pro Max - 2026-09-10 at 17 26 44

</details>


FLGMwt (approved):
LGTM! 

seems odd that we want to entirely remove the `not_user_id` feature but I guess it's better than having three different "user-y" options in the filter modal.

PR #4054, opened 2026-09-10
MOB-1744 ExploreV2 AdvancedSearch subjects in header

sepeterson:
<details>
<summary>
screenshots
</summary>
Simulator Screenshot - iPhone 17 - 2026-09-11 at 09 14 30
Simulator Screenshot - iPhone 17 - 2026-09-11 at 09 11 41

</details>


jtklein (approved):
The behaviour change works well on my end, I can see projects in Advanced Search Explore header. However, I think we would need a bit more padding between the subject and the tabs. Design shows 16px between bottom of taxa image and tabs number text. I think you have 8. This also applies to non-Advanced Search I think.

sepeterson:
> I think we would need a bit more padding between the subject and the tabs. Design shows 16px between bottom of taxa image and tabs number text. I think you have 8. This also applies to non-Advanced Search I think.

Thanks for reviewing! changed it to 10 bc Abhas made [this ticket](https://linear.app/inaturalist/issue/MOB-1700/reduce-myobs-header-spacing) which specs 10px between the visuals and asked us to be consistent with it in ExploreV2 as well.

PR #4055, opened 2026-09-11
expose local-only / logged out language switching

FLGMwt:
Closes MOB-1711 (and incidentally MOB-1378)

The language switcher will now display for logged out users.

Confirmed in the ticket w/ Product: we _do_ want account settings to override logged-out locale on login.

This endpoint is available unauth'd so I switched it to unconditionally use the unauth'd useQuery.

For some reason I can't tell, we were also doing a manual disk cache of locales. I went ahead and removed that because it doesn't seem justified. This was tracked as a "don't do that w/ the zustand store" in [MOB-1378](https://linear.app/inaturalist/issue/MOB-1378) and looking into this, I further confirmed that it was unnecessary.

### DIFF
--- a/src/components/Settings/LanguageSetting.tsx
+++ b/src/components/Settings/LanguageSetting.tsx
@@ -1,5 +1,4 @@
 import fetchAvailableLocales from "api/translations";
-import { getJWT } from "components/LoginSignUp/AuthenticationService";
 import {
   Button,
   Heading4,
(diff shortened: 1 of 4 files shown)


FLGMwt:
@.jtklein raised the good point that this unlocks the ability to (attempt to) respect device locale on first app launch. I've raised this to Product in the ticket as a potential follow up.

sepeterson (approved):
Awesome cleanup in LanguageSettings! All went well during a quick smoke test

PR #4056, opened 2026-09-11
 remove extra list bottom padding

FLGMwt:
closes MOB-1719

two separate fixes:
- `InfiniteScrollLoadingWheel` renders a `View` with a `testId` even when `hideLoadingWheel`. This change removes the styling from that `View` so it doesn't unconditionally add padding.
- We still had extra padding in GridView with that change due to a `TAB_BAR_HEIGHT = 80` added to to paddingBottom. _as far as I can tell_, this is vestigial from when we actually needed to account for an absolutely positioned bottom tab bar.

~I have an open question on the ticket about whether we want to retain _some_ padding because our floating controls / overlays obscure the bottom of the list. But this implements ticket as written so ready for review.~ To account for controls on list view only, we're maintaining some artificial space. 

https://github.com/user-attachments/assets/d96a4a68-e68c-408a-a30b-7f3c96859489

### DIFF
--- a/src/components/SharedComponents/FlashList/InfiniteScrollLoadingWheel.tsx
+++ b/src/components/SharedComponents/FlashList/InfiniteScrollLoadingWheel.tsx
@@ -23,7 +23,7 @@ const InfiniteScrollLoadingWheel = ( {
     ? "h-[128px] py-16"
     : "h-64 py-16";
   if ( hideLoadingWheel ) {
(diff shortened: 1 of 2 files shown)


sepeterson:
Bottom row on Group Photos is partially obscured on these changes

Simulator Screenshot - iPhone 17 - 2026-09-11 at 13 48 23

FLGMwt:
image

Good catch, @sepeterson , fixed!

sepeterson (approved):
Looks great!! thanks for fixing!

PR #4058, opened 2026-09-11
Traditional projects: integration tests parity prototype

jtklein:
Closes MOB-1516

I tried to lay it out in sequential commits that make sense but it got a bit messy at the end. Anyways, maybe doesn't also need a very rigorous review as long as the feature is not completed yet.


tests/integration/navigation/AddToProjects.test.js
```diff
@@ -0,0 +1,200 @@
...
+  inatjs.users.projects.mockResolvedValue( makeResponse( [mockProject] ) );
+} );
+
+beforeEach( () => {
+  setStoreStateLayout( {
+    isDefaultMode: false,
+  } );
+  useStore.setState( {
+    featureFlagConfig: {
+      [FeatureFlag.TraditionalProjectsEnabled]: true,
```
sepeterson, line 96:
```suggestion
      ...state.featureFlagConfig,
      [FeatureFlag.TraditionalProjectsEnabled]: true,
```

I'd propose that it's probably best to keep the other default flags for integration tests

sepeterson (approved):
Good stuff! 1 suggestion but not a blocker

jtklein:
Will update the PR with the FF helper after #4062

PR #4061, opened 2026-09-14
MOB-1782: reusable component and visual consistency for ExploreV2 saved search and camera toats

sepeterson:
<details>
<summary>
screenshots
</summary>
IMG_0267
IMG_0268
IMG_0270
IMG_0269

</details>


src/components/SharedComponents/Toast.tsx
```diff
@@ -1,67 +1,94 @@
...
-      Animated.delay( VISIBLE_DURATION ),
-      Animated.timing( opacity, {
-        toValue: 0,
-        duration: FADE_OUT_DURATION,
-        useNativeDriver: true,
-      } ),
-    ] );
+    opacity.set(
+      withSequence(
+        withTiming( 1, { duration: FADE_IN_DURATION } ),
```
jtklein, line 47: Welcome to the dark side.

src/components/SharedComponents/Toast.tsx
```diff
@@ -1,67 +1,94 @@
...
-    } );
-
-    return ( ) => animation.stop( );
-  }, [opacity, text] );
+  const isDark = variant === "dark";
+  const containerClassName = classnames(
+    "flex-row self-center items-center gap-2 rounded-lg",
+    isDark
+      ? "bg-darkGray/50 p-2"
+      : "bg-white px-[10px] py-[7px] max-w-[220px]",
```
jtklein: px-[10px] py-[7px]  Can we not just make p-2 as well?

src/components/Explore/ExploreV2/screens/ExploreResults.tsx
```diff
@@ -413,15 +413,18 @@ const ExploreResults = ( ) => {
...
               state.activeTab === OBSERVATIONS_TAB && showMap
                 ? "bottom-[140px]"
                 : "bottom-[82px]",
             )}
             pointerEvents="box-none"
           >
             {saveToast && (
-              <View className="shrink">
+              <View
+                className="absolute left-0 right-0 items-center"
```
jtklein, line 425: I do like and am grateful for the refactor to re-use existing UI here. I humbly suggest we go one step further and re-use the same placement of toast. So, I'd make it appear in the top center of the relevant container. So, for this case I'd say it should appear just below the Explore Tabs. Then behaviour is same as in camera where it also doesn't appear next to the button.
If you prefer yours I'd suggest we move camera also next to the button.
But probably all of this is @abhasmisraraj 's call.

jtklein (approved):
As I said inline, grateful we re-use existing UI were it makes sense.
This is a great PR. I didn't see we had unit tests for Toasts, so it's good we were able to add more usage to it.

src/components/SharedComponents/Toast.tsx
sepeterson: Sure! I think this looks fine

Simulator Screenshot - iPhone 16 Pro Max - 2026-09-15 at 11 37 25

PR #4062, opened 2026-09-14
Explorev2 on

sepeterson:
turn explore v2 on, handle in tests


tests/integration/navigation/ExploreV2EntryPoints.test.js
```diff
@@ -121,7 +112,7 @@ beforeEach( ( ) => {
     isAllAddObsOptionsMode: true,
   } );
   zustandStorage.setItem( "exploreV2ObservationsLayout", "grid" );
-  enableExploreV2( );
+  setStoreStateFeatureFlags( { exploreV2Enabled: true } );
```
FLGMwt, line 115: It seems to me that if we're anointing V2, this should be a straight delete and it's the V1 paths that should be exceptionally overriding Feature Flags.

FLGMwt (approved):
yaaaaaaas lessgooooooo

tests/integration/navigation/ExploreV2EntryPoints.test.js
sepeterson: We're in kind of a confusing state since our feature flag setup is still in motion but doesn't it make sense that tests that require a specific feature flag specify it rather than depending on a remote response or a particular set of defaults?

Either way, I think this is something we can defer to the cleanup of ExploreV1

FLGMwt: that's what I mean though. why is this call necessary? are other units tests still asserting on V1?

meanwhile: I'm responsible for this, I should have set up "test all the flags" logic for jest. I'll think about that.

PR #4063, opened 2026-09-15
scientific names font bug

FLGMwt:
MOB-1526

oof, event w/ agent assist, this was still one of the more annoying issues I've dealt with _in my programming career_.

Learning: based, I think on render size, RN / iOS font resolution can say "hrmmm, you asked for Lato w/ italics. but at this size, I see you have a Lato bold w/ italics. I think that's better so I'll use that without asking"

Solution: programmatically edit our Lato fonts so their internal family name metadata is unique and matches their file name convention. This lets us target _exactly_ the fontface we want and doesn't let RN/iOS get clever about font resolution because every font is a unique isolated, use-case-scoped "family". 

We now have to handle weight/italic configuration that would have been handled by a family. So we have a className resolver, similar to twMerge that evaluates any attempted font configuration and resolves to a _single_ className for the appropriate fontface (say, `font-Lato-Italic`).

and also: we don't have any attempted style prop weight/italic overrides, but to prevent this from silently failing, the last commit adds a typography props type which inherits everything from TextProps but _forbids_ those potential problem props. <- This I'd be okay with dropping

agent docs includes documentation for this, including a script stub for if we need to update Lato in the future. explicitly not introduced to /scripts but I'm open to changing that (I tried a node version of the ttf manipulation, but it got funky / would need more work).


sepeterson:
thanks for taking this on! stuff like this is always such a headache

I noticed that if I just replace the tailwind styling on FontComponent in ScientificName with an inline style like `style={{ fontFamily: "Lato", fontWeight: "300", fontStyle: "italic" }}` , it achieves non-bold italics text.

italics

I'm wondering if just applying `fontFamily: "Lato"` via tailwind could save us the extra resolver and splitting up the font into more families. However I *did* also confirm that this breaks android fonts but I think we would just need to [set up a font resource to fix it](https://developer.android.com/guide/topics/resources/font-resource)

FLGMwt:
> thanks for taking this on! stuff like this is always such a headache
> 
> I noticed that if I just replace the tailwind styling on FontComponent in ScientificName with an inline style like `style={{ fontFamily: "Lato", fontWeight: "300", fontStyle: "italic" }}` , it achieves non-bold italics text.
> 
> italics
> I'm wondering if just applying `fontFamily: "Lato"` via tailwind could save us the extra resolver and splitting up the font into more families. However I _did_ also confirm that this breaks android fonts but I think we would just need to [set up a font resource to fix it](https://developer.android.com/guide/topics/resources/font-resource)

@sepeterson Well I wish I could unlearn that 😅

I'll need to test more broadly, but supposing that does work, we'd need to somehow enforce "only do font stuff in styles" to enforce this font-specific workaround.

Since we _mostly_ do styling with tailwind with rare one-off inlines, I'd rather bake the workaround in to the "system" than enforce an exception to use inlines for one styling category.

What do you think?

fwiw: at one point I did have the "resolver" happen at ScientificNames only but that felt risky acknowledging that we're fixing a "problem" with the fonts but only applying a fix at the case where we currently see it

sepeterson:
> Since we _mostly_ do styling with tailwind with rare one-off inlines, I'd rather bake the workaround in to the "system" than enforce an exception to use inlines for one styling category.

Yes! Sorry I wasn't trying to suggest we use inline styles as a workaround, rather reconfigure tailwind to apply those styles. We could collapse the existing [fontFamily config](https://github.com/inaturalist/iNaturalistReactNative/blob/1120d4603f96ad3bd8dfcc7325c443f09bb9ba10/tailwind.config.js#L40) to just `Lato: ["Lato"],` and handle bold/italics through the built-in tailwind [font weight](https://v2.tailwindcss.com/docs/font-weight) and [font style](https://v2.tailwindcss.com/docs/font-style)

FLGMwt:
> Yes! Sorry I wasn't trying to suggest we use inline styles as a workaround, rather reconfigure tailwind to apply those styles. We could collapse the existing [fontFamily config](https://github.com/inaturalist/iNaturalistReactNative/blob/1120d4603f96ad3bd8dfcc7325c443f09bb9ba10/tailwind.config.js#L40) to just `Lato: ["Lato"],` and handle bold/italics through the built-in tailwind [font weight](https://v2.tailwindcss.com/docs/font-weight) and [font style](https://v2.tailwindcss.com/docs/font-style)

Ah, I see. I'm baffled why that wouldn't hit the same family selection issue since that basically _reinforces_ what the rename is explicitly preventing. Maybe nativewind is actually doing something similar to the PR's workaround?

Thanks for looking into this! @sepeterson 

I'll give it another look on Monday when I'm fresh ; D marking draft until then

FLGMwt:
Closing, much simpler patch discovered for bugfix in RN 86

PR #4064, opened 2026-09-15
Merge beta into main

jtklein:


package.json
```diff
@@ -148,7 +148,7 @@
     "react-native-vision-camera": "^4.7.3",
     "react-native-volume-manager": "^2.0.8",
     "react-native-webview": "^13.15.0",
-    "react-native-worklets": "0.9.1",
+    "react-native-worklets": "0.7.0",
```
sepeterson, line 151: Looks like prior to the 0.9.1 upgrade, this was `^0.7.0` so the actually installed version was `0.7.4` whereas for this branch we have `0.7.0` in the lockfile.

Not too concerned about it but worth noting

sepeterson (approved):

package.json
jtklein: Yeah, I did that part in a hurry on PTO, so since this PR is the version that is tested now, I am more confident in the downgrade than going back to 0.7.4 without testing. And I have re-opened the ticket for 0.9 so I would guess it wouldn't take us a long time to get back up.

PR #4071, opened 2026-09-17
tsify the zustand root

FLGMwt:
closes MOB-1779

uses the existing slice types (or "slicies" types if you prefer my commit typo) to type the root. This should give us informed types when using the `useStore` without having the provide them ourselves.

The following screenshares demonstrate the before and after of this compared with `main` vs a `useStore` use with the annotation removed in the followup PR: https://github.com/inaturalist/iNaturalistReactNative/pull/4073/changes/00ade142dcc471a182a030a1bbb5a3dbe094bdb8

Note that in both cases: the firebase store value is unannotated, but receives types in the latter

type information hovering behavior before:

https://github.com/user-attachments/assets/b584789a-12f6-4ba8-a42a-7f00d294d5a2

type information hovering behavior after:

https://github.com/user-attachments/assets/cf520f66-36a9-4387-a66b-44318ad0d90b


websoty (approved):

src/stores/useStore.ts
```diff
@@ -85,12 +110,15 @@ const useStore = create( persist(
...
-      }, {} ) ),
+      layout: ( ( Object.keys( state.layout ) as ( keyof LayoutSlice["layout"] )[] ).reduce(
+        ( memo: Record<string, unknown>, key ) => {
+          if ( typeof ( state.layout[key] ) !== "function" ) {
+            memo[key] = state.layout[key];
+          }
+          return memo;
+        },
+        {},
+      ) ),
```
sepeterson, line 121: sincere apologies this PR is great as is but I couldn't help myself. We can eliminate the cast here if we switch to `Object.fromEntries`

Please feel totally free to discard if you want to keep changes type-only! I just see a challenge every time a ts cast pops up 🤓 

```suggestion
      layout: Object.fromEntries(
        Object.entries( state.layout ).filter( ( [, value] ) => typeof value !== "function" ),
      ),
```

sepeterson (approved):
Great stuff!

src/stores/useStore.ts
FLGMwt: argh! I swear I go back and forth on `.fromEntries` in ts-land. It used to be my _only_ thing for this sort of object math (to a fault), but I've gotten unlucky with inference / narrowing but I've been forced into enough casts that I wrote it off here. Totally makes sense though since this doesn't require any strict upstream key'd types, only js types.

I'll pull this into a future ts fix after I resolve my mistaken use of stacked PRs ; P

PR #4075, opened 2026-09-17
Completing ts zustand simplification

FLGMwt:
completes removing the explicit `useStore` type annotations in favor of deferring to the now baked-in slice typing on the store introduced in #4071. 

Lone exception: MyObservationsResults.tsx demonstrated that the upload slice has incomplete types for at least: `setStartUploadObservations`, `addToUploadQueue`, `addTotalToolbarIncrements`, `startManualSync`, `startAutomaticSync`. Leaving those alone for now.

PR #4076, opened 2026-09-18
type functions in `UploadObservationsSlice` and `SyncObservationsSlice`

FLGMwt:
#4075 unsurfaced missing types on `UploadObservationsSlice` and `SyncObservationsSlice`: we were typing the serializable values, but not the functions. This makes a distinction.

With that, the `MyObservationsResults` `useStore` calls are now free of ts errors w/o annotations.

I also missed `useFeatureFlag` which is _actually_ the last one of these


src/stores/createSyncObservationsSlice.ts
```diff
@@ -22,7 +22,21 @@ export interface SyncObservationsSlice {
   syncingStatus: SyncingStatus;
 }
 
-const DEFAULT_STATE: SyncObservationsSlice = {
+export interface SyncObservationsSlice extends SyncObservationsState {
```
jtklein: In previously migrated file with such big state and actions we used a `export type ObservationFlowSlice = ObservationFlowState & ObservationFlowActions;` syntax here with two different interfaces.
I don't mind which way we do it (& vs extends) but I would prefer consistency.

src/stores/createSyncObservationsSlice.ts
```diff
@@ -6,13 +6,13 @@ export const BEGIN_AUTOMATIC_SYNC = "begin-automatic-sync";
 export const MANUAL_SYNC_IN_PROGRESS = "manual-sync-progress";
 export const AUTOMATIC_SYNC_IN_PROGRESS = "automatic-sync-progress";
 
-type SyncingStatus = typeof SYNC_PENDING
+export type SyncingStatus = typeof SYNC_PENDING
```
jtklein: Don't see that imported somewhere.

src/stores/createSyncObservationsSlice.ts
```diff
@@ -6,13 +6,13 @@ export const BEGIN_AUTOMATIC_SYNC = "begin-automatic-sync";
...
   | typeof BEGIN_MANUAL_SYNC
   | typeof BEGIN_AUTOMATIC_SYNC
   | typeof MANUAL_SYNC_IN_PROGRESS
   | typeof AUTOMATIC_SYNC_IN_PROGRESS;
 
-export interface SyncObservationsSlice {
+export interface SyncObservationsState {
```
jtklein: Don't see that imported somewhere.

src/stores/createUploadObservationsSlice.ts
```diff
@@ -20,7 +20,7 @@ interface TotalUploadProgress {
   totalProgress: number;
 }
 
-export interface UploadObservationsSlice {
+export interface UploadObservationsState {
```
jtklein: Don't see that imported somewhere

src/stores/createUploadObservationsSlice.ts
```diff
@@ -34,7 +34,27 @@ export interface UploadObservationsSlice {
   uploadStatus: UploadStatus;
 }
 
-const DEFAULT_STATE: UploadObservationsSlice = {
+export interface UploadObservationsSlice extends UploadObservationsState {
```
jtklein: Same comment as above

jtklein: Would you like to extend the types into the fct params in Lines 74-116.

jtklein (approved):
LGTM, only minor comments.

FLGMwt:
> LGTM, only minor comments.

@jtklein thanks for the review! addressed in last commit

PR #4077, opened 2026-09-18
useShallow on layout prefs to prevent appwide rerenders

FLGMwt:
wrapping the layoutPrefs in a useShallow improves list performance nearly 2x

-----

I dismissed this potential as a minor opportunity when I was doing localObs stuff, I'm so sorry!

every consumer of useLayout gets a rerender triggered when any zustand store value changes. 

normally, useStore would skip a useStore listener call on shallow equality ("I don't care about that key being updated"), but layout's selector spreads a new object. This means that all useStore triggers across the app are incidentally also useLayoutStore triggers (the root reducer calls all slice reducers, the spread creates a new object every time).

ObsList is affected by this particularly because scroll index is stored in zustand. Scrolls trigger zustand updates which, because of useLayout, trigger FlashList rerenders.

---
Fix:

wrapping useLayout's selector with a useShallow applies the shallow equality check of just the arg object's keys equality to trigger a rerender. A component that only calls useLayout will no longer rerender when myObsOffsetchanges.

Rough video:  https://github.com/user-attachments/assets/3fdeb8a8-c3e2-4f9d-9f11-bab998d2b3de

warm restart before, without useShallow 1: 0:10 - FlashList useBenchmark over 3 runs reports JS FPS Avg: 18.1
warm restart before, without useShallow 2: 0:38 - FlashList useBenchmark over 3 runs reports JS FPS Avg: 17.6

warm restart after, with useShallow 1: 1:06  - FlashList useBenchmark over 3 runs reports JS FPS Avg: 33
warm restart after, with useShallow 2: 1:36  - FlashList useBenchmark over 3 runs reports JS FPS Avg: 32.2

framerate probably can't capture result fully but setup is:

- logged into staging-lite as mr10k
- at _least_ 1.5k are scrolled / loaded
- any scroll shouldn't trigger a "endPage" fetch
- all pages are warm, I've toggled multiple times before starting any tests

Test: relaunch myobslist with `useBenchmark(retries:3)` from flashlist

Condition A: `const useLayoutPrefs = ( ) => useStore( selector );`
Condition B: `const useLayoutPrefs = ( ) => useStore( useShallow( selector ) );`

W/ useShallow, we get almost _twice_ the FPS.

fixes: MOB-1795 (just created)


jtklein (approved):
All right now I am convinced that "All persisted layout state must be in this one object" which then in turn needs special handling can be moved past.
So, instead of relying on object spreading in persist logic and here, there is no shame in explicitly re-state in several places which keys we need, and keep the default store access zustand gives us.
Then we wouldn't have had this problem here, I think.
Anyways, LGTM

PR #4072, opened 2026-09-17
MOB-1600: comment on potential disagreement sheet

sepeterson:
Add and wire a comment button to match the other suggestion sheet. I wasn't sure whether to align the comment preview with the taxon thumbnail or the radio buttons but I think it makes more sense with the taxon info.

<details>
<summary>screenshots</summary>
Simulator Screenshot - iPhone 17 Pro - 2026-09-18 at 10 00 51
Simulator Screenshot - iPhone 17 Pro - 2026-09-18 at 10 00 13

</details>

Issue #2153, opened 2024-09-23
Can't tap agree after dismissing agree bottom sheet using background


kueda:
**Describe the bug**
When you dismiss the agree bottom sheet by tapping the background (not the x button), you can't tap the same agree button again.

**To Reproduce**
1. Sign in and view an observation with an agree button
2. Tap the agree button to bring up the bottom sheet
3. Tap on the background to dismiss the bottom sheet
4. Tap on the same agree button again
5. Nothing happens

**Expected behavior**
Something should happen, and that something is that the bottom sheet should appear again.

**Context (please complete the following information):**
 - Device: iPhone SE (3rd gen)
 - OS: iOS 17.6.1
 - App version: 0.47.0 (116)

**Additional context**
Reported by Tony, who was experiencing a similar bug in which he couldn't tap anything after dismissing the bottom sheet.

PR #4080, opened 2026-09-21
Stop stale route params reopening the suggest ID sheet

rainhead:
Hello! I finally got frustrated enough with bottom sheet handling today to spend some time with Claude investigating. Because we've run into a few different issues with different sheets, I encouraged it to find structural fixes to the underlying problems. This PR is one of those.

I have experience with native apps, several decades of experience with JavaScript, and some experience with React, but React Native is mostly new to me. Besides reviewing the work myself, I got adversarial reviews from Fable 5.1 and @coderabbitai.

What follows is an analysis from Claude Code.

## What a user sees today

Add an identification from Suggestions, close the sheet, then leave a comment on the observation — and the Suggest ID sheet comes back on its own, with the comment you had typed into it gone. Backgrounding and reopening the app does it too, as does a network reconnect. Nothing the user did asked for the sheet again.

Here is the whole of it, on `main`:

1. Open an observation, tap Suggest ID, pick *Apis mellifera*. The sheet appears.
2. Tap the X. The sheet closes.
3. Add an ordinary comment to the observation.
4. The Suggest ID sheet is back, offering *Apis mellifera* again.

Step 3 is doing nothing more sinister than refetching the observation. Any refetch will do: `useFocusEffect` invalidates on focus, `focusManager` refetches when the app foregrounds, `onlineManager` refetches on reconnect, and both comment and identification mutations refetch on success.

## Why

Two facts meet badly.

`identTaxonId`, `identAt` and `identTaxonFromVision` mean *the user just picked a taxon*. That is an event. They were left sitting in the route as state, and only a successful submit ever cleared them — closing the sheet did not.

Meanwhile [the effect that consumed them](https://github.com/inaturalist/iNaturalistReactNative/blob/e64f0ba300b8fdc0be11941c905606e7731253d9/src/components/ObsDetailsDefaultMode/IdentificationSheets.tsx) depended on `hasPotentialDisagreement`, which is rebuilt whenever `observationShown` is replaced. `observationShown` was replaced on **every** refetch, because `Observation.mapApiToRealm` builds a new object each call and the dispatch was keyed on `isRefetching` flipping rather than on the data changing. So an unrelated refetch invalidated the callback, re-ran the effect, found the params still sitting there, and restarted a flow the user had already finished.

The comment vanishing is the same effect re-running: its `SET_NEW_IDENTIFICATION` passes no `body`, so it rebuilds the pending identification without whatever the user had typed.

## What changes

**Route params are consumed as one-shot events.** New hook [`useRouteEvent`](https://github.com/inaturalist/iNaturalistReactNative/blob/e64f0ba300b8fdc0be11941c905606e7731253d9/src/sharedHooks/useRouteEvent.ts) reads the params, clears them, and *then* hands them to its caller — so an event is delivered exactly once per navigation that carries it. "One-shot" here just means the params cannot be read a second time; the hook clears them itself, so no exit path has to remember to. That last part matters: the first version of this fix called a `clearIdentParams()` helper on three separate exit paths, and a fourth exit path added later would have quietly reintroduced the bug.

The hook keys on the params object's identity. React Navigation builds a new params object for every `navigate`, `popTo` and `setParams` and keeps it stable otherwise, so that identity is exactly "a navigation happened". Worth knowing when writing tests against it: a mock whose `setParams` mutates the params object in place will not behave like the real router, and two of the tests here had to be corrected for that.

**The disagreement question is settled once, by which sheet opens.** The Suggest ID sheet only appears when the identification does not disagree, so `onSuggestId` no longer needs to ask. It used to re-derive the answer on press, against whatever observation object was current by then. In principle that could flip and open the disagreement sheet over a sheet that was still up, submitting nothing — no test here demonstrates that happening, and removing the second derivation makes it moot either way. It also orphaned `identTaxon` and `CLEAR_SUGGESTED_TAXON`, which nothing read, so those are gone.

**`observationShown` is only replaced when there is something new.** React Query's structural sharing keeps `remoteObservation`'s identity stable when the data has not changed, so that identity is the honest signal. A refetch that settles with nothing new — an identical response, or a failure that leaves the previous data in place — now dispatches [`ACTIVITY_ITEM_SETTLED`](https://github.com/inaturalist/iNaturalistReactNative/blob/e64f0ba300b8fdc0be11941c905606e7731253d9/src/components/ObsDetailsSharedComponents/hooks/useObsDetailsSharedLogic.ts), a new action that does nothing but lower the pending-activity spinner. That separation is load-bearing: on someone else's observation this effect is the only thing that lowers the spinner, because `handleIdentificationMutationSuccess` and `handleCommentMutationSuccess` only dispatch directly for your own observations. Without it, a flaky refetch would leave the spinner up.

**Out-of-order taxon fetches.** Picking a taxon that is not in Realm starts a network fetch. Pick again before it lands and both flows would dispatch, last to resolve winning. Each event now takes a flow id and only the latest may act; unmounting invalidates whatever is in flight. A failed fetch is reported rather than swallowed, since nothing will retry it once the event is consumed.

## What this does *not* fix

This started from a report of the Suggest ID sheet staying up after an identification submitted successfully. **This PR does not claim to fix that**, and the investigation contradicted the premise I began with.

That symptom looked like an orphaned `@gorhom/bottom-sheet` portal — the sheet's content outliving its owner because the sheet is closed by unmounting it rather than by telling it to close. I reproduced that failure exactly once, in an instrumented harness, and then could not reproduce it again in 36 controlled passes. A rate that low is consistent with the original "sometimes", but it means I could not validate any fix against it.

The structural answer there is to stop unmounting sheets to close them — always mount them and drive a `visible` prop, with `onDismiss` required so a user-initiated close always reports back. `BottomSheetModal.handleDismiss` and `handlePortalOnUnmount` both early-return without removing the portal, so that change makes the failure unreachable by construction rather than by measurement. That is a rewrite of a file #4072 is actively changing, so it belongs in its own PR, after this one and after #4072. Whether to do it at all is not decided here.

Prior attempts at this family, for whoever picks that up: #3333 (ID with a comment leaving the sheet in a bad state), #3576 ("the sheet could linger or reappear after navigation because cleanup didn't run reliably when we were returning null" — fixed in the wrappers, never in `IdentificationSheets`), #3301 and #2153 (parent state not told when a sheet is swiped away). `SuggestIDSheet` still does not pass `onDismiss` to `BottomSheetV2`, so the prop #3333 added is unused on that sheet.

## Merge order

#4072 should land first. It is in the same file and fixes two things this PR deliberately leaves alone: `SHOW_POTENTIAL_DISAGREEMENT_SHEET` not clearing `showSuggestIdSheet`, and `onChangeIdentBody` dropping the `vision` flag. Neither conflicts with the code here. The test file will conflict trivially — both PRs append a `describe` at the end, and #4072's tests will need the `useNavigation` mock this PR adds at module level.

## Tests

Three new behaviour-level tests in [IdentificationSheets.test.js](https://github.com/inaturalist/iNaturalistReactNative/blob/e64f0ba300b8fdc0be11941c905606e7731253d9/tests/unit/components/ObsDetailsDefaultMode/IdentificationSheets.test.js), two of which were red before this change:

- `does not reopen the suggest ID sheet the user already closed`
- `keeps a comment the user already typed`
- `ignores an earlier taxon fetch that resolves after a later one`

Each was verified to fail without its corresponding fix, not merely to pass with it. A fourth, `closes the sheet and leaves it closed when the observation refetches`, passed before and after; it is here because it is what ruled out this bug as the explanation for the originally reported one.

26 suites / 94 tests pass across everything ObsDetails-related. No new eslint errors (one pre-existing `preserve-manual-memoization` warning on `onAgree`, present on `main`). `lint:tsc` output for the touched files is identical to the pre-change baseline.

## No screenshots

The repo asks for a visual summary when a PR changes what the user sees. This one changes the *absence* of a behaviour — a sheet that no longer reappears, a comment that no longer vanishes. There is no new visual vocabulary and no still frame that would convey it. A recording of the before-state would be the genuinely useful artifact, but reproducing it needs the real Suggestions → ObsDetails flow against a live account, which I could not do. Saying so rather than skipping the requirement silently.

## Two unrelated findings

Raising separately, not addressed here:

- This repo does not build on Xcode 27. Pod resource bundles carry `IPHONEOS_DEPLOYMENT_TARGET` of 9.0–12.0 against a 15.0 floor; a `post_install` hook in the Podfile is the fix.
- A Debug build needs `.env.staging` and `ios/GoogleService-Info.staging.plist` specifically, not `.env`/`GoogleService-Info.plist`, and `[FIRApp configure]` aborts at launch if the plist is missing or its API key is not exactly 39 characters. None of that is in the README.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rainhead:
Superseded by #4082. Renaming this branch to follow the `<issue>-<description>` convention closed this PR, and GitHub doesn't allow changing a PR's head branch, so the same change is reopened there with the renamed branch and `Closes #4081`. The code is identical. 🤖

